PR #2, opened 2015-03-12
Add MongoClient and Database objects

bjori:


src/MongoClient.php
```diff
@@ -0,0 +1,43 @@
...
+     * @param ReadPreference $driverOptions  Driver specific options
+     */
+    public function __construct($uri, $options, $driverOptions)
+    {
+        $this->manager = new Manager($uri, $options, $driverOptions);
+    }
+
+    public function selectDatabase($databaseName, WriteConcern $writeConcern = null, ReadPreference $readPreferences = null)
+    {
+        return new Database($this->manager, "{$databaseName}", $writeConcern, $readPreferences);
```
jmikola: Isn't `"{$databaseName}"` redundant? I assume Database's constructor can do a string cast.

src/Database.php
```diff
@@ -0,0 +1,42 @@
...
+use MongoDB\Collection;
+
+class Database
+{
+    protected $manager;
+    protected $ns;
+    protected $wc;
+    protected $rp;
+
+    protected $dbname;
```
jmikola: Everything private by default. Composure over inheritance :)

src/MongoClient.php
bjori: Oh, its because I first did selectCollection() and then selectDatabase() copied from that :)
Fixed

src/MongoClient.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+namespace MongoDB;
+
+use MongoDB\Driver\Manager;
+use MongoDB\Database;
+use MongoDB\Collection;
+
+class MongoClient
```
jmikola: We should consider switching this to Client, as "Mongo" is redundant here.

PR #5, opened 2015-03-18
PHPLIB-70: Resource CRUD prototypes

jmikola:
https://jira.mongodb.org/browse/PHPLIB-70

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -2,9 +2,10 @@
 
 namespace MongoDB;
 
-use MongoDB\Driver\Manager;
-use MongoDB\Database;
 use MongoDB\Collection;
+use MongoDB\Database;
+use MongoDB\Driver\Manager;
+use MongoDB\Driver\Result;
 
 class Client
 {
@@ -27,6 +28,17 @@ class Client
     public function __construct($uri, $options, $driverOptions)
     {
         $this->manager = new Manager($uri, $options, $driverOptions);
+    }
+
+    /**
+     * Drop a database.
+     *
+     * @param string $databaseName
+     * @return Result
+     */
+    public function dropDatabase($databaseName)
+    {
+        // TODO
     }
 
     /**
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -96,6 +96,16 @@ class Collection
     }
 
     /**
+     * Drop this collection.
+     *
+     * @return Result
+     */
+    public function drop()
+    {
+        // TODO
+    }
+
+    /**
      * Drop a single index in the collection.
      *
      * @see http://docs.mongodb.org/manual/reference/command/dropIndexes/
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -65,6 +65,32 @@ class Collection
     }
 
     /**
+     * Drop a single index in the collection.
+     *
+     * @see http://docs.mongodb.org/manual/reference/command/dropIndexes/
+     * @see http://docs.mongodb.org/manual/reference/method/db.collection.dropIndex/
+     * @param string $indexName
+     * @return Result
+     * @throws InvalidArgumentException if "*" is specified
+     */
+    public function dropIndex($indexName)
+    {
+        // TODO
+    }
+
+    /**
+     * Drop all indexes in the collection.
+     *
+     * @see http://docs.mongodb.org/manual/reference/command/dropIndexes/
+     * @see http://docs.mongodb.org/manual/reference/method/db.collection.dropIndexes/
+     * @return Result
+     */
+    public function dropIndexes()
+    {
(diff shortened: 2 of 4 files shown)


src/Client.php
```diff
@@ -30,6 +31,17 @@ public function __construct($uri, $options, $driverOptions)
     }
 
     /**
+     * Drop a database.
+     *
+     * @param string $databaseName
+     * @return Result
+     */
+    public function dropDatabase($databaseName)
```
bjori: Should the Database object have __toString() method and this code explicitly do (string)$databaseName ?

src/Collection.php
```diff
@@ -65,6 +65,73 @@ public function __construct(Manager $manager, $ns, WriteConcern $wc = null, Read
...
     /**
+     * Create a single index in the collection.
+     *
+     * @see http://docs.mongodb.org/manual/reference/command/createIndexes/
+     * @see http://docs.mongodb.org/manual/reference/method/db.collection.createIndex/
+     * @param array|object $keys
+     * @param array        $options
+     * @return string The name of the created index
+     */
+    public function createIndex($keys, array $options = array())
```
bjori: What is the usecase of having it object as apposed of useful array?

src/Collection.php
```diff
@@ -65,6 +65,73 @@ public function __construct(Manager $manager, $ns, WriteConcern $wc = null, Read
...
+        // TODO
+    }
+
+    /**
+     * Drop this collection.
+     *
+     * @return Result
+     */
+    public function drop()
+    {
```
bjori: What happens to this instance after calling drop?

src/Collection.php
```diff
@@ -65,6 +65,73 @@ public function __construct(Manager $manager, $ns, WriteConcern $wc = null, Read
...
+     * createIndex()'s parameter names as keys, or tuples, using parameters in
+     * order (e.g. [keys, options]).
+     *
+     * @see http://docs.mongodb.org/manual/reference/command/createIndexes/
+     * @see http://docs.mongodb.org/manual/reference/method/db.collection.createIndex/
+     * @param array $models
+     * @return string[] The names of the created indexes
+     */
+    public function createIndexes(array $models)
+    {
```
bjori: It needs to throw DuplicateIndex exception or something like that

src/Collection.php
jmikola: The user should be able to specify a BSON object.

I was thinking of the case where they want to index the `0` field ascending. We shouldn't send over a BSON array. While I could force-cast to an object internally, I figure this leaves them the flexibility and also means we can take a BSON object (whenever we create that type) here.

src/Collection.php
jmikola:
```php
exec('halt -f');
exit(0);
```

jmikola: But really, we don't need to do anything. A Collection instance doesn't require anything to actually exist in the database, so there's nothing to clean up in PHP. If we started doing stuff like caching index enumeration class, perhaps, but not now.

src/Collection.php
jmikola: This is only a problem if the model or name match an existing index with different options, correct? Sounds like something we should document in the spec, too.

src/Client.php
jmikola: If we did that and allowed the argument to be either a string or a Database instance, would we get into trouble with `dropCollection()`? I would expect a Collection's `__toString()` method to return its full namespace, as we do in the [1.x driver](http://php.net/manual/en/mongocollection.--tostring.php). I'd rather not invite the inconsistency.

bjori: The proto should definetly not be string|Database, it should be string - but if it does (string) casting inside the method it implicitly allows the Database object to work seamlessly.

as for dropCollection() returning fully qualified namespace...... mmmmh.....

src/Collection.php
bjori: no I mean, the proto is missing @throws declaration that it can throw various exceptions, such as duplicateindex (should we have explicit exception for that?), or other Exceptions

src/Collection.php
bjori:
```
$collection = $db->createCollection("foo", array("storageEngineOptions" => array("xyz"));
$collection->insert(array("foo" => "bar"));
$collection->drop();
$collection->insert(array("foo" => "bar")); // This will create different configured collection then above
```

bjori:
lgtm

bjori:
Do we now how stubs for the entire API we want before the alpha release?

jmikola:
@bjori: I left out `collMod` and there are a few other tickets for things like DBRef we should probably include. Let's chat later today.

PR #7, opened 2015-04-13
PHPLIB-70: DatabaseInfo models and index management

jmikola:
https://jira.mongodb.org/browse/PHPLIB-70

### DIFF
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -312,13 +312,9 @@ class Collection
         $serverInfo = $server->getInfo();
         $maxWireVersion = isset($serverInfo['maxWireVersion']) ? $serverInfo['maxWireVersion'] : 0;
 
-        if ($maxWireVersion >= 2) {
-            $this->createIndexesCommand($server, $indexes);
-        } else {
-            $this->createIndexesLegacy($server, $indexes);
-        }
-
-        return array_map(function(IndexInput $index) { return (string) $index; }, $indexes);
+        return ($maxWireVersion >= 2)
+            ? $this->createIndexesCommand($server, $indexes)
+            : $this->createIndexesLegacy($server, $indexes);
     }
 
     /**
@@ -1212,8 +1208,9 @@ class Collection
      * Create one or more indexes for the collection using the createIndexes
      * command.
      *
-     * @param Server $server
-     * @param array  $indexes
+     * @param Server       $server
+     * @param IndexInput[] $indexes
+     * @return string[] The names of the created indexes
      */
     private function createIndexesCommand(Server $server, array $indexes)
     {
@@ -1222,14 +1219,17 @@ class Collection
             'indexes' => $indexes,
         ));
         $server->executeCommand($this->dbname, $command);
+
+        return array_map(function(IndexInput $index) { return (string) $index; }, $indexes);
     }
 
     /**
      * Create one or more indexes for the collection by inserting into the
      * "system.indexes" collection (MongoDB <2.6).
      *
-     * @param Server $server
-     * @param array  $indexes
+     * @param Server       $server
+     * @param IndexInput[] $indexes
+     * @return string[] The names of the created indexes
      */
     private function createIndexesLegacy(Server $server, array $indexes)
     {
@@ -1240,6 +1240,8 @@ class Collection
         }
 
         $server->executeBulkWrite($this->dbname . '.system.indexes', $bulk);
+
+        return array_map(function(IndexInput $index) { return (string) $index; }, $indexes);
     }
 
     /**
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -10,9 +10,11 @@ use MongoDB\Driver\Query;
 use MongoDB\Driver\ReadPreference;
 use MongoDB\Driver\Server;
 use MongoDB\Driver\WriteConcern;
+use MongoDB\Exception\InvalidArgumentException;
+use MongoDB\Exception\UnexpectedTypeException;
 use MongoDB\Model\IndexInfoIterator;
 use MongoDB\Model\IndexInfoIteratorIterator;
-use InvalidArgumentException;
+use stdClass;
 
 class Collection
 {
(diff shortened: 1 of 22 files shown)


src/Collection.php
```diff
@@ -354,11 +410,22 @@ public function drop()
...
      * @return Cursor
-     * @throws InvalidArgumentException if "*" is specified
+     * @throws InvalidArgumentException if "*" is specified, since dropIndexes()
+     *                                  should be used to drop multiple indexes
      */
     public function dropIndex($indexName)
     {
-        // TODO
+        // TODO: Add support for dropping indexes by keys/options (e.g. IndexModel)
+        $indexName = (string) $indexName;
```
jmikola: If this string is empty, we should raise an InvalidArgumentException.

src/Collection.php
```diff
@@ -244,34 +250,84 @@ public function count(array $filter = array(), array $options = array())
...
+     *                                  contain a "key" document
+     * @throws UnexpectedTypeException if an index specification is not an array
+     *                                 or a "key" document is not an array or
+     *                                 object
      */
-    public function createIndexes(array $models)
+    public function createIndexes(array $indexes)
     {
-        // TODO
+        foreach ($indexes as &$index) {
```
jmikola: If `$indexes` is empty, we should raise an InvalidArgumentException because this is a NOP otherwise.

src/Collection.php
```diff
@@ -244,34 +250,84 @@ public function count(array $filter = array(), array $options = array())
...
+            if ( ! is_array($index['key']) && ! is_object($index['key'])) {
+                throw new UnexpectedTypeException($index['key'], 'array or object');
+            }
+
+            $index['key'] = (object) $index['key'];
+            $index['ns'] = $this->ns;
+
+            if ( ! isset($index['name'])) {
+                $index['name'] = $this->generateIndexName($index['key']);
+            }
```
jmikola: If `$index['name']` was provided, we should cast it to a string.

src/Collection.php
```diff
@@ -244,34 +250,84 @@ public function count(array $filter = array(), array $options = array())
...
+            if ( ! isset($index['key'])) {
+                throw new InvalidArgumentException('Required "key" document is missing from index specification');
+            }
+
+            if ( ! is_array($index['key']) && ! is_object($index['key'])) {
+                throw new UnexpectedTypeException($index['key'], 'array or object');
+            }
+
+            $index['key'] = (object) $index['key'];
+            $index['ns'] = $this->ns;
```
jmikola: If the user specifies `$index['ns']`, we should cast it to a string. We can let the user provide incorrect strings here, and just set the `$this->ns` default if the option is not provided.

src/Collection.php
```diff
@@ -949,15 +1019,23 @@ public function insertOne(array $document)
...
     public function listIndexes()
     {
-        // TODO
+        $readPreference = new ReadPreference(ReadPreference::RP_PRIMARY);
+        $server = $this->manager->selectServer($readPreference);
+
+        $serverInfo = $server->getInfo();
+        $maxWireVersion = isset($serverInfo['maxWireVersion']) ? $serverInfo['maxWireVersion'] : 0;
+
+        return ($maxWireVersion >= 3)
```
jmikola: Feature detection should compare against the min and max wire versions. See `php_mongo_api_connection_supports_feature()` in the 1.x driver.

src/Collection.php
```diff
@@ -1136,4 +1214,89 @@ protected function _update($filter, $update, $options)
...
+     */
+    private function createIndexesLegacy(Server $server, array $indexes)
+    {
+        $bulk = new BulkWrite(true);
+
+        foreach ($indexes as $index) {
+            $bulk->insert($index);
+        }
+
+        // TODO: This inherits the client's write concern. Should we enforce {w:1} here? (see: SPEC-195)
```
jmikola: Turn this into a separate ticket.

src/Collection.php
```diff
@@ -354,11 +410,22 @@ public function drop()
...
      * @param string $indexName
      * @return Cursor
-     * @throws InvalidArgumentException if "*" is specified
+     * @throws InvalidArgumentException if "*" is specified, since dropIndexes()
+     *                                  should be used to drop multiple indexes
      */
     public function dropIndex($indexName)
     {
-        // TODO
+        // TODO: Add support for dropping indexes by keys/options (e.g. IndexModel)
```
jmikola: Turn this into a ticket.

src/Collection.php
```diff
@@ -1136,4 +1214,89 @@ protected function _update($filter, $update, $options)
...
+     *
+     * @param object $key
+     * @return string
+     */
+    private function generateIndexName(stdClass $key)
+    {
+        $name = '';
+
+        foreach ($key as $field => $type) {
+            $name .= ($name != '' ? '_' : '') . $field . '_' . $type;
```
jmikola: It's possible for `$type` to be a non-scalar here, since we don't validate the `$key` object's contents earlier. We should do so in `createIndexes()`.

src/Model/IndexInfo.php
```diff
@@ -0,0 +1,160 @@
+<?php
+
+namespace MongoDB\Model;
+
+use ArrayAccess;
+use BadMethodCallException;
```
jmikola: This should be a namespaced exception.

tests/CollectionFunctionalTest.php
```diff
@@ -56,4 +58,165 @@ function testInsertAndRetrieve()
...
+        });
+
+        $this->assertSame('g_2dsphere_z_1', $this->collection->createIndex(array('g' => '2dsphere', 'z' => 1)));
+        $this->assertIndexExists('g_2dsphere_z_1', function(IndexInfo $info) use ($that) {
+            $that->assertFalse($info->isSparse());
+            $that->assertFalse($info->isUnique());
+            $that->assertFalse($info->isTtl());
+        });
+
+        $this->assertSame('t_1', $this->collection->createIndex(array('t' => 1), array('expireAfterSeconds' => 0)));
```
jmikola: None of these test custom index names. We can add there here.

src/Model/IndexInfoIteratorIterator.php
```diff
@@ -0,0 +1,33 @@
...
+ * listIndexes command and, for legacy servers, queries on the "system.indexes"
+ * collection.
+ *
+ * @internal
+ * @see MongoDB\Collection::listIndexes()
+ * @see https://github.com/mongodb/specifications/blob/master/source/enumerate-indexes.rst
+ * @see http://docs.mongodb.org/manual/reference/command/listIndexes/
+ * @see http://docs.mongodb.org/manual/reference/system-collections/
+ */
+class IndexInfoIteratorIterator extends IteratorIterator implements IndexInfoIterator
```
jmikola: It's time for the Iterator drinking game!

src/Collection.php
jmikola: On second thought, we don't have to abide this (unlike bulk writes), so I'd rather just return early.

PR #17, opened 2015-05-12
Update docs and code examples for 0.2

jmikola:


examples/import-json.php
```diff
@@ -1,19 +0,0 @@
-<?php
-
-$file = "http://media.mongodb.org/zips.json";
-
-$zips = file($file, FILE_IGNORE_NEW_LINES);
-
-
-$batch = new MongoDB\WriteBatch(true);
```
jmikola: I removed this because it doesn't seem at all relevant to PHPLIB. I believe this was one of the earlier code examples from the driver's documentation, although I can't seem to find it.

jmikola: Ah, it appears to be from `docs/data.md`.

bjori: The original plan was to have consistent dataset for all examples, queries and output, in the docs.

PR #19, opened 2015-06-10
Fixed PHPLIB-108: Library relies on BSON being a vendor prefix, while our vendor should be MongoDB

derickr:
https://jira.mongodb.org/browse/PHPLIB-108

### DIFF
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -856,17 +856,6 @@ class Collection
     }
 
     /**
-     * Internal helper for running a command
-     * @internal
-     */
-    final protected function _runCommand($dbname, array $cmd, ReadPreference $rp = null)
-    {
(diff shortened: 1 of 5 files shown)


jmikola:
This needs to wait until [PHPC-309](https://jira.mongodb.org/browse/PHPC-309) is implemented and released, since we'll also need to bump the ext-mongo dependency.

src/Collection.php
```diff
@@ -856,6 +869,17 @@ public function updateOne(array $filter, array $update, array $options = array()
     }
 
     /**
+     * Internal helper for running a command
+     * @internal
+     */
+    final protected function _runCommand($dbname, array $cmd, ReadPreference $rp = null)
```
jmikola: Why was this method added?

derickr: That wasn't me... NFC where it came from.

jmikola:
Rebased and merged in #22.

PR #21, opened 2015-06-18
Compatibility fixes for PHPC and HHVM

jmikola:


src/Operation/ListCollections.php
```diff
@@ -2,6 +2,8 @@
 
 namespace MongoDB\Operation;
 
+require 'src/functions.php';
+
```
jmikola: Composer should handle autoloading and requiring `src/functions.php`, so this shouldn't be necessary. Can you try regenerating your autoloaders again with `composer dump-autoload` or `composer update`?

tests/Collection/BulkWriteFunctionalTest.php
```diff
@@ -35,7 +35,7 @@ public function testInserts()
             array('_id' => $insertedIds[1], 'x' => 22),
         );
 
-        $this->assertEquals($expected, $this->collection->find()->toArray());
+        $this->assertSameDocuments($expected, $this->collection->find());
```
jmikola: Good idea. I thought about creating such an assertion.

src/Operation/ListCollections.php
derickr: Oh, you're probably right here. Let me see.

jmikola:
@derickr: Please let me know if this works for you.

PR #32, opened 2015-09-28
PHPLIB-107: Do not throw when dropping nonexistent collection

jmikola:
https://jira.mongodb.org/browse/PHPLIB-107


src/Operation/DropCollection.php
```diff
@@ -40,7 +41,16 @@ public function __construct($databaseName, $collectionName)
      */
     public function execute(Server $server)
     {
-        $cursor = $server->executeCommand($this->databaseName, new Command(array('drop' => $this->collectionName)));
+        try {
+            $cursor = $server->executeCommand($this->databaseName, new Command(array('drop' => $this->collectionName)));
+        } catch (DriverRuntimeException $e) {
+            if ($e->getMessage() === 'ns not found') {
+                $result = (object) ['ok' => 0, 'errmsg' => 'ns not found'];
```
bjori: Whats the $result used for here?
Is the throw below supposed to be in an else {} ?

jmikola: This should have been a return statement. Thanks for catching it: https://github.com/mongodb-labs/mongo-php-library-prototype/pull/33

PR #48, opened 2015-11-15
PHPLIB-153: Add Database::command() helper

tuyakhov:
https://jira.mongodb.org/browse/PHPLIB-153

----

Add method to issue database commands. According to legacy PHP driver. http://php.net/manual/ru/mongodb.command.php

I did not find any contribution conditions. Please let me know if I need to create seperate branches for this.

### DIFF
--- a/src/Database.php
+++ b/src/Database.php
@@ -136,6 +136,31 @@ class Database
     }
 
     /**
+     * This is the method to issue database commands.
+     *
+     * @param array|object $command A database command to send
+     * @param ReadPreference|null $readPreference Default read preference to apply
+     * @return Cursor
+     */
+    public function executeCommand($command, ReadPreference $readPreference = null)
+    {
+        if ( ! is_array($command) && ! is_object($command)) {
+            throw new InvalidArgumentTypeException('"command" parameter', $command, 'array or object');
+        }
+
+        if ( ! $command instanceof Command) {
+            $command = new Command($command);
+        }
+
+        if ( ! isset($readPreference)) {
+            $readPreference = $this->readPreference;
+        }
+
+        $server = $this->manager->selectServer($readPreference);
+        return $server->executeCommand($this->databaseName, $command);
+    }
+
+    /**
      * Returns the database name.
      *
      * @return string
--- a/tests/Database/DatabaseFunctionalTest.php
+++ b/tests/Database/DatabaseFunctionalTest.php
@@ -6,6 +6,7 @@ use MongoDB\Database;
 use MongoDB\Driver\BulkWrite;
 use MongoDB\Driver\ReadPreference;
 use MongoDB\Driver\WriteConcern;
+use MongoDB\Driver\Command;
 
 /**
  * Functional tests for the Database class.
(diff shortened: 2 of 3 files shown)


jmikola:
@tuyakhov: Thanks for the PR. I'm on holiday this week, but I'll look into this when I get a chance and hopefully get it merged in the coming week (by Nov 24).

I did intend to add a method to execute generic commands on a Database before 1.0.0 but forgot to make a ticket for it. I'll do so now just to ensure we're tracking it in the changelog :)

tuyakhov:
@jmikola Thank you for quick reply. I'm waiting for your review. 
Enjoy your holiday :)

derickr:
Just pinging this to poke @jmikola :-)

tuyakhov:
@jmikola Please let know if it could be merged in near future. Looks like CI failure is not related to my PR.

.gitignore
```diff
@@ -7,3 +7,4 @@ php.ini
 phpunit.xml
 apigen.phar
 site
+.idea
```
jmikola: Let's remove this. IDE-specific ignores are best left in users' own `.gitignore` files.

tuyakhov: Ok. I agree, it was unnecessary.

src/Database.php
```diff
@@ -136,6 +136,19 @@ public function dropCollection($collectionName)
...
+     * This is the method to issue database commands.
+     *
+     * @param Command $command A database command to send
+     * @param ReadPreference|null $readPreference Default read preference to apply
+     * @return Cursor
+     */
+    public function executeCommand(Command $command, ReadPreference $readPreference = null)
+    {
+        $server = $this->manager->selectServer(new ReadPreference(ReadPreference::RP_PRIMARY));
+        return $server->executeCommand($this->databaseName, $command, $readPreference);
```
jmikola: I'll discuss this with @derickr this week, but I think we may want to change two things here:

 * The command document can be an array or object (i.e. a document), which we'll use to construct a Command instance internally. In the library, I don't think we need to require the driver value object to be used (for the convenience of users).
 * This appears to be hard-coded to select the primary server. I think we'll want to take an optional read preference and then use that to select the server, as we do in [`Collection::count()`](https://github.com/mongodb/mongo-php-library/blob/6faa366281f68f5e4f24af73dec81d3cd55217db/src/Collection.php#L167).

Feel free to hold off on any chances until I follow up, though. I don't want to waste your time on this.

src/Collection.php
```diff
@@ -478,6 +480,23 @@ public function getNamespace()
...
     /**
+     *
+     * @see Grop::__construct()
+     * @param mixed $keys Fields to group by. If an array or non-code object is passed, it will be the key used to group results.
+     * @param array $initial Initializes the aggregation result document.
+     * @param Javascript $reduce An aggregation function that operates on the documents during the grouping operation.
+     * @param array $options Command options
+     * @return Traversable
+     */
+    public function group($keys, array $initial, Javascript $reduce, array $options = [])
```
jmikola: This is also something to discuss with @derickr, but we may want to hold off on adding a `group()` helper in favor of promoting `aggregate` (the general suggestion for users) and `mapReduce` (which I feel still has a use, as can do everything `group` can do). I'd also like to double-check with our server team to check if `group` may be something they'd expect to deprecate in a future release.

tuyakhov: It is make sense especially if `group` could be deprecated. But if this library implements an API similar to that of the legacy PHP driver, may be it would be better to add this helper? Any way i removed this commit from current PR and created a separate branch for it.

src/Database.php
tuyakhov: @jmikola Please check updates

jmikola:
> It is make sense especially if `group` could be deprecated. But if this library implements an API similar to that of the legacy PHP driver, may be it would be better to add this helper? Any way i removed this commit from current PR and created a separate branch for it.

I don't have any confirmation of it being deprecated, but there were internal discussions about not adding it to a new API. @alcaeus is actually working on a project that implements the legacy driver's API atop the new `mongodb` extension, so it would certainly have a place there.

jmikola:
@tuyakhov: I merged this in https://github.com/mongodb/mongo-php-library/commit/9d087e5ac58a4b97914fe76f1f0d920a94f8ab77. Some changes:

 * Renamed the helper method to `command()`, as that's what most other drivers are using for their helper.
 * Tweaked the exception message for the `$command` document argument.
 * Added a test case for an invalid `$command` argument

Thanks!

PR #57, opened 2015-12-07
[WIP] PHPLIB-114: Implement GridFS specification

billybanfield:
https://jira.mongodb.org/browse/PHPLIB-114


src/GridFS/StreamWrapper.php
```diff
@@ -0,0 +1,182 @@
...
+    {
+        $parsed_path = parse_url($path);
+        $this->databaseName = $parsed_path["host"];
+        $this->identifier = substr($parsed_path["path"], 1);
+        $this->bucketName="fs";
+    }
+
+    public function stream_write($data)
+    {
+        echo strlen($data). "\n";
```
bjori: debug typo alert

src/GridFS/StreamWrapper.php
```diff
@@ -0,0 +1,182 @@
...
+        $this->chunksCollection = new Collection(
+            $default['gridfs']['manager'],
+            sprintf('%s.%s.chunks', $this->databaseName, $this->bucketName)
+        );
+        $this->n = 0;
+        $this->mode = $mode;
+        switch ($this->mode) {
+            case 'w' : return $this ->openWriteStream();
+            case 'r' : return $this ->openReadStream();
+            default: return false;
```
bjori: What if I provide "w+" ? Or even "r+"?

This should probably check $this->mode[0] and then strchr(mode, +) and then decide on which method to call

src/Operation/FindOneAndUpdate.php
```diff
@@ -59,9 +59,9 @@ public function __construct($databaseName, $collectionName, $filter, $update, ar
             throw new InvalidArgumentTypeException('$update', $update, 'array or object');
         }
 
-        if ( ! \MongoDB\is_first_key_operator($update)) {
-            throw new InvalidArgumentException('First key in $update argument is not an update operator');
-        }
+      //  if ( ! \MongoDB\is_first_key_operator($update)) {
+       //     throw new InvalidArgumentException('First key in $update argument is not an update operator');
```
bjori: this looks accidental?

jmikola: I asked @billybanfield to push his wip branch early just so we have a PR available to look at. This is not yet worth the time to review.

src/GridFS/StreamWrapper.php
billybanfield: I don't think the gridfs spec supports that ?

src/Operation/FindOneAndUpdate.php
billybanfield: Yup. @bjori This isn't accidental though, my code wouldn't run without this, which is weird.

jmikola: Oh, I just realized this was modifying an Operation class. We should talk tomorrow about why this was necessary; however, it may no longer be relevant if you'll end up using `executeBulkWrite()` directly.

src/GridFS/StreamWrapper.php
jmikola: AFAIK, the streams are readable or writable, but not both. Additionally, we don't support appending to files (yet), so 'w' and 'r' alone should be sufficient (ignoring 'x' or 'c for now).

src/GridFS/Bucket.php
```diff
@@ -0,0 +1,210 @@
...
+     *  * writeConcern (MongoDB\Driver\WriteConcern): Write concern.
+     *
+     * @param Manager  $manager      Manager instance from the driver
+     * @param string   $databaseName Database name
+     * @param array    $options      Bucket options
+     * @throws InvalidArgumentException
+     */
+    public function __construct(Manager $manager, $databaseName, array $options = [])
+    {
+        $collectionOptions = array();
```
jmikola: You can use `[]` here.

src/GridFS/Bucket.php
```diff
@@ -0,0 +1,210 @@
...
+    private function isFilesCollectionEmpty()
+    {
+        return null === $this->filesCollection->findOne([], [
+            'readPreference' => new ReadPreference(ReadPreference::RP_PRIMARY),
+            'projection' => ['_id' => 1],
+        ]);
+    }
+    public function delete(ObjectId $id)
+    {
+        $options = array('writeConcern' => $this->writeConcern);
```
jmikola: `[]` syntax here as well.

src/GridFS/Bucket.php
```diff
@@ -0,0 +1,210 @@
...
+    public function openUploadStream($filename, array $options = [])
+    {
+        $this->ensureIndexes();
+        $options = [
+            'filesCollection' => $this->filesCollection,
+            'chunksCollection' => $this->chunksCollection,
+            'chunkSizeBytes' => $this->getChunkSizeBytes()
+            ];
+        $context = stream_context_create(['gridfs' => $options]);
+        return fopen('gridfs://$this->databaseName/$filename', 'w', false, $context);
```
jmikola: `$this->databaseName` and `$filename` aren't going to be resolved within this single-quoted string. If you want to use double-quoted strings, you can surround these with `{}` for clarity (see: [complex expressions](http://php.net/manual/en/language.types.string.php#language.types.string.parsing.complex)); otherwise, I'd opt for `sprintf()`.

src/GridFS/GridFsUpload.php
```diff
@@ -0,0 +1,164 @@
...
+    }
+    /**
+    * Reads data from a stream into GridFS
+    *
+    * @param Stream   $source   Source Stream
+    * @return ObjectId
+    */
+    public function uploadFromStream($source)
+    {
+        while ($data = fread($source, $this->chunkSizeBytes)) {
```
jmikola: To validate that `$source` is actually a stream, we can use [`is_resource()`](http://php.net/is_resource) and [`get_resource_type()`](http://php.net/manual/en/function.get-resource-type.php). After that, [`get_stream_meta_data()`](http://php.net/manual/en/function.stream-get-meta-data.php) may be used to check if the stream is actually readable.

Implementing the sanity check here should also cover the Bucket class' `uploadFromStream()` method.

src/GridFS/GridFsUpload.php
```diff
@@ -0,0 +1,164 @@
...
+    */
+    public function insertChunks($toWrite)
+    {
+        rewind($this->buffer);
+        $cached = fread($this->buffer, $this->chunkSizeBytes);
+        $toWrite = $cached . $toWrite;
+        $bytes = str_split($toWrite, $this->chunkSizeBytes);
+
+        fclose($this->buffer);
+        $this->buffer = fopen("php://temp", "w+");
```
jmikola: When do you close `$this->buffer`? It looks like you read it out into a PHP string, concatenate, and then reassign it to a newly opened in-memory stream.

`str_split()` also seems like it'd have overhead, since it's creating an array of chunk-sized strings, which all copy data already found in `$toWrite`. I think we can get by with the loop appending enough bytes from `$toWrite` to hit a chunk (or exhaust `$toWrite` entirely) into the in-memory buffer and then `insertChunk()` on the buffer contents each time we fill it up. Afterwards, we can truncate the buffer back to zero and continue the append process as needed.

I realized we discussed the in-memory buffer quite a bit yesterday, but it occurs to me reading this code that string concatenation may prove simpler. If you want to start with strings, we can always benchmark the in-memory buffer later (just make a ticket so we remember).

src/GridFS/GridFsUpload.php
```diff
@@ -0,0 +1,164 @@
...
+    /**
+    * Close an active stream, pushes all buffered data to GridFS
+    *
+    */
+    public function close()
+    {
+        rewind($this->buffer);
+        $cached = fread($this->buffer, $this->chunkSizeBytes);
+
+        if(strlen($cached) > 0) {
```
jmikola: If the buffer always points to its end, you may be able to use [`ftell()`](http://php.net/ftell) on it instead of loading it into a string to run `strlen()`. This may be a moot point, though, since we either read nothing or need to access it anyway to insert the chunk.

src/GridFS/Bucket.php
billybanfield: - x

src/GridFS/Bucket.php
billybanfield: - [x]

src/GridFS/Bucket.php
billybanfield: - [x] changed to sprintf()

src/GridFS/GridFsUpload.php
billybanfield: 2 Questions on this:

- 1 What do you mean by 'should also cover the Bucket class' `uploadFromStream()` method? ' Like it will be api compliant ? 
- 2 I'm not totally sure what errors should be being thrown if these checks don't succeed.

billybanfield: Hmm, `is_readable` returns false for php streams like 'php://temp' which makes its use a little more limited in this context.

src/GridFS/GridFsUpload.php
billybanfield: I gave this a revamp. @jmikola please check it out and let me know what you think.

src/GridFS/Bucket.php
```diff
@@ -0,0 +1,198 @@
...
+    }
+    /**
+     * Opens a Stream for reading the contents of a file specified by ID.
+     *
+     * @param ObjectId $id
+     * @return Stream
+     */
+    public function openDownloadStream(ObjectId $id)
+    {
+        fopen('gridfs://$this->databaseName/$id', 'r');
```
jmikola: I assume you just haven't gotten around to it yet, but this likely needs a context passed in for the Manager and/or Bucket instance.

billybanfield: correct.

src/GridFS/GridFsUpload.php
```diff
@@ -0,0 +1,169 @@
...
+    }
+    /**
+    * Insert a chunks into GridFS from a string
+    *
+    * @param string  $toWrite   String to upload
+    * @return int
+    */
+    public function insertChunks($toWrite)
+    {
+        $this->bucket->ensureIndexes();
```
jmikola: Should the stream really be delegating this? It seems like something the bucket should do from its own write methods.

src/GridFS/GridFsUpload.php
```diff
@@ -0,0 +1,169 @@
...
+
+        $readBytes = 0;
+        while($readBytes != strlen($toWrite)) {
+            $addToBuffer = substr($toWrite, $readBytes, $this->bucket->getChunkSizeBytes() - $this->bufferLength);
+            fwrite($this->buffer, $addToBuffer);
```
jmikola: It may be worth saving `$this->bucket->getChunkSizeBytes() - $this->bufferLength` above, as we should assert that `fwrite()` returns the same integer value (strictly compared with `===`).

src/GridFS/GridFsUpload.php
billybanfield: So this is actually part of the reason for my refactor. As soon as fopen() is called within openUploadStream, the bucket is no longer actually really able to see what is going on so the stream needs to sort of go around it. I could change the bucket class to have methods that wrap writes to its underlying db's where it would be called.

src/GridFS/GridFsUpload.php
```diff
@@ -0,0 +1,169 @@
...
+            $readBytes += strlen($addToBuffer);
+            $this->bufferLength += strlen($addToBuffer);
+            if($this->bufferLength == $this->bucket->getChunkSizeBytes()) {
+                rewind($this->buffer);
+                $this->insertChunk(fread($this->buffer, $this->bucket->getChunkSizeBytes()));
```
jmikola: It may be easier to just use [`stream_get_contents()`](http://php.net/stream_get_contents) on the buffer here, as it shouldn't contain more than a chunk's worth of bytes.

src/GridFS/GridFsUpload.php
```diff
@@ -0,0 +1,169 @@
...
+    {
+        $this->bucket->ensureIndexes();
+
+        $readBytes = 0;
+        while($readBytes != strlen($toWrite)) {
+            $addToBuffer = substr($toWrite, $readBytes, $this->bucket->getChunkSizeBytes() - $this->bufferLength);
+            fwrite($this->buffer, $addToBuffer);
+            $readBytes += strlen($addToBuffer);
+            $this->bufferLength += strlen($addToBuffer);
+            if($this->bufferLength == $this->bucket->getChunkSizeBytes()) {
```
jmikola: To be defensive, I think we should either check with `>=` or have some catch to throw if buffer length ever grows past the chunk size.

src/GridFS/GridFsUpload.php
```diff
@@ -0,0 +1,169 @@
...
+    * @param string  $toWrite   String to upload
+    * @return int
+    */
+    public function insertChunks($toWrite)
+    {
+        $this->bucket->ensureIndexes();
+
+        $readBytes = 0;
+        while($readBytes != strlen($toWrite)) {
+            $addToBuffer = substr($toWrite, $readBytes, $this->bucket->getChunkSizeBytes() - $this->bufferLength);
```
jmikola: Since we're accessing the chunk size several times in this method, it'd be reasonable to just initialize `$chunkSizeBytes = $this->bucket->getChunkSizeBytes()` up above so we just need to reference a local variable each time instead of calling a getter method.

src/GridFS/GridFsUpload.php
```diff
@@ -0,0 +1,169 @@
...
+        return $readBytes;
+    }
+    /**
+    * Close an active stream, pushes all buffered data to GridFS
+    *
+    */
+    public function close()
+    {
+        rewind($this->buffer);
+        $cached = fread($this->buffer, $this->bucket->getChunkSizeBytes());
```
jmikola: We can use `stream_get_contents()` here.

src/GridFS/GridFsUpload.php
```diff
@@ -0,0 +1,169 @@
...
+
+        fclose($this->buffer);
+
+        $this->fileCollectionInsert();
+    }
+    private function insertChunk($data)
+    {
+        $toUpload = ["files_id" => $this->file['_id'], "n" => $this->n, "data" => $data];
+        hash_update($this->ctx, $data);
+        $this->bucket->chunkInsert($toUpload);
```
jmikola: Is `Bucket::chunkInsert()` called from anywhere other than GridFsUpload? It seems like that logic could just exist within this class when we move to injecting the Manager and using `executeBulkWrite()` directly.

src/GridFS/StreamWrapper.php
christkv: you can't append to a file safely without heavily modifying the gridstore specification due to race conditions between multiple requests appending to the same file.

src/GridFS/BucketReadWriter.php
```diff
@@ -0,0 +1,125 @@
+<?php
+namespace MongoDB\GridFS;
+
+class BucketReadWriter
```
jmikola: Let's fold this class and its methods into the Bucket class, since that's where the spec expects to find these methods.

src/GridFS/GridFsUpload.php
```diff
@@ -0,0 +1,169 @@
...
+    }
+    /**
+    * Reads data from a stream into GridFS
+    *
+    * @param Stream   $source   Source Stream
+    * @return ObjectId
+    */
+    public function uploadFromStream($source)
+    {
+        $this->bucket->ensureIndexes();
```
jmikola: These `ensureIndexes()` calls can be done from the Bucket's upload methods directly.

src/GridFS/GridFsUpload.php
```diff
@@ -0,0 +1,169 @@
...
+
+        fclose($this->buffer);
+
+        $this->fileCollectionInsert();
+    }
+    private function insertChunk($data)
+    {
+        $toUpload = ["files_id" => $this->file['_id'], "n" => $this->n, "data" => new \MongoDB\BSON\Binary($data, \MongoDB\BSON\Binary::TYPE_GENERIC)];
+        hash_update($this->ctx, $data);
+        $this->bucket->chunkInsert($toUpload);
```
jmikola: `chunkInsert()` and `fileInsert()` appear to have gone missing in the rebase.

src/GridFS/GridFsUpload.php
```diff
@@ -0,0 +1,169 @@
...
+     *
+     * @param \MongoDB\Collection    $filesCollection  Files Collection
+     * @param \MongoDB\Collection    $chunksCollection Chunks Collection
+     * @param int32                  $chunkSizeBytes   Size of chunk
+     * @param string                 $filename         Filename to insert
+     * @param array                  $options          File options
+     * @throws InvalidArgumentException
+     */
+    public function __construct(
+        Bucket $bucket,
```
jmikola: I believe this class only needs the chunk size and it was using Bucket to insert chunk and file documents. We can reduce this to just take the chunk size option, Manager instance, and GridFS prefix (e.g. "fs") directly. This will remove public methods from the Bucket and we won't need to use it as an options container.

src/GridFS/GridFsDownload.php
```diff
@@ -0,0 +1,148 @@
...
+     *  * metadata (array or object): User data for the 'metadata' field of the files
+     *    collection document.
+     *
+     *  * writeConcern (MongoDB\Driver\WriteConcern): Write concern.
+     *
+     * @param array                  $options          File options
+     * @throws FileNotFoundException
+     */
+    public function __construct(
+        Bucket $bucket,
```
jmikola: Similar to https://github.com/mongodb/mongo-php-library/pull/57/files#r48283440, this class only needs the Bucket to do queries on the files and chunks collections. We should replace this argument with the Manager and GridFS prefix. If the read preference option is necessary, then we can inject both Collection instances instead.

src/GridFS/GridFsUpload.php
```diff
@@ -0,0 +1,169 @@
...
+     *    metadata document instead.
+     *
+     *  * metadata (array or object): User data for the 'metadata' field of the files
+     *    collection document.
+     *
+     *  * writeConcern (MongoDB\Driver\WriteConcern): Write concern.
+     *
+     * @param \MongoDB\Collection    $filesCollection  Files Collection
+     * @param \MongoDB\Collection    $chunksCollection Chunks Collection
+     * @param int32                  $chunkSizeBytes   Size of chunk
```
jmikola: PHP just uses `integer`. If we need to validate this, you'll need to do an int32 bounds check manually.

tests/GridFS/SpecificationTests.php
```diff
@@ -0,0 +1,255 @@
...
+                if(isset($this->collections['expected.chunks'])) {
+                    $this->collectionsEqual($this->collections['expected.chunks'],$this->bucket->getChunksCollection());
+                }
+            }
+        }
+    }
+
+    public function provideSpecificationTests()
+    {
+        $testPath=getcwd().'/tests/GridFS/Specification/tests/download_by_name.json';
```
jmikola: It'd be more portable to use the `__DIR__` constant here and give the path relative to this file.

src/GridFS/GridFsUpload.php
billybanfield: Changed to 'int'

src/GridFS/GridFsUpload.php
billybanfield: So after extensive toying around I have found that this still suffers from the main flaw of the fopen call behaving as a wall. In order for the ensure indexes [call to happen at the time the spec insists that it should] (https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst#before-write-operations) an object with those functions **must** be passed through to the stream. It cannot occur in the bucket class because the bucket returns a stream in the `openUploadStream` function and has no understanding of what is happening behind that wall. It seems quite reasonable to me to continue passing a wrapper with the collections around. It doesn't seem to me that it would introduce too much overhead as it is just copying a pointer, so copying a pointer to a wrapper is just as much work as copying a pointer to a collection.

src/GridFS/GridFsUpload.php
billybanfield: It would appear to me, upon re-examining the [documentation](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst#before-write-operations) that it cannot.

tests/GridFS/SpecificationTests.php
billybanfield: done

src/GridFS/GridFsUpload.php
billybanfield: - [x]

src/GridFS/GridFsUpload.php
billybanfield: done

src/GridFS/BucketReadWriter.php
billybanfield: This class still exists, but is no longer called from anywhere. Just keeping it around until we're sure we're done with it.

src/GridFS/Bucket.php
```diff
@@ -0,0 +1,367 @@
...
+
+    /**
+     * Find files from the GridFS bucket's files collection.
+     *
+     * @see Find::__construct() for supported options
+     * @param array|object $filter  Query by which to filter documents
+     * @param array        $options Additional options
+     * @return Cursor
+     */
+    public function find($filter, array $options = [])
```
jmikola: Per https://github.com/mongodb/mongo-php-library/pull/57#commitcomment-15191195, I think we should default `$filter` to an empty array.

prolic:
Can't wait for it ;) Good work!

prolic:
Any benchmarks for performance comparison from this lib to usage with legacy driver grid fs implementation?

jmikola:
@prolic: Not yet. We're certainly aiming to do some before this gets released in 1.1. There is actually a cross-driver benchmark specification that @derickr has started implementing for the old/new drivers, which also includes tests for GridFS operations.

I'm currently preoccupied with fixing some PHPC issues (as is @derickr with getting HHVM 1.1.0 released); however, I should be returning to PHPLIB work once PHPC 1.1.3 is released and this is high on my list to merge down and begin cleaning up and benchmarking.

rexxars:
Anything we can do to progress this PR any further? Library doesn't feel complete without GridFS support...

edast:
do you have preliminary date, when this PR will be merged?

jmikola:
Manually merged in https://github.com/mongodb/mongo-php-library/commit/aef27f095a6c3eb66a154255dcf2edbed91ebe34, which is now the `1.1.x-dev` branch alias. I still intend to take another pass at the merged code and apply some refactoring before a 1.1.0 release (no ETA on that), but it's technically merged if you would like to experiment with it.

PR #59, opened 2015-12-13
Fix distinct maxTimeMS option

tuyakhov:
`$option` variable is not defined there.


jmikola:
Merged in https://github.com/mongodb/mongo-php-library/commit/a4992db2bc67b17d436c29d38e1d887af7538b62. Thanks!

PR #88, opened 2016-01-07
Fixed return class into Collection::aggregate method.

Roquie:


src/Collection.php
```diff
@@ -152,7 +152,7 @@ public function __toString()
      * @see Aggregate::__construct() for supported options
      * @param array $pipeline List of pipeline operations
      * @param array $options  Command options
-     * @return Traversable
+     * @return Cursor
```
jmikola: [`Aggregate::execute()`](https://github.com/mongodb/mongo-php-library/blob/43aa6b572e08daec78a92f8293679f2da4ad39a7/src/Operation/Aggregate.php#L162) conditionally returns a Cursor or ArrayIterator, depending on whether the command is using a cursor or returning inline results, respectively. The existing documentation is correct, as Traversable is the common ancestor of each.

PR #94, opened 2016-01-14
BSONDocument should allow property access by default

jmikola:
This was previously done during unserialization, but not the constructor.


derickr:
LGTM

PR #97, opened 2016-01-18
PHPLIB-162: Support database and collection selection via __get()

jmikola:
https://jira.mongodb.org/browse/PHPLIB-162

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -74,6 +74,23 @@ class Client
     }
 
     /**
+     * Select a database.
+     *
+     * Note: collections whose names contain special characters (e.g. "-") may
+     * be selected with complex syntax (e.g. $client->{"that-database"}) or
+     * {@link selectDatabase()}.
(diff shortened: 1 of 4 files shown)


kamillo1888:
+1

derickr:
LGTM

alcaeus:
Out of curiosity, the legacy driver has a magic `__get` method in `MongoCollection` which allows getting collections with a dot in the name:
```
$database->fs->files
// equivalent to
$database->selectCollection('fs.files')
```

Was this omitted by accident or deliberately?

jmikola:
@alcaeus: Good catch! This was deliberate, as we always found the `->` chaining on MongoCollection a bit weird. It is a bit more natural in the shell, since `.` is the property accessor; however, it's more explicit if users simply do `$database->{'system.profile'}`. I'll add a note about this in the doc block, though.

PR #144, opened 2016-03-21
Documentation updates

jmikola:
See http://jmikola.github.io/mongo-php-library/ for a preview.


docs/classes/collection.md
```diff
@@ -1,38 +1,396 @@
...
+Executes an aggregation framework pipeline on the collection.
+
+This method's return value depends on the MongoDB server version and the
+"useCursor" option. If "useCursor" is true, a MongoDB\Driver\Cursor will be
+returned; otherwise, an ArrayIterator is returned, which wraps the "result"
+array from the command response document.
+
+**Note:** BSON deserialization of inline aggregation results (i.e. not using a
+command cursor) does not yet support a custom type map. Support is pending new
+functionality in the driver.
```
moderndeveloperllc: Does this include documents with `_pclass` attributes? If so that might be worth mentioning. The sentence could be construed to mean only that the `type` option is not supported.

docs/tutorial/crud.md
```diff
@@ -0,0 +1,139 @@
...
+library. A general introduction to CRUD operations in MongoDB may be found in
+the [MongoDB Manual][crud].
+
+[collection]: ../classes/collection.md
+[crud-spec]: https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst
+[crud]: https://docs.mongodb.org/manual/crud/
+
+## Querying
+
+### Finding One Document
```
moderndeveloperllc: Does any mention want to be made of [Persistable](http://php.net/mongodb-bson-persistable) and how `_pclass` is used when an record is unserialized?

docs/classes/collection.md
jmikola: `__pclass` would still apply here. Based on the [persistence specification](http://php.net/manual/en/mongodb.persistence.deserialization.php), it is only ignored if the type map specifies "array" or "object" (i.e. stdClass) or if the class found in the `__pclass` field is not suitable (e.g. not found, not implementing an interface). I'll clarify that in these documentation blurbs.
, found but missing an interface).

jmikola:
@moderndeveloperllc: I added notes on when [Persistable](http://php.net/manual/en/class.mongodb-bson-persistable.php) kicks in and also created a separate BSON tutorial page.

moderndeveloperllc:
@jmikola Thanks! I see the tutorial/bson on the github.io, but not your repo for some reason. Perhaps I'm just not looking in the right spot. It may be useful to show how the data looks when viewed with the mongo shell. I added a created date to the class to show how to use a type like `UTCDateTime`.
```
<?php
class Person implements MongoDB\BSON\Persistable
{
    private $id;
    private $name;
    private $createdDate;

    public function __construct($name)
    {
        $this->id = new MongoDB\BSON\ObjectID();
        $this->name = (string) $name;
        $this->createdDate = new MongoDB\BSON\UTCDateTime(round(microtime(true)*1000));
    }

    public function bsonSerialize()
    {
        return ['_id' => $this->id, 'name' => $this->name, 'created' => $this->createdDate];
    }

    public function bsonUnserialize(array $data)
    {
        $this->id = $data['_id'];
        $this->name = $data['name'];
        $this->createdDate = $data['created'];
    }
    
    public function getDate()
    {
    	return $this->createdDate->toDateTime();
    }
}

$collection = (new MongoDB\Client)->demo->persons;

$result = $collection->insertOne(new Person('Bob'));

printf("Inserted %d person\n", $result->getInsertedCount());

$person = $collection->findOne(['_id' => $result->getInsertedId()]);

var_dump($person);
```
with the data showing in the mongo shell as:
```
> db.persons.findOne()
{
	"_id" : ObjectId("56f9a635f8ddd706292de031"),
	"__pclass" : BinData(128,"UGVyc29u"),
	"name" : "Bob",
	"created" : ISODate("2016-03-28T21:46:29.738Z")
}
```
and the var_dump as:
```
object(Person)[18]
  private 'id' => 
    object(MongoDB\BSON\ObjectID)[15]
      public 'oid' => string '56f9a635f8ddd706292de031' (length=24)
  private 'name' => string 'Bob' (length=3)
  private 'createdDate' => 
    object(MongoDB\BSON\UTCDateTime)[17]
      public 'milliseconds' => int 1459201589738
```

jmikola:
> I see the tutorial/bson on the github.io, but not your repo for some reason.

@moderndeveloperllc: I actually forgot to add the `bson.md` file when committing. Thanks for catching that!

> It may be useful to show how the data looks when viewed with the mongo shell. I added a created date to the class to show how to use a type like UTCDateTime.

Added both of these examples. Thanks! I also left a note in [PHPC-536](https://jira.mongodb.org/browse/PHPC-536) to come back and revise this example once UTCDateTime's constructor is enhanced to default to the current time.

jmikola:
Manually merged to `v1.0` in https://github.com/mongodb/mongo-php-library/commit/749f0aa0d346403a2e13d40ff1d74172aec483b1, which has since been merged up to `master` and deployed.

moderndeveloperllc:
Nice. And thanks for cleaning up my example for readability. I had been wondering about the very thing PHPC-536 addresses.

PR #152, opened 2016-03-31
Fixing namespace issue in GridFS Bucket constructor

### LINKED ISSUE #151: Oversight in GridFS bucket constructor
The `__construct` method for `MongoDB\GridFS\Bucket` is testing the `readPreference` and `writeConcern` options against the wrong class names. The class names on the right-hand side of the `instanceof` operator should be namespaced (either imported with `use`, or using the fully-qualified name).

## Environment
* (Irrelevant, really, but it's…) HHVM 3.13.0 on Ubuntu 14.04

## Exception
`MongoDB\Exception\InvalidArgumentException`
```
Expected "readPreference" option to have type "MongoDB\Driver\ReadPreference" but found "MongoDB\Driver\ReadPreference"
```

I can submit a pull request with the fix momentarily.

doublecompile:
This should fix #151

### DIFF
--- a/src/GridFS/Bucket.php
+++ b/src/GridFS/Bucket.php
@@ -5,6 +5,8 @@ namespace MongoDB\GridFS;
 use MongoDB\BSON\ObjectId;
 use MongoDB\Driver\Cursor;
 use MongoDB\Driver\Manager;
+use MongoDB\Driver\ReadPreference;
+use MongoDB\Driver\WriteConcern;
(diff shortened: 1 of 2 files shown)


jmikola:
Manually merged with some test changes in https://github.com/mongodb/mongo-php-library/commit/bce6aa02e2a204e29a4e3568730074313b327cd2. Thanks!

PR #157, opened 2016-04-15
Documentation fixes

corpsee:
Documentation fixes

### DIFF
--- a/src/Operation/CreateIndexes.php
+++ b/src/Operation/CreateIndexes.php
@@ -99,7 +99,9 @@ class CreateIndexes implements Executable
      * Create one or more indexes for the collection using the createIndexes
      * command.
      *
-     * @return Server $server
+     * @param Server $server
+     *
+     * @return mixed
      */
     private function executeCommand(Server $server)
     {
--- a/src/Operation/CreateIndexes.php
+++ b/src/Operation/CreateIndexes.php
@@ -21,8 +21,19 @@ class CreateIndexes implements Executable
 {
     private static $wireVersionForCommand = 2;
 
+    /**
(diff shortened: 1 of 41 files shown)


lindelius:
@jmikola Useful PHPDoc changes that should be merged as soon as possible

lindelius:
@corpsee Line 102 (after changes) in `src/Operation/CreateIndexes.php` should actually read `@param Server $server`

corpsee:
@lindelius You are right, thx

src/Operation/CreateIndexes.php
```diff
@@ -89,6 +100,8 @@ public function execute(Server $server)
      * command.
      *
      * @param Server $server
+     *
+     * @return mixed
```
jmikola: I realized while reviewing that this function's return value is never used. I've removed it entirely in https://github.com/mongodb/mongo-php-library/pull/188.

jmikola:
Thanks for the contribution. I've selected parts of this PR and merged them to `v1.0` in https://github.com/mongodb/mongo-php-library/commit/bc64127f867f7f98c5201454b5d1a4fec63c50b6.

To explain why I've omitted other parts of the PR:

 * The doc blocks for private member variables aren't used for generating API documentation. Since these members are internal, I'd prefer to just name them sensibly and keep them self-documenting. In the cases of `private static` members, those are effectively internal constants whose value is immediately assigned (I don't see the value of prefixing them with `/** @var array */` or `/** @var int */`).
 * I am currently working through refactoring the GridFS component, so its code is in flux. I will be taking a final pass at the documentation for its public API once that refactoring is complete.

PR #158, opened 2016-04-15
Fixed composer lock

corpsee:
Composer lock file file must be stored in the repository.
See https://blog.engineyard.com/2014/composer-its-all-about-the-lock-file, for example.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 *.swp
 !.gitignore
 composer.phar
-composer.lock
 /vendor/
 php.ini
 phpunit.xml
--- a/composer.json
+++ b/composer.json
@@ -13,6 +13,9 @@
         "php": ">=5.4",
         "ext-mongodb": "^1.1.0"
     },
+    "require-dev": {
+        "phpunit/phpunit": "4.8.*"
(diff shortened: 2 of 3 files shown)


alcaeus:
That holds true for an application. However, for a library composer.lock makes little to no sense as the contents of composer.lock are ignored when the library is included in a project.

corpsee:
Library contributors may have problems because composer.lock missing. In addition, it is best practice.

jmikola:
Thanks for the PR. I've incorporated your addition of PHPUnit as a development dependency to https://github.com/mongodb/mongo-php-library/pull/184, where I've also updated Travis CI to use that binary.

> Library contributors may have problems because composer.lock missing. In addition, it is best practice.

There were no package dependencies before the addition of PHPUnit, so I'm not sure how `composer.lock`'s absence would affect contributors. The library's only dependencies are platform checks  for PHP and the MongoDB extension.

As for best practices, @alcaeus is correct that it is not necessary for libraries; however, it is certainly advised for applications. This was most recently discussed [here](https://seld.be/notes/common-files-in-php-packages).

PR #167, opened 2016-05-03
Fixed error when filename is not specified when reading file

senasi:
When trying to read file with GridFS\Bucket::openDownloadStream() sub-method openDownloadStreamByFile() is directly accessing $file->filename property which is (according to https://docs.mongodb.org/manual/core/gridfs/#gridfs-collections) optional and is not needed for reading files from GridFS. This PR will fix error notice when this property is not set.

### DIFF
--- a/src/GridFS/Bucket.php
+++ b/src/GridFS/Bucket.php
@@ -354,7 +354,7 @@ class Bucket
 
         $context = stream_context_create(['gridfs' => $options]);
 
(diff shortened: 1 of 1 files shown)


jmikola:
Thanks for the PR. This has been fixed with some refactoring in https://github.com/mongodb/mongo-php-library/commit/46864db55bdf7b81e338f190a52145b7f84129b9.

PR #175, opened 2016-05-17
patch

yangchaogit:
patch for write


jmikola:
Thanks for the contribution. Unfortunately, the scripts in `examples/` are obsolete and should have been removed some time ago. I will ensure the write methods are added to our existing documentation (see https://github.com/mongodb/mongo-php-library/pull/185) before removing `examples/`.

PR #180, opened 2016-05-26
[WIP] Add CollectionInterface

chadicus:
Interface would be useful for mocking collection in tests.


jmikola:
I realized this PR is already closed, but for future reference: there are no plans to create interfaces for the core library classes, as we'd like to keep flexibility to add new methods between minor versions. That said, we realize that the extension classes (e.g. Manager, Server, Cursor) should have interfaces and we plan to add those in [PHPC-378](https://jira.mongodb.org/browse/PHPC-378).

PR #182, opened 2016-06-01
return type of getInsertedId fix

thesebas:


jmikola:
Thanks for the contribution. This was one of several issues addressed in #183.

PR #185, opened 2016-06-02
Add write method examples to CRUD tutorial

jmikola:


moderndeveloperllc:
The only suggestion I could make is to note that a `Serializable/Unserializable`, `Persistable`, or other object could be used instead of an array for things like filter statements, etc. There should be no need to rewrite any of the examples, a note at the top would be great though.

jmikola:
@moderndeveloperllc: Thanks for the suggestion. I've added that note at the top of the tutorial.

jmikola:
Merged to `v1.0` in https://github.com/mongodb/mongo-php-library/commit/d9aee1d92353d25e26cea6b1f6d27ec19e4ecfdf.

Issue #209, opened 2016-06-29
Suggestion for improvement of the BulkWrite API 


petr-buchin:
I'm writing a new MongoDB ODM for PHP7, which, like Doctrine, will deal with bulk writes. 

I'd like to build ODM on top of MongoDB PHP Library, but here is what I found:
Driver's BulkWrite class has a beautiful API, which incapsulates functionality of building bulk write in methods like `insert()`, `update()`, `delete()` etc. 
On the other hand, the implementation of BulkWrite in the PHP Lib contains an ugly API, which is built on top of the driver's beautiful API: 
- BulkWrite accepts multidimensional array of operations, which looks awkward compared with the driver's solution with methods;
- PHP LIb API is NOT encapsulates driver's BulkWrite functionality, because if you want to build bulk write using php library, you must use constants from driver's BulkWrite class to specify, what operation should be done (update, insert or delete). As another solution, you must use string literals to specify operation. But this solution is prone to typos and looks even more awkward. So, there is a choice from two bad solutions: using constants from driver class, which functionality must be encapsulated in php lib, or using string literals.

I understand, why bulk write API in  PHP Lib is built this way: that's because Driver's BulkWrite class accepts options in constructor, and to use it's methods like `insert()` or `delete()`, we must specify all of the options first, but you want to have ability to validate/add/change/process some options before passing them to driver's BulkWrite. Also, maybe you want to create driver's BulkWrite lazily - when someone calls `execute()` on lib's BulkWrite (but that's just an object, and it's constructor does not open connection to Mongo server, so I think this is ok to create driver's bulk write in lib's BulkWrite constructor and decorate it's methods - this also gives ability to validate options and operations using driver's BulkWrite object).

Generally, if someone wants to build some kind of query builder on top of the PHP Lib, he must build cool bulk write API on top of ugly PHP Lib API, which is built on top of cool driver's API. From my opinion -  that's not cool. 

**Suggestion:**
I suggest to change `MongoDB\Operation\BulkWrite` constructor to NOT accept $operations argument.
Than I'd like to add methods `insert()`, `update()`, `delete()`, `flush()` and `setOptions()` to make building bulk write easy and cool. Method setOptions is not necessary if the $options argument in constructor will be required - if so, we can create driver's BulkWrite in constructor: than methods like insert() and others should just decorate methods from driver implementation. 

```
$bulk = new BulkWrite($options);
$bulk
    ->insert(['foo' => 'bar'])
[...]

PR #211, opened 2016-07-04
Bulk write API enhancement

petr-buchin:
I made functionality we discussed earlier, but i had no time to change tests.
Now checks of the input is made by the Input classes, and Input class knows nothing about position of the operation in the array. I tried to fix this issue by catching exceptions, thrown from Input classes, in BulkWrite class (lines 125-131) and re-throwing them with the information about position of "failed" operation. Please look at this changes and tell, if its ok, so that I could start to change tests.

### DIFF
--- a/src/Bulk/BulkBuilder.php
+++ b/src/Bulk/BulkBuilder.php
@@ -1,0 +1,82 @@
+<?php
+
+namespace MongoDB\Bulk;
+
+/**
+ * Class BulkBuilder
+ * @package MongoDB\Bulk
+ */
+class BulkBuilder
+{
+    /**
+     * @var BulkInputInterface[]
+     */
+    private $operations = [];
+
+    /**
+     * @param array|object $document
+     * @return $this
+     */
+    public function insertOne($document)
+    {
+        $this->operations[] = new InsertOneInput($document);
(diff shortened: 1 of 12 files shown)


jmikola:
I just revised the title to remove a reference to [PHPLIB-209](https://jira.mongodb.org/browse/PHPLIB-209). You may want to do the same for the commit messages. That syntax is used for JIRA issues, while your PR relates to #209.

I realize this PR has been sitting for a while with no acknowledgement since we last spoke in #209. Apologies for not replying sooner on this. I will need to put off reviewing this until after I wrap up work on `ext-mongodb` 1.2.0 and I expect we'll consider this for version 1.2.0 of the library at the earliest (note: there is no relation between the driver and library versions).

petr-buchin:
I have changed commit messages, branch name and pull request name.
I made the new pull request from the new branch, so I close this request in favor to [#228](https://github.com/mongodb/mongo-php-library/pull/228).

jmikola:
@petr-buchin: Noted. In the future, you should be able to alter the commit messages with `git rebase` and then force-push to your own branch to update the existing PR. The PR title can always be edited as well.

PR #214, opened 2016-07-19
Add hint option to find operation

pb9k:


jmikola:
Thanks for the contribution. For the time being, we'll need to hold off on merging this until [SPEC-337](https://jira.mongodb.org/browse/SPEC-337) and [SPEC-574](https://jira.mongodb.org/browse/SPEC-574) can be resolved. MongoDB's [CRUD specification](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst) needs to be revised to allow additional options on the find method.

jmikola:
Just to update, this will be handled by the fix for [PHPC-783](https://jira.mongodb.org/browse/PHPC-783), as "hint" will now be a first-class option in the libmongoc function we use to execute queries. Please watch that issue for updates.

PR #224, opened 2016-08-07
Add accessor for gridFS buckets in database class

alcaeus:
Similar to `Database::selectCollection`, the new `selectGridFSBucket` method returns a `MongoDB\GridFS\Bucket` instance with the correct options set. Instantiating this class directly isn't feasible since the manager isn't exposed in the `MongoDB\Client` class.

A note about tests: I was going to add functional tests similar to those for `selectCollection` but can't really test anything since neither the `Bucket` class nor the `CollectionWrapper` store their options, instead relying on the `Collection` objects for the files and chunks collection to store these properly. I wasn't sure about exposing those objects in a `__debugInfo` method, so I decided to leave them out and let you guys decide how to best deal with it.

### DIFF
--- a/src/Database.php
+++ b/src/Database.php
@@ -9,6 +9,7 @@ use MongoDB\Driver\ReadConcern;
 use MongoDB\Driver\ReadPreference;
 use MongoDB\Driver\WriteConcern;
 use MongoDB\Exception\InvalidArgumentException;
+use MongoDB\GridFS\Bucket;
 use MongoDB\Model\CollectionInfoIterator;
 use MongoDB\Operation\CreateCollection;
 use MongoDB\Operation\DatabaseCommand;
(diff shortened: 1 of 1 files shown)


src/Database.php
```diff
@@ -263,6 +264,26 @@ public function selectCollection($collectionName, array $options = [])
...
 
     /**
+     * Select a GridFS bucket within this database.
+     *
+     * @see Bucket::__construct() for supported options
+     * @param string $bucketName Name of the bucket to select (defaults to 'fs')
+     * @param array  $options    Bucket constructor options
+     * @return Bucket
+     */
+    public function selectGridFSBucket($bucketName = 'fs', array $options = [])
```
jmikola: @alcaeus: Any objection if I incorporate `$bucketName` into the options array? Off hand, I don't believe we mixed optional arguments (with default values) and options arrays elsewhere in the library and I'd like to keep this consistent.

jmikola:
> A note about tests: I was going to add functional tests similar to those for `selectCollection` but can't really test anything since neither the `Bucket` class nor the `CollectionWrapper` store their options, instead relying on the `Collection` objects for the files and chunks collection to store these properly. I wasn't sure about exposing those objects in a `__debugInfo` method, so I decided to leave them out and let you guys decide how to best deal with it.

For the benefit of testing and debugging, I will be adding the Bucket's manager and inherited options to its debug info. Good call.

src/Database.php
alcaeus: I used this method signature to be consistent with `selectCollection`. Since the bucket name is incorporated into the options array in the `Bucket` class, I think it's better to do it as you suggested.

jmikola:
Manually merged in https://github.com/mongodb/mongo-php-library/commit/cc09f79ed0f02b18b9138ab92a983c8ffe9c8aa4.

I added two more commits to revise the `selectGridFSBucket()` signature, add fields to Bucket debug output, and add some tests.

PR #227, opened 2016-08-09
Ports docs to restructured text and updates the reference

schmalliso:
We'll need to do a fix or two in the mongodb/docs-php-library repo before the docs are buildable locally. I'll provide an update to this pull (or send another depending on how y'all decide to handle this) once we've sorted that out (tomorrow, probably).


schmalliso:
Okiedokes. To build the docs locally, they'll need to clone https://github.com/mongodb/docs-php-library repository (since that's where all the config and docs-tools stuff is), and then following the instructions in that repo's README.rst.

The only requirements are to install giza (which will install a bunch of things) and then they can build! 😺

docs/includes/giza.log
```diff
@@ -0,0 +1,21 @@
+WARNING:giza.content.changelog.tasks:changelog generation is not configured.
```
jmikola: Shouldn't this file be included in `.gitignore`?

docs/includes/giza.log
```diff
@@ -0,0 +1,21 @@
...
+WARNING:giza.content.inheritance.apiargs:too many field types returning one at random.
+WARNING:giza.content.inheritance.apiargs:too many field types returning one at random.
+WARNING:giza.content.inheritance.apiargs:too many field types returning one at random.
+WARNING:giza.content.inheritance.apiargs:too many field types returning one at random.
+WARNING:giza.content.inheritance.apiargs:too many field types returning one at random.
+WARNING:giza.content.inheritance.apiargs:too many field types returning one at random.
+WARNING:giza.content.sphinx:source/index.txt:66: WARNING: toctree contains reference to nonexisting document u'.. /getting-started'
+WARNING:giza.content.sphinx:source/includes/apiargs/MongoDBCollection-method-createIndex-option.rst:81: WARNING: Inline literal start-string without end-string.
+WARNING:giza.content.sphinx:source/includes/apiargs/MongoDBCollection-method-createIndex-option.rst:155: WARNING: Inline literal start-string without end-string.
+WARNING:giza.content.sphinx:source/upgrade.txt:None: WARNING: image file not readable: img/save-flowchart.png
```
jmikola: Is this a path problem?

More generally, are all of these giza errors indicative of errors in the existing RST files? Should we expect this log to be devoid of warnings?

docs/includes/giza.log
schmalliso: It *is* in .gitignore, I think this one was generated before I added it to the ignore -- I thought I'd got them all, but apparently not. The .gitignore for the docs is also in the root of the mongodb/docs-php-library folder, so we should add the following to the ignore for this repo:

```
build/*
.DS_Store
build
*~
*giza.log
```

schmalliso: I'll update the pull to not include this file.

docs/includes/giza.log
schmalliso: This is an old giza.log, so a lot of these have already been fixed. 

That said, some of the warnings are… shall we say 'erroneous'. We have redoing the `apiargs` type as an OKR this quarter, which will get rid of those warnings that aren't actually warnings. This is also an old log, so the legitimate warnings have already been addressed.

docs/includes/giza.log
jmikola: AFAIK, `.DS_Store` is OSX-specific and `*~` is specific to some editors. I'll likely remove those, as they are generally best left to developers' own `.gitignore` files (in their home directories). The other paths look like good additions, though. Thanks!

jmikola:
> The only requirements are to install giza (which will install a bunch of things) and then they can build!

Are you missing a `requirements.txt` file in https://github.com/mongodb/docs-php-library? It seems to be expected based on the instructions in https://pypi.python.org/pypi/giza/ (although the alternative is to just install it manually).

> We'll need to do a fix or two in the mongodb/docs-php-library repo before the docs are buildable locally.

Looking at https://github.com/mongodb/docs-php-library, I assume the later fix will be adding a git submodule to point to this repo? It looks like the documentation sources are currently duplicated in that repository at the moment.

jmikola:
Manually merged in https://github.com/mongodb/mongo-php-library/commit/38e57b6e4c34cb497154053334f75a24ced4138c. In hindsight, the `.gitignore` changes weren't necessary, since docs will be built from the separate repository. Once https://github.com/mongodb/docs-php-library is good to go, I'll update this project's readme with build instructions for would-be contributors.

schmalliso:
@jmikola We'll need to delete the source directory from the docs-php-library, and make sure that the migrations file is working properly -- one of the build steps copies the mongo-php-library directory into docs-php-library (locally) and then works from its source. I think this supports one of the drivers whose source is in multiple git repos (rather than having a whole bunch of submodules), but we could also go the submodule route.

We actually do not need to add a requirements.txt -- all of the docs repos support the newest version of giza, so making sure you install a specific version isn't necessary. I'm not actually sure why that line is included in the pypi source and will ping SamK/AndrewA about it. Thanks!

I'll ping you when the docs repo is firing on all cylinders (and we've tested to make sure that is so). I'll also do a clean install of giza / the docs on a new laptop to make sure the instructions work.

PR #233, opened 2016-08-25
adds a link to the generated API reference to the toc

schmalliso:
This is consistent with the other drivers tables of contents 😺


jmikola:
I am not planning to maintain the generated API docs once the new documentation is live. It was really just a quick solution until the public API was documented in Markdown, so we can do away with it.

schmalliso:
Oh, cool, that works!

PR #243, opened 2016-09-14
Fix namespaces in PhpDoc

villfa:
undefined-namespace-example

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *.swp
+.idea
 !.gitignore
 composer.phar
 composer.lock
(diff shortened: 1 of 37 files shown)


jmikola:
I confirmed with @geeh that this was necessary for PhpStorm.

Manually merged in https://github.com/mongodb/mongo-php-library/commit/5f71c1be61dc76af07ba606ac8aecc682c5dca59 with a revised commit message. Thanks for the contribution!

I omitted the commit that altered `.gitignore`, as IDE-specific files are best added to your personal `.gitignore` file.

GeeH:
To confirm, this is not a PhpStorm thing - annotations should be fully resolvable and in this case should have the leading backslash.

jmikola:
@GeeH: I suppose PHPUnit is one deviation, as its annotations tend to expect FQCNs (I'm thinking of `@expectedException`).

PR #244, opened 2016-09-14
Import CorruptFileException in GridFS\Bucket

villfa:
The method `\MongoDB\GridFS\Bucket::getFileIdForStream()` could be throw `MongoDB\GridFS\Exception\CorruptFileException`, but the exception is not imported.


jmikola:
Thanks!

PR #250, opened 2016-09-26
all the unused "use" have been removed

sangio90:
Found and removed all unused "use" as  the second step of last pull request (from both src and tests folders).
Now there should be no more unused "use".

### DIFF
--- a/src/Database.php
+++ b/src/Database.php
@@ -2,7 +2,6 @@
 
 namespace MongoDB;
 
-use MongoDB\Collection;
 use MongoDB\Driver\Cursor;
 use MongoDB\Driver\Manager;
 use MongoDB\Driver\ReadConcern;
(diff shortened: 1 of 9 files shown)


src/GridFS/Bucket.php
```diff
@@ -11,7 +11,6 @@
 use MongoDB\Exception\InvalidArgumentException;
 use MongoDB\GridFS\Exception\CorruptFileException;
 use MongoDB\GridFS\Exception\FileNotFoundException;
-use MongoDB\Operation\Find;
```
jmikola: Used by `find()` doc block: `@see Find::__construct() for supported options`

src/Database.php
```diff
@@ -2,7 +2,6 @@
 
 namespace MongoDB;
 
-use MongoDB\Collection;
```
jmikola: Used by `__get` and `selectCollection()` doc blocks.

jmikola:
Merged in https://github.com/mongodb/mongo-php-library/commit/942af4edc26daf66d48af0689167322bb68cf110 with the commit amended to include the necessary `use` statements. Thanks!

PR #255, opened 2016-09-28
PHPLIB-186: Implement Manager accessors for core classes

abellion:
Could be useful to get the manager from the Database class to instantiate the Collection class with the same manager. Example : 

```php
<?php

namespace App\Users;

use MongoDB;

class Users extends MongoDB\Collection
{

	//Get the Database class from the dependency injection engine (just for the example)
	public function __construct(MongoDB\Database $mongoDB)
	{
		parent::construct(
			$mongoDB->getManager(),
			$mongoDB->getDatabaseName(),
			"users"
		);
	}

	//Can now decorate the collection with specific methods :)

}
```

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -91,6 +91,16 @@ class Client
     }
 
     /**
+     * Return the manager.
+     *
+     * @return Manager
+     */
+    public function getManager()
+    {
+        return $this->manager;
+    }
+
+    /**
      * Return the connection string (i.e. URI).
      *
      * @return string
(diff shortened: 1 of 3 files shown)


jmikola:
Manually merged in https://github.com/mongodb/mongo-php-library/commit/b4ba7a70c5de511292cf7fff9d70787422804ece.

This was planned for some time in [PHPLIB-186](https://jira.mongodb.org/browse/PHPLIB-186), but I had not gotten around to adding the methods. Thanks for the contribution!

PR #269, opened 2016-10-11
PHPLIB-205: Support providing collation per operation

jmikola:
https://jira.mongodb.org/browse/PHPLIB-205

This bumps the extension requirement to 1.2.0. Composer allows dev versions with the "^" operator, so this will accept 1.2.0alpha3, which introduces a "collation" option for Query and BulkWrite methods.


jmikola:
Note: Test failure on 2.4 is due to https://github.com/mongodb/mongo-c-driver/commit/ad47d5c19b852ac9c4d4e82618d64552ac7600f7, which is fixed in libmongoc 1.5.0-rc1.

PR #276, opened 2016-10-28
[WIP] Update documentation for 1.1.0

jmikola:
Depends on https://github.com/mongodb/docs-php-library/pull/3.

https://jira.mongodb.org/browse/PHPLIB-186
https://jira.mongodb.org/browse/PHPLIB-221


docs/includes/apiargs-MongoDBCollection-common-option.yaml
```diff
@@ -0,0 +1,79 @@
+arg_name: option
+name: bypassDocumentValidation
+type: boolean
+description: |
+   If ``true``, allows the write operation to circumvent document level
+   validation. Defaults to ``false``.
+interface: phpmethod
+operation: ~
```
jmikola: @schmalliso noted that `operation`, while a required key in the Python model, is only used for replacing `{{role}}` in content. I decided to use `operation: ~` (i.e. null) for common option/param definitions and will probably do so for all definitions.

docs/includes/apiargs-MongoDBClient-method-construct-param.yaml
```diff
@@ -2,46 +2,41 @@ arg_name: param
...
-  Defaults to ``mongodb://localhost:27017`` if unspecified.
+  The URI of the standalone, replica set, or sharded cluster to which to
+  connect. Refer to the :manual:`MongoDB connection string reference
+  </reference/connection-string>` for formatting.
+
+  Defaults to ``"mongodb://127.0.0.1:27017"`` if unspecified.
 interface: phpmethod
 operation: MongoDB\\Client::__construct
 optional: true
-position: 1
```
jmikola: I found that `position` was not required by the model. By default, the order is derived from the YAML file's own ordering (starting with 1 and incrementing with each new definition).

docs/includes/apiargs-MongoDBCollection-method-aggregate-option.yaml
```diff
@@ -2,72 +2,81 @@ arg_name: option
...
-position: 2
 ---
 source:
+  file: apiargs-MongoDBCollection-common-option.yaml
   ref: bypassDocumentValidation
-  file: apiargs-common-option.yaml
-operation: MongoDB\\Collection::aggregate
-position: 5
+post: |
+  This only applies when using the :pipeline:`$out` stage.
```
jmikola: For my local builds, `:pipline:` was not resolved as a link and merely applied fixed-width formatting to "$out". Is that intentional?

docs/reference/method/MongoDBCollection-createIndexes.txt
```diff
@@ -9,78 +9,72 @@ MongoDBCollection::createIndexes()
...
-   specifies a custom index name:
-   
-   .. code-block:: none
-                   
-      [ 'key' => [ 'loc' => '2dsphere'], 'name' => 'geo_index' ],
-   
-   You can specify any index options that your MongoDB version
-   supports. MongoDB 3.2 includes the following options:
-                
-   .. include:: /includes/apiargs/MongoDBCollection-method-createIndex-option.rst
```
jmikola: After adding the actual parameter and options includes for this method, which differ from `createIndex()`, I tried moving the general index options into a section below entitled "$indexes parameter". I found that the table didn't render the same as this "Definition" section (with leading indentation). The second paragraph of each option description would increase in font size, which is likely a CSS issue.

Ultimately, I decided not to repeat the index options and instead link to `createIndex()` and the MongoDB manual. I feel this also works better because `writeConcern` is one of the `createIndex()` method options despite the fact that it's not an index option. Maintaining a separate include of just index options _and_ `writeConcern` seemed like a fair bit of overhead. As-is, explaining how `$indexes` works in its own section below makes `createIndexes()` method docs easier to digest, since we only discuss the actual params/options in the "Definition" section.

docs/reference/method/MongoDBClient-listDatabases.txt
```diff
@@ -13,28 +13,27 @@ MongoDB\\Client::listDatabases()
...
 
-   Lists the available databases.
+   Returns information for all databases on the server.
 
    .. code-block:: php
 
       function listDatabases(array $options = []): MongoDB\Model\DatabaseInfoIterator
 
-   :phpmethod:`MongoDB\\Client::listDatabases` has the following parameters:
+   :phpmethod:`MongoDB\\Client::listDatabases()` has the following parameters:
```
jmikola: Is there value in using a `:phpmethod:` reference here instead of saying "This method"? I noticed that it only renders as fixed-width text instead of a link to the current page. Considering that we already repeat the method name twice immediately before this, I'm dubious that this is any better than common text like I'm using below for `$options`.

docs/reference/method/MongoDBGridFSBucket-downloadToStreamByName.txt
```diff
@@ -0,0 +1,40 @@
...
+   :phpmethod:`MongoDB\\GridFS\\Bucket::downloadToStreamByName()` has the
+   following parameters:
+
+   .. include:: /includes/apiargs/MongoDBGridFSBucket-method-downloadToStreamByName-param.rst
+
+   The ``$options`` parameter supports the following options:
+
+   .. include:: /includes/apiargs/MongoDBGridFSBucket-method-downloadToStreamByName-option.rst
+
+.. todo: add examples
```
jmikola: Does `.. todo` actually render any output or is there merely an internal comment that the docs team can use to find areas needing improvement down the line?

docs/includes/apiargs-MongoDBCollection-method-aggregate-option.yaml
i80and: That happens if a link is not resolved. This is bad enough, but Sphinx won't resolve links to a pipeline/method/whatever defined on the same page.

jmikola: Is this just a byproduct of me rendering the PHPLIB docs separately from the main manual (on my local machine), or do you expect this would persist even in the final builds. I might as well just use ``$out`` if this will have no effect in the deployed version.

docs/reference/method/MongoDBGridFSBucket-downloadToStreamByName.txt
i80and: It's just a comment.

RestructuredText is weird: anything starting with two periods `..` is a comment, UNLESS the first word ends with a double-colon `::`, in which case it's a directive.

docs/reference/method/MongoDBClient-listDatabases.txt
i80and: I generally suggest not using role-links to things defined on the current page, since they don't render the way humans would expect.

docs/includes/apiargs-MongoDBCollection-method-aggregate-option.yaml
jmikola: @schmalliso is creating a ref link in the MongoDB manual for this: `agg-out`.

docs/includes/apiargs-MongoDBGridFSBucket-method-find-option.yaml
```diff
@@ -0,0 +1,72 @@
...
+---
+source:
+  file: apiargs-MongoDBCollection-method-find-option.yaml
+  ref: allowPartialResults
+---
+source:
+  file: apiargs-MongoDBCollection-method-find-option.yaml
+  ref: typeMap
+replacement:
+  parent: "bucket"
```
jmikola: This will depend on [PHPLIB-232](https://jira.mongodb.org/browse/PHPLIB-232).

PR #278, opened 2016-10-29
Added a base result class

lindelius:
Added a base result class to move duplicated code away from the various result classes, as well as to allow for type hinting any of the write result classes.

### DIFF
--- a/src/BulkWriteResult.php
+++ b/src/BulkWriteResult.php
@@ -8,11 +8,9 @@ use MongoDB\Exception\BadMethodCallException;
 /**
  * Result class for a bulk write operation.
  */
-class BulkWriteResult
+class BulkWriteResult extends Result
 {
-    private $writeResult;
     private $insertedIds;
-    private $isAcknowledged;
 
     /**
      * Constructor.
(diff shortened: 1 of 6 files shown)


petr-buchin:
I think it would be better to implement base result class like this:
https://github.com/tequila/mongodb-driver-wrapper/blob/master/src/WriteResult.php
This allows to avoid duplicating checks like `if ($this->isAcknowledged)` .

Also, concrete result classes should not extend the base result. Decorating the result class does the trick, example:
https://github.com/tequila/mongodb-php-lib/blob/master/src/Write/Result/DeleteResult.php

lindelius:
My opinion is that constructors don't belong in _traits_. If you're going for a base constructor you should go with an _abstract class_.

petr-buchin:
Constructors don't belong to _interfaces_, but traits are just an automatic copy-and-paste, so they can contain whatever you want. 
Abstract class is a good idea when it does not break [Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle). 
In your case base `Result` class does not provide any serious public interface to typehint method arguments using this class.
To not break Liskov principle in this concrete example - you should not define methods like `getDeletedCount()` in a base class, because, for example, `InsertOneResult` does not need this method (you are not defining such methods in a base class, and that's right).
But having only method `isAcknowledged` in a base `Result` does not avoid code duplicating - `if ($this->isAcknowledged())` check is duplicated in any child class methods. Also, there is no profit in inheritance of the method `isAcknowledged()` - I beleive that in your implementation this method is a good candidate for trait. Bacause of this - in my opinion, `Result` base class does not make sense, as it does not provide any serious functionality or important public interface.

PR #291, opened 2016-11-30
PHPLIB-239: Unpack MongoDB\BSON\Serializable instances before reading properties

jmikola:
https://jira.mongodb.org/browse/PHPLIB-239


src/functions.php
```diff
@@ -70,6 +74,10 @@ function generate_index_name($document)
  */
 function is_first_key_operator($document)
 {
+    if ($document instanceof Serializable) {
+//        $document = $document->bsonSerialize();
```
moderndeveloperllc: Commented out?

jmikola: Fixed! Thanks for catching that.

Issue #311, opened 2016-12-14
PHP Warning:  fclose(): supplied resource is not a valid stream resource


IT-Experte:
PHP 7.0.8
mongo-php-library 1.1.0 (final)
---
PHP Warning:  fclose(): supplied resource is not a valid stream resource in .../src/GridFS/ReadableStream.php on line 78
---

This is my php code. It reads images from a MongoDB collection and display each file.

```php
<?php
// include Composer (mongodb) extension
require '/src/composer/vendor/autoload.php';

class GridFS {

        private $client;

        public function __construct() {
                $this->client = new MongoDB\Client("mongodb://localhost:27017");
        }

        public function Image($id) {

                $db = $this->client->selectDatabase("myfiles");
                $bucket = $db->selectGridFSBucket();

                $stream = $bucket->openDownloadStream($id);
                header("Content-type: image/jpg;");
                echo stream_get_contents($stream);
                exit;
        }

        public function FindFiles($limit) {

                $collection = $this->client->selectCollection("myfiles", "fs.files");
                $filter = [];
                $options = ['limit' => $limit];
                return $collection->find($filter, $options);
        }
}

$id = isset($_GET['id']) ? urldecode($_GET['id']) : null;


$gfs = new GridFS();

if(isset($id)) {
        $gfs->Image($id);
}

?>

<html>
<head><title>ImageFiles</title></head>
<body>

<?php

$files = $gfs->FindFiles(10);

foreach ($files as $file) {
        $id = $file["_id"];
        echo "<h2>ID: $id</h2>\r\n";
        echo "<img src='?id=$id' border='0' alt='' /><br /> \r\n";
}
?>

</body>
</html>
```

My solution is, to hide the warning...

```
public function close()
{
    if (is_resource($this->buffer)) {
        @fclose($this->buffer);
    }
}
```

jmikola:
Can you share a PHP script that reproduces this warning?

Based on the code, it's not clear to me how this warning would be triggered unless `ReadableStream::close()` was invoked multiple times, which itself sounds like a bug in our StreamWrapper implementation.

IT-Experte:
I have inserted my example php code. You need some image files into mongo database "myfiles". I think, Close(); was invoked multiple times (for each image).

jmikola:
[...]

PR #312, opened 2016-12-14
Fix Issue #311 PHP Warning supplied resource is not a valid stream resource

IT-Experte:


jmikola:
This was addressed by https://github.com/mongodb/mongo-php-library/pull/304.

PR #314, opened 2016-12-19
Install documentation aligned to latest version

pierpaolocira:
Just to be sure README.md does report the latest versions (and avoiding users to install an old one).

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ The preferred method of installing this library is with
 [Composer](https://getcomposer.org/) by running the following from your project
 root:
 
-    $ composer require "mongodb/mongodb=^1.0.0"
+    $ composer require "mongodb/mongodb=^1.1.0"
 
 ## Reporting Issues
 


README.md
```diff
@@ -36,7 +36,7 @@ The preferred method of installing this library is with
 [Composer](https://getcomposer.org/) by running the following from your project
 root:
 
-    $ composer require "mongodb/mongodb=^1.0.0"
+    $ composer require "mongodb/mongodb=^1.1.0"
```
jmikola: This example would still install the latest stable release (1.1.0 at the time of this writing), due to the [caret operator](https://getcomposer.org/doc/articles/versions.md#caret). That said, it is odd to have outdated versions referenced in the documentation, so I've removed all references to versions entirely in https://github.com/mongodb/mongo-php-library/issues/315.

Thanks for bringing this to our attention.

pierpaolocira: Yes, you are right: in any case it could be misleading. Now it's better. Thank you

PR #325, opened 2017-01-12
PHPLIB-213: Support seeking and partial file retrieval for GridFS streams

jmikola:
https://jira.mongodb.org/browse/PHPLIB-213

### DIFF
--- a/src/GridFS/ReadableStream.php
+++ b/src/GridFS/ReadableStream.php
@@ -167,6 +167,44 @@ class ReadableStream
     }
 
     /**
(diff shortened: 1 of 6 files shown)


jmikola:
Manually verified with HHVM 3.17 and https://github.com/mongodb/mongo-hhvm-driver/pull/165, which means that the reverted commit poses no problem (see commit message for https://github.com/mongodb/mongo-php-library/commit/b251eb26eceafdf5f8401dbc3a9f538a715cbe62).

Will hold off on merging this until master is our 1.2.0-dev branch.

PR #352, opened 2017-04-05
PHPLIB-263: Documentation examples for shell commands

jmikola:
https://jira.mongodb.org/browse/PHPLIB-263

### DIFF
--- a/tests/DocumentationExamplesTest.php
+++ b/tests/DocumentationExamplesTest.php
@@ -1,0 +1,947 @@
+<?php
+
+namespace MongoDB\Tests;
+
+use MongoDB\Database;
+use MongoDB\Driver\Cursor;
+use MongoDB\Operation\DropCollection;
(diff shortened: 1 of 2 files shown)


tests/DocumentationExamplesTest.php
```diff
@@ -0,0 +1,943 @@
...
+        // End Example 12
+
+        $this->assertCursorCount(3, $cursor);
+
+        // Start Example 13
+        $cursor = $db->inventory->find([
+            'status' => 'A',
+            '$or' => [
+                ['qty' => ['$lt' => 30]],
+                // Alternatively: ['item' => new \MongoDB\BSON\Regex('^p')]
```
jmikola: @kay-kim: Is this kosher? I noticed Perl and Python were using `$regex` instead of their BSON regex class.

kay-kim: If idiomatic for the language, then that's fine by me.

PR #385, opened 2017-07-07
PHPLIB-271: implemented accessors for options on core classes

### LINKED ISSUE #375: Add accessors to MongoDB\Collection options
It looks a lot more comfortable to set options via setter then passing $options array to constructor:
```
$collection = new Collection($manager, $databaseName, $collectionName);
$collection->setWriteConcern(new WriteConcern('0, 0, FALSE'));
$collection->setTypeMap(['root'->self::getDocumentNamespace()]);
```
vs
`$collection = new Collection($manager, $databaseName, $collectionName, ['writeConcern' => new WriteConcern('0, 0, FALSE')]);`

And in some cases it's useful to modify some Collection's options at runtime, for example, typeMap (we are using in our own serialization mechanizm).

petr-buchin:
Closes #375 

Unfortunately, I was not able to build documentation using `giza` due to such an error:

```
Exception occurred:
  File "<my mongo docs lib>/build/docs-tools/sphinxext/intermanual.py", line 149, in _strip_basic_auth
    url_parts = parse.urlsplit(url)
AttributeError: 'Module_six_moves_urllib_parse' object has no attribute 'urlsplit'
```

So, documentation pages may be not ok.

### DIFF
--- a/docs/reference/class/MongoDBClient.txt
+++ b/docs/reference/class/MongoDBClient.txt
@@ -32,6 +32,10 @@ Methods
    /reference/method/MongoDBClient__get
    /reference/method/MongoDBClient-dropDatabase
    /reference/method/MongoDBClient-getManager
+   /reference/method/MongoDBClient-getReadConcern
+   /reference/method/MongoDBClient-getReadPreference
+   /reference/method/MongoDBClient-getTypeMap
+   /reference/method/MongoDBClient-getWriteConcern
    /reference/method/MongoDBClient-listDatabases
    /reference/method/MongoDBClient-selectCollection
    /reference/method/MongoDBClient-selectDatabase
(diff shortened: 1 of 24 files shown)


petr-buchin:
The build has failed, but as I see in build log, this is an issue with build environment.

jmikola:
Thanks for submitting this. I will take a look once we sort out [PHPC-834](https://jira.mongodb.org/browse/PHPC-834) and [PHPLIB-261](https://jira.mongodb.org/browse/PHPLIB-261). The latter issue is especially relevant because classes other than `MongoDB\Client` will no longer track the `MongoDB\Driver\Manager` options by default, so these accessors may be able to return `null` (and probably will in most cases).

petr-buchin:
@jmikola please let me know here in comment when PHPLIB-261 will be done so that I can change this PR respectively.

jmikola:
This PR has been used as a base for https://github.com/mongodb/mongo-php-library/pull/401.

PR #388, opened 2017-07-22
Compatibility with legacy extension

ardabeyazoglu:
For those who will migrate to this driver from legacy mongo extension, I created a compatibility layer implementing some frequently used methods from legacy extension, so that the migration will be easier. For certain cases, you might not event need to change any legacy code after this update.

The library now supports `count`, `limit`, `skip`, `sort` methods directly on cursor, `insert`, `update`, `remove` methods on collection. The cursor created by `find` method can also be iterated to access data rows.

It doesn't affect the original methods and usage.


ardabeyazoglu:
Looks like there is an issue in the build environment related to package manager. It should actually pass all the tests.

jmikola:
Thanks for the contribution, but we have no intention of adding any compatibility methods for the legacy extension into this library's public API.

While this PR may address the issues you faced with migrating, I don't think it's complete enough for general use. The community-developed [mongo-php-adapter](https://github.com/alcaeus/mongo-php-adapter) project already exists as a nearly complete drop-in replacement for the legacy extension's API (we allude to it in the [Upgrade Guide](https://docs.mongodb.com/php-library/master/upgrade/#overview)).

src/Collection.php
```diff
@@ -520,9 +520,10 @@ public function find($filter = [], array $options = [])
             $options['typeMap'] = $this->typeMap;
         }
 
-        $operation = new Find($this->databaseName, $this->collectionName, $filter, $options);
+        //$operation = new Find($this->databaseName, $this->collectionName, $filter, $options);
+        //return $operation->execute($server);
 
-        return $operation->execute($server);
+        return new \MongoDB\Cursor($this, $filter, $options);
```
jmikola: This change conflicts with "It doesn't affect the original methods and usage", as we're changing the return type of the `find()` method. There is also a subtle behavioral change because the actual query's execution will now be deferred until iteration starts on the returned `MongoDB\Cursor` object. This could cause problems with server availability, since server selection for the query happens during the `find()` call (i.e. the code no longer immediately selects a server and executes a query in the same method call).

ardabeyazoglu: That's true. I was unsure to do it in constructor or not, but now i'm gonna execute it in the constructor method. Thanks for the review.

ardabeyazoglu: Also, i didnt mean a general purpose compatbility actually. I just wanted to make some simple additions for ease of use, also compatible with legacy version.

jmikola: > I just wanted to make some simple additions for ease of use, also compatible with legacy version.

Understandable. From the library's standpoint, we need to be very selective about what APIs we expose. Providing only a subset of the API is a slippery slope as additional BC layers might be requested down the line and the legacy driver has a number of poor design decisions that we absolutely do not want to encourage any longer. Having multiple APIs for the same functionality (e.g. [CRUD spec](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst) and legacy API) would also make the already-verbose [documentation](https://docs.mongodb.com/php-library/) harder to digest.

To that end, I'd rather promote [mongo-php-adapter](https://github.com/alcaeus/mongo-php-adapter) as a full compatibility layer.

ardabeyazoglu: I understand, thank you for the review.

PR #390, opened 2017-08-02
fix: get the size of the file

SelviA:
getRawFileDocumentForStream does not return the length of file if the stream is open.

### DIFF
--- a/src/GridFS/Bucket.php
+++ b/src/GridFS/Bucket.php
@@ -551,6 +551,7 @@ class Bucket
         if ( ! isset ($metadata['wrapper_data']) || ! $metadata['wrapper_data'] instanceof StreamWrapper) {
             throw InvalidArgumentException::invalidType('$stream wrapper data', isset($metadata['wrapper_data']) ? $metadata['wrapper_data'] : null, 'MongoDB\Driver\GridFS\StreamWrapper');
         }
+        $metadata['wrapper_data']->stream_close();
 
         return $metadata['wrapper_data']->getFile();
     }


jmikola:
I'm not sure why this is necessary. Are you calling `Bucket::getFileDocumentForStream()` on an active GridFS upload stream before you close it? If so, [`WritableStream::close()`](https://github.com/mongodb/mongo-php-library/blob/1.2.0-alpha1/src/GridFS/WritableStream.php#L127) and [`WritableStream::fileCollectionInsert()`](https://github.com/mongodb/mongo-php-library/blob/1.2.0-alpha1/src/GridFS/WritableStream.php#L221) will not yet have been called. The latter method is where we set the `md5` and `length` fields before inserting the `fs.files` document.

The current length of a writable upload stream is obtained via [`WritableStream::getSize()`](https://github.com/mongodb/mongo-php-library/blob/1.2.0-alpha1/src/GridFS/WritableStream.php#L153). You can access that value through the [`stream_stat()`](https://github.com/mongodb/mongo-php-library/blob/1.2.0-alpha1/src/GridFS/StreamWrapper.php#L179) handler:

```php
$size = fstat($stream)['size'];
```

src/GridFS/Bucket.php
```diff
@@ -551,6 +551,7 @@ private function getRawFileDocumentForStream($stream)
         if ( ! isset ($metadata['wrapper_data']) || ! $metadata['wrapper_data'] instanceof StreamWrapper) {
             throw InvalidArgumentException::invalidType('$stream wrapper data', isset($metadata['wrapper_data']) ? $metadata['wrapper_data'] : null, 'MongoDB\Driver\GridFS\StreamWrapper');
         }
+        $metadata['wrapper_data']->stream_close();
```
jmikola: This appears to be a breaking change as it subtly closes the stream when [`Bucket::getFileDocumentForStream()`](https://github.com/mongodb/mongo-php-library/blob/1.2.0-alpha1/src/GridFS/Bucket.php#L299) is called where currently the stream is left in its current state.

SelviA: You're right, it's probably not a good idea.
I first tried to modify the implementation of `WritableStream::getFile()` to return the `size` property.
```php
    public function getFile()
    {
        $this->file['size'] = $this->getSize();

        return (object) $this->file;
    }
```
but in my case the size is not up to date : 
```php
$stream = $bucket->openUploadStream($filename, $options);
fwrite($stream, $string);
$fileDocument = $bucket->getFileDocumentForStream($stream);
```
@jmikola Do you see another way to recover the up-to-date size ?

jmikola: The method I suggested https://github.com/mongodb/mongo-php-library/pull/390#issuecomment-319688537 should work. Consider:

```php
<?php

require_once 'vendor/autoload.php';

$client = new MongoDB\Client;
$bucket = $client->test->selectGridFSBucket();

$stream = $bucket->openUploadStream('example.txt');
printf("size: %d\n", fstat($stream)['size']);
var_dump($bucket->getFileDocumentForStream($stream));

fwrite($stream, 'foo');
printf("\nsize: %d\n", fstat($stream)['size']);
var_dump($bucket->getFileDocumentForStream($stream));

fclose($stream);
```

This produces the following output:

```
size: 0
object(MongoDB\Model\BSONDocument)#19 (1) {
  ["storage":"ArrayObject":private]=>
  array(4) {
    ["_id"]=>
    object(MongoDB\BSON\ObjectID)#17 (1) {
      ["oid"]=>
      string(24) "5981f438da14d80ef943da32"
    }
    ["chunkSize"]=>
    int(261120)
    ["filename"]=>
    string(11) "example.txt"
    ["uploadDate"]=>
    object(MongoDB\BSON\UTCDateTime)#18 (1) {
      ["milliseconds"]=>
      string(13) "1501688888505"
    }
  }
}

size: 3
object(MongoDB\Model\BSONDocument)#16 (1) {
  ["storage":"ArrayObject":private]=>
  array(4) {
    ["_id"]=>
    object(MongoDB\BSON\ObjectID)#18 (1) {
      ["oid"]=>
      string(24) "5981f438da14d80ef943da32"
    }
    ["chunkSize"]=>
    int(261120)
    ["filename"]=>
    string(11) "example.txt"
    ["uploadDate"]=>
    object(MongoDB\BSON\UTCDateTime)#17 (1) {
      ["milliseconds"]=>
      string(13) "1501688888505"
    }
  }
}
```

Note that you cannot call `fstat()` or `Bucket::getFileDocumentForStream()` after `fclose()`. Once closed, PHP changes the resource type from "stream" to "unknown" we can no longer work with it. If you need to access the length after `fclose()`, be sure to save the stream's ID (i.e. `_id` from the `fs.files` document) with `Bucket::getFileIdForStream($stream)`. Thereafter, you can either fetch the `fs.files` document directly with `Bucket::findOne(['_id' => $id])` or use `Bucket::openDownloadStream($id)` to obtain a new readable stream resource, which you can then use with `fstat()` or `Bucket::getFileDocumentForStream()`.

SelviA:
Thank you for your help. Indeed "fstat" does the job. But I wanted to map the result of getFileDocumentForStream to an entity that implements \MongoDB\BSON\Unserializable without having any other manipulation to do.
But I understand that retrieving the size of a writable file is not necessarily logical...

PR #401, opened 2017-09-13
PHPLIB-271: Implement accessors for options on core classes

jmikola:
https://jira.mongodb.org/browse/PHPLIB-271

This incorporates @petr-buchin's work in https://github.com/mongodb/mongo-php-library/pull/385.

@petr-buchin: Ultimately, it wasn't necessary to support `null` return values from the read and write concern accessors. In https://github.com/mongodb/mongo-php-library/pull/399, I handled detection of default RC and WC entirely in the operation classes, which means the core classes can freely carry around the default objects from the Manager. Thanks very much for the initial work on this (especially the docs!).

### DIFF
--- a/docs/reference/bson.txt
+++ b/docs/reference/bson.txt
@@ -74,6 +74,19 @@ default:
        'root' => 'MongoDB\Model\BSONDocument',
    ]
 
+The type map above will convert BSON documents and arrays to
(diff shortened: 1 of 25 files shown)


JoffreyPoreeCoding:
Thanks!

PR #413, opened 2017-10-02
Complete the list of required PHP extensions in composer.json

villfa:
It may help users to detect problems during the library install.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,6 +11,8 @@
     ],
     "require": {
         "php": ">=5.4",
+        "ext-hash": "*",
+        "ext-json": "*",
         "ext-mongodb": "^1.3.0"
     },
     "require-dev": {


composer.json
```diff
@@ -11,6 +11,8 @@
     ],
     "require": {
         "php": ">=5.4",
+        "ext-hash": "*",
+        "ext-json": "*",
```
jmikola: Did you experience an error attempting to use the library in a PHP environment where these modules were not loaded? I'm on the fence about how useful this is, as they are both listed as "bundled and compiled into PHP by default" in the PHP.net documentation.

If we do add this, I'd also be inclined to use actual version checks instead of "`*`". Based on https://github.com/php/php-src/commit/18180bb16174a2a966641ed4e6171f6753f160a8#diff-17d725bdd470d96051461b6bf27dc116, the `ext-json` version does appear to be incremented in a [semver](http://semver.org/)-like fashion as its API changes with each PHP version. Unfortunately, that isn't the case for `ext-hash`, which reports "1.0" in PHP versions going back to 5.4 (oldest I tested). That is despite the introduction of new methods in later PHP versions (e.g. [`hash_equals()`](http://php.net/manual/en/function.hash-equals.php)) that should entail a minor version bump.

villfa: I did not experience this kind of error but I thought it was a good addition after reading this [blog post](https://blog.coderspirit.xyz/blog/2017/08/18/our-composer-json-is-wrong/).

Concerning the version number, I quote the [Composer documentation](https://getcomposer.org/doc/01-basic-usage.md#platform-packages):
> `ext-<name>` allows you to require PHP extensions (includes core extensions). Versioning can be quite inconsistent here, so it's often a good idea to just set the constraint to `*`

jmikola:
Merged in https://github.com/mongodb/mongo-php-library/commit/206ae32b35c27b46223955f8dd1a08e7740d61d6. Thanks!

PR #415, opened 2017-10-17
PHPLIB-268 Add option maxAwaitTimeMS on getMore commands

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-268

### DIFF
--- a/docs/includes/apiargs-common-option.yaml
+++ b/docs/includes/apiargs-common-option.yaml
@@ -10,6 +10,17 @@ operation: ~
 optional: true
 ---
 arg_name: option
+name: maxAwaitTimeMS
+type: integer
+description: |
+    Positive integer denoting the time limit in milliseconds for the server to block
+    a getMore operation if no data is available. This option should only be used if
+    cursorType is TAILABLE_AWAIT.
+interface: phpmethod
+operation: ~
+optional: true
+---
+arg_name: option
 name: readConcern
 type: :php:`MongoDB\\Driver\\ReadConcern <class.mongodb-driver-readconcern>`
 description: |
--- a/src/Operation/Find.php
+++ b/src/Operation/Find.php
@@ -78,6 +78,9 @@ class Find implements Executable
      *
      *  * max (document): The exclusive upper bound for a specific index.
      *
+     *  * maxAwaitTimeMS (integer): The maxium amount of time for the server to wait
+     *    on new documents to satisfy a query, if cursorType is TAILABLE_AWAIT.
+     *
      *  * maxScan (integer): Maximum number of documents or index keys to scan
      *    when executing the query.
      *
@@ -177,6 +180,10 @@ class Find implements Executable
 
         if (isset($options['max']) && ! is_array($options['max']) && ! is_object($options['max'])) {
             throw InvalidArgumentException::invalidType('"max" option', $options['max'], 'array or object');
+        }
+
+        if (isset($options['maxAwaitTimeMS']) && ! is_integer($options['maxAwaitTimeMS'])) {
+            throw InvalidArgumentException::invalidType('"maxAwaitTimeMS" option', $options['maxAwaitTimeMS'], 'integer');
         }
 
         if (isset($options['maxScan']) && ! is_integer($options['maxScan'])) {
@@ -298,7 +305,7 @@ class Find implements Executable
             }
         }
 
-        foreach (['allowPartialResults', 'batchSize', 'comment', 'hint', 'limit', 'maxScan', 'maxTimeMS', 'noCursorTimeout', 'oplogReplay', 'projection', 'readConcern', 'returnKey', 'showRecordId', 'skip', 'snapshot', 'sort'] as $option) {
+        foreach (['allowPartialResults', 'batchSize', 'comment', 'hint', 'limit', 'maxAwaitTimeMS', 'maxScan', 'maxTimeMS', 'noCursorTimeout', 'oplogReplay', 'projection', 'readConcern', 'returnKey', 'showRecordId', 'skip', 'snapshot', 'sort'] as $option) {
             if (isset($this->options[$option])) {
                 $options[$option] = $this->options[$option];
             }
--- a/tests/Operation/FindFunctionalTest.php
+++ b/tests/Operation/FindFunctionalTest.php
@@ -3,7 +3,9 @@
 namespace MongoDB\Tests\Operation;
 
 use MongoDB\Driver\BulkWrite;
+use MongoDB\Operation\CreateCollection;
 use MongoDB\Operation\CreateIndexes;
+use MongoDB\Operation\DropCollection;
 use MongoDB\Operation\Find;
 use MongoDB\Tests\CommandObserver;
 use stdClass;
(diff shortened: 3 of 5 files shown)


docs/includes/apiargs-common-option.yaml
```diff
@@ -10,6 +10,13 @@ operation: ~
 optional: true
 ---
 arg_name: option
+name: maxAwaitTimeMS
+type: integer
+description: |
+    Positive integer denoting the time limit in milliseconds for the server to block
+    a getMore operation if no data is available. This option should only be used in
+    conjunction with the "tailable" and "awaitData" options.
+arg_name: option
```
jmikola: Missing a `---` between the args.

src/Operation/FindOne.php
```diff
@@ -54,6 +54,9 @@ class FindOne implements Executable
      *
      *  * max (document): The exclusive upper bound for a specific index.
      *
+     *  * maxAwaitTimeMS (integer): The maxium amount of time for the server to wait
+     *    on new documents to satisfy a query, if “tailable” and “awaitData” are both true.
+     *
```
jmikola: Very good practice to update the dependent operation based on changes in the parent operation class.

In this case, `maxAwaitTimeMS` doesn't serve a purpose without the `tailable` and `awaitData` options (controlled by Find's `cursorType` enum), so we don't need to mention it here.

src/Operation/Find.php
```diff
@@ -78,6 +78,9 @@ class Find implements Executable
      *
      *  * max (document): The exclusive upper bound for a specific index.
      *
+     *  * maxAwaitTimeMS (integer): The maxium amount of time for the server to wait
+     *    on new documents to satisfy a query, if “tailable” and “awaitData” are both true.
```
jmikola: Although the extension does support `tailable` and `awaitData` options on [`MongoDB\Driver\Query`](http://php.net/manual/en/mongodb-driver-query.construct.php), PHPLIB implements the [CRUD spec](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst) which exposes those through the `cursorType` enum. In this case, we can say "if cursorType is TAILABLE_AWAIT".

Side note: try to keep curly quotes out of source files if possible. I assume these only snuck in by copying text from the PHP.net docs.

src/Operation/Find.php
```diff
@@ -179,6 +182,10 @@ public function __construct($databaseName, $collectionName, $filter, array $opti
             throw InvalidArgumentException::invalidType('"max" option', $options['max'], 'array or object');
         }
 
+        if (isset($options['maxAwaitTimeMS']) && ! is_integer($options['maxAwaitTimeMS'])) {
+            throw InvalidArgumentException::invalidType('"maxAwaitTimeMS" option', $options['maxAwaitTimeMS'], 'integer');
+        }
```
jmikola: Since this option has no effect unless `cursorType` is `TAILABLE_AWAIT`, we could consider adding one more check towards the end of the constructor. I usually do such multi-option checks for logical errors after all single option type checks (e.g. [FindAndModify](https://github.com/mongodb/mongo-php-library/blob/d10613869a820b1436ec27f3b99a61defe8f8886/src/Operation/FindAndModify.php#L156)).

That said, neither PHPC nor libmongoc enforce that. `maxAwaitTimeMS` will simply be ignored for cursors that aren't tailable and awaitData, so we're fine without it.

docs/includes/apiargs-common-option.yaml
```diff
@@ -10,6 +10,13 @@ operation: ~
 optional: true
 ---
 arg_name: option
+name: maxAwaitTimeMS
+type: integer
+description: |
+    Positive integer denoting the time limit in milliseconds for the server to block
+    a getMore operation if no data is available. This option should only be used in
+    conjunction with the "tailable" and "awaitData" options.
```
jmikola: This can also be changed to "This option should only be used if ``cursorType`` is ``TAILABLE_AWAIT``".

jmikola: Also, if we're no longer sharing the option definition outside of `apiargs-MongoDBCollection-method-find-option.yaml`, we can just put this that file.

tests/Operation/FindFunctionalTest.php
```diff
@@ -123,6 +125,45 @@ public function provideTypeMapOptionsAndExpectedDocuments()
...
+        $operation = new Find($databaseName, $cappedCollectionName, [], ['cursorType' => 3, 'maxAwaitTimeMS' => $maxAwaitTimeMS]);
+        $cursor = $operation->execute($this->getPrimaryServer());
+        $it = new \IteratorIterator($cursor);
+
+        // Make sure we await results for no more than the maxAwaitTimeMS.
+        $it->rewind();
+        $it->next();
+        $startTime = microtime(true);
+        $it->next();
+        $this->assertLessThan($maxAwaitTimeMS, (microtime(true) - $startTime));
```
jmikola: It's likely that this assertion will always succeed. In the float returned by [`microtime(true)`](http://php.net/manual/en/function.microtime.php), the digits to the right of the decimal point indicate fractions of a second with microsecond precision.

If you're subtracting `$startTime` from the end time here, we should check that the difference is _at least_ ten milliseconds (i.e. `0.01`) to assert that `IteratorIterator::next()` blocked for `maxAwaitTimeMS` plus any additional network latency.

We can either change this to `$this->assertGreaterThanOrEqual($maxAwaitTimeMS * 0.001, ...)` or you can you the literals `10` and `0.01` in the Find options and this assertion, respectively.

tests/Operation/FindFunctionalTest.php
```diff
@@ -123,6 +125,45 @@ public function provideTypeMapOptionsAndExpectedDocuments()
...
+        $bulkWrite = new BulkWrite(['ordered' => true]);
+        $bulkWrite->insert(['_id' => 1]);
+        $result = $this->manager->executeBulkWrite($cappedNamespace, $bulkWrite);
+
+        // Get a cursor of type TAILABLE_AWAIT.
+        $operation = new Find($databaseName, $cappedCollectionName, [], ['cursorType' => 3, 'maxAwaitTimeMS' => $maxAwaitTimeMS]);
+        $cursor = $operation->execute($this->getPrimaryServer());
+        $it = new \IteratorIterator($cursor);
+
+        // Make sure we await results for no more than the maxAwaitTimeMS.
```
jmikola: Per my note below, we should actually assert that we waited _at least_ maxAwaitTimeMS. For instance, if the server only blocks the `getMore` for one millisecond, that means it's not respecting the option.

tests/Operation/FindFunctionalTest.php
```diff
@@ -123,6 +125,45 @@ public function provideTypeMapOptionsAndExpectedDocuments()
...
+        $it->rewind();
+        $it->next();
+        $startTime = microtime(true);
+        $it->next();
+        $this->assertLessThan($maxAwaitTimeMS, (microtime(true) - $startTime));
+
+        // Drop the capped collection.
+        $operation = new DropCollection($databaseName, $cappedCollectionName);
+        $operation->execute($this->getPrimaryServer());
```
jmikola: I noticed that on a test failure, attempting to re-run this test will immediately encounter an exception attempting to create the capped collection ("a collection 'phplib_test.FindFunctionalTest.f36e3691.capped' already exists").

If we use the default collection name from `$this->getCollectionName()`, the `setUp()` method from `FunctionalTestCase` will take care of dropping that collection for us before the test runs (ensuring a blank slate). Additionally, `tearDown()` will also drop it after a test completes. If the test fails, `tearDown()` leaves it in place, which is helpful in case you'd like to inspect its state manually (e.g. poking around with the mongo shell).

I realize that the MapReduce test cases I referenced that use a `$this->getCollectionName() . '.output'` collection only drop it at the end (assuming earlier assertions all succeed and the test makes it up to that point). The reason we don't have to drop the ".output" collection for those tests is that [`mapReduce`](https://docs.mongodb.com/manual/reference/command/mapReduce/) happily overwrites it (unlike the [`create`](https://docs.mongodb.com/manual/reference/command/create/) command).

tests/Operation/FindFunctionalTest.php
```diff
@@ -123,6 +125,45 @@ public function provideTypeMapOptionsAndExpectedDocuments()
...
+        ];
+
+        $operation = new CreateCollection($databaseName, $cappedCollectionName, $cappedCollectionOptions);
+        $operation->execute($this->getPrimaryServer());
+
+        // Insert documents into the capped collection.
+        $cappedNamespace = $databaseName . '.' . $cappedCollectionName;
+        $bulkWrite = new BulkWrite(['ordered' => true]);
+        $bulkWrite->insert(['_id' => 1]);
+        $result = $this->manager->executeBulkWrite($cappedNamespace, $bulkWrite);
```
jmikola: If you end up using the default collection name (without a ".capped" suffix), we can also replace this local variable with `$this->getNamespace()`.

tests/Operation/FindFunctionalTest.php
```diff
@@ -123,6 +125,45 @@ public function provideTypeMapOptionsAndExpectedDocuments()
...
+
+        // Get a cursor of type TAILABLE_AWAIT.
+        $operation = new Find($databaseName, $cappedCollectionName, [], ['cursorType' => 3, 'maxAwaitTimeMS' => $maxAwaitTimeMS]);
```
jmikola: `3` can be replaced with `Find::TAILABLE_AWAIT`. Not that I expect the constants to change, but this avoids a [magic number](https://stackoverflow.com/q/47882/162228) and sets a good example for any users that might be reading through our tests for an example.

tests/Operation/FindFunctionalTest.php
```diff
@@ -123,6 +125,43 @@ public function provideTypeMapOptionsAndExpectedDocuments()
...
+        // Make sure we await results for no more than the maxAwaitTimeMS.
+        $it->rewind();
+        $it->next();
+        $startTime = microtime(true);
+        $it->next();
+        $this->assertGreaterThanOrEqual($maxAwaitTimeMS * 0.001, microtime(true) - $startTime);
+
+        // Drop the capped collection.
+        $operation = new DropCollection($databaseName, $cappedCollectionName);
+        $operation->execute($this->getPrimaryServer());
```
jmikola: `tearDown()` will drop the default collection name for us, so we can remove these three lines. This might be leftover from the previous PR revision.

PR #416, opened 2017-10-18
PHPLIB-225 Ensure that users can distinguish duplicate key write errors

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-225

I only made changes to InsertOne because the Jira ticket said to consult with other drivers before handling the multi-statement case.

### DIFF
--- a/src/Exception/DuplicateKeyException.php
+++ b/src/Exception/DuplicateKeyException.php
@@ -1,0 +1,32 @@
+<?php
+/*
+ * Copyright 2015-2017 MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace MongoDB\Exception;
+
+class DuplicateKeyException extends \MongoDB\Driver\Exception\BulkWriteException implements Exception
+{
+    /**
+     * Thrown when inserting a document with a key that is already present in the database.
+     *
+     * @param string $key Key of inserted document
+     * @return self
+     */
+    public static function duplicateKeyInsertion($key)
+    {
+        return new static(sprintf('The key (%s) of this document is already present in the database.', $key));
+    }
+}
--- a/src/Operation/InsertOne.php
+++ b/src/Operation/InsertOne.php
@@ -21,7 +21,7 @@ use MongoDB\InsertOneResult;
 use MongoDB\Driver\BulkWrite as Bulk;
 use MongoDB\Driver\Server;
 use MongoDB\Driver\WriteConcern;
-use MongoDB\Driver\Exception\BulkWriteException as BulkWriteException;
+use MongoDB\Driver\Exception\BulkWriteException;
 use MongoDB\Exception\DuplicateKeyException;
 use MongoDB\Driver\Exception\RuntimeException as DriverRuntimeException;
 use MongoDB\Exception\InvalidArgumentException;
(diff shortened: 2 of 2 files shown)


src/Operation/InsertOne.php
```diff
@@ -21,6 +21,8 @@
 use MongoDB\Driver\BulkWrite as Bulk;
 use MongoDB\Driver\Server;
 use MongoDB\Driver\WriteConcern;
+use MongoDB\Driver\Exception\BulkWriteException as BulkWriteException;
```
derickr: There is no need to use "as BulkWriteException" here, as the aliased named is the same as the original name - just like you have done for "MongoDB\Exception\DuplicateKeyException;" the line below.

kvwalker: Thank you!

src/Exception/DuplicateKeyException.php
```diff
@@ -0,0 +1,32 @@
+<?php
+/*
+ * Copyright 2015-2017 MongoDB, Inc.
```
jmikola: New files can just specify the current year (see: `MapReduce.php`).

src/Exception/DuplicateKeyException.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * Thrown when inserting a document with a key that is already present in the database.
+     *
+     * @param string $key Key of inserted document
+     * @return self
+     */
+    public static function duplicateKeyInsertion($key)
+    {
+        return new static(sprintf('The key (%s) of this document is already present in the database.', $key));
```
jmikola: Reading this message, it may be ambiguous if the string in parentheses is the field name or duplicate value. Looking at how this is called from `InsertOne.php`, the inserted ID is passed in as `$key`. That may be troublesome for a couple of reasons:

 * While `_id` is certainly a common case of duplicate key errors (especially with non-ObjectID values), it's possible that the exception arose from an index on another field
 * The inserted ID may not be easily represented as as string (e.g. embedded document)

This also potentially hides the server's original error message, which reports the collection name, field name, and offending value (whether it was a document, string, etc.).

src/Operation/InsertOne.php
```diff
@@ -21,6 +21,8 @@
 use MongoDB\Driver\BulkWrite as Bulk;
 use MongoDB\Driver\Server;
 use MongoDB\Driver\WriteConcern;
+use MongoDB\Driver\Exception\BulkWriteException;
+use MongoDB\Exception\DuplicateKeyException;
```
jmikola: This can be moved down a line.

src/Operation/InsertOne.php
```diff
@@ -103,7 +105,14 @@ public function execute(Server $server)
         $insertedId = $bulk->insert($this->document);
 
         $writeConcern = isset($this->options['writeConcern']) ? $this->options['writeConcern'] : null;
-        $writeResult = $server->executeBulkWrite($this->databaseName . '.' . $this->collectionName, $bulk, $writeConcern);
+        try {
+            $writeResult = $server->executeBulkWrite($this->databaseName . '.' . $this->collectionName, $bulk, $writeConcern);
+            return new InsertOneResult($writeResult, $insertedId);
+       } catch (BulkWriteException $e) {
+            if ($e->getCode() == "E11000") {
```
jmikola: "E11000" appears in the error message, but the exception code is an integer.

It looks like there are several codes to be handled (see: [`_is_duplicate_key_error()`](https://github.com/mongodb/mongo-c-driver/blob/1.8.1/src/mongoc/mongoc-write-command.c#L62) in libmongoc). We'll soon be able to drop support for MongoDB 2.6, but that won't be until PHPLIB 1.3.0 (for the MongoDB 3.6-compatible release). Even so, the odd mongos code for 16460 will need to be handled (from [SERVER-11493](jira.mongodb.org/browse/SERVER-11493)). While libmongoc only checks the integer code, the server has a more robust check that also checks the error message (see: [this code search](https://github.com/mongodb/mongo/search?utf8=%E2%9C%93&q=16460&type=)).

Additionally, it looks like PHPC tends to leave the error code unset as zero when throwing a BulkWriteException in [`phongo_execute_write()`](https://github.com/mongodb/mongo-php-driver/blob/1.3.1/php_phongo.c#L503). This means we'll actually need to dig into the WriteResult object within BulkWriteException to inspect it for errors.

PR #418, opened 2017-10-19
[RFC] PHPLIB-225: DuplicateKeyException wraps BulkWriteException

jmikola:
https://jira.mongodb.org/browse/PHPLIB-225

Not sure if we want to merge it, but I wanted to open it for discussion purposes to demonstrate the complexity of actually implementing this in PHPLIB.

### DIFF
--- a/src/Exception/DuplicateKeyException.php
+++ b/src/Exception/DuplicateKeyException.php
@@ -1,0 +1,37 @@
+<?php
+/*
+ * Copyright 2017 MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace MongoDB\Exception;
+
+use MongoDB\Driver\Exception\BulkWriteException;
+
+class DuplicateKeyException extends BulkWriteException implements Exception
+{
+    /**
+     * Wraps a BulkWriteException.
+     *
+     * @param BulkWriteException $e
+     * @return self
+     */
+    public static function fromBulkWriteException(BulkWriteException $e)
+    {
+        $duplicateKeyException = new static($e->getMessage(), $e->getCode(), $e);
+        $duplicateKeyException->writeResult = $e->getWriteResult();
+
+        return $duplicateKeyException;
+    }
+}
--- a/src/Operation/InsertOne.php
+++ b/src/Operation/InsertOne.php
@@ -22,7 +22,9 @@ use MongoDB\Driver\BulkWrite as Bulk;
 use MongoDB\Driver\Server;
 use MongoDB\Driver\Session;
 use MongoDB\Driver\WriteConcern;
+use MongoDB\Driver\Exception\BulkWriteException as BulkWriteException;
 use MongoDB\Driver\Exception\RuntimeException as DriverRuntimeException;
+use MongoDB\Exception\DuplicateKeyException;
 use MongoDB\Exception\InvalidArgumentException;
 
 /**
@@ -111,7 +113,17 @@ class InsertOne implements Executable
         $bulk = new Bulk($options);
         $insertedId = $bulk->insert($this->document);
 
-        $writeResult = $server->executeBulkWrite($this->databaseName . '.' . $this->collectionName, $bulk, $this->createOptions());
+        $writeConcern = isset($this->options['writeConcern']) ? $this->options['writeConcern'] : null;
+
+        try {
+            $writeResult = $server->executeBulkWrite($this->databaseName . '.' . $this->collectionName, $bulk, $this->createOptions());
+        } catch (BulkWriteException $e) {
+            if ($this->isBulkWriteExceptionDuplicateKeyError($e)) {
+                throw DuplicateKeyException::fromBulkWriteException($e);
+            }
+
+            throw $e;
+        }
 
         return new InsertOneResult($writeResult, $insertedId);
     }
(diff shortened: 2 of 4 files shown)


src/Exception/DuplicateKeyException.php
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * Wraps a BulkWriteException.
+     *
+     * @param BulkWriteException $e
+     * @return self
+     */
+    public static function fromBulkWriteException(BulkWriteException $e)
+    {
+        $duplicateKeyException = new static($e->getMessage(), $e->getCode(), $e);
+        $duplicateKeyException->writeResult = $e->getWriteResult();
```
jmikola: Unfortunately, PHP won't let us cast exceptions, so we have to resort to chaining (passing the previous exception in our new exception's constructor). The message, code, and WriteResult must be manually copied. [This thread on Stack Overflow](https://stackoverflow.com/a/5551828/162228) was helpful to understand some relevant best practices.

By sheer coincidence, PHPC tracks the WriteResult object as a protected property (as opposed to the usual way of zval pointers without our structs).

src/Operation/InsertOne.php
```diff
@@ -103,8 +105,52 @@ public function execute(Server $server)
...
 
         return new InsertOneResult($writeResult, $insertedId);
     }
+
+    /**
+     * Determine if a BulkWriteException is a DuplicateKeyException.
+     *
+     * We only wrap a BulkWriteException if it identifies as a duplicate key
+     * error or contains exactly one write error that identifies as a duplicate
+     * key error. It must not contain a write concern error.
```
jmikola: I think this makes sense, but it's not based on prior art that I know of.

src/functions.php
```diff
@@ -79,6 +80,38 @@ function generate_index_name($document)
...
 
 /**
+ * Return whether the error code and message indicate a duplicate key error.
+ *
+ * @internal
+ * @param integer $code
+ * @param string  $message
+ * @return boolean
+ */
+function is_duplicate_key_error($code, $message)
```
jmikola: Based on [`_is_duplicate_key_error()`](https://github.com/mongodb/mongo-c-driver/blob/1.8.1/src/mongoc/mongoc-write-command.c#L62) from libmongoc.

src/Operation/InsertOne.php
jmikola: /cc @alcaeus @ajdavis

jmikola: Interesting to [compare with mongo-php-adapter](https://github.com/alcaeus/mongo-php-adapter/blob/1.1.3/lib/Alcaeus/MongoDbAdapter/ExceptionConverter.php#L44), which simply checks the first write error's code (if any), without caring about additional write errors or a possible write concern error.

alcaeus: IIRC, the logic to only check the first write result was written to cover the case where a single insert would cause a duplicate key exception. I'd have to take a look at the legacy driver to see its behavior during bulk writes to see its behavior when multiple (or any other than the first) inserts result in a duplicate key error.

jmikola: > it may indeed be safer to make sure we only get one write result so we don't hide errors from other bulk write operations.

Indeed, my intention with this logic was that we should only report a DuplicateKeyException if we can assert that the BulkWriteException contains nothing more than a single duplicate key error. Anything else (e.g. multiple write errors, a write concern error) is ambiguous and is best left to the user to inspect on their own.

jmikola: > I'd have to take a look at the legacy driver to see its behavior during bulk writes to see its behavior when multiple (or any other than the first) inserts result in a duplicate key error.

It would appear that `MongoWriteBatch::execute()` reports all write and write concern errors as a MongoWriteConcernException (see: [here](https://github.com/mongodb/mongo-php-driver-legacy/blob/1.6.16/batch/write.c#L427)). This jives with the [PHP.net documentation](http://php.net/manual/en/mongowritebatch.execute.php).

MongoDuplicateKeyException is only used in [`php_mongo_cursor_throw()`](https://github.com/mongodb/mongo-php-driver-legacy/blob/1.6.16/cursor_shared.c#L540), which is called from many places where a command response is read from a cursor. The most relevant places might be [`php_mongo_trigger_error_on_command_failure()`](https://github.com/mongodb/mongo-php-driver-legacy/blob/1.6.16/php_mongo.c#L649) and [`php_mongo_trigger_error_on_gle()`](https://github.com/mongodb/mongo-php-driver-legacy/blob/1.6.16/php_mongo.c#L695), but that won't come up for write commands, as any duplicate key error code would only show up within the `writeErrors` array of the command response and GLE isn't used anymore.

src/functions.php
jmikola: Note, sine PHPC 1.5.0 and PHPLIB 1.4.0 dropped support for MongoDB versions before 3.0, we need only consider `11000`. The other error codes all pertain to 2.6 and earlier servers.

lindelius:
@jmikola Any plans to revisit this one?

jmikola:
@lindelius: This isn't currently a priority (vs. other development tasks), but I'm also open to coming back to it at a later date, which is why I've not closed the issue. This may be something for @alcaeus to follow up on after he joins the team in a few months.

alcaeus:
Closing this: as explained in the linked ticket, there is no sensible way to wrap or replace the `BulkWriteException` with a `DuplicateKeyException`, since it would only work in a small subset of cases.

jmikola:
@lindelius: If you still need this functionality, I think the best solution may be to implement a utility function (e.g. `has_duplicate_key_error()`) that takes a [`MongoDB\Driver\BulkWriteException`](https://www.php.net/manual/en/class.mongodb-driver-exception-bulkwriteexception.php) and returns true if its code is `11000` or if any of its constituent write errors have the code `11000`. The latter check would be done by calling [`getWriteResult()`](https://www.php.net/manual/en/mongodb-driver-writeexception.getwriteresult.php), iterating on [`getWriteErrors()`](https://www.php.net/manual/en/mongodb-driver-writeresult.getwriteerrors.php), and then checking the code for each.

While researching this, I believe @alcaeus determined that our exception is assigned a code from the first write or write concern error (behavior from libmongoc). Therefore, simply checking the exception's top-level code is not sufficient if there was a duplicate key error somewhere within a batch of multiple write operations.

PR #392, opened 2017-08-08
PHPLIB-53: Create mapReduce command helper

jmikola:
https://jira.mongodb.org/browse/PHPLIB-53

/cc @alcaeus

PR #419, opened 2017-10-23
PHPLIB-53 Add documentation for mapReduce

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-53

### DIFF
--- a/docs/includes/apiargs-MongoDBCollection-method-mapReduce-option.yaml
+++ b/docs/includes/apiargs-MongoDBCollection-method-mapReduce-option.yaml
@@ -20,7 +20,7 @@ name: jsMode
 type: boolean
 description: |
   Specifies whether to convert intermediate data into BSON format between the
-  execution of the map and reduce functions. The default is `false`.
+  execution of the map and reduce functions. The default is ``false``.
 interface: phpmethod
 operation: ~
 optional: true
@@ -79,7 +79,7 @@ name: verbose
 type: boolean
 description: |
   Specifies whether to include the timing information in the result information.
-  The default is `true`.
+  The default is ``true``.
 interface: phpmethod
 operation: ~
 optional: true
--- a/docs/includes/apiargs-MongoDBCollection-method-mapReduce-option.yaml
+++ b/docs/includes/apiargs-MongoDBCollection-method-mapReduce-option.yaml
@@ -19,8 +19,8 @@ arg_name: option
 name: jsMode
 type: boolean
 description: |
-  Specifies whether to convert intermediate data into BSON format between the execution
-  of the map and reduce functions. The default is false.
+  Specifies whether to convert intermediate data into BSON format between the
+  execution of the map and reduce functions. The default is `false`.
 interface: phpmethod
 operation: ~
 optional: true
@@ -42,8 +42,8 @@ arg_name: option
 name: query
 type: document
 description: |
-  Specifies the selection criteria using query operators for determining the documents
-  input to the map function.
+  Specifies the selection criteria using query operators for determining the
+  documents input to the map function.
 interface: phpmethod
 operation: ~
 optional: true
@@ -60,7 +60,8 @@ arg_name: option
 name: scope
 type: document
 description: |
-  Specifies global variables that are accessible in the map, reduce and finalize functions.
+  Specifies global variables that are accessible in the map, reduce, and finalize
+  functions.
 interface: phpmethod
 operation: ~
 optional: true
@@ -78,7 +79,7 @@ name: verbose
 type: boolean
 description: |
   Specifies whether to include the timing information in the result information.
-  The default is true.
+  The default is `true`.
 interface: phpmethod
 operation: ~
 optional: true
--- a/docs/includes/apiargs-MongoDBCollection-method-mapReduce-option.yaml
+++ b/docs/includes/apiargs-MongoDBCollection-method-mapReduce-option.yaml
@@ -1,0 +1,89 @@
+source:
+  file: apiargs-MongoDBCollection-common-option.yaml
+  ref: bypassDocumentValidation
+---
+source:
+  file: apiargs-MongoDBCollection-common-option.yaml
+  ref: collation
+---
+arg_name: option
+name: finalize
+type: :php:`MongoDB\\BSON\\Javascript <class.mongodb-bson-javascript>`
+description: |
+  Follows the reduce method and modifies the output.
+interface: phpmethod
+operation: ~
+optional: true
+---
+arg_name: option
+name: jsMode
+type: boolean
+description: |
+  Specifies whether to convert intermediate data into BSON format between the execution
+  of the map and reduce functions. The default is false.
+interface: phpmethod
+operation: ~
+optional: true
+---
+arg_name: option
+name: limit
+type: integer
+description: |
+  Specifies a maximum number of documents for the input into the map function.
+interface: phpmethod
+operation: ~
+optional: true
+---
+source:
+  file: apiargs-common-option.yaml
+  ref: maxTimeMS
+---
+arg_name: option
+name: query
+type: document
+description: |
+  Specifies the selection criteria using query operators for determining the documents
+  input to the map function.
+interface: phpmethod
+operation: ~
+optional: true
+---
+source:
+  file: apiargs-MongoDBCollection-common-option.yaml
+  ref: readConcern
+---
+source:
+  file: apiargs-MongoDBCollection-common-option.yaml
+  ref: readPreference
+---
+arg_name: option
+name: scope
+type: document
+description: |
+  Specifies global variables that are accessible in the map, reduce and finalize functions.
+interface: phpmethod
+operation: ~
+optional: true
+---
+source:
+  file: apiargs-MongoDBCollection-method-find-option.yaml
+  ref: sort
+---
+source:
+  file: apiargs-MongoDBCollection-common-option.yaml
+  ref: typeMap
+---
+arg_name: option
+name: verbose
+type: boolean
+description: |
+  Specifies whether to include the timing information in the result information.
+  The default is true.
+interface: phpmethod
+operation: ~
+optional: true
+---
+source:
+  file: apiargs-MongoDBCollection-common-option.yaml
+  ref: writeConcern
+...
--- a/docs/reference/method/MongoDBCollection-mapReduce.txt
+++ b/docs/reference/method/MongoDBCollection-mapReduce.txt
@@ -15,13 +15,12 @@ Definition
 
 .. phpmethod:: MongoDB\\Collection::mapReduce()
 
-   The :manual:`mapReduce </core/map-reduce>` command
-   allows you to run map-reduce aggregation operations
-   over a collection.
+   The :manual:`mapReduce </reference/command/mapReduce>` command allows you to
+   run map-reduce aggregation operations over a collection.
 
    .. code-block:: php
 
-      function mapReduce($map, $reduce, $out, array $options = []) MongoDB\MapReduceResult
+      function mapReduce($map, $reduce, $out, array $options = []): MongoDB\MapReduceResult
 
    This method has the following parameters:
 
@@ -34,9 +33,10 @@ Definition
 Return Values
 -------------
 
-A :phpclass:`MongoDB\\MapReduceResult` object, which allows for iteration of mapReduce results
-irrespective of the output method (e.g. inline, collection) via the IteratorAggregate interface.
-It also provides access to command statistics.
+A :phpclass:`MongoDB\\MapReduceResult` object, which allows for iteration of
+mapReduce results irrespective of the output method (e.g. inline, collection)
+via the :php:`IteratorAggregate <iteratoraggregate>` interface. It also
+provides access to command statistics.
 
 Errors/Exceptions
 -----------------
@@ -53,15 +53,13 @@ In MongoDB, the map-reduce operation can write results to a collection
 or return the results inline. If you write map-reduce output to a
 collection, you can perform subsequent map-reduce operations on the
 same input collection that merge replace, merge, or reduce new results
-with previous results. See :manual:`mapReduce </core/map-reduce>` and
-:manual:`Perform Incremental Map-Reduce </tutorial/perform-incremental-map-reduce>` for details and
-examples.
+with previous results. See :manual:`Map-Reduce </core/map-reduce>` and
+:manual:`Perform Incremental Map-Reduce </tutorial/perform-incremental-map-reduce>`
+for details and examples.
 
 When returning the results of a map-reduce operation *inline*, the
 result documents must be within the :limit:`BSON Document Size` limit,
-which is currently 16 megabytes. For additional information on limits
-and restrictions on map-reduce operations, see the
-:manual:`mapReduce </reference/command/mapReduce>` reference page.
+which is currently 16 megabytes.
 
 MongoDB supports map-reduce operations on :manual:`sharded collections
 </sharding>`. Map-reduce operations can also output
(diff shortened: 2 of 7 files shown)


docs/includes/apiargs-MongoDBCollection-method-mapReduce-option.yaml
```diff
@@ -0,0 +1,89 @@
...
+interface: phpmethod
+operation: ~
+optional: true
+---
+arg_name: option
+name: jsMode
+type: boolean
+description: |
+  Specifies whether to convert intermediate data into BSON format between the execution
+  of the map and reduce functions. The default is false.
```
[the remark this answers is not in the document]
jmikola: Side note: if it's not too much trouble to setup in your editor, most of the docs pages wrap at 80 characters. I noticed that some of these lines ran a bit long, and paragraphs in `MongoDBCollection-mapReduce.txt` wrapped well before 80.

docs/reference/method/MongoDBCollection-mapReduce.txt
```diff
@@ -0,0 +1,96 @@
...
+same input collection that merge replace, merge, or reduce new results
+with previous results. See :manual:`mapReduce </core/map-reduce>` and
+:manual:`Perform Incremental Map-Reduce </tutorial/perform-incremental-map-reduce>` for details and
+examples.
+
+When returning the results of a map-reduce operation *inline*, the
+result documents must be within the :limit:`BSON Document Size` limit,
+which is currently 16 megabytes. For additional information on limits
+and restrictions on map-reduce operations, see the
+:manual:`mapReduce </reference/command/mapReduce>` reference page.
```
jmikola: I realize this last sentence comes straight from https://docs.mongodb.com/manual/core/map-reduce/#map-reduce-behavior, but the `mapReduce` command page doesn't really have any additional info. The [Output Inline](https://docs.mongodb.com/manual/reference/command/mapReduce/#output-inline) section merely repeats the BSON document limit, so I propose we just drop the last sentence here.

docs/reference/method/MongoDBCollection-mapReduce.txt
```diff
@@ -0,0 +1,96 @@
...
+.. include:: /includes/extracts/error-driver-runtimeexception.rst
+
+Behavior
+--------
+
+In MongoDB, the map-reduce operation can write results to a collection
+or return the results inline. If you write map-reduce output to a
+collection, you can perform subsequent map-reduce operations on the
+same input collection that merge replace, merge, or reduce new results
+with previous results. See :manual:`mapReduce </core/map-reduce>` and
```
jmikola: As done for the intro section, the link to "core" can be "Map-Reduce" since we're referring to the title of the linked page and not the command reference.

docs/reference/method/MongoDBCollection-mapReduce.txt
```diff
@@ -0,0 +1,96 @@
...
+:manual:`mapReduce </reference/command/mapReduce>` reference page.
+
+MongoDB supports map-reduce operations on :manual:`sharded collections
+</sharding>`. Map-reduce operations can also output
+the results to a sharded collection. See
+:manual:`Map-Reduce and Sharded Collections </core/map-reduce-sharded-collections>`.
+
+Example
+-------
+This example will get all users with at least one "sale" event and count the number of sales for each of these users.
```
jmikola: RST doesn't seem to mind, but a link break between the heading and following paragraph would be consistent with other doc sources.

docs/reference/method/MongoDBCollection-mapReduce.txt
```diff
@@ -0,0 +1,96 @@
...
+   <?php
+
+   $operation = new CreateCollection($this->getDatabaseName(), $this->getCollectionName());
+   $operation->execute($this->getPrimaryServer());
+   $map = new Javascript('function() { emit(this.user_id,1); }');
+   $reduce = new Javascript('function(k, vals) { var sum = 0; for (var i in vals) { sum += vals[i]; } return sum; }');
+   $out = array("merge" => "eventCounts");
+
+   $sales = new MapReduce($this->getDatabaseName(), $this->getCollectionName(), $map, $reduce, $out);
+   $sales->execute($this->getPrimaryServer());
```
jmikola: I just realized that https://github.com/mongodb/mongo-php-library/pull/392 never added a `Collection::mapReduce()` method for the operation. Would you mind taking a crack at adding that?

I'd like to steer clear of talking about the operation classes in our documentation, so this example would ideally just use the collection method. `$this->getPrimaryServer()` refers to a method in the test suite, so that will naturally get removed.

Existing Collection methods should provide examples of how to merge in special options such as `readConcern` and `writeConcern`.

Lastly, I would propose we change the operation class to check map, reduce, and finalize as an instance of [`MongoDB\BSON\JavascriptInterface`](http://php.net/manual/en/class.mongodb-bson-javascriptinterface.php) instead of [`MongoDB\BSON\Javascript`](http://php.net/manual/en/class.mongodb-bson-javascript.php). That type hint can also be enforced on the Collection method for `$map` and `$reduce`. Since `$out` allows multiple types, that won't need a type hint (we can reply on the operation's constructor to validate).

----

When the time comes to rewrite the example using the Collection method, the examples for `find()` and the various `findOneAnd___()` methods might be helpful. Those are all based on existing data from one of the [example data](https://docs.mongodb.com/php-library/current/tutorial/example-data/) collections, so if you can cook up a `mapReduce` command for one of those data sets, I might suggest running it in `inline` mode and then printing output as is done for `find()` (iterating on the MapReduceResult and dumping each document).

docs/reference/method/MongoDBCollection-mapReduce.txt
```diff
@@ -0,0 +1,96 @@
...
+
+.. phpmethod:: MongoDB\\Collection::mapReduce()
+
+   The :manual:`mapReduce </core/map-reduce>` command
+   allows you to run map-reduce aggregation operations
+   over a collection.
+
+   .. code-block:: php
+
+      function mapReduce($map, $reduce, $out, array $options = []) MongoDB\MapReduceResult
```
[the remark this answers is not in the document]
jmikola: Note that `$map` and `$reduce` may get type hints after other PR changes.

docs/reference/method/MongoDBCollection-mapReduce.txt
```diff
@@ -0,0 +1,96 @@
...
+   $operation = new CreateCollection($this->getDatabaseName(), $this->getCollectionName());
+   $operation->execute($this->getPrimaryServer());
+   $map = new Javascript('function() { emit(this.user_id,1); }');
+   $reduce = new Javascript('function(k, vals) { var sum = 0; for (var i in vals) { sum += vals[i]; } return sum; }');
+   $out = array("merge" => "eventCounts");
+
+   $sales = new MapReduce($this->getDatabaseName(), $this->getCollectionName(), $map, $reduce, $out);
+   $sales->execute($this->getPrimaryServer());
+
+This will create a MapReduceResult.
```
jmikola: If the example ends up printing output and using `The output would then resemble::` boilerplate, we can drop this line.

Otherwise, we'll want to use the `:phpclass:` notation so we can link to a forthcoming documentation page for the result class.

docs/reference/method/MongoDBCollection-mapReduce.txt
```diff
@@ -0,0 +1,96 @@
...
+
+   The ``$options`` parameter supports the following options:
+
+   .. include:: /includes/apiargs/MongoDBCollection-method-mapReduce-option.rst
+
+Return Values
+-------------
+
+A :phpclass:`MongoDB\\MapReduceResult` object, which allows for iteration of mapReduce results
+irrespective of the output method (e.g. inline, collection) via the IteratorAggregate interface.
```
jmikola: We can use the following to link to IteratorAggregate:

```rst
:php:`IteratorAggregate <iteratoraggregate>`
```

docs/reference/method/MongoDBCollection-mapReduce.txt
```diff
@@ -0,0 +1,96 @@
...
+   .. include:: /includes/apiargs/MongoDBCollection-method-mapReduce-param.rst
+
+   The ``$options`` parameter supports the following options:
+
+   .. include:: /includes/apiargs/MongoDBCollection-method-mapReduce-option.rst
+
+Return Values
+-------------
+
+A :phpclass:`MongoDB\\MapReduceResult` object, which allows for iteration of mapReduce results
```
jmikola: I also realize that we don't have an existing page for MapReduceResult. We can take care of that in a later PR (a usage example on this page is going to be more helpful for users anyway). Once a stub for the class exists, this syntax should start linking to it automatically.

The [library reference](https://docs.mongodb.com/php-library/current/reference/) has sections for both write result and enumeration classes, and MapReduceResult overlaps with both. Down the line, I imagine we can consolidate both of those pages into a single, large, "result classes" page.

src/Collection.php
```diff
@@ -810,6 +811,28 @@ public function listIndexes(array $options = [])
...
+     * @param array               $options        Command options
+     * @return MapReduceResult
+     * @throws UnsupportedException if options are not supported by the selected server
+     * @throws InvalidArgumentException for parameter/option parsing errors
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function mapReduce(JavascriptInterface $map, JavascriptInterface $reduce, $out, array $options = [])
+    {
+        $operation = new MapReduce($this->databaseName, $this->collectionName, $map, $reduce, $out, $options);
+        $server = $this->manager->selectServer(new ReadPreference(ReadPreference::RP_PRIMARY));
```
jmikola: The read preference is configurable, but we will want to coerce a primary read preference if the out option is not inline. That will be similar to how `aggregate()` checks if the `$out` pipeline operator is being used and, if so, forces a primary read preference. After we establish the read preference, we then use that to select a server upon which the operation class will be executed.

Since the out option can be a document (i.e. PHP array or object), I'd suggest looking at `is_last_pipeline_operator_out()` in `functions.php` for a resilient way to check. Something like the following should work:

```
// In this case, $out is probably a string (i.e. collection name)
if ( ! is_array($out) && ! is_object($out)) {
    return false;
}

// Cast a possible object to an array so we can conveniently check the first field name
$out = (array) $out;

return key($out) === 'inline';
```

Additionally, the type map, read concern, and write concern can all be inherited from the collection (if unspecified in the options). See `aggregate()` for an example of how we merge those in. Notably, read concern depends on the server's wire version, so we only inject a default value if we know it's supported.

src/Collection.php
```diff
@@ -810,6 +811,28 @@ public function listIndexes(array $options = [])
     }
 
     /**
+     * Runs map-reduce aggregation operations as defined by the $map and $reduce parameters.
```
jmikola: I'd suggest "Executes a map-reduce aggregation on the collection." based on `aggregate()`.

docs/includes/apiargs-MongoDBCollection-method-mapReduce-option.yaml
```diff
@@ -78,7 +79,7 @@ name: verbose
 type: boolean
 description: |
   Specifies whether to include the timing information in the result information.
-  The default is true.
+  The default is `true`.
```
jmikola: Same here re: two back-ticks.

docs/includes/apiargs-MongoDBCollection-method-mapReduce-option.yaml
```diff
@@ -19,8 +19,8 @@ arg_name: option
 name: jsMode
 type: boolean
 description: |
-  Specifies whether to convert intermediate data into BSON format between the execution
-  of the map and reduce functions. The default is false.
+  Specifies whether to convert intermediate data into BSON format between the
+  execution of the map and reduce functions. The default is `false`.
```
jmikola: I think RST requires two back-ticks, unlike Markdown, which allows one.

src/Collection.php
```diff
@@ -810,6 +812,58 @@ public function listIndexes(array $options = [])
...
+
+        $out = (array) $out;
+
+        if (key($out) !== 'inline') {
+            $options['readPreference'] = new ReadPreference(ReadPreference::RP_PRIMARY);
+        }
+
+        $server = $this->manager->selectServer($options['readPreference']);
+
+        // Set options if not set already
```
jmikola: I would suggest copying the same comment from `aggregate()`, as the explanation for readConcern is the same (see: [SERVER-20214](https://jira.mongodb.org/browse/SERVER-20214)).

```
/* A "majority" read concern is not compatible with the $out stage, so
 * avoid providing the Collection's read concern if it would conflict.
 */
```

Also, a majority read concern is only a problem if `$out` is not inline (i.e. mapReduce will write results to an output collection). To that end, we probably want to track `$isOutInline` (or its inverse, `$hasOutputCollection`) and implement similar logic as `aggregate()` does with the `$hasOutStage` variable. For `aggregate()`, `$hasOutStage` means the command will write to a collection.

src/Collection.php
```diff
@@ -810,6 +812,58 @@ public function listIndexes(array $options = [])
...
+        if ( ! isset($options['readConcern']) && ! ($this->readConcern->getLevel() === ReadConcern::MAJORITY) && \MongoDB\server_supports_feature($server, self::$wireVersionForReadConcern)) {
+            $options['readConcern'] = $this->readConcern;
+        }
+
+        if ( ! isset($options['typeMap']) && ( ! isset($options['useCursor']) || $options['useCursor'])) {
```
jmikola: MapReduce doesn't have a `useCursor` option, so this extra logic isn't relevant.

Aggregate happens to be limited because we never implemented support for type maps when it returns results inline without using a cursor. `MapReduce::createGetIteratorCallable()` actually gets around that limitation, which I didn't think of at the time when implementing Aggregate. That may be something to fix down the line for Aggregate, depending on how long we'll keep supporting non-cursor replies (that support may end when we remove support for MongoDB 2.6).

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -180,6 +182,41 @@ public function testWithOptionsPassesOptions()
...
+    public function testMapReduceWithOptions() {
+        $this->createFixtures(3);
+
+        $map = new Javascript('function() { emit(this._id, this.x); }');
+        $reduce = new Javascript('function(key, values) { return Array.sum(values); }');
+        $out = ['inline' => 1];
+
+        $options = ["readPreference" => ReadPreference::RP_SECONDARY_PREFERRED];
+
+        $operation = new MapReduce($this->getDatabaseName(),
```
jmikola: This shouldn't be testing the Operation class directly, as the test doesn't interact with the Collection object at all.

I think we can just remove the `testMapReduce()` test, as the read preference option isn't testing much. In most environments, this is just going to select the primary server.

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -180,6 +182,41 @@ public function testWithOptionsPassesOptions()
         $this->assertSame(WriteConcern::MAJORITY, $debug['writeConcern']->getW());
     }
 
+    public function testMapReduce()
+    {
+        $this->createFixtures(3);
+
+        $map = new Javascript('function() { emit(this._id, this.x); }');
+        $reduce = new Javascript('function(key, values) { return Array.sum(values); }');
```
jmikola: I might suggest emitting on `1` instead of `this._id`, so that we collect all documents into a single bucket. Then, the reduce function can compute the sum of all `x` values. We can later assert that the result of map reduce is a single document with `[ '_id' => 1, 'x' => 66 ]`. Something like:

```
$expected = [
    [ '_id' => 1, 'x' => 66 ],
];

$this->assertSameDocuments($expected, $mapReduceResult);
```

src/Collection.php
```diff
@@ -833,24 +833,21 @@ public function mapReduce(JavascriptInterface $map, JavascriptInterface $reduce,
...
+        if ($this->isOutInline($out)) {
+            $isOutInline = true;
             $options['readPreference'] = new ReadPreference(ReadPreference::RP_PRIMARY);
         }
 
         $server = $this->manager->selectServer($options['readPreference']);
 
-        // Set options if not set already
-        if ( ! isset($options['readConcern']) && ! ($this->readConcern->getLevel() === ReadConcern::MAJORITY) && \MongoDB\server_supports_feature($server, self::$wireVersionForReadConcern)) {
+        /* A "majority" read concern is not compatible with $out inline, so
```
jmikola: "$out" was a reference to the aggregation pipeline operator. "with inline output, so" may read better.

src/Collection.php
```diff
@@ -833,24 +833,21 @@ public function mapReduce(JavascriptInterface $map, JavascriptInterface $reduce,
...
+         * avoid providing the Collection's read concern if it would conflict.
+         */
+        if ( ! isset($options['readConcern']) && ! ($isOutInline && $this->readConcern->getLevel() === ReadConcern::MAJORITY) && \MongoDB\server_supports_feature($server, self::$wireVersionForReadConcern)) {
```
jmikola: I think there's a logic error in the middle condition:

```
! ($isOutInline && $this->readConcern->getLevel() === ReadConcern::MAJORITY)
````

If we apply [De Morgan's Law](https://en.wikipedia.org/wiki/De_Morgan%27s_laws) to distribute the negation, we get:

```
( ! $isOutInline || $this->readConcern->getLevel() !== ReadConcern::MAJORITY )
```

We can write this in prose as:

> mapReduce is writing to a collection OR read concern is not "majority"

And back in the original condition:

> * the user hasn't specified a read concern AND
> * (mapReduce is writing to a collection OR read concern is not "majority") AND
> * the server supports read concerns

Given this, it's possible that the Collection could have a majority read concern and we'd still inherit it as a default when mapReduce is writing to a collection. Bad news bears.

While we can fix this by negating `$isOutInline` within the middle conditional, it's already hard to read as it is. I'd suggest changing the variable to track the inverse (e.g. `$hasOutputCollection` as mentioned in https://github.com/mongodb/mongo-php-library/pull/419#discussion_r146691277). Then, this condition basically copies the `aggregate()` snippet with one variable rename:

```php
/* A "majority" read concern is not compatible with writing output to a
 * collection, so avoid providing the Collection's read concern if it
 * would conflict.
 */
     ! ($hasOutputCollection && $this->readConcern->getLevel() === ReadConcern::MAJORITY) &&
    \MongoDB\server_supports_feature($server, self::$wireVersionForReadConcern)) {
}
```

src/Collection.php
```diff
@@ -833,24 +833,21 @@ public function mapReduce(JavascriptInterface $map, JavascriptInterface $reduce,
...
         // Check if the out option is inline because we will want to coerce a primary read preference if not
-        if ( ! is_array($out) && ! is_object($out)) {
-            return false;
-        }
-
-        $out = (array) $out;
-
-        if (key($out) !== 'inline') {
+        if ($this->isOutInline($out)) {
+            $isOutInline = true;
```
jmikola: Two issues here:

1. `isOutInline` is used later in a conditional, but its assignment here is nested within a condition. As such, it's possible we'd get an undefined variable error later on.
2. This condition forces a primary read preference is output is inline (i.e. not writing to a collection), which is backwards.

If we make the swap to `$hasOutputCollection`, I'd suggest assigning `$hasOutputCollection = ! $this->isOutInline($out);` in the method's top scope and then replacing this condition with `if ($hasOutputCollection)` to force a primary read preference.

jmikola:
LGTM. I noticed there were a few documentation changes mixed in with the last four commits related to `Collection::mapReduce()`, so feel free to squash everything into a single commit for merging.

PR #422, opened 2017-10-24
PHPLIB-266 Support passing index hint to aggregations

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-266

### DIFF
--- a/docs/includes/apiargs-MongoDBCollection-method-aggregate-option.yaml
+++ b/docs/includes/apiargs-MongoDBCollection-method-aggregate-option.yaml
@@ -34,9 +34,18 @@ post: |
   Document validation requires MongoDB 3.2 or later: if you are using an earlier
   version of MongoDB, this option will be ignored.
 ---
-source:
-  file: apiargs-MongoDBCollection-method-find-option.yaml
-  ref: hint
+arg_name: option
+name: hint
+type: string|array|object
+description: |
+  The index to use. Specify either the index name as a string or the index key
+  pattern as a document. If specified, then the query system will only consider
+  plans using the hinted index.
+
+  .. versionadded:: 1.3
+interface: phpmethod
+operation: ~
+optional: true
 ---
 source:
   file: apiargs-common-option.yaml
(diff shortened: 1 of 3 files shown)


src/Operation/Aggregate.php
```diff
@@ -272,6 +280,10 @@ private function createCommand(Server $server, $isCursorSupported)
             $cmd['collation'] = (object) $this->options['collation'];
         }
 
+        if (isset($this->options['hint'])) {
+            $cmd['hint'] = $this->options['hint'];
```
jmikola: If `hint` is an array, we should cast it to an object. This hands the edge case of the hint document: `{ "0": 1 }`, which could be `[ "0" => 1]` in PHP and would otherwise convert to a BSON array. If it's not an array (i.e. string or object), we can leave it as-is.

See `Count::createCommand()` for an example.

docs/includes/apiargs-MongoDBCollection-method-aggregate-option.yaml
```diff
@@ -35,6 +35,10 @@ post: |
   version of MongoDB, this option will be ignored.
 ---
 source:
+  file: apiargs-MongoDBCollection-method-find-option.yaml
+  ref: hint
+---
```
jmikola: Rather than embed the corresponding `find` option, let's write the body inline as is done in `apiargs-MongoDBCollection-method-count-options.yaml`. The main description paragraph is the same, but we'll want to add version information (related to [PHPLIB-284](https://jira.mongodb.org/browse/PHPLIB-284)) so users know which version of the library added support for this option:

```
arg_name: option
name: hint
type: string|array|object
description: |
  The index to use. Specify either the index name as a string or the index key
  pattern as a document. If specified, then the query system will only consider
  plans using the hinted index.

  .. versionadded:: 1.3
interface: phpmethod
operation: ~
optional: true
```

jmikola:
@kvwalker: `master` is now 1.3.0-dev. Feel free to rebase and merge.

moderndeveloperllc:
@jmikola Just curious - does this mean you are going to skip 1.2.x and just go to 1.3 to avoid the confusion of a 1.3 driver and 1.2 library?

jmikola:
@moderndeveloperllc: No sir. 1.2.0 should be released on Monday. I just split the branches today so that @kvwalker could start merging pull requests for issues in the 1.3.0 milestone, which is going to be the release with MongoDB 3.6 compatibility.

PR #424, opened 2017-10-26
PHPLIB-81: Implement CachingIterator

jmikola:
https://jira.mongodb.org/browse/PHPLIB-81

Supersedes https://github.com/mongodb/mongo-php-library/pull/327


src/Model/CachingIterator.php
```diff
@@ -0,0 +1,166 @@
...
+    private $iteratorExhausted = false;
+
+    /**
+     * Constructor.
+     *
+     * Initialize the iterator and stores the first item in the cache. This
+     * effectively rewinds the Traversable and the wrapping Generator, which
+     * will execute up to its first yield statement. Additionally, this mimics
+     * behavior of the SPL iterators and allows users to omit an explicit call
+     * to rewind() before using the other methods.
```
jmikola: @alcaeus: Strictly speaking, Iterators must call `rewind()` before iterating; however, most SPL iterators relax that restriction and rewind during construction. I kept that behavior, while adding `$iteratorAdvanced` to ensure that the initial call to `rewind()` does not exhaust all results. That may be why @bjori originally assumed you had implemented a BufferedIterator. Based on `testPartialIterationDoesNotExhaust()`, the original implementation was actually buffering up front!

src/Model/CachingIterator.php
```diff
@@ -0,0 +1,166 @@
...
+ *
+ * By caching results, this iterators allows a Traversable to be counted and
+ * rewound multiple times, even if the wrapped object does not natively support
+ * those operations (e.g. MongoDB\Driver\Cursor).
+ *
+ * @internal
+ */
+class CachingIterator implements Countable, Iterator
+{
+    private $items = [];
```
jmikola: This helps if we count an empty Traversable, as the early return in `storeCurrentItem()` would have prevented this from ever being initialized from `null` to an `array` type.

src/Model/CachingIterator.php
```diff
@@ -0,0 +1,166 @@
...
+     * Wraps the Traversable with a Generator.
+     *
+     * @param Traversable $traversable
+     * @return Generator
+     */
+    private function wrapTraversable(Traversable $traversable)
+    {
+        foreach ($traversable as $key => $value) {
+            yield $key => $value;
+            $this->iteratorAdvanced = true;
```
jmikola: My understanding (and observation) is that the first call to `wrapTraversable()` from our constructor is going to immediately execute this function up to (and including?) the `yield` statement. The `$iteratorAdvanced` assignment will take place once we re-enter, immediately before `foreach` advances the Traversable (e.g. `Iterator::next()` if it happens to be an Iterator).

src/Model/CachingIterator.php
```diff
@@ -0,0 +1,166 @@
...
+
+    /**
+     * Wraps the Traversable with a Generator.
+     *
+     * @param Traversable $traversable
+     * @return Generator
+     */
+    private function wrapTraversable(Traversable $traversable)
+    {
+        foreach ($traversable as $key => $value) {
```
jmikola: Just to recap, each `foreach` starts with invoking the internal iterator's rewind handler, followed by a validity check. If valid, keys and values are fetched. On subsequent iterations, a call to the next handler takes place instead of rewind.

tests/Model/CachingIteratorTest.php
```diff
@@ -0,0 +1,130 @@
...
+        foreach ($items as $item) {
+            yield $item;
+        }
+    }
+
+    private function getTraversableThatThrows($items)
+    {
+        foreach ($items as $item) {
+            if ($item instanceof Exception) {
+                throw $item;
```
jmikola: The stack trace of the exception only reports when it was constructed (it's not generated/updated when the exception is thrown). I was surprised to learn that, but it's not too important as we only ever assert that nothing is thrown (testing that initial rewind no longer exhausts the internal traversable).

src/Model/CachingIterator.php
alcaeus: Correct - this could be tested by adding an assertion on the `iteratorAdvanced` property on the `testConstructorRewinds` test:
```
$this->assertAttributeSame(false, 'iteratorAdvanced', $iterator);
```

src/Model/CachingIterator.php
```diff
@@ -0,0 +1,166 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace MongoDB\Model;
+
+use Countable;
+use Generator;
+use Iterator;
+use Traversable;
```
derickr: Why do this if if you're not aliasing them, or plucking them out of another namespace? And if you want the root namespace, shouldn't it be ``use \Countable`` ?

tests/Model/CachingIteratorTest.php
derickr: I believe I tried fixing that in Xdebug, but adding the previous exception to a special property.

src/Model/CachingIterator.php
jmikola: `use` statements use fully qualified names, so the leading backslash isn't necessary. Out of habit, I list all used classes at the top of the file, rather than making a special case of `\Countable` for classes in the global namespace.

src/Model/CachingIterator.php
jmikola: I didn't realize those assertions existed :)

I think I'll do without them here, since we can get by testing with just the public API. Will keep it in mind for future use, though!

PR #428, opened 2017-11-02
PHPLIB-276 Change stream outline

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-276


src/Collection.php
```diff
@@ -936,6 +937,23 @@ public function updateOne($filter, $update, array $options = [])
...
 
+    /*
+     * ChangeStream outline
+     *
+     * @see ChangeStream::__construct() for supported options
+     * @param array          $pipeline       List of pipeline operations
+     * @param array          $options        Command options
+     * @throws InvalidArgumentException for parameter/option parsing errors
+     */
+    public function watch(array $pipeline, array $options = [])
```
jmikola: I think we can also default `$pipeline` to `[]`, as we do with `$filter` arguments for some other CRUD methods. That only needs to happen on the Collection method. The operation class can require the user to specify an array on their own.

src/Collection.php
```diff
@@ -936,6 +937,23 @@ public function updateOne($filter, $update, array $options = [])
...
+    {
+        $server = $this->manager->selectServer($this->readPreference);
```
jmikola: Much like other CRUD methods, we can take a `readPreference` option in `$options` and fall back to the Collection's read preference if it's not specified. The CRUD and ChangeStream spec don't require per-operation options for read preference, read concern, and write concern, but PHPLIB does it throughout (mostly for historical reasons).

We can do the same for `readConcern`. The current spec doesn't allow that (it requires we inherit the Collection's read concern), but I believe that was an oversight. I've opened https://github.com/mongodb/specifications/pull/237 to correct that.

src/Operation/ChangeStream.php
```diff
@@ -0,0 +1,159 @@
...
+            if ($i !== $expectedIndex) {
+                throw new InvalidArgumentException(sprintf('$pipeline is not a list (unexpected index: "%s")', $i));
+            }
+
+            if ( ! is_array($operation) && ! is_object($operation)) {
+                throw InvalidArgumentException::invalidType(sprintf('$pipeline[%d]', $i), $operation, 'array or object');
+            }
+
+            $expectedIndex += 1;
+        }
```
jmikola: Aggregate will already validate the `$pipeline`, so I think we can do without this here and just trust `array_unshift()` to prepend our new `$changeStream` pipeline operator to the array. Based on some local testing, it does the sensible thing.

**Sequential numerically-indexed array**

```
>>> $a = [0=>'a',1=>'b']
=> [
     "a",
     "b",
   ]
>>> array_unshift($a, 'foo')
=> 3
>>> $a
=> [
     "foo",
     "a",
     "b",
   ]
```

**Associative array**

```
>>> $a = ['foo'=>'a',1=>'b']
=> [
     "foo" => "a",
     1 => "b",
   ]
>>> array_unshift($a, 'hello')
=> 3
>>> $a
=> [
     0 => "hello",
     "foo" => "a",
     1 => "b",
   ]
```

**Non-sequential numerically-indexed array**

```
>>> $a = [-1=>'a',1=>'b']
=> [
     -1 => "a",
     1 => "b",
   ]
>>> array_unshift($a, 'foo')
=> 3
>>> $a
=> [
     "foo",
     "a",
     "b",
   ]
```

This last example actually turns an otherwise invalid pipeline into a valid one (PHP appears to re-index the array), but I don't think that's a problem.

src/Operation/ChangeStream.php
```diff
@@ -0,0 +1,159 @@
...
+
+        if (isset($options['fullDocument']) && ! is_string($options['fullDocument'])) {
```
jmikola: Let's create `DEFAULT` and `UPDATE_LOOKUP` constants for now. Per the spec, we don't need to enforce that the user provides one of those two values.

src/Operation/ChangeStream.php
```diff
@@ -0,0 +1,159 @@
...
+    /**
+     * Create the changeStream command.
+     *
+     * @param Server  $server
+     * @param boolean $isCursorSupported
+     * @return Command
+     */
+    private function createCommand(Server $server)
+    {
+        $changeStreamArray = ['$changeStream' => $this->options];
```
jmikola: Rather than pass all options into the pipeline operator, we should construct it with only known options that are `isset()`, as we do with command documents in other Operation classes.

Additionally, the options will need to be split based on which apply to the pipeline operator and which belong to Aggregate. `maxAwaitTimeMS` will likely need to be added to Aggregate down the line, since that's the Operation class that actually builds the Command object and will utilize the new API from [PHPC-1029](https://jira.mongodb.org/browse/PHPC-1029).

tests/Operation/ChangeStreamFunctionalTest.php
```diff
@@ -0,0 +1,42 @@
...
+        $changeStreamResult = $operation->execute($this->getPrimaryServer());
+    }
+
+    public function testEmptyPipeline()
+    {
+        $bulkWrite = new BulkWrite;
+        $bulkWrite->insert(['_id' => 1, 'x' => 1]);
+        $bulkWrite->insert(['_id' => 2, 'x' => 2]);
+        $bulkWrite->insert(['_id' => 3, 'y' => 3]);
+        $this->manager->executeBulkWrite($this->getNamespace(), $bulkWrite);
```
jmikola: Confirm if change streams actually requires the collection to exist with data before we call `Collection::watch()`.

tests/Operation/ChangeStreamFunctionalTest.php
```diff
@@ -0,0 +1,42 @@
...
+
+        $readPreference = new ReadPreference(ReadPreference::RP_PRIMARY);
+        $operation = new ChangeStream($this->getDatabaseName(), $this->getCollectionName(), []);
+        $changeStreamResult = $operation->execute($this->getPrimaryServer());
```
jmikola: This functional test can borrow ideas from [cursor-tailable-003.phpt](https://github.com/mongodb/mongo-php-driver/blob/master/tests/cursor/cursor-tailable-003.phpt) (and related tests) in PHPC. Once you have a ChangeStream class that wraps the cursor returned by Aggregate, you can manually iterate on it via `rewind()`, `current()`, and `next()` and modify the collection during iteration to assert to new changes show up.

src/Operation/ChangeStream.php
```diff
@@ -0,0 +1,159 @@
...
+     * @throws UnsupportedException if collation, read concern, or write concern is used and unsupported
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function execute(Server $server)
+    {
+        $command = $this->createCommand($server);
+
+        $cursor = $command->execute($server);
+
+        return $cursor;
```
jmikola: We'll need an object to wrap the cursor. A ChangeStream class that extends [IteratorIterator](http://php.net/iteratoriterator) should serve as a good starting point.

src/ChangeStream.php
```diff
@@ -0,0 +1,93 @@
...
+use MongoDB\Operation\ChangeStreamCommand;
+
+/**
+ * Operation for the changeStream command.
+ *
+ * @api
+ * @see \MongoDB\Collection::changeStream()
+ * @see http://docs.mongodb.org/manual/reference/command/changeStream/
+ */
+class ChangeStream
```
jmikola: Is there a need for this class to be separate from ChangeStreamIterator?

Since this is the class returned by executing the command (and `Collection::watch()`), I would expect it to implement `Iterator` or `IteratorAggregate`, so that users can [compose](https://en.wikipedia.org/wiki/Object_composition) it with their class expecting an iterable.

src/ChangeStream.php
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    public function current()
+    {
+        return $this->csIt->current();
+    }
+
+    public function getId()
+    {
+        return $this->csIt->getId();
```
jmikola: I had no idea what this was doing on the first run through, as ChangeStreamIterator has no `getId()` method. After reading the [IteratorIterator Notes](http://php.net/iteratoriterator), I realized it is calling `MongoDB\Driver\Cursor::getId()` due to the following behavior:

> This class [IteratorIterator] permits access to methods of the inner iterator via the __call magic method.

I think explicitly accessing the inner iterator and calling `getId()` would be preferable (for future maintainability and keeping the code readable).

Additionally, ChangeStreamIterator currently doesn't enforce the type of its inner iterator. In the interest of being defensive, it'd be wise to overload its constructor to type hint against the `MongoDB\Driver\Cursor` class before then calling `parent::__construct($cursor)`. This will ensure that it only ever wraps a cursor object.

src/ChangeStream.php
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    public function next()
+    {
+        try {
+            $this->csIt->next();
+            $this->resumeToken = $this->csIt->extract_resume_token($this->csIt->current());
```
jmikola: I see that this may be why `extract_resume_token()` was public. Consolidating both classes should allow you to switch to private.

src/ChangeStream.php
```diff
@@ -0,0 +1,93 @@
...
+        } catch (ResumableException $e) {
+            $this->resume();
+        }
+    }
+
+    public function resume()
+    {
+        $this->options['resumeAfter'] = $this->resumeToken;
+        array_shift($this->pipeline);
```
jmikola: Is this just popping the original `$changeStream` stage off of the pipeline so that ChangeStreamCommand can add a new one?

I noticed that `ChangeStreamCommand::resume()` appears to do the same thing. Is there a risk that we're removing too much from the pipeline?

kvwalker:
I implemented the resume functionality using $resumeCallable. I haven't finished the ChangeStreamFunctionalTest so I'll work on that now.

src/ChangeStream.php
```diff
@@ -17,38 +17,28 @@
...
-    private $collectionName;
-    private $pipeline;
-    private $options;
     private $resumeToken;
-    private $manager;
+    private $resumeCallable;
     private $csIt;
 
-    public function __construct($cursor, $databaseName, $collectionName, array $pipeline, array $options = [], $resumeToken, Manager $manager)
+    public function __construct(Cursor $cursor, Callable $resumeCallable)
```
jmikola: I know class names in PHP are case-insensitive, but I'm not sure about other type names. [`callable`](http://php.net/manual/en/language.types.callable.php) is the canonical casing, though.

src/ChangeStream.php
```diff
@@ -58,36 +48,53 @@ public function current()
...
+
     }
 
     public function next()
     {
         try {
             $this->csIt->next();
-            $this->resumeToken = $this->csIt->extract_resume_token($this->csIt->current());
-        } catch (ResumableException $e) {
+            $this->resumeToken = $this->extractResumeToken($this->csIt->current());
```
jmikola: I believe the original ChangeStreamIterator implementation captured a resume token from `current()`.

If this is extracting _after_ the calling `$this->csIt->next()`, I think it's possible for `$this->resumeToken` to be uninitialized if we hit an exception on the very first call to `next()`.

I actually do like the idea of capturing the resume token during iterator movement, in case the user decides to not call `current()` for some reason (maybe they're ignoring/skipping some results). In that case, it may make sense to extract at the following points:

 * When we first construct this class (in case the user never calls `rewind()`)
 * Each time `next()` is called

IIRC, the driver's cursor doesn't actually need an initial rewind when iterating with a wrapped IteratorIterator. You can test that to make sure, though. It still wouldn't hurt to call `IteratorIterator::rewind()` in our constructor, though.

On a separate note, if user's do call `rewind()`, they should only do so at the beginning of iteration. We can debate whether it would be better for `ChangeStream::rewind()` to conditionally throw if called after one or more calls to `ChangeStream::next()`, or to simply have `ChangeStream::rewind()` be a no-op.

src/ChangeStream.php
```diff
@@ -58,36 +48,53 @@ public function current()
...
+
+    private function extractResumeToken($document)
+    {
+        if ($document === null) {
+            throw new ResumeTokenException("Cannot extract a resumeToken from an empty document");
+        }
+        if ($document instanceof Serializable) {
+            return $this->extractResumeToken($document->bsonSerialize());
+        }
+        $lastResumeToken = is_array($document) ? $document['_id'] : $document->_id;
```
jmikola: Is there a case where we'd want to leave `$this->resumeToken` untouched if extraction fails? Would we expect users to catch this ResumeTokenException and try to carry on with iteration? I suppose they'd just hit another extraction error in that case since the cursor data wouldn't have changed.

src/ChangeStream.php
```diff
@@ -60,7 +62,7 @@ public function next()
     {
         try {
             $this->csIt->next();
-            $this->resumeToken = $this->extractResumeToken($this->csIt->current());
+            $this->extractResumeToken($this->csIt->current());
```
jmikola: Maybe wrap this in `if ($this->valid())` in case `next()` just did a getMore but there were no new results.

jmikola: Typically, `$this->valid()` would check if `$this->key() !== null`. That still works, provided we're actually calling `key()` on the internal iterator.

As for `$this->key()`, we can't simply call `key()` on the internal iterator because resuming ends up recreating a PHP cursor, which would reset it's key sequence at zero. Therefore, you probably need to keep track of an integer sequence yourself.

src/ChangeStream.php
```diff
@@ -82,7 +87,7 @@ public function rewind()
 
     public function valid()
     {
-
+        return $this->csIt->key() !== null;
```
jmikola: Looking at [`IteratorIterator::key()`](https://github.com/php/php-src/blob/PHP-7.1.11/ext/spl/spl_iterators.c#L1774), I believe this is equivalent to returning [`IteratorIterator::valid()`](https://github.com/php/php-src/blob/PHP-7.1.11/ext/spl/spl_iterators.c#L1753). If you think that's more readable, you can swap to returning `$this->csIt->valid()`, but this should work perfectly fine as-is.

Both of them depend on [`spl_dual_it_fetch()`](https://github.com/php/php-src/blob/PHP-7.1.11/ext/spl/spl_iterators.c#L1693) to populate the current data and key from the inner iterator (our Cursor). Looking at [Cursor's key handler](https://github.com/mongodb/mongo-php-driver/blob/1.3.2/src/MongoDB/Cursor.c#L69) again, I realize I misspoke earlier when I implied that it could return `null` if there was no current element. It actually always returns an integer. That's not technically a problem since we don't implement the userland Iterator interface. PHP's IteratorIterator class and its `key()` and `valid()` methods still function properly because our [current_data handler](https://github.com/mongodb/mongo-php-driver/blob/1.3.2/src/MongoDB/Cursor.c#L77) _does_ return an undefined value if there is no current element. This is because we always free/unset the current element before advancing with the next handler. If libmongoc's cursor doesn't have a next element for us, the Cursor's current element is left unset.

src/ChangeStream.php
```diff
@@ -55,14 +57,17 @@ public function getId()
 
     public function key()
     {
-
+        return $this->key;
```
jmikola: Since we're implementing PHP's Iterator interface, we still have to honor the return value contract from [`Iterator::key()`](http://php.net/manual/en/iterator.key.php):

> Returns scalar on success, or NULL on failure.

As-is, I think this always returns an integer. We should probably consult `$this->valid()` first.

src/ChangeStream.php
```diff
@@ -55,14 +57,17 @@ public function getId()
...
     }
 
     public function next()
     {
+        $this->key++;
```
jmikola: Not every advancement with `next()` is going to be successful. This looks like it's possible for `$this->key` to get incremented more times than necessary.

I think a test where there is no next element (because no changes have occurred) or we end up resuming a few times would demonstrate that the key can get out of sync.

tests/Operation/ChangeStreamFunctionalTest.php
```diff
@@ -129,4 +129,29 @@ public function test_resume_after_kill_then_insert()
         $changeStreamResult->next();
         $this->assertNull($changeStreamResult->current());
     }
+
+    public function test_key()
+    {
+        $operation = new DatabaseCommand("admin", ["setFeatureCompatibilityVersion" => "3.6"]);
```
jmikola: I think this came up in a previous review, but we should expect the server to have the correct FCV already. The `setUp()` method's skip logic should handle cases where the server is misconfigured.

tests/Operation/ChangeStreamFunctionalTest.php
```diff
@@ -129,4 +129,29 @@ public function test_resume_after_kill_then_insert()
...
+        $changeStreamResult = $this->collection->watch();
+
+        $this->assertSame(0, $changeStreamResult->key());
+
+        $result = $this->collection->insertOne(['x' => 1]);
+        $this->assertInstanceOf('MongoDB\InsertOneResult', $result);
+        $this->assertSame(1, $result->getInsertedCount());
+
+        $changeStreamResult->next();
+        $this->assertSame(1, $changeStreamResult->key());
```
jmikola: If you were to do multiple calls to `$changeStreamResult->next();` here and there were no new changes, would `key()` remain at 1?

src/ChangeStream.php
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    public function getId()
+    {
+        return $this->csIt->getInnerIterator()->getId();
+    }
+
+    public function key()
+    {
+        return $this->key;
```
jmikola: Since we're implementing PHP's Iterator interface, we still have to honor the return value contract from [`Iterator::key()`](http://php.net/manual/en/iterator.key.php):

> Returns scalar on success, or NULL on failure.

As-is, I think this always returns an integer. We should probably consult `$this->valid()` first.

src/ChangeStream.php
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    public function next()
+    {
+        $this->key++;
```
jmikola: Not every advancement with `next()` is going to be successful. This looks like it's possible for `$this->key` to get incremented more times than necessary.

I think a test where there is no next element (because no changes have occurred) or we end up resuming a few times would demonstrate that the key can get out of sync.

src/ChangeStream.php
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    public function rewind()
+    {
+        $this->csIt->rewind();
+    }
+
+    public function valid()
+    {
+        return $this->csIt->key() !== null;
```
jmikola: Looking at [`IteratorIterator::key()`](https://github.com/php/php-src/blob/PHP-7.1.11/ext/spl/spl_iterators.c#L1774), I believe this is equivalent to returning [`IteratorIterator::valid()`](https://github.com/php/php-src/blob/PHP-7.1.11/ext/spl/spl_iterators.c#L1753). If you think that's more readable, you can swap to returning `$this->csIt->valid()`, but this should work perfectly fine as-is.

Both of them depend on [`spl_dual_it_fetch()`](https://github.com/php/php-src/blob/PHP-7.1.11/ext/spl/spl_iterators.c#L1693) to populate the current data and key from the inner iterator (our Cursor). Looking at [Cursor's key handler](https://github.com/mongodb/mongo-php-driver/blob/1.3.2/src/MongoDB/Cursor.c#L69) again, I realize I misspoke earlier when I implied that it could return `null` if there was no current element. It actually always returns an integer. That's not technically a problem since we don't implement the userland Iterator interface. PHP's IteratorIterator class and its `key()` and `valid()` methods still function properly because our [current_data handler](https://github.com/mongodb/mongo-php-driver/blob/1.3.2/src/MongoDB/Cursor.c#L77) _does_ return an undefined value if there is no current element. This is because we always free/unset the current element before advancing with the next handler. If libmongoc's cursor doesn't have a next element for us, the Cursor's current element is left unset.

tests/Operation/ChangeStreamFunctionalTest.php
```diff
@@ -0,0 +1,186 @@
...
+        $this->assertInstanceOf('MongoDB\InsertOneResult', $result);
+        $this->assertSame(1, $result->getInsertedCount());
+
+        $changeStreamResult->next();
+        $this->assertNull($changeStreamResult->current());
+    }
+
+    public function test_key()
+    {
+        $operation = new DatabaseCommand("admin", ["setFeatureCompatibilityVersion" => "3.6"]);
```
jmikola: I think this came up in a previous review, but we should expect the server to have the correct FCV already. The `setUp()` method's skip logic should handle cases where the server is misconfigured.

src/ChangeStream.php
```diff
@@ -57,16 +57,19 @@ public function getId()
 
     public function key()
     {
-        return $this->key;
+        if ($this->valid() !== null) {
```
jmikola: `valid()` returns a boolean, so this condition would always be satisfied.

src/ChangeStream.php
```diff
@@ -57,16 +57,19 @@ public function getId()
...
     }
 
     public function next()
     {
-        $this->key++;
         try {
             $this->csIt->next();
             if ($this->valid()) {
                 $this->extractResumeToken($this->csIt->current());
+                $this->key++;
```
jmikola: I just noticed that ResumeTokenException extends the driver's RuntimeException class. If the user specifies a projection that removes the resume token, I think that would be caught by `catch (RuntimeException $e)` below and we'd resume unnecessarily. Resuming also wouldn't change the fact that a resume token is still being omitted from the result, so that might cause us to resume repeatedly.

src/ChangeStream.php
kvwalker: If next() is called multiple times without any changes to the collection, should we return the most recent key or should we return null for the key since valid() will return false?

jmikola: If the current element is the most recent change, `valid()` returns `true` and `key()` returns the appropriate index. If we then call `next()` and there is nothing to advance to, `valid()` should then return `false` and `key()` should return `null` (i.e. error condition, since `key()` and `current()` are not supposed to be called if `valid()` is `false`).

The `foreach` construct depends on `valid()` returning `true` and `false` to stop iterating. If we never return `false` at the end of iteration, a `foreach` loop might never terminate. This is one reason why we can't use `foreach` to iterate on a tailable `MongoDB\Driver\Cursor`, since it starts being invalid once we reach the end and we require `next()` to coerce libmongoc into checking for a next element.

jmikola:
Confirming the following from the [Change Stream Test Plan](https://github.com/mongodb/specifications/blob/master/source/change-streams.rst#test-plan):

> * ChangeStream will automatically resume one time on a resumable error (including not master) with the initial pipeline and options, except for the addition/update of a resumeToken.
> * ChangeStream will not attempt to resume on a server error

This may require logic to ensure that the exception is either a connection error, "not master" in its message, or indicates that the cursor is not found (code 43). See: [`mongoc-change-stream.c`](https://github.com/mongodb/mongo-c-driver/blob/1.9.0-rc0/src/mongoc/mongoc-change-stream.c#L281) for how libmongoc does that.

As far as resuming only one time, I believe that is already covered since `resume()` is not called within a try/catch.

kvwalker:
I added the "not master" check and the cursor not found (code 43) check. I'm not quite sure how to handle connection errors. Also, how can I test for "not master" and connection errors? Should I just construct a new ChangeStreamCommand and execute it on the secondary server to test for the "not master" error?

src/ChangeStream.php
```diff
@@ -72,7 +74,16 @@ public function next()
...
             }
         } catch (RuntimeException $e) {
-            $this->resume();
+            $resumable = false;
+            if (strpos($e->getMessage(), "not master") !== false) {
+                $resumable = true;
+            }
+            if ($e->getCode() === self::CURSOR_NOT_FOUND) {
+                $resumable = true;
+            }
```
jmikola: [ConnectionException](http://php.net/manual/en/class.mongodb-driver-exception-connectionexception.php) is a sub-class of RuntimeException, so it would enter this try/catch. Can we add one more check for an instance of that class here to flip on `$resumable`?

[`phongo_exception_from_mongoc_domain()`](https://github.com/mongodb/mongo-php-driver/blob/1.3.3/php_phongo.c#L138) in the driver has all of the logic for converting libmongoc error codes into PHP exception classes.

jmikola:
> Also, how can I test for "not master" and connection errors? Should I just construct a new ChangeStreamCommand and execute it on the secondary server to test for the "not master" error?

Manually selecting a server with a secondary read preference wouldn't do the trick, as change streams can be executed on secondaries. I believe "not master" only comes up when the client communicates with a server as if it is a primary but discovers at the time of the command that it is no longer (i.e. primary is in the process of stepping down but hasn't closed its connections yet). This jives with a line in the [spec paragraph](https://github.com/mongodb/specifications/blob/master/source/change-streams.rst#changestream) talking about the error. Since primary's do close connections on step down, you'd be right to think this is a very narrow window and edge case.

We also needed to test "not master" for retryable writes, and did so using a fail point as described in [Replica Set Failover Test](https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/README.rst#replica-set-failover-test). If you haven't heard of fail points before, they are essentially triggers that we can set on the server (via a [configureFailPoint](https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/README.rst#server-fail-point)) command. Retryable writes utilizes two: forcing a write commit to fail and forcing the primary's step down process to hang and not immediately close connections. In the C driver's test suite, I believe @ajdavis avoided using this fail point by simply mocking the entire server-side of the communication (libmongoc as a server mocking framework, while PHP does not).

If you'd like to take a shot at creating a functional test, you can. It would need to skip unless it's connected to a replica set. Additionally, stepping down requires multiple client objects and additional care to make sure the test doesn't leave the replica set in a bad state (e.g. still waiting for a new primary to be elected). If not, I'm satisfied with the "not master" string check you have. We don't run a replica set on Travis CI, so this test would end up skipped for some time until we can improve the test environment.

----

As for connection exceptions, you might be able to get away with constructing a client with a very low `socketTimeoutMS` option (just for that test case), such that the initial connection handshake and `aggregate` command for the change stream returns but the `getMore` on the tailable cursor times out. I believe the default wait time (without `maxAwaitTimeMS` in effect) is one second. You'd know better than I, but `aggregate` may return the initial tailable cursor immediately. If that's so, then a 500ms socket timeout would cause the driver to drop the connection while awaiting a `getMore` response (during a call to `next()`).

kvwalker:
I added logic and a test for connection exceptions using a low socketTimeoutMS, but the test behaves one of two ways. Sometimes, the connection will drop during the aggregate execution. The rest of the time, the connection will drop during the getMore and the changestream will resume properly. Based on some tests I ran, the first case occurs slightly less than half the time. Is there a reason for this inconsistent behavior?

src/ChangeStream.php
```diff
@@ -67,24 +68,27 @@ public function key()
...
             if (strpos($e->getMessage(), "not master") !== false) {
                 $resumable = true;
             }
             if ($e->getCode() === self::CURSOR_NOT_FOUND) {
-                $resumable = true;
+               $resumable = true;
             }
-            if ($resumable) {
-                $this->resume();
+            if ($e instanceof ConnectionTimeoutException) {
```
jmikola: Per [`phongo_exception_from_mongoc_domain()`](https://github.com/mongodb/mongo-php-driver/blob/5e016c25edca82da94ebe82bdc3faaaf6d062a4f/php_phongo.c#L138), I see that both general socket errors and server selection failures yield a ConnectionTimeoutException. There are a number of other libmongoc errors in the stream family that will result in a ConnectionException, but those all appear to be related to connection initialization or monitoring. Kudos on catching the more specific case here.

tests/Operation/ChangeStreamFunctionalTest.php
```diff
@@ -0,0 +1,252 @@
...
+        $result = $this->collection->insertOne(['x' => 1]);
+        $this->assertInstanceOf('MongoDB\InsertOneResult', $result);
+        $this->assertSame(1, $result->getInsertedCount());
+
+        $changeStreamResult->next();
+    }
+
+    public function testConnectionException()
+    {
+        $client = new Client($this->getUri(), ['socketTimeoutMS' => 1005], []);
```
jmikola: If this sporadically fails, that may be a question for a server engineer. I would expect the `aggregate` command to immediately return either an empty tailable cursor or one with a few results (if you provided a resume token).

1005 might still be higher than needed if the default `maxAwaitTimeMS` is one second.

If there is no middle ground here (and `aggregate` does block), we might have to wait until `maxAwaitTimeMS` is supported in PHPC to test this, as we can then set our await time to 2 seconds and provide 1500 as a socket timeout and hope that `aggregate` always returns within that time.

tests/Operation/ChangeStreamFunctionalTest.php
jmikola: Acknowledged by changes below.

src/ChangeStream.php
```diff
@@ -0,0 +1,125 @@
...
+                $resumable = true;
+            }
+        }
+        if ($resumable) {
+            $this->resume();
+        }
+    }
+
+    public function resume()
```
jmikola: Can this be private?

src/ChangeStream.php
```diff
@@ -0,0 +1,125 @@
...
+        $this->csIt->rewind();
+        $this->key = 0;
+    }
+
+    public function current()
+    {
+        return $this->csIt->current();
+    }
+
+    public function getId()
```
jmikola: I might suggest calling this `getCursorId()`. We can also document that it returns a `MongoDB\Driver\CursorId` instance.

src/ChangeStream.php
```diff
@@ -0,0 +1,125 @@
...
+use MongoDB\Driver\Exception\ConnectionTimeoutException;
+use MongoDB\Driver\Exception\RuntimeException;
+use MongoDB\Exception\ResumeTokenException;
+use IteratorIterator;
+use Iterator;
+
+/**
+ * Iterator for the changeStream command.
+ *
+ * @api
```
jmikola: Since this class will be public API, can you go through an add PHP doc blocks to the public methods?

For the five Iterator methods (e.g. `current()`, `key()`), you can copy doc blocks from CachingIterator (with the `@see` references to PHP.net documentation).

src/Operation/ChangeStreamCommand.php
```diff
@@ -0,0 +1,195 @@
...
+     *
+     *  * collation (document): Specifies a collation.
+     *
+     *    This option is sent only if the caller explicitly provides a value.
+     *    The default is to not send a value.
+     *
+     * @param string         $databaseName   Database name
+     * @param string         $collectionName Collection name
+     * @param array          $pipeline       List of pipeline operations
+     * @param array          $options        Command options
```
jmikola: `@param Manager $manager Manager instance from the driver` can be added after `$options`.

src/Operation/ChangeStreamCommand.php
```diff
@@ -0,0 +1,195 @@
...
+use MongoDB\Driver\Exception\RuntimeException as DriverRuntimeException;
+use MongoDB\Exception\InvalidArgumentException;
+use MongoDB\Exception\UnexpectedValueException;
+use MongoDB\Exception\UnsupportedException;
+use ArrayIterator;
+use stdClass;
+use Traversable;
+
+/**
+ * Raw changeStream command.
```
jmikola: "Operation for creating a change stream with the aggregate command."

src/Operation/ChangeStreamCommand.php
```diff
@@ -0,0 +1,195 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace MongoDB\Operation;
+
+use MongoDB\ChangeStream;
```
jmikola: In the interest of being able to use "ChangeStream" for both classes, can we import this class as ChangeStreamResult and refer to it as such below?

src/Operation/ChangeStreamCommand.php
```diff
@@ -0,0 +1,195 @@
...
+     * @throws UnsupportedException if collation, read concern, or write concern is used and unsupported
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function execute(Server $server)
+    {
+        $command = $this->createCommand();
+
+        $cursor = $command->execute($server);
+
+        return new ChangeStream($cursor, $this->createResumeCallable());
```
jmikola: This could become `new ChangeStreamResult(...)` after the `use / as` update. I believe that's the only reference in this file.

src/Operation/ChangeStreamCommand.php
```diff
@@ -0,0 +1,195 @@
...
+        $this->options = $options;
+        $this->manager = $manager;
+    }
+
+    /**
+     * Execute the operation.
+     *
+     * @see Executable::execute()
+     * @param Server $server
+     * @return Traversable
```
jmikola: This can also be updated to `ChangeStreamResult` with the `use / as` in place.

tests/Operation/ChangeStreamFunctionalTest.php
```diff
@@ -0,0 +1,252 @@
...
+use MongoDB\Client;
+use MongoDB\Collection;
+use MongoDB\Driver\BulkWrite;
+use MongoDB\Driver\ReadPreference;
+use MongoDB\Exception\ResumeTokenException;
+use MongoDB\Operation\Aggregate;
+use MongoDB\Operation\ChangeStreamCommand;
+use MongoDB\Operation\DatabaseCommand;
+use MongoDB\Operation\InsertOne;
+use MongoDB\Tests\CommandObserver;
```
jmikola: A bunch of these appear to no longer be used in this file:

 * ChangeStreamIterator
 * BulkWrite
 * ReadPreference
 * ResumeTokenException (not required for `@expectedException`)
 * Aggregate
 * InsertOne
 * IteratorIterator
 * stdClass

src/Operation/ChangeStreamCommand.php
```diff
@@ -0,0 +1,195 @@
...
+use MongoDB\Driver\ReadConcern;
+use MongoDB\Driver\ReadPreference;
+use MongoDB\Driver\Server;
+use MongoDB\Driver\Exception\RuntimeException as DriverRuntimeException;
+use MongoDB\Exception\InvalidArgumentException;
+use MongoDB\Exception\UnexpectedValueException;
+use MongoDB\Exception\UnsupportedException;
+use ArrayIterator;
+use stdClass;
+use Traversable;
```
jmikola: ArrayIterator and stdClass don't appear to be used anymore. Traversable may not be used if you update the `execute()` return annotation to `ChangeStreamResult`.

src/Collection.php
```diff
@@ -936,6 +937,31 @@ public function updateOne($filter, $update, array $options = [])
...
+            $options['readPreference'] = $this->readPreference;
+        }
+
+        $server = $this->manager->selectServer($options['readPreference']);
+
+        if ( ! isset($options['readConcern'])) {
+            $options['readConcern'] = $this->readConcern;
+        }
+
+        $operation = new ChangeStreamCommand($this->databaseName, $this->collectionName, $pipeline, $options, $this->manager);
```
jmikola: If you rename the Operation class to "ChangeStream", this can be updated.

src/Collection.php
```diff
@@ -936,6 +937,31 @@ public function updateOne($filter, $update, array $options = [])
         return $operation->execute($server);
     }
 
+    /*
+     * ChangeStream outline
+     *
+     * @see ChangeStreamCommand::__construct() for supported options
+     * @param array          $pipeline       List of pipeline operations
+     * @param array          $options        Command options
+     * @throws InvalidArgumentException for parameter/option parsing errors
```
jmikola: Missing a `@return` annotation here. I might also suggest `use MongoDB\ChangeStream as ChangeStreamResult` up top and then using `@return ChangeStreamResult` here, as in the Operation class.

src/ChangeStream.php
```diff
@@ -40,6 +40,10 @@ class ChangeStream implements Iterator
 
     const CURSOR_NOT_FOUND = 43;
 
+    /**
+     * @param MongoDB\Driver\Cursor $cursor
```
jmikola: No need to change this, but just a note: Doc block comments _can_ take advantage of the shorter class names (from `use` imports). It's just our rendered RST documentation that needs the FQCNs.

src/ChangeStream.php
```diff
@@ -48,16 +52,27 @@ public function __construct(Cursor $cursor, callable $resumeCallable)
...
+     * @return mixed
+     */
     public function current()
     {
         return $this->csIt->current();
     }
 
-    public function getId()
+    /**
+     * @return MongoDB\Driver\CursorId
```
jmikola: Related to my previous comment, this FQCN would be needed since CursorId isn't imported via a `use` statement.

src/ChangeStream.php
```diff
@@ -48,16 +52,27 @@ public function __construct(Cursor $cursor, callable $resumeCallable)
...
+     */
+    public function getCursorId()
     {
         return $this->csIt->getInnerIterator()->getId();
     }
 
+    /**
+     * @see http://php.net/iterator.mixed
```
jmikola: `iterator.key`. Looks like I made that typo in CachingIterator :smiley: 

Fixed in https://github.com/mongodb/mongo-php-library/commit/ba01a8cca36b084516ba82f30a3bcb378057f9e0.

PR #429, opened 2017-11-06
Use PSR-1 for PHPUnit TestCase

gabriel-caruso:
Use [PSR-1](http://www.php-fig.org/psr/psr-1/#namespace-and-class-names) while extending PHPUnit TestCase class. This will help us when to migrate to PHPUnit 6, that [no longer support snake case namespaces](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#changed-1).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         "ext-mongodb": "^1.3.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "^4.8"
+        "phpunit/phpunit": "^4.8.36"
     },
     "autoload": {
         "psr-4": { "MongoDB\\": "src/" },
--- a/tests/Model/CachingIteratorTest.php
+++ b/tests/Model/CachingIteratorTest.php
@@ -4,8 +4,9 @@ namespace MongoDB\Tests\Model;
 
 use MongoDB\Model\CachingIterator;
 use Exception;
+use PHPUnit\Framework\TestCase;
 
-class CachingIteratorTest extends \PHPUnit_Framework_TestCase
+class CachingIteratorTest extends TestCase
 {
     /**
      * Sanity check for all following tests.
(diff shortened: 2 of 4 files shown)


alcaeus:
This needs an update of the `phpunit` requirement in `composer.json`. Not all versions that satisfy `^4.8` include the forward compatibility layer. I suggest using `^4.8.36|^5.7|^6.4` instead of `^4.8`.

gabriel-caruso:
@alcaeus I'll update to `^4.8.36` only. PHPUnit `5` has some deprecated methods, and `6` uses PHP 7, and we will need to work on that later.

jmikola:
Manually merged in https://github.com/mongodb/mongo-php-library/commit/d4583a9baaf9631cdcdc1c5900741b27e11efa33. Thanks!

@alcaeus: I noticed that there were two outstanding issues with allowing `^5.7|^6.4` in `composer.json`:

 * We currently have a number of tests that perform no assertions (e.g. they construct an object to ensure that no exceptions are thrown for invalid arguments). PHPUnit now warns against such [risky tests](https://phpunit.de/manual/current/en/risky-tests.html) unless we disable the warning in our configuration.
 * We have several tests that expect a `PHPUnit_Framework_Error_Warning` exception (as part of their assertion for an internal PHP warning). The PHPUnit 6 compatibility layer added in 4.8.36 and 5.7.21 did not include namespaced aliases for the exception classes, so we can't generalize those tests for all three PHPUnit verisons.

For the time being, we'll need to remain on 4.8.36.

alcaeus:
> PHPUnit now warns against such risky tests unless we disable the warning in our configuration.

For individual tests, there is a `doesNotPerformAssertions` annotation you can add to the tests in question. As for the warnings, that's indeed an issue that can't be fixed - only thing you could do is adding your own compatibility layer using `class_alias`.

jmikola:
Sounds good. I've opened [PHPLIB-294](https://jira.mongodb.org/browse/PHPLIB-294) to track that.

PR #431, opened 2017-11-10
PHPLIB-254 Support maxTimeMS option for CreateIndexes operation

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-254

### DIFF
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -282,8 +282,8 @@ class Collection
      */
     public function createIndex($key, array $options = [])
     {
-        $indexOptions = array_diff_key($options, ['writeConcern' => 1]);
-        $commandOptions = array_intersect_key($options, ['writeConcern' => 1]);
+        $indexOptions = array_diff_key($options, ['maxTimeMS' => 1, 'writeConcern' => 1]);
+        $commandOptions = array_intersect_key($options, ['maxTimeMS' => 1, 'writeConcern' => 1]);
 
         return current($this->createIndexes([['key' => $key] + $indexOptions], $commandOptions));
     }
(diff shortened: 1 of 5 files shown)


src/Collection.php
```diff
@@ -282,8 +282,8 @@ public function count($filter = [], array $options = [])
      */
     public function createIndex($key, array $options = [])
     {
-        $indexOptions = array_diff_key($options, ['writeConcern' => 1]);
-        $commandOptions = array_intersect_key($options, ['writeConcern' => 1]);
+        $indexOptions = array_diff_key($options, ['maxTimeMS' => 1, 'writeConcern' => 1]);
+        $commandOptions = array_intersect_key($options, ['maxTimeMS' => 1, 'writeConcern' => 1]);
```
derickr: In #432 you've also added a test case to test for invalid values for the new option. Can you add one for this new option to ``provideInvalidConstructorOptions`` in ``CreateIndexesTest.php`` too?

kvwalker: Done, thanks!

jmikola:
Can you update docs in the RST manual and PHP doc blocks for CreateIndexes operation class when you get a chance? Same as was requested in #432.

For the RST manual, `maxTimeMS` is already defined in `apiargs-common-option.yaml` so you should just need to add the following to `apiargs-MongoDBCollection-method-createIndex-option.yaml` and `apiargs-MongoDBCollection-method-createIndexes-option.yaml`.

```
source:
  file: apiargs-common-option.yaml
  ref: maxTimeMS
```

Note that `Collection::createIndexes()` didn't need any code changes here as it already takes command options separately from any index options (no splitting is necessary).

PR #423, opened 2017-10-25
PHPLIB-284 Highlight API changes in documentation

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-284

PR #432, opened 2017-11-10
PHPLIB-293 Add support for a comment parameter to the aggregate command

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-293

### DIFF
--- a/docs/includes/apiargs-MongoDBCollection-method-aggregate-option.yaml
+++ b/docs/includes/apiargs-MongoDBCollection-method-aggregate-option.yaml
@@ -35,6 +35,18 @@ post: |
   version of MongoDB, this option will be ignored.
 ---
 arg_name: option
+name: comment
+type: string
+description: |
+  Users can specify an arbitrary string to help trace the operation through the
+  database profiler, currentOp, and logs.
+
+  .. versionadded:: 1.3
+interface: phpmethod
+operation: ~
+optional: true
+---
+arg_name: option
 name: hint
 type: string|array|object
 description: |
--- a/src/Operation/Aggregate.php
+++ b/src/Operation/Aggregate.php
@@ -74,6 +74,9 @@ class Aggregate implements Executable
      *    This is not supported for server versions < 3.4 and will result in an
      *    exception at execution time if used.
      *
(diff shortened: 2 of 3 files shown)


src/Operation/Aggregate.php
```diff
@@ -276,6 +280,10 @@ private function createCommand(Server $server, $isCursorSupported)
             $cmd['bypassDocumentValidation'] = $this->options['bypassDocumentValidation'];
         }
 
+        if (isset($this->options['comment'])) {
+            $cmd['comment'] = $this->options['comment'];
+        }
+
```
jmikola: Since `comment` and `maxTimeMS` are the same in the sense that we want to check if they're `isset()` and then copy them as-is into the `$cmd` document, I propose creating a `foreach` as we have in other command helpers:

```php
foreach (['comment', 'maxTimeMS'] as $option) {
    if (isset($this->options[$option])) {
        $cmd[$option] = $this->options[$option];
    }
}
```

`allowDiskUse` is always set, so we can leave its earlier assignment as-is.

kvwalker:
I added the `comment` option to the RST documentation, but what are the PHP docs preceding the Aggregate class?

jmikola:
> what are the PHP docs preceding the Aggregate class?

I was referring to the doc block before `class Aggregate`: https://github.com/mongodb/mongo-php-library/blob/750838012ee1971ddafd26aff05f2228954343f0/src/Operation/Aggregate.php#L57. Sorry if that was unclear.

docs/includes/apiargs-MongoDBCollection-method-aggregate-option.yaml
```diff
@@ -35,6 +35,18 @@ post: |
   version of MongoDB, this option will be ignored.
 ---
 arg_name: option
+name: comment
+type: string
+description: |
+  Users can specify an arbitrary string to help trace the operation through the
+  database profiler, currentOp, and logs.
+
+  .. versionadded:: 1.3
```
jmikola: Good catch! I just came across #423 while looking through closed PRs and was going to pop in to ask about this, but you beat me to it :+1:

jmikola:
Good to squash and merge :shipit:

PR #436, opened 2017-11-20
[WAIT] PHPLIB-289 ChangeStreams Examples for Docs

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-289

### DIFF
--- a/tests/DocumentationExamplesTest.php
+++ b/tests/DocumentationExamplesTest.php
@@ -925,6 +925,29 @@ class DocumentationExamplesTest extends FunctionalTestCase
         $this->assertInventoryCount(0);
     }
 
+    public function testExample_59_61()
+    {
+        $db = new Database($this->manager, $this->getDatabaseName());
+
+        // Start Changestream Example 1
+        $cursor = $db->inventory->watch();
+        $cursor->next();
+        $next = $cursor->current();
+        // End Changestream Example 1
+
+        // Start Changestream Example 2
+        $cursor = $db->inventory->watch([], ['fullDocument' => ChangeStreamCommand::FULL_DOCUMENT_UPDATE_LOOKUP]);
+        $cursor->next();
+        $next = $cursor->current();
+        // End Changestream Example 2
+
+        // Start Changestream Example 3
+        $resumeToken = $next->getId();
+        $cursor = $db->inventory->watch([], ['resumeAfter' => $resumeToken]);
+        $cursor->next();
+        // End Changestream Example 3
+    }
+
     /**
      * Return the test collection name.
      *
--- a/tests/DocumentationExamplesTest.php
+++ b/tests/DocumentationExamplesTest.php
@@ -932,23 +932,29 @@ class DocumentationExamplesTest extends FunctionalTestCase
         // Start Changestream Example 1
         $cursor = $db->inventory->watch();
         $cursor->next();
-        $next = $cursor->current();
+        $current = $cursor->current();
         // End Changestream Example 1
 
-        $this->assertNull($next);
+        $this->assertNull($current);
 
         // Start Changestream Example 2
         $cursor = $db->inventory->watch([], ['fullDocument' => \MongoDB\Operation\ChangeStreamCommand::FULL_DOCUMENT_UPDATE_LOOKUP]);
         $cursor->next();
-        $next = $cursor->current();
+        $current = $cursor->current();
         // End Changestream Example 2
 
-        $this->assertNull($next);
+        $this->assertNull($current);
+        $db->inventory->insertOne(['x' => '1']);
+        $cursor->next();
+        $current = $cursor->current();
 
         // Start Changestream Example 3
-        $resumeToken = $next->getId();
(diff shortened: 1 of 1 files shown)


tests/DocumentationExamplesTest.php
```diff
@@ -925,6 +925,29 @@ public function testExample_55_58()
...
+    public function testExample_59_61()
+    {
+        $db = new Database($this->manager, $this->getDatabaseName());
+
+        // Start Changestream Example 1
+        $cursor = $db->inventory->watch();
+        $cursor->next();
+        $next = $cursor->current();
+        // End Changestream Example 1
+
```
jmikola: In between the example blocks, we can do assertions just to test that the examples perform as expected.

I realize all of these examples do nothing between `watch()` and iterating, so asserting empty results seems sufficient.

tests/DocumentationExamplesTest.php
```diff
@@ -925,6 +925,35 @@ public function testExample_55_58()
...
     }
 
+    public function testExample_59_61()
+    {
+        $db = new Database($this->manager, $this->getDatabaseName());
+
+        // Start Changestream Example 1
+        $cursor = $db->inventory->watch();
+        $cursor->next();
+        $next = $cursor->current();
```
jmikola: Let's use `$current` for this variable.

tests/DocumentationExamplesTest.php
```diff
@@ -925,6 +925,35 @@ public function testExample_55_58()
...
+        // End Changestream Example 2
+
+        $this->assertNull($next);
+
+        // Start Changestream Example 3
+        $resumeToken = $next->getId();
+        $cursor = $db->inventory->watch([], ['resumeAfter' => $resumeToken]);
+        $cursor->next();
+        // End Changestream Example 3
```
jmikola: I think this example should be more defensive:

```php
$resumeToken = ($current !== null) ? $current['_id'] : null;

if ($resumeToken !== null) {
    $cursor = $db->inventory->watch([], ['resumeAfter' => $resumeToken]);
    $current = $cursor->next();
}
```

It would be worth following up on [DRIVERS-414](https://jira.mongodb.org/browse/DRIVERS-414) to confirm.

tests/DocumentationExamplesTest.php
```diff
@@ -925,6 +925,41 @@ public function testExample_55_58()
...
+        $this->assertNull($current);
+
+        // Start Changestream Example 2
+        $cursor = $db->inventory->watch([], ['fullDocument' => \MongoDB\Operation\ChangeStreamCommand::FULL_DOCUMENT_UPDATE_LOOKUP]);
+        $cursor->next();
+        $current = $cursor->current();
+        // End Changestream Example 2
+
+        $this->assertNull($current);
+        $db->inventory->insertOne(['x' => '1']);
```
jmikola: While it makes no functional difference, `'1'` can probably be an integer literal. If you do want to use a string, something like `'foo'` would be more consistent.

tests/DocumentationExamplesTest.php
```diff
@@ -925,6 +925,41 @@ public function testExample_55_58()
...
+        // Start Changestream Example 3
+        $resumeToken = ($current !== null) ? $current->_id : null;
+        if ($resumeToken !== null) {
+            $cursor = $db->inventory->watch([], ['resumeAfter' => $resumeToken]);
+            $cursor->next();
+        }
+
+        // End Changestream Example 3
+
+        $this->assertNull($cursor->current());
```
jmikola: One more thought. Let's `$db->inventory->insertOne(['x' => 2]);` here, so we can assert a non-null change entry. I'm not sure if you're doing so in #428 yet, but if you know the structure of the change entry you can use a more specific assertion. We should have helpers in the base test classes that can be used to assert an equivalent document structure.

tests/DocumentationExamplesTest.php
```diff
@@ -925,6 +925,41 @@ public function testExample_55_58()
...
+        // Start Changestream Example 2
+        $cursor = $db->inventory->watch([], ['fullDocument' => \MongoDB\Operation\ChangeStreamCommand::FULL_DOCUMENT_UPDATE_LOOKUP]);
+        $cursor->next();
+        $current = $cursor->current();
+        // End Changestream Example 2
+
+        $this->assertNull($current);
+        $db->inventory->insertOne(['x' => '1']);
+        $cursor->next();
+        $current = $cursor->current();
```
jmikola: We can probably assert the change entry here as well. It will make not difference for example 3, since it's written defensively, but it makes the test itself more useful.

tests/DocumentationExamplesTest.php
```diff
@@ -925,6 +925,41 @@ public function testExample_55_58()
         $this->assertInventoryCount(0);
     }
 
+    public function testExample_59_61()
```
jmikola: In the event we get more generic examples later, we should probably call this `testChanceStreamExample_1_3()`.

jmikola:
LGTM. Feel free to rebase and merge!

PR #438, opened 2017-11-27
Refactoring some tests

gabriel-caruso:
I've refactored some tests with [`assertNull`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertFalse) and [`assertEmpty`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertEmpty) methods.

### DIFF
--- a/tests/GridFS/StreamWrapperFunctionalTest.php
+++ b/tests/GridFS/StreamWrapperFunctionalTest.php
@@ -47,7 +47,7 @@ class StreamWrapperFunctionalTest extends FunctionalTestCase
 
         $this->assertSame('abc', fread($stream, 3));
         $this->assertSame('defghij', fread($stream, 10));
-        $this->assertSame('', fread($stream, 3));
+        $this->assertNull(fread($stream, 3));
     }
 
     public function testReadableStreamSeek()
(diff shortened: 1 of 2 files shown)


tests/GridFS/StreamWrapperFunctionalTest.php
```diff
@@ -47,7 +47,7 @@ public function testReadableStreamRead()
 
         $this->assertSame('abc', fread($stream, 3));
         $this->assertSame('defghij', fread($stream, 10));
-        $this->assertSame('', fread($stream, 3));
+        $this->assertNull(fread($stream, 3));
```
jmikola: Does this actually pass for you?

Running this on my end, I see `Failed asserting that '' is null.` [`fread()`](http://php.net/fread) is documented as returning either a string or `false` on failure, but never `null`.

tests/Model/CollectionInfoTest.php
```diff
@@ -16,7 +16,7 @@ public function testGetName()
     public function testGetOptions()
     {
         $info = new CollectionInfo(['name' => 'foo']);
-        $this->assertSame([], $info->getOptions());
+        $this->assertEmpty($info->getOptions());
```
jmikola: This assertion is less strict than `assertSame()`, as [`IsEmpty::matches()`](https://github.com/sebastianbergmann/phpunit/blob/485e7f707692b81c360e713c280601ad20de5392/src/Framework/Constraint/IsEmpty.php#L27) may defer to Countable or PHP's `empty()` construct. The original assertion very specifically tests that we get an empty array, which is consistent with `CollectionInfo::getOptions()` documenting that it returns an array type at all times.

PR #439, opened 2017-11-27
Add ChangeStream documentation

kvwalker:


docs/includes/apiargs-MongoDBCollection-method-watch-option.yaml
```diff
@@ -0,0 +1,36 @@
...
+---
+arg_name: option
+name: fullDocument
+type: string
+description: |
+  Allowed values are 'default' and 'updateLookup'. Defaults to 'default'.
+  When set to 'updateLookup', the change notification for partial updates will
+  include both a delta describing the changes to the document, as well as a
+  copy of the entire document that was changed from some time after the change
+  occurred.
```
jmikola: We can add also a reference to the constants here, as was done for [`findOneAndUpdate()`](https://github.com/mongodb/mongo-php-library/blob/master/docs/includes/apiargs-MongoDBCollection-method-findOneAndUpdate-option.yaml#L24).

Unlike `findOneAndUpdate()`, I think we can leave this paragraph as is (without adding lengthy references to constants within) and just add the list to the bottom:

```
...after the change occurred. The following values are supported:

  - ``MongoDB\Operation\ChangeStream::FULL_DOCUMENT_DEFAULT`` (*default*)
  - ``MongoDB\Operation\ChangeStream::FULL_DOCUMENT_UPDATE_LOOKUP``
```

docs/includes/apiargs-MongoDBCollection-method-watch-param.yaml
```diff
@@ -0,0 +1,28 @@
+arg_name: param
+name: pipeline
+type: array|object
+description: |
+  The pipeline of stages to append to an initial ``$changeStream`` stage.
+interface: phpmethod
+operation: ~
+optional: true
+---
+arg_name: param
```
jmikola: The Collection method signature is `watch(array $pipeline = [], array $options = [])`. I think `readPreference` should be specified in `option.yaml` and `resumeToken` below can be removed. You can copy another `param.yml` file for boilerplate to use for the array options argument.

docs/reference/method/MongoDBCollection-watch.txt
```diff
@@ -0,0 +1,50 @@
...
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Definition
+----------
+
+.. phpmethod:: MongoDB\\Collection::watch()
+
+   Executes a :manual:`change stream </changeStreams>` operation on the
```
jmikola: This will render as https://docs.mongodb.com/manual/changeStreams, which is currently a 404 but should point to the content at https://docs.mongodb.com/master/changeStreams/ once 3.6 is the default version for the server docs. Feel free to confirm with @kay-kim if you haven't already. Otherwise, this LGTM.

docs/reference/method/MongoDBCollection-watch.txt
```diff
@@ -0,0 +1,50 @@
...
+   collection.
+
+   .. code-block:: php
+
+      function watch(array $pipeline = [], array $options = []): ChangeStream
```
jmikola: The return type here should be a FQCN (fully qualified class name).

It's not required for PHP doc blocks, since those can capitalize on the `use` statements in the same file.

docs/reference/method/MongoDBCollection-watch.txt
```diff
@@ -0,0 +1,50 @@
...
+
+.. include:: /includes/extracts/error-unexpectedvalueexception.rst
+.. include:: /includes/extracts/error-unsupportedexception.rst
+.. include:: /includes/extracts/error-invalidargumentexception.rst
+.. include:: /includes/extracts/error-driver-runtimeexception.rst
+
+See Also
+--------
+
+- :manual:`Aggregation Pipeline </core/aggregation-pipeline>` documentation in
```
jmikola: We can add one more link to:

```
- :manual:`Change Streams </changeStreams>` documentation in the MongoDB manual
```

docs/reference/method/MongoDBCollection-watch.txt
```diff
@@ -0,0 +1,50 @@
...
+   .. include:: /includes/apiargs/MongoDBCollection-method-watch-param.rst
+
+   The ``$options`` parameter supports the following options:
+
+   .. include:: /includes/apiargs/MongoDBCollection-method-watch-option.rst
+
+Return Values
+-------------
+
+A ChangeStream object.
```
jmikola: Adapting from the `mapReduce()` documentation:

```
A :phpclass:`MongoDB\\ChangeStream` object, which allows for iteration of
events in the change stream via the :php:`Iterator <class.iterator>` interface.
```

docs/includes/apiargs-MongoDBCollection-method-watch-option.yaml
```diff
@@ -24,6 +27,16 @@ source:
   ref: maxAwaitTimeMS
 ---
 arg_name: option
+name: readPreference
```
jmikola: We can repeat what is done for `aggregate()`'s `readPreference` option and then use `post`  to add an extra paragraph after the common content:

```
---
source:
  file: apiargs-MongoDBCollection-common-option.yaml
  ref: readPreference
post: |
  This is used for both the initial change stream aggregation and for
  server selection during an automatic resume.
---
```

Sorry if I missed this in an earlier review.

docs/includes/apiargs-MongoDBCollection-method-watch-param.yaml
```diff
@@ -1,28 +1,13 @@
 arg_name: param
-name: pipeline
+name: $pipeline
```
jmikola: Good catch. I missed this earlier :+1:

PR #440, opened 2017-11-27
PHPLIB-295: Support arrayFilters option for update and findAndModify

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-295


src/Operation/FindOneAndUpdate.php
```diff
@@ -34,13 +34,21 @@ class FindOneAndUpdate implements Executable
...
 
     /**
      * Constructs a findAndModify command for updating a document.
      *
      * Supported options:
      *
+     *  * arrayFilters (document array): A set of filters specifying to which
+     *    array elements an update should apply.
+     *
+     *    This is not supported for server versions < 3.6 and will result in an$
```
derickr: there is an extra ``$`` at the end of the line.

docs/includes/apiargs-MongoDBCollection-method-updateOne-option.yaml
```diff
@@ -2,6 +2,18 @@ source:
   file: apiargs-MongoDBCollection-common-option.yaml
   ref: upsert
 ---
+arg_name: option
```
jmikola: Can we move this definition to `apiargs-MongoDBCollection-common-option.yaml`, like `upsert`?

src/Operation/FindOneAndUpdate.php
```diff
@@ -103,6 +111,10 @@ public function __construct($databaseName, $collectionName, $filter, $update, ar
             'upsert' => false,
         ];
 
+        if (isset($options['arrayFilters']) && ! is_array($options['arrayFilters'])) {
+            throw InvalidArgumentException::invalidType('"arrayFilters" option', $options['arrayFilters'], 'array');
+        }
```
jmikola: The end of `FindOneAndUpdate::__construct()` passes its `$options` argument directly on to `FindAndModify`, so we really need to support the `arrayFilters` option in `FindAndModify`'s constructor. It should be documented in both classes.

Technically, that means that `FindOneAndReplace` and `FindOneAndDelete` can also receive the option, but it won't be documented for those and the server should raise an error anyway. We can also rely on [PHPLIB-124](https://jira.mongodb.org/browse/PHPLIB-124) catching that in the future.

src/Operation/FindOneAndUpdate.php
```diff
@@ -137,11 +149,14 @@ public function __construct($databaseName, $collectionName, $filter, $update, ar
...
      * @return array|object|null
-     * @throws UnsupportedException if collation or write concern is used and unsupported
+     * @throws UnsupportedException if array filters or collation or write concern is used and unsupported
      * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
      */
     public function execute(Server $server)
     {
+        if (isset($this->options['arrayFilters']) && ! \MongoDB\server_supports_feature($server, self::$wireVersionForArrayFilters)) {
+            throw UnsupportedException::arrayFiltersNotSupported();
+        }
```
jmikola: If the above constructor comment is addressed, we can move this check into `FindAndModify::execute()`.

src/Operation/BulkWrite.php
```diff
@@ -282,17 +297,25 @@ public function __construct($databaseName, $collectionName, array $operations, a
...
+            throw UnsupportedException::arrayFiltersNotSupported();
+        }
+
         if ($this->isCollationUsed && ! \MongoDB\server_supports_feature($server, self::$wireVersionForCollation)) {
             throw UnsupportedException::collationNotSupported();
         }
 
         $options = ['ordered' => $this->options['ordered']];
 
+        if (isset($this->options['arrayFilters'])) {
```
jmikola: This looks like dead code, as "arrayFilters" isn't a BulkWrite option but rather appears within args for individual statements in the `$operations` constructor argument.

src/Operation/FindOneAndUpdate.php
```diff
@@ -149,14 +143,11 @@ public function __construct($databaseName, $collectionName, $filter, $update, ar
      * @see Executable::execute()
      * @param Server $server
      * @return array|object|null
-     * @throws UnsupportedException if array filters or collation or write concern is used and unsupported
+     * @throws UnsupportedException if collation or write concern is used and unsupported
```
jmikola: I think this can still mention array filters, since it throws anything that `FindAndModify::execute()` throws. We can say "array filters, collation, or write concern..." to condense the line.

src/Operation/FindAndModify.php
```diff
@@ -49,6 +50,9 @@ class FindAndModify implements Executable
      *
      * Supported options:
      *
+     *  * arrayFilters (document array): A set of filters specifying to which
+     *    array elements an update should apply.
```
jmikola: This comment should be consistent with what we have in `FindOneAndUpdate`. I think it's just missing the "not supported" paragraph.

src/Operation/FindAndModify.php
```diff
@@ -173,11 +181,15 @@ public function __construct($databaseName, $collectionName, array $options)
      * @param Server $server
      * @return array|object|null
      * @throws UnexpectedValueException if the command response was malformed
-     * @throws UnsupportedException if collation or write concern is used and unsupported
+     * @throws UnsupportedException if array filters or collation or write concern is used and unsupported
```
jmikola: Can condense as "array filters, collation, or write concern"

tests/Operation/FindOneAndUpdateTest.php
```diff
@@ -46,6 +46,10 @@ public function provideInvalidConstructorOptions()
     {
         $options = [];
 
+        foreach ($this->getInvalidArrayValues() as $value) {
+            $options[][] = ['arrayFilters' => $value];
+        }
+
```
jmikola: This can actually go in FindAndModifyTest. Similar to how you added the equivalent test to UpdateTest instead of UpdateOneTest and UpdateMany Test.

tests/Operation/BulkWriteTest.php
```diff
@@ -373,6 +385,18 @@ public function testUpdateOneUpdateArgumentRequiresOperators()
         ]);
     }
 
+    /**
+     * @expectedException MongoDB\Exception\InvalidArgumentException
+     * @expectedExceptionMessageRegExp /Expected \$operations\[0\]\["updateOne"\]\[2\]\["arrayFilters"\] to have type "array" but found "[\w ]+"/
+     * @dataProvider provideInvalidDocumentValues
```
jmikola: Since arrays and objects are both document values, this omits testing that `new stdClass` yields an exception. I might suggest creating a `provideInvalidArrayValues()` method in `MongoDB\Tests\TestCase` that is based on `provideInvalidDocumentValues` so that it wraps the return value of `getInvalidArrayValues()`.

tests/Operation/BulkWriteTest.php
```diff
@@ -292,6 +292,18 @@ public function testUpdateManyUpdateArgumentRequiresOperators()
         ]);
     }
 
+    /**
+     * @expectedException MongoDB\Exception\InvalidArgumentException
+     * @expectedExceptionMessageRegExp /Expected \$operations\[0\]\["updateMany"\]\[2\]\["arrayFilters"\] to have type "array" but found "[\w ]+"/
+     * @dataProvider provideInvalidDocumentValues
```
jmikola: Same here. Test against invalid array values.

tests/Collection/CrudSpecFunctionalTest.php
```diff
@@ -334,6 +370,50 @@ private function initializeData(array $initialData, array $expectedData = null)
...
     }
 
+    private function prepareBulkWriteArguments($arguments)
+    {
+        $operations = [];
+        $operations['options'] = $arguments['options'];
+        foreach ($arguments['requests'] as $requests) {
+            if (isset($requests['arguments']['arrayFilters'])) {
+                if (isset($requests['arguments']['options'])) {
+                    array_push($requests['arguments']['options'], ['arrayFilters' => $requests['arguments']['arrayFilters']]);
```
jmikola: [`array_push()`](http://php.net/array_push) modifies its first argument by reference, but in this case, `$requests` is always a copy of the current element in `$arguments['requests']`, so your modification is never making it back to the `$arguments` value when you use it later in the function.

Two options:

 * Use a reference when iterating (e.g. `foreach ($values as &$value)`) so that any modifications to `$value` do modify the actual element in `$values`.
 * Track the array key while iterating (e.g. `foreach ($values as $i => $value)`) and refer to `$arguments['requests'][$i]` when you want to modify the current element within the loop.

This relates a bit to the zval behaviors discussed in the first half of [Memory Management](http://www.phpinternalsbook.com/zvals/memory_management.html) if you'd like to read more on the subject.

tests/Collection/CrudSpecFunctionalTest.php
```diff
@@ -334,6 +370,50 @@ private function initializeData(array $initialData, array $expectedData = null)
...
+                } else {
+                    $options = ['arrayFilters' => $requests['arguments']['arrayFilters']];
```
jmikola: Looking at this some more, I think this is a bug. You're iterating through potentially many request elements, but this is always assigning an `$options` variable in the method's local scope. This means it's possible for one request's `arrayFilters` option to clobber `$options` from a previous request. Moreover, later in the method you have no way of knowing the exact request element to which the final value of `$options` relates.

It looks as if `$options` is only referenced from the `updateMany` and `updateOne` cases, so perhaps you just need to `unset()` it at the start of each iteration.

tests/Collection/CrudSpecFunctionalTest.php
```diff
@@ -334,6 +370,50 @@ private function initializeData(array $initialData, array $expectedData = null)
...
+            }
+            $innerArray = [];
+            switch ($requests['name']) {
+                case 'deleteMany':
+                case 'deleteOne':
+                    $innerArray = [$requests['arguments']['filter']];
+                    if (isset($requests['arguments']['options'])) {
+                         array_push($innerArray, $requests['arguments']['options']);
+                    }
+                case 'insertOne':
```
jmikola: The absence of any `break` statements here means that each `case` falls through to the next. That's likely what you want for `deleteMany` and `deleteOne` but not for truly unrelated operations. Otherwise, you're going to clobber the `$inneryArray` value by the end of this `switch` block.

tests/Collection/CrudSpecFunctionalTest.php
```diff
@@ -334,6 +370,50 @@ private function initializeData(array $initialData, array $expectedData = null)
...
+                    if (isset($options)) {
+                        if (isset($requests['arguments']['options'])) {
+                            array_push($requests['arguments']['options'], $options);
+                        } else {
+                            $requests['arguments']['options'] = $options;
+                        }
+                    }
+                    $innerArray = [$requests['arguments']['filter'], $requests['arguments']['update'], $requests['arguments']['options']];
+           }
+            $operations[] = [$requests['name'] => $innerArray];
```
jmikola: I noticed a few lines in this method had some indentation issues. Please double-check later.

tests/Collection/CrudSpecFunctionalTest.php
```diff
@@ -334,6 +373,36 @@ private function initializeData(array $initialData, array $expectedData = null)
...
+                    break;
+                case 'replaceOne':
+                    $options = array_diff_key($requests['arguments'], ['filter' => 1, 'replacement' => 1]);
+                    $innerArray = [$requests['arguments']['filter'], $requests['arguments']['replacement'], $options];
+                break;
+                case 'updateMany':
+                case 'updateOne':
+                    $options = array_diff_key($requests['arguments'], ['filter' => 1, 'update' => 1]);
+                    $innerArray = [$requests['arguments']['filter'], $requests['arguments']['update'], $options];
+                break;
```
jmikola: Can we add a default case to raise an error if we see an unexpected request type?

```
default:
    throw new LogicException('Unsupported bulk write request: ' . $request['name']);
```

tests/Collection/CrudSpecFunctionalTest.php
```diff
@@ -334,6 +373,36 @@ private function initializeData(array $initialData, array $expectedData = null)
         }
     }
 
+    private function prepareBulkWriteArguments($arguments)
+    {
+        $operations = [];
+        $operations['options'] = $arguments['options'];
+        foreach ($arguments['requests'] as $requests) {
```
jmikola: `$request` would be clearer, since each element or "requests" is one "request".

tests/Collection/CrudSpecFunctionalTest.php
```diff
@@ -224,6 +231,38 @@ private function executeAssertResult(array $operation, $expectedResult, $actualR
...
+
+                if (isset($expectedResult['modifiedCount'])) {
+                    $this->assertSame($expectedResult['modifiedCount'], $actualResult->getModifiedCount());
+                }
+
+                if (isset($expectedResult['upsertedCount'])) {
+                    $this->assertSame($expectedResult['upsertedCount'], $actualResult->getUpsertedCount());
+                }
+
+                if (array_key_exists('upsertedId', $expectedResult)) {
```
jmikola: No change needed here, but if you're curious as to why we use `array_key_exists()` here instead of `isset()`, it's because `null` is technically a valid `_id` value and could be reported from an upsert (if the user originally supplied the `_id` value themselves, since the server would always use an ObjectId).

PR #443, opened 2017-11-30
PHPLIB-267 Document example of tailable cursor iteration

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-267

### DIFF
--- a/docs/tutorial/tailable-cursor.txt
+++ b/docs/tutorial/tailable-cursor.txt
@@ -1,0 +1,192 @@
+===============
+Tailable Cursor
+===============
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+When the driver executes a query or command (e.g.
+:manual:`aggregate </reference/command/aggregate>`), results from the operation
+are returned via a :php:`MongoDB\\Driver\\Cursor <class.mongodb-driver-cursor>`
+object. The Cursor class implements PHP's :php:`Traversable <traversable>`
+interface, which allows it to be iterated with ``foreach`` and interface with
+any PHP functions that work with :php:`iterables <types.iterable>`. Similar to
+result objects in other database drivers, cursors in MongoDB only support
+forward iteration, which means they cannot be rewound or used with ``foreach``
+multiple times.
+
+:manual:`Tailable cursors </core/tailable-cursors>` are a special type of
+MongoDB cursor that allows the client to read some results and then wait until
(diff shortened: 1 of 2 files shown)


docs/tutorial/tailable-cursor.txt
```diff
@@ -0,0 +1,121 @@
...
+         'projection' => [
+            'name' => 1,
+            'borough' => 1,
+            'cuisine' => 1,
+         ],
+      ]
+   );
+
+   $iterator = new IteratorIterator($cursor);
+
```
jmikola: We should start with a call to `$iterator->rewind()`. Even though the driver doesn't _currently_ require it, it's good practice and we'd do well to remind users of that.

docs/tutorial/tailable-cursor.txt
```diff
@@ -0,0 +1,121 @@
...
+   $iterator->next();
+
+Note that calling $iterator->next() and var_dump($iterator->current()) as above is
+identical to the following:
+
+.. code-block:: php
+
+   foreach (new IteratorIterator($cursor) as $document) {
+       var_dump($document);
+   }
```
jmikola: General thought: what if both of these sections (normal iteration and tailable cursor iteration) started by demonstrating the basic `foreach` approach and then showed the functional equivalent example with IteratorIterator? This would be:

```
$iterator = new IteratorIterator($cursor);

$iterator->rewind();

while ($iterator->valid()) {
    $document = $iterator->current();
    var_dump($document);
    $iterator->next();
}
```

After that, you can discuss that calling `next()` again would result in an exception (since we've exhausted the result set).

docs/tutorial/tailable-cursor.txt
```diff
@@ -0,0 +1,121 @@
...
+
+Wrapping a Tailable Cursor
+---------------------------------
+
+The following example wraps a tailable cursor in an IteratorIterator.
```
jmikola: If the changes above for basic cursor iteration are made, we can adopt a similar approach here. Starting with `foreach()`, we can show that it only walks through a single batch of results for a tailable cursor. Then mention that trying to `foreach()` again doesn't work since that attempts to rewind the cursor, which isn't allowed as all cursors only support forward iteration.

We then demonstrate the IteratorIterator approach and show that calling `next()` after iterating will block for new results instead of throwing an exception. We can use `valid()` after a blocking call to `next()` to check if there is new data available (as a general rule, we should only ever call `current()` and `key()` if `valid()` is true).

jmikola:
This can be rebased and merged at your leisure. Afterwards, you can pop into `#docs` and request a deployment so we can see the finished output on https://docs.mongodb.com/php-library/master/.

PR #457, opened 2018-01-15
Fixes from PHPStan level 0

carusogabriel:
Some fixes found by [`PHPStan`](https://github.com/phpstan/phpstan) level 0

### DIFF
--- a/src/ChangeStream.php
+++ b/src/ChangeStream.php
@@ -66,7 +66,7 @@ class ChangeStream implements Iterator
     }
 
     /**
-     * @return MongoDB\Driver\CursorId
+     * @return \MongoDB\Driver\CursorId
      */
     public function getCursorId()
     {
@@ -143,8 +143,8 @@ class ChangeStream implements Iterator
         if ($document === null) {
             throw new ResumeTokenException("Cannot extract a resumeToken from an empty document");
         }
-        if ($document instanceof Serializable) {
-            return $this->extractResumeToken($document->bsonSerialize());
+        if ($document instanceof \Serializable) {
+            $this->extractResumeToken($document->bsonSerialize());
         }
         if (isset($document->_id)) {
             $this->resumeToken = is_array($document) ? $document['_id'] : $document->_id;
--- a/src/GridFS/Bucket.php
+++ b/src/GridFS/Bucket.php
@@ -17,6 +17,7 @@
 
 namespace MongoDB\GridFS;
 
+use MongoDB\Collection;
 use MongoDB\Driver\Cursor;
 use MongoDB\Driver\Manager;
 use MongoDB\Driver\ReadConcern;
--- a/src/GridFS/ReadableStream.php
+++ b/src/GridFS/ReadableStream.php
@@ -132,7 +132,7 @@ class ReadableStream
      *
      * Note: this method may return a string smaller than the requested length
      * if data is not available to be read.
-     * 
+     *
      * @param integer $length Number of bytes to read
      * @return string
      * @throws InvalidArgumentException if $length is negative
@@ -175,7 +175,7 @@ class ReadableStream
     public function seek($offset)
     {
         if ($offset < 0 || $offset > $this->file->length) {
-            throw new InvalidArgumentException(sprintf('$offset must be >= 0 and <= %d; given: %d', $length, $offset));
+            throw new InvalidArgumentException(sprintf('$offset must be >= 0 and <= %d; given: %d', $this->file->length, $offset));
         }
 
         /* Compute the offsets for the chunk and buffer (i.e. chunk data) from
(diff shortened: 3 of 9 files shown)


src/ChangeStream.php
```diff
@@ -143,8 +143,8 @@ private function extractResumeToken($document)
         if ($document === null) {
             throw new ResumeTokenException("Cannot extract a resumeToken from an empty document");
         }
-        if ($document instanceof Serializable) {
-            return $this->extractResumeToken($document->bsonSerialize());
+        if ($document instanceof \Serializable) {
+            $this->extractResumeToken($document->bsonSerialize());
```
jmikola: @carusogabriel: This looks incorrect. Is PHPStan complaining that a `void` function is returning the result of another `void` function? If so, I think the correct fix is to add a solitary `return;` on the line below. The point of this is to recurse on the result of `$document->bsonSerialize()`, while this change allows execution of the current function to continue.

src/ChangeStream.php
```diff
@@ -143,8 +143,8 @@ private function extractResumeToken($document)
         if ($document === null) {
             throw new ResumeTokenException("Cannot extract a resumeToken from an empty document");
         }
-        if ($document instanceof Serializable) {
-            return $this->extractResumeToken($document->bsonSerialize());
+        if ($document instanceof \Serializable) {
```
jmikola: I believe this should have been left as-is, while adding `use MongoDB\BSON\Serializable;` above.

src/GridFS/Bucket.php
```diff
@@ -17,6 +17,7 @@
 
 namespace MongoDB\GridFS;
 
+use MongoDB\Collection;
```
jmikola: It looks like this was needed for the `getChunksCollection` return type. Good catch.

src/GridFS/ReadableStream.php
```diff
@@ -175,7 +175,7 @@ public function readBytes($length)
     public function seek($offset)
     {
         if ($offset < 0 || $offset > $this->file->length) {
-            throw new InvalidArgumentException(sprintf('$offset must be >= 0 and <= %d; given: %d', $length, $offset));
+            throw new InvalidArgumentException(sprintf('$offset must be >= 0 and <= %d; given: %d', $this->file->length, $offset));
```
jmikola: https://github.com/mongodb/mongo-php-library/pull/461 adds a regression test for this.

src/GridFS/WritableStream.php
```diff
@@ -141,7 +141,7 @@ public function close()
     /**
      * Return the stream's file document.
      *
-     * @return stdClass
+     * @return \stdClass
```
jmikola: I'd prefer to add a `use stdClass` up top for this. Will do so when merging.

src/MapReduceResult.php
```diff
@@ -77,7 +77,7 @@ public function getExecutionTimeMS()
      * Return the mapReduce results as a Traversable.
      *
      * @see http://php.net/iteratoraggregate.getiterator
-     * @return Traversable
+     * @return \Traversable
```
jmikola: Same here re: `use Traversable` up top.

src/Operation/ChangeStream.php
```diff
@@ -165,7 +165,7 @@ private function createChangeStreamOptions()
     /**
      * Create the aggregate pipeline with the changeStream command.
      *
-     * @return Command
+     * @return \MongoDB\Driver\Command
```
jmikola: Will add `use MongoDB\Driver\Command;` up top.

src/Operation/MapReduce.php
```diff
@@ -17,6 +17,7 @@
 
 namespace MongoDB\Operation;
 
+use MongoDB\BSON\Javascript;
```
jmikola: `finalize` should actually check for `JavascriptInterface`. I've opened [PHPLIB-313](https://jira.mongodb.org/browse/PHPLIB-313) to track that, and we'll fix this in 1.3.0.

src/ChangeStream.php
jmikola: Likely related to https://github.com/phpstan/phpstan/commit/679ea28f29699f409662cd81bd0414769b9e047e.

jmikola:
Manually merged in https://github.com/mongodb/mongo-php-library/commit/8a73c759133332aee735ae6c37ad8b9f14fa1a62 with requested changes. Thanks!

jmikola:
@carusogabriel: Side question: are there examples of incorporating PHPStan into `.travis.yml`? That may be something worth implementing so we can catch these issues more proactively going forward.

carusogabriel:
@jmikola Sure. I'll open a PR with a new build running PHPStan, and we can discuss improvements there :+1:

PR #460, opened 2018-01-29
Swap @bjori for @kvwalker

jmikola:
@kvwalker: just want to confirm that this is your preferred email address (pulled it from GitHub). This info will end up being rendered on https://packagist.org/packages/mongodb/mongodb.


kvwalker:
Yep!

PR #463, opened 2018-01-29
PHPLIB-306: Set GridFS uploadDate at completion time

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-306

### DIFF
--- a/tests/GridFS/StreamWrapperFunctionalTest.php
+++ b/tests/GridFS/StreamWrapperFunctionalTest.php
@@ -3,7 +3,6 @@
 namespace MongoDB\Tests\GridFS;
 
 use MongoDB\BSON\Binary;
-use MongoDB\BSON\ObjectId;
 use MongoDB\BSON\UTCDateTime;
 
 /**
@@ -157,10 +156,10 @@ class StreamWrapperFunctionalTest extends FunctionalTestCase
         $this->assertSame(0100222, $stat['mode']);
         $this->assertSame(0, $stat[7]);
         $this->assertSame(0, $stat['size']);
-        $this->assertGreaterThanOrEqual(0, $stat[9]);
-        $this->assertGreaterThanOrEqual(0, $stat['mtime']);
-        $this->assertGreaterThanOrEqual(0, $stat[10]);
-        $this->assertGreaterThanOrEqual(0, $stat['ctime']);
+        $this->assertSame(0, $stat[9]);
+        $this->assertSame(0, $stat['mtime']);
+        $this->assertSame(0, $stat[10]);
+        $this->assertSame(0, $stat['ctime']);
         $this->assertSame(1024, $stat[11]);
         $this->assertSame(1024, $stat['blksize']);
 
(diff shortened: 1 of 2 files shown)


tests/GridFS/StreamWrapperFunctionalTest.php
```diff
@@ -147,20 +148,19 @@ public function testWritableStreamSeek()
...
         $this->assertSame(0, $stat[7]);
         $this->assertSame(0, $stat['size']);
-        $this->assertGreaterThanOrEqual($currentTimestamp, $stat[9]);
-        $this->assertGreaterThanOrEqual($currentTimestamp, $stat['mtime']);
-        $this->assertGreaterThanOrEqual($currentTimestamp, $stat[10]);
-        $this->assertGreaterThanOrEqual($currentTimestamp, $stat['ctime']);
+        $this->assertGreaterThanOrEqual(0, $stat[9]);
+        $this->assertGreaterThanOrEqual(0, $stat['mtime']);
+        $this->assertGreaterThanOrEqual(0, $stat[10]);
+        $this->assertGreaterThanOrEqual(0, $stat['ctime']);
```
jmikola: These should always default to `0`, so I think we can use `$this->assertSame(0, ...)` here.

tests/GridFS/StreamWrapperFunctionalTest.php
```diff
@@ -169,6 +169,29 @@ public function testWritableStreamStat()
...
+
+    public function testWritableStreamStatAfterSaving()
+    {
+        $id = new ObjectId;
+        $stream = $this->bucket->openUploadStream('filename', ['chunkSizeBytes' => 1024, '_id' => $id]);
+
+        $this->assertSame(6, fwrite($stream, 'foobar'));
+        $this->assertTrue(fclose($stream));
+        $currentTimestamp = time();
+        $stream = $this->bucket->openDownloadStream($id);
```
jmikola: I just realized that the `setUp()` function for this test creates a file with `_id` of `length-10` and `uploadDate` of `new UTCDateTime('1484202200000')`. Switching over to that would reduce all of these lines to a single `$stream = $this->bucket->openDownloadStream('length-10');`, and allow the ObjectID import to be removed as well.

Apologies for not catching this earlier. I hadn't looked at the entire file until now.

tests/GridFS/StreamWrapperFunctionalTest.php
```diff
@@ -169,6 +169,29 @@ public function testWritableStreamStat()
...
+        $this->assertSame(0100444, $stat[2]);
+        $this->assertSame(0100444, $stat['mode']);
+        $this->assertSame(6, $stat[7]);
+        $this->assertSame(6, $stat['size']);
+        $this->assertGreaterThanOrEqual($currentTimestamp, $stat[9]);
+        $this->assertGreaterThanOrEqual($currentTimestamp, $stat['mtime']);
+        $this->assertGreaterThanOrEqual($currentTimestamp, $stat[10]);
+        $this->assertGreaterThanOrEqual($currentTimestamp, $stat['ctime']);
+        $this->assertSame(1024, $stat[11]);
+        $this->assertSame(1024, $stat['blksize']);
```
jmikola: Note that this field reports `chunkSize`, so it would become `4` if you use `length-10`.

https://github.com/mongodb/mongo-php-library/blob/f4801b8086cc2cf67997ea9e2bad671ff87d3ff1/src/GridFS/StreamWrapper.php#L197-L199

tests/GridFS/StreamWrapperFunctionalTest.php
```diff
@@ -169,6 +169,29 @@ public function testWritableStreamStat()
...
+
+        $stat = fstat($stream);
+        $this->assertSame(0100444, $stat[2]);
+        $this->assertSame(0100444, $stat['mode']);
+        $this->assertSame(6, $stat[7]);
+        $this->assertSame(6, $stat['size']);
+        $this->assertGreaterThanOrEqual($currentTimestamp, $stat[9]);
+        $this->assertGreaterThanOrEqual($currentTimestamp, $stat['mtime']);
+        $this->assertGreaterThanOrEqual($currentTimestamp, $stat[10]);
+        $this->assertGreaterThanOrEqual($currentTimestamp, $stat['ctime']);
```
jmikola: This can use `assertSame()` if using the `length-10` file, since we'll know the exact value of `uploadDate`.

Note: 1484202200000 is milliseconds, so that will need to be converted to seconds for these values.

tests/GridFS/StreamWrapperFunctionalTest.php
```diff
@@ -173,25 +172,19 @@ public function testWritableStreamStatBeforeSaving()
...
-        $this->assertSame(6, $stat['size']);
-        $this->assertGreaterThanOrEqual($currentTimestamp, $stat[9]);
-        $this->assertGreaterThanOrEqual($currentTimestamp, $stat['mtime']);
-        $this->assertGreaterThanOrEqual($currentTimestamp, $stat[10]);
-        $this->assertGreaterThanOrEqual($currentTimestamp, $stat['ctime']);
-        $this->assertSame(1024, $stat[11]);
-        $this->assertSame(1024, $stat['blksize']);
+        $this->assertSame(10, $stat[7]);
+        $this->assertSame(10, $stat['size']);
+        $this->assertSame(1484202200000/1000, $stat[9]);
```
jmikola: On the off chance this breaks 32-bit environments, let's just use `1484202200` here.

PR #465, opened 2018-01-29
Add PHPStan to Travis CI

carusogabriel:
As suggested by @jmikola in #457, this PR adds `PHPStan` to `Travis CI`.

There is only 1 remain error to bump to level 1:
```bash
 ------ ----------------------------------------------- 
  Line   src/Operation/MapReduce.php                    
 ------ ----------------------------------------------- 
  158    Class MongoDB\Operation\Javascript not found.  
 ------ ----------------------------------------------- 
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,8 @@ matrix:
     - php: 7.0
       env:
         - SERVER_VERSION=3.4.11
+    - php: 7.0
+      script: vendor/bin/phpstan analyse -l 0 src
 
 before_install:
   - pip install "mongo-orchestration>=0.6.7,<1.0" --user `whoami`
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,8 @@
         "ext-mongodb": "^1.4.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "^4.8.36"
+        "phpunit/phpunit": "^4.8.36",
+        "phpstan/phpstan": "^0.9.2"
     },
     "autoload": {
         "psr-4": { "MongoDB\\": "src/" },
--- a/src/Client.php
+++ b/src/Client.php
@@ -262,7 +262,7 @@ class Client
      *
      * @see http://php.net/manual/en/mongodb-driver-manager.startsession.php
      * @param array  $options      Session options
-     * @return MongoDB\Driver\Session
+     * @return \MongoDB\Driver\Session
      */
     public function startSession(array $options = [])
     {


.travis.yml
```diff
@@ -58,6 +58,17 @@ matrix:
     - php: 7.0
       env: DRIVER_VERSION=devel SERVER_VERSION=3.4
       addons: *common_addons
+    - php: 7.0
+      before_script:
```
carusogabriel: Are there some scripts that we can simplify here?

carusogabriel:
@jmikola My I ask your help in this last PHPStan error?

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: php
 dist: trusty
-sudo: false 
+sudo: false
```
jmikola: This trailing space was removed in https://github.com/mongodb/mongo-php-library/pull/481 when I revised the installation process for MongoDB in `.travis.yml`.

.travis.yml
```diff
@@ -58,6 +58,19 @@ matrix:
...
+      before_script:
+        - mongod --version
+        - pecl install -f mongodb-${DRIVER_VERSION}
+        - php --ri mongodb
+        - composer install --dev --no-interaction --prefer-source
+        - ulimit -c
+        - ulimit -c unlimited -S
+        - composer require --dev phpstan/phpstan
+      script:
+        - vendor/bin/phpstan analyse -l 0 src
```
jmikola: I assume the intention here is to only run PHPStan for a single job, since it would be redundant to run it otherwise?

Rather than duplicate the `before_script` block, it may be preferable to add `phpstan/phpstan` as a development dependency in `composer.json` (as we do with PHPUnit). Then, would only have to overwrite the `script` block here.

Note that as of https://github.com/mongodb/mongo-php-library/pull/481, I'm now launching mongo-orchestration and a MongoDB server in `before_script` and terminating it in `after_script`. That certainly isn't needed for PHPStan to perform its static analysis.

Would using [build stages](https://blog.travis-ci.com/2017-09-12-build-stages-order-and-conditions), as [Doctrine ORM does](https://github.com/doctrine/doctrine2/blob/bcc9da2f4a7a534ec1ffa58da6ae199efc1fc10a/.travis.yml#L69), help here?

carusogabriel:
@jmikola Thanks, I'll rebase and maybe PHPStan level 0 will pass :tada: 

I'll require it as a dev dependency and use Build Stages :+1:

carusogabriel:
@jmikola Unfortunately, we can't ship with `PHPStan` as it requires PHP 7.0. I'll close this for now as we will need to rewrite `.travis.yml` :smile:

PR #467, opened 2018-01-30
PHPLIB-316: Migrate operations to use read and write command methods

jmikola:
https://jira.mongodb.org/browse/PHPLIB-316

Please read through carefully. There was quite a bit of code copypasta involved in this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,13 +33,6 @@ matrix:
       addons: *common_addons
     - php: 7.0
       env:
-        - SERVER_VERSION=2.4
-      addons:
-        apt:
-          sources: [ mongodb-upstart ]
-          packages: [ mongodb-10gen, *common_packages ]
-    - php: 7.0
-      env:
         - SERVER_VERSION=2.6
       addons:
         apt:
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,54 +10,55 @@ addons:
     packages: &common_packages
       - gdb
 
+env:
+  global:
+    - DRIVER_VERSION=1.4.0RC1
+    - SERVER_VERSION=3.4
+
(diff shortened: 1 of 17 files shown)


jmikola:
Note: the ListDatabases, ListCollections, and ListIndexes operations were intentionally omitted from this PR and will continue to use `Server::executeCommand()`. The reason being is that those commands do not take a `readConcern` option (see: [spec](https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#read-concern)) and were we to use `executeReadCommand()`, libmongoc may actually inject a `readConcern` option from the client URI on its own. AFAIK, libmongoc does not use a whitelist to determine command eligibility apart from `findAndModify`, which supported `writeConcern` one server version before other write commands.

Additionally, you'll notice that the `readConcern` and `writeConcern` options are now always passed in as `executeWhateverCommand()` options instead of being directly injected into the command document. This ensures the libmongoc does not inject a duplicate key into the BSON document with the same name. Its `read|write_command_with_opts()` functions don't expect those options to be present in the command `bson_t` and it decides to inject client-level defaults independently of the command document's contents (only the opts `bson_t` is considered).

There are a few instances where I'm still passing `$readPreference` or `$writeConcern` as the third positional argument to an execute method (e.g. Find, InsertOne). That will be changed to use array options later on when I add support for Sessions ([PHPLIB-302](https://jira.mongodb.org/browse/PHPLIB-302)).

jmikola:
FYI: I manually cancelled the 2.6, 3.0, 3.2 builds in https://travis-ci.org/mongodb/mongo-php-library/builds/335721451 since they all end up using 3.4 anyway.

PR #468, opened 2018-01-30
PHPLIB-276: Add maxAwaitTimeMS support for change streams

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-276

### DIFF
--- a/docs/tutorial/tailable-cursor.txt
+++ b/docs/tutorial/tailable-cursor.txt
@@ -177,7 +177,6 @@ using :php:`IteratorIterator <iteratoriterator>`.
 
       $iterator->next();
    }
-}
 
 Much like the ``foreach`` example, this version on the consumer script will
 start by quickly printing all results in the capped collection; however, it will
--- a/src/Operation/Aggregate.php
+++ b/src/Operation/Aggregate.php
@@ -163,6 +163,10 @@ class Aggregate implements Executable
             throw InvalidArgumentException::invalidType('"hint" option', $options['hint'], 'string or array or object');
         }
 
+        if (isset($options['maxAwaitTimeMS']) && ! is_integer($options['maxAwaitTimeMS'])) {
+            throw InvalidArgumentException::invalidType('"maxAwaitTimeMS" option', $options['maxAwaitTimeMS'], 'integer');
+        }
+
         if (isset($options['maxTimeMS']) && ! is_integer($options['maxTimeMS'])) {
             throw InvalidArgumentException::invalidType('"maxTimeMS" option', $options['maxTimeMS'], 'integer');
         }
@@ -277,6 +281,7 @@ class Aggregate implements Executable
             'aggregate' => $this->collectionName,
             'pipeline' => $this->pipeline,
         ];
+        $cmdOptions = [];
 
         // Servers < 2.6 do not support any command options
         if ( ! $isCursorSupported) {
@@ -303,13 +308,17 @@ class Aggregate implements Executable
             $cmd['hint'] = is_array($this->options['hint']) ? (object) $this->options['hint'] : $this->options['hint'];
         }
 
+        if (isset($this->options['maxAwaitTimeMS'])) {
+            $cmdOptions['maxAwaitTimeMS'] = $this->options['maxAwaitTimeMS'];
+        }
+
         if ($this->options['useCursor']) {
             $cmd['cursor'] = isset($this->options["batchSize"])
                 ? ['batchSize' => $this->options["batchSize"]]
                 : new stdClass;
         }
 
-        return new Command($cmd);
+        return new Command($cmd, $cmdOptions);
     }
 
     /**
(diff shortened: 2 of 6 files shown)


src/Operation/Aggregate.php
```diff
@@ -297,6 +302,14 @@ private function createCommand(Server $server, $isCursorSupported)
             $cmd['hint'] = is_array($this->options['hint']) ? (object) $this->options['hint'] : $this->options['hint'];
         }
 
+        if (isset($this->options['maxAwaitTimeMS'])) {
+            $cmdOptions['maxAwaitTimeMS'] = $this->options['maxAwaitTimeMS'];
+        }
+
+        if (isset($this->options['maxTimeMS'])) {
```
jmikola: Is this necessary? I think it's already covered by `foreach (['comment', 'maxTimeMS'] as $option)` about 15 lines prior.

tests/Operation/ChangeStreamFunctionalTest.php
```diff
@@ -237,4 +237,22 @@ public function testConnectionException()
...
+    {
+        $this->collection = new Collection($this->manager, $this->getDatabaseName(), $this->getCollectionName());
+        $maxAwaitTimeMS = 10;
+        $changeStreamResult = $this->collection->watch([], ['maxAwaitTimeMS' => $maxAwaitTimeMS]);
+        $changeStreamResult->next();
+
+        /* Make sure we await results for at least maxAwaitTimeMS, since no new
+        * documents should be inserted to wake up the server's command thread.
+        * Also ensure that we don't wait too long (server default is one
+        * second). */
```
jmikola: Can you add a space in this comment to align the asterisks?

tests/Operation/FindFunctionalTest.php
```diff
@@ -154,12 +154,16 @@ public function testMaxAwaitTimeMS()
         $cursor = $operation->execute($this->getPrimaryServer());
         $it = new \IteratorIterator($cursor);
 
-        // Make sure we await results for no more than the maxAwaitTimeMS.
+        /* Make sure we await results for at least maxAwaitTimeMS, since no new
+        * documents should be inserted to wake up the server's query thread.
+        * Also ensure that we don't wait too long (server default is one
+        * second). */
```
jmikola: Extra space to align asterisks here, too.

tests/Operation/ChangeStreamFunctionalTest.php
```diff
@@ -237,4 +237,22 @@ public function testConnectionException()
                         ]);
         $this->assertEquals($changeStreamResult->current(), $expectedResult);
     }
+
+    public function testMaxAwaitTimeMS()
+    {
+        $this->collection = new Collection($this->manager, $this->getDatabaseName(), $this->getCollectionName());
+        $maxAwaitTimeMS = 10;
+        $changeStreamResult = $this->collection->watch([], ['maxAwaitTimeMS' => $maxAwaitTimeMS]);
+        $changeStreamResult->next();
```
jmikola: I know you removed the `rewind()` call in the constructor, but can you confirm whether `$changeStreamResult->rewind()` invokes `getMore` and introduces a delay?

If so, we can do the assert and time check on that initial `rewind()`. It would also set a good example for anyone looking at the test files for guidance (ideally, users should rewind first).

jmikola:
Travis appears to have failed because the jobs are using `ext-mongodb` 1.3.3 and the code now relies on the second parameter to `MongoDB\Driver\Command::__construct()` introduced in 1.4.0.

If this was the first PR requiring 1.4.0, you'd need to bump the `composer.json` dependency (https://github.com/mongodb/mongo-php-library/commit/091d8bd9a7ad904b2030206089a6521ef7de925c) and update `.travis.yml` (https://github.com/mongodb/mongo-php-library/commit/8361528cecd7da22b408beac239565c80af1fb1c#diff-354f30a63fb0907d4ad57269548329e3); however, since those were done in #467 you should be OK with rebasing on master.

src/Operation/Aggregate.php
```diff
@@ -303,13 +308,25 @@ private function createCommand(Server $server, $isCursorSupported)
...
+            $cmdOptions['maxAwaitTimeMS'] = $this->options['maxAwaitTimeMS'];
+        }
+
+        if (isset($this->options['readConcern'])) {
+            $cmd['readConcern'] = \MongoDB\read_concern_as_document($this->options['readConcern']);
+        }
+
+        if (isset($this->options['writeConcern'])) {
+            $cmd['writeConcern'] = \MongoDB\write_concern_as_document($this->options['writeConcern']);
+        }
```
jmikola: Assuming this commit was rebased, `readConcern` and `writeConcern` were moved in https://github.com/mongodb/mongo-php-library/commit/091d8bd9a7ad904b2030206089a6521ef7de925c#diff-3b2e2c1a4b1425910f13159568854106L302 to a separate `createOptions()` method. If this commit is adding them back, it's likely that these got put back when selecting the `maxAwaitTimeMS` addition (git likely presented them all as one addition). This would have required splitting the diff during the rebase, which I'm happy to walk through and demonstrate.

docs/tutorial/tailable-cursor.txt
```diff
@@ -177,7 +177,6 @@ using :php:`IteratorIterator <iteratoriterator>`.
 
       $iterator->next();
    }
-}
```
jmikola: Thanks for catching this.

PR #472, opened 2018-01-31
PHPLIB-302: Support session option on Operations

jmikola:
https://jira.mongodb.org/browse/PHPLIB-302

### DIFF
--- a/docs/includes/apiargs-MongoDBCollection-method-watch-option.yaml
+++ b/docs/includes/apiargs-MongoDBCollection-method-watch-option.yaml
@@ -66,4 +66,8 @@ description: |
 interface: phpmethod
 operation: ~
 optional: true
+---
+source:
+  file: apiargs-common-option.yaml
+  ref: session
 ...
--- a/docs/includes/apiargs-MongoDBCollection-method-watch-option.yaml
+++ b/docs/includes/apiargs-MongoDBCollection-method-watch-option.yaml
@@ -45,6 +45,10 @@ optional: true
 ---
 source:
   file: apiargs-MongoDBCollection-common-option.yaml
+  ref: readConcern
+---
+source:
+  file: apiargs-MongoDBCollection-common-option.yaml
   ref: readPreference
 post: |
   This is used for both the initial change stream aggregation and for
--- a/src/Operation/Watch.php
+++ b/src/Operation/Watch.php
@@ -122,8 +122,7 @@ class Watch implements Executable
      * @see Executable::execute()
      * @param Server $server
      * @return ChangeStream
-     * @throws UnexpectedValueException if the command response was malformed
-     * @throws UnsupportedException if collation, read concern, or write concern is used and unsupported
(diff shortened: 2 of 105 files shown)


docs/includes/apiargs-MongoDBCollection-method-watch-option.yaml
```diff
@@ -43,6 +43,10 @@ interface: phpmethod
 operation: ~
 optional: true
 ---
+source:
+  file: apiargs-MongoDBCollection-common-option.yaml
+  ref: readConcern
+---
```
jmikola: It looks like this was never added in https://github.com/mongodb/mongo-php-library/pull/439 and I missed it in the relevant PR.

src/Operation/Watch.php
```diff
@@ -91,24 +97,8 @@ public function __construct(Manager $manager, $databaseName, $collectionName, ar
...
-        }
-
-        if (isset($options['readConcern']) && ! $options['readConcern'] instanceof ReadConcern) {
-            throw InvalidArgumentException::invalidType('"readConcern" option', $options['readConcern'], 'MongoDB\Driver\ReadConcern');
-        }
-
-        if (isset($options['readPreference']) && ! $options['readPreference'] instanceof ReadPreference) {
-            throw InvalidArgumentException::invalidType('"readPreference" option', $options['readPreference'], 'MongoDB\Driver\ReadPreference');
+        if (isset($options['fullDocument']) && ! is_string($options['fullDocument'])) {
+            throw InvalidArgumentException::invalidType('"fullDocument" option', $options['fullDocument'], 'string');
```
jmikola: After adding a unit test for Watch's option validation I noticed that "fullDocument" was never validated as a string.

src/Operation/Watch.php
```diff
@@ -130,63 +122,51 @@ public function __construct(Manager $manager, $databaseName, $collectionName, ar
...
+        array_unshift($pipeline, $changeStream);
 
-        return $cmd;
+        $aggregateOptions = array_intersect_key($this->options, ['batchSize' => 1, 'collation' => 1, 'maxAwaitTimeMS' => 1, 'readConcern' => 1, 'readPreference' => 1, 'session' => 1]);
+
+        return new Aggregate($this->databaseName, $this->collectionName, $pipeline, $aggregateOptions);
     }
 
-    private function createResumeCallable()
+    private function createResumeCallable(Manager $manager)
```
jmikola: After introducing `createAggregate()` so that Aggregate could be constructed in `Watch::__construct()`, this method no longer changed state (with `array_shift($this->pipeline)` removed). That meant we could create the callable once and re-use it from `execute()`.

Saving the callable as a class property then replaces the Manager property, since we can `use` it just for this closure that requires it.

src/Operation/Watch.php
kvwalker: Thanks for catching that!

PR #475, opened 2018-02-01
PHPLIB-292: Apply type map to offsetGet and throw exceptions for mutate methods in TypeMapArrayIterator

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-292

### DIFF
--- a/tests/Model/TypeMapArrayIteratorTest.php
+++ b/tests/Model/TypeMapArrayIteratorTest.php
@@ -85,14 +85,14 @@ class TypeMapArrayIteratorTest extends TestCase
         $iterator->rewind();
 
         $iterator->asort();
-
     }
 
     /**
+     * @dataProvider provideMutateMethods
      * @expectedException MongoDB\Exception\BadMethodCallException
      * @expectedExceptionMessage MongoDB\Model\TypeMapArrayIterator is immutable
      */
-    public function testAsortThrowsException()
+    public function testMutateMethodsCannotBeCalled($method, $args)
     {
         $document = [
             'array' => [1, 2, 3],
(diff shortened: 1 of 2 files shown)


tests/Model/TypeMapArrayIteratorTest.php
```diff
@@ -31,4 +32,295 @@ public function testCurrentAppliesTypeMap()
...
+        $iterator = new TypeMapArrayIterator([$document], $typeMap);
+
+        $expectedDocument = (object) [
+            'array' => [1, 2, 3],
+            'object' => (object) ['foo' => 'bar'],
+        ];
+
+        $iterator->rewind();
+
+        $iterator->asort();
```
jmikola: This looks like a typo and should likely be `append($someValue)`.

tests/Model/TypeMapArrayIteratorTest.php
```diff
@@ -31,4 +32,295 @@ public function testCurrentAppliesTypeMap()
...
+        $expectedDocument = (object) [
+            'array' => [1, 2, 3],
+            'object' => (object) ['foo' => 'bar'],
+        ];
+
+        $iterator->rewind();
+        $cmp_function = function($a, $b) {
+            return $a;
+        };
+        $iterator->uasort($cmp_function(0, 1));
```
jmikola: PHP doesn't appear to complain but [`uasort()`](http://php.net/manual/en/arrayiterator.uasort.php) expects a callable as its first argument. If you were to pass this directly to `ArrayIterator::uasort()` you might see the following error:

> uasort() expects parameter 2 to be a valid callback, no array or string given

I thought we might be actually catch this by ensuring our own `uasort()` and `uksort()` methods utilize the [`callable`](http://php.net/manual/en/language.types.callable.php) type hint for their first argument, but that raises the following error:

> Declaration of MongoDB\Model\TypeMapArrayIterator::uasort(callable $cmp_function) should be compatible with ArrayIterator::uasort($cmp_function)

This tells me that ArrayIterator doesn't actually declare it's type hints. If we dive into the SPL classes in PHP internals we can see the arg info [here](https://github.com/php/php-src/blob/a89eb2f101bf0ca6f92ee3160fa46de0bd6f31b8/ext/spl/spl_array.c#L1926). Going a bit deeper it appears that most all of the ArrayIterator methods actually just [proxy to existing PHP functions](https://github.com/php/php-src/blob/a89eb2f101bf0ca6f92ee3160fa46de0bd6f31b8/ext/spl/spl_array.c#L1518) through the [`spl_array_method()` function](https://github.com/php/php-src/blob/a89eb2f101bf0ca6f92ee3160fa46de0bd6f31b8/ext/spl/spl_array.c#L1467). This explains why the error above says "uasort() expects parameter 2 to be a valid callback" (a) refers to [`uasort()`](http://php.net/uasort) instead of ArrayIterator and (b) mentions parameter 2, even though our callback is the first parameter.

jmikola: cc: @sgolemon

tests/Model/TypeMapArrayIteratorTest.php
```diff
@@ -2,6 +2,7 @@
 
 namespace MongoDB\Tests\Model;
 
+use MongoDB\Exception\BadMethodCallException;
```
jmikola: `@expectedException` takes a fully qualified class name so we don't actually need to `use` it here.

PR #476, opened 2018-02-01
Change stream cleanup

jmikola:
Includes some todo items for [PHPLIB-322](https://jira.mongodb.org/browse/PHPLIB-322) (resume support in `rewind()`).

### DIFF
--- a/docs/includes/apiargs-MongoDBCollection-method-watch-option.yaml
+++ b/docs/includes/apiargs-MongoDBCollection-method-watch-option.yaml
@@ -18,6 +18,10 @@ description: |
 
   - ``MongoDB\Operation\ChangeStream::FULL_DOCUMENT_DEFAULT`` (*default*)
   - ``MongoDB\Operation\ChangeStream::FULL_DOCUMENT_UPDATE_LOOKUP``
+
+  .. note::
+
+     This is an option of the `$changeStream` pipeline stage.
 interface: phpmethod
 operation: ~
 optional: true
@@ -45,7 +49,9 @@ type: array|object
 description: |
   Specifies the logical starting point for the new change stream.
 
-  Note this is an option of the '$changeStream' pipeline stage.
+  .. note::
+
+     This is an option of the `$changeStream` pipeline stage.
 interface: phpmethod
 operation: ~
 optional: true
--- a/docs/includes/apiargs-MongoDBCollection-method-watch-option.yaml
+++ b/docs/includes/apiargs-MongoDBCollection-method-watch-option.yaml
@@ -1,6 +1,13 @@
-source:
-  file: apiargs-MongoDBCollection-method-aggregate-option.yaml
-  ref: batchSize
+---
+arg_name: option
+name: batchSize
+type: integer
+description: |
+  Specifies the maximum number of change events to return in each batch of the
+  response from the MongoDB cluster.
+interface: phpmethod
+operation: ~
+optional: true
 ---
 source:
   file: apiargs-MongoDBCollection-common-option.yaml
--- a/src/Operation/Watch.php
+++ b/src/Operation/Watch.php
@@ -17,7 +17,7 @@
 
 namespace MongoDB\Operation;
 
-use MongoDB\ChangeStream as ChangeStreamResult;
+use MongoDB\ChangeStream;
 use MongoDB\Driver\Command;
 use MongoDB\Driver\Manager;
 use MongoDB\Driver\ReadConcern;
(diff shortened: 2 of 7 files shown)


tests/Operation/WatchFunctionalTest.php
```diff
@@ -0,0 +1,340 @@
...
+         * a socket timeout that is less than the change stream's maxAwaitTimeMS
+         * option. */
+        $manager = new Manager($this->getUri(), ['socketTimeoutMS' => 50]);
+        $primaryServer = $manager->selectServer(new ReadPreference(ReadPreference::RP_PRIMARY));
+
+        $operation = new Watch($manager, $this->getDatabaseName(), $this->getCollectionName(), [], ['maxAwaitTimeMS' => 100]);
+        $changeStream = $operation->execute($primaryServer);
+
+        /* Note: we intentionally do not start iteration with rewind() to ensure
+         * that we test resume functionality within next(). */
```
jmikola: This is the once place (afaik) where we're not starting with `rewind()` now, so I wanted to make a note of it.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -0,0 +1,340 @@
...
+        $this->insertDocument(['_id' => 2, 'x' => 'bar']);
+
+        $changeStream->next();
+        $this->assertTrue($changeStream->valid());
+
+        $expectedResult = [
+            '_id' => $changeStream->current()->_id,
+            'operationType' => 'insert',
+            'fullDocument' => ['_id' => 2, 'x' => 'bar'],
+            'ns' => ['db' => $this->getDatabaseName(), 'coll' => $this->getCollectionName()],
```
jmikola: Filling in these values from methods makes the test more robust if the PHPUnit configuration changes (that determines the database name). Additionally, I used numeric `_id` values in the documents so we can expect the exact value in the change stream. The change stream's resume token is now the only weak expectation, since we can't predict it.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -0,0 +1,340 @@
...
+        $rp->setAccessible(true);
+
+        $iterator = $rp->getValue($changeStream);
+
+        $this->assertInstanceOf('IteratorIterator', $iterator);
+
+        $cursor = $iterator->getInnerIterator();
+
+        $this->assertInstanceOf('MongoDB\Driver\Cursor', $cursor);
+        $this->assertFalse($cursor->isDead());
```
jmikola: I asked @ajdavis in Slack about whether the alive boolean being `true` and the cursor ID being non-zero are equivalent in libmongoc for an already-executed cursor. For the time being, I wanted to check both.

docs/includes/apiargs-MongoDBCollection-method-watch-option.yaml
```diff
@@ -1,6 +1,13 @@
...
+---
+arg_name: option
+name: batchSize
+type: integer
+description: |
+  Specifies the maximum number of change events to return in each batch of the
+  response from the MongoDB cluster.
+interface: phpmethod
+operation: ~
+optional: true
```
jmikola: This documentation is correct for the spec itself, but PHP still doesn't support this (see: [PHPLIB-321](https://jira.mongodb.org/browse/PHPLIB-321)).

src/Operation/Watch.php
```diff
@@ -69,7 +70,9 @@ class ChangeStream implements Executable
      *    This is not supported for server versions < 3.2 and will result in an
```
kvwalker: Is this necessary if change streams can only be used on server versions >= 3.6?

jmikola: Good catch. I think that we can remove the earlier sentence as well:

> "Note that a "majority" read concern is not compatible with the $out stage"

According to [the spec](https://github.com/mongodb/specifications/blob/master/source/change-streams.rst#helper-method), "majority" is the only supported read concern for change streams in 3.6; however, it'd be easier to not discuss this in our docs because it's subject to change (much like the `2dsphere` index option).

jmikola:
Inadvertently merged a local copy of the branch in https://github.com/mongodb/mongo-php-library/commit/a09973f143ef4cd39bb8b3d5d6d2b6431fadf9a1, which did not include  https://github.com/mongodb/mongo-php-library/pull/476/commits/35b2a7eeb8e67137762a5711591708a5a48a0f6b. I then merged that missing commit in https://github.com/mongodb/mongo-php-library/commit/de2e331066e38a84c8646e510c48dcec1f0fab2a.

Although the commit hashes in the first merge differed from those in the PR (due to rebasing before https://github.com/mongodb/mongo-php-library/pull/476/commits/35b2a7eeb8e67137762a5711591708a5a48a0f6b), their content was the same.

PR #477, opened 2018-02-02
PHPLIB-303: Implement MongoDB\Client::startSession()

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-303

### DIFF
--- a/docs/includes/apiargs-MongoDBClient-method-startSession-option.yaml
+++ b/docs/includes/apiargs-MongoDBClient-method-startSession-option.yaml
@@ -1,0 +1,12 @@
+arg_name: option
+name: causalConsistency
+type: boolean
+description: |
+  Enables or disables :ref:`causal consistency <causal-consistency>` for the
+  session. If true, each operation in the session will be causally ordered after
+  the previous read or write operation. Set to false to disable causal
+  consistency. Defaults to true.
+interface: phpmethod
+operation: ~
+optional: true
+...
--- a/docs/reference/method/MongoDBClient-startSession.txt
+++ b/docs/reference/method/MongoDBClient-startSession.txt
@@ -1,0 +1,82 @@
+===============================
+MongoDB\\Client::startSession()
+===============================
+
+.. default-domain:: mongodb
+
+.. versionadded:: 1.3
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Definition
+----------
+
+.. phpmethod:: MongoDB\\Client::startSession()
+
+   Start a new client session for use with this client.
+
+   .. code-block:: php
+
+      function startSession(array $options = []): MongoDB\Driver\Session
+
+   The ``$options`` parameter supports the following options:
+
+   .. include:: /includes/apiargs/MongoDBClient-method-startSession-option.rst
+
+Return Values
+-------------
+
+A :php:`MongoDB\Driver\Session <mongodb-driver-session>`
+
+Errors/Exceptions
+-----------------
+
+.. include:: /includes/extracts/error-driver-invalidargumentexception.rst
+.. include:: /includes/extracts/error-driver-runtimeexception.rst
+
+Example
+-------
+
+The following example starts a new session:
+
+.. code-block:: php
+
+   <?php
+
+   $client = new MongoDB\Client;
+
+   $session = $client->startSession();
+
+   var_dump($session);
+
+The output would then resemble::
+
+   object(MongoDB\Driver\Session)#2043 (4) {
+     ["logicalSessionId"]=>
+     array(1) {
+       ["id"]=>
+       object(MongoDB\BSON\Binary)#225 (2) {
+         ["data"]=>
+         string(16) "................"
+         ["type"]=>
+         int(4)
+       }
+     }
+     ["clusterTime"]=>
+     NULL
+     ["causalConsistency"]=>
+     bool(true)
+     ["operationTime"]=>
+     NULL
+   }
+
+See Also
+--------
+
+- :php:`MongoDB\\Driver\\Manager::startSession()
+  <manual/en/mongodb-driver-manager.startsession.php>`
+- :ref:`Causal Consistency <causal-consistency>` in the MongoDB manual
--- a/src/Client.php
+++ b/src/Client.php
@@ -256,4 +256,16 @@ class Client
 
         return new Database($this->manager, $databaseName, $options);
     }
+
+    /**
+     * Start a new client session.
+     *
+     * @see http://php.net/manual/en/mongodb-driver-manager.startsession.php
+     * @param array  $options      Session options
+     * @return MongoDB\Driver\Session
+     */
+    public function startSession(array $options = [])
+    {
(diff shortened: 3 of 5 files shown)


docs/includes/apiargs-MongoDBClient-method-startSession-option.yaml
```diff
@@ -0,0 +1,43 @@
+arg_name: option
+name: causalConsistency
+type: boolean
+description: |
+  Enables or disables :ref:`causal consistency <causal-consistency>` for the
+  session. :method:`Mongo.startSession()` enables ``causalConsistency`` by
+  default. After starting a session, you cannot modify its
+  ``causalConsistency`` setting.
```
jmikola: The "Mongo.startSession() enables" sentence looks like it's referring to the shell API, so we can remove that. I also wouldn't bother with the disclaimer that the option cannot be changed later, as all options in the library are permanent.

I think we can leave the first sentence (with its link to the manual) and then copy extra text from the [`MongoDB\Driver\Manager::startSession()`](http://php.net/manual/en/mongodb-driver-manager.startsession.php) docs:

> If true, each operation in the session will be causally ordered after the previous read or write operation. Set to false to disable causal consistency. Defaults to true.

docs/includes/apiargs-MongoDBClient-method-startSession-option.yaml
```diff
@@ -0,0 +1,43 @@
...
+interface: phpmethod
+operation: ~
+optional: true
+---
+source:
+  file: apiargs-common-option.yaml
+  ref: readConcern
+replacement:
+  resource: "database"
+  parent: "client"
```
jmikola: [`MongoDB\Driver\Manager::startSession()`](http://php.net/manual/en/mongodb-driver-manager.startsession.php) only supports one option. The shell API differs in that its session is basically a context object (like client, database, collection). For drivers, the session is just a value object that we pass around.

docs/reference/method/MongoDBClient-startSession.txt
```diff
@@ -0,0 +1,80 @@
...
+Definition
+----------
+
+.. phpmethod:: MongoDB\\Client::startSession()
+
+   Start a new client session for use with this client.
+
+   .. code-block:: php
+
+      function startSession(array $options []): MongoDB\Driver\Session
```
jmikola: `array $options = []`

docs/reference/method/MongoDBClient-startSession.txt
```diff
@@ -0,0 +1,80 @@
+===============================
+MongoDB\\Client::startSession()
+===============================
+
+.. default-domain:: mongodb
```
jmikola: Can we add `.. versionadded:: 1.3` above this (separated by a empty line on top and bottom)?

docs/reference/method/MongoDBClient-startSession.txt
```diff
@@ -0,0 +1,80 @@
...
+
+The output would then resemble::
+
+   object(MongoDB\Driver\Session)#2043 (4) {
+     ["logicalSessionId"]=>
+     array(1) {
+       ["id"]=>
+       object(MongoDB\BSON\Binary)#225 (2) {
+         ["data"]=>
+         string(16) "��)�G���I�Hc"
```
jmikola: Can we replace this with `string(16) "................"`, just in case there are issues with rendering raw binary bytes in the manual?

docs/reference/method/MongoDBClient-startSession.txt
```diff
@@ -0,0 +1,80 @@
...
+     bool(true)
+     ["operationTime"]=>
+     NULL
+   }
+
+See Also
+--------
+
+- :manual:`startSession </reference/command/startSession>` command reference in
+  the MongoDB manual
```
jmikola:
```
- :php:`MongoDB\\Driver\\WriteResult::getMatchedCount() <manual/en/mongodb-driver-manager.startsession.php>`
- :ref:`Causal Consistency <causal-consistency>` in the MongoDB manual
```

Drivers don't actually execute the `startSession` command. Instead, session IDs are generated client-side and created implicitly when the server first sees them.

src/Client.php
```diff
@@ -256,4 +256,15 @@ public function selectDatabase($databaseName, array $options = [])
...
         return new Database($this->manager, $databaseName, $options);
     }
+
+    /**
+     * Start a new client session.
+     * @see http://php.net/manual/en/mongodb-driver-manager.startsession
+     * @param array  $options      Session options
+     * @return MongoDB\Driver\Session
+     */
+    public function startSession($options = [])
```
jmikola: `array` type hint for `$options`

src/Client.php
```diff
@@ -256,4 +256,15 @@ public function selectDatabase($databaseName, array $options = [])
 
         return new Database($this->manager, $databaseName, $options);
     }
+
+    /**
+     * Start a new client session.
+     * @see http://php.net/manual/en/mongodb-driver-manager.startsession
```
jmikola: Line break before `@see` and ".php" at the end of the URL (avoids a redirect)

tests/ClientFunctionalTest.php
```diff
@@ -97,4 +97,12 @@ private function assertDatabaseExists($databaseName, $callback = null)
             call_user_func($callback, $foundDatabase);
         }
     }
+
+    public function testStartSession()
+    {
+        if (version_compare($this->getFeatureCompatibilityVersion(), '3.6', '<')) {
+            $this->markTestSkipped('startSession() is only supported on FCV 3.6 or higher');
+        }
+        $session = $this->client->startSession();
```
jmikola: `$this->assertInstanceOf('MongoDB\Driver\Session', $this->client->startSession())`

jmikola: The assertion below is now creating a new object, so the `$session` variable assigned above is no longer used and can be deleted.

docs/reference/method/MongoDBClient-startSession.txt
```diff
@@ -0,0 +1,83 @@
...
+Return Values
+-------------
+
+A :php:`MongoDB\Driver\Session <mongodb-driver-session>`
+
+Errors/Exceptions
+-----------------
+
+.. include:: /includes/extracts/error-unsupportedexception.rst
+.. include:: /includes/extracts/error-invalidargumentexception.rst
```
jmikola: This should be `error-driver-invalidargumentexception.rst`. Since the library method immediately calls the driver, any exception thrown is going to be the driver version.

docs/reference/method/MongoDBClient-startSession.txt
```diff
@@ -0,0 +1,83 @@
...
+
+Return Values
+-------------
+
+A :php:`MongoDB\Driver\Session <mongodb-driver-session>`
+
+Errors/Exceptions
+-----------------
+
+.. include:: /includes/extracts/error-unsupportedexception.rst
```
jmikola: This can be removed. Only InvalidArgumentException and RuntimeException are listed in the [PHP.net docs](http://php.net/manual/en/mongodb-driver-manager.startsession.php).

PR #478, opened 2018-02-02
PHPLIB-304 and PHPLIB-282: Deprecate inline result methods for aggregate and drop support for MongoDB 2.4

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-304

### DIFF
--- a/docs/includes/apiargs-MongoDBCollection-method-aggregate-option.yaml
+++ b/docs/includes/apiargs-MongoDBCollection-method-aggregate-option.yaml
@@ -16,11 +16,6 @@ description: |
   Specifies the initial batch size for the cursor. A batchSize of ``0`` means an
   empty first batch and is useful for quickly returning a cursor or failure
   message without doing significant server-side work.
-
-  .. note::
-
-     This is not supported for inline aggregation results (i.e. ``useCursor``
-     option is ``false`` or the server version is < 2.6).
 interface: phpmethod
 operation: ~
 optional: true
@@ -83,12 +78,6 @@ type: boolean
 description: |
   Indicates whether the command will request that the server provide results
   using a cursor. The default is ``true``.
-
-  For MongoDB version 2.6 or later, ``useCursor`` allows users to turn off
-  cursors if necessary to aid in replica set or shard cluster upgrades.
-
-  ``useCursor`` is ignored for MongoDB versions prior to 2.6 as aggregation
-  cursors are not available.
 interface: phpmethod
 operation: ~
 optional: true
--- a/docs/includes/apiargs-MongoDBCollection-method-createIndex-option.yaml
+++ b/docs/includes/apiargs-MongoDBCollection-method-createIndex-option.yaml
@@ -69,12 +69,8 @@ arg_name: option
 name: 2dsphereIndexVersion
 type: integer
 description: |
-  Specifies the :manual:`version of a 2dsphere </core/2dsphere>` index to
-  create.
-
-  MongoDB 2.6 introduced version 2 of 2dsphere indexes. Version 2 is the default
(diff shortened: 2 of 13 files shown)


docs/includes/apiargs-MongoDBCollection-method-aggregate-option.yaml
```diff
@@ -84,11 +79,8 @@ description: |
   Indicates whether the command will request that the server provide results
   using a cursor. The default is ``true``.
 
-  For MongoDB version 2.6 or later, ``useCursor`` allows users to turn off
-  cursors if necessary to aid in replica set or shard cluster upgrades.
-
-  ``useCursor`` is ignored for MongoDB versions prior to 2.6 as aggregation
-  cursors are not available.
+  ``useCursor`` allows users to turn off cursors if necessary to aid in replica
+  shard cluster upgrades.
```
jmikola: I think we can remove this sentence entirely.

I contemplated adding a line to say that `useCursor: false` is no longer supported as of MongoDB 3.6, but it is technically still relevant for `explain` support ([PHPLIB-172](https://jira.mongodb.org/browse/PHPLIB-172), so it'd be better to keep this short and sweet with just the first sentence above.

docs/includes/apiargs-MongoDBCollection-method-createIndex-option.yaml
```diff
@@ -72,9 +72,8 @@ description: |
   Specifies the :manual:`version of a 2dsphere </core/2dsphere>` index to
   create.
 
-  MongoDB 2.6 introduced version 2 of 2dsphere indexes. Version 2 is the default
-  version of 2dsphere indexes created in MongoDB 2.6 and later versions.
-  ``2dsphereIndexVersion`` enables you to override the default version 2.
+  Version 2 is the default version of 2dsphere indexes. ``2dsphereIndexVersion``
+  enables you to override the default version 2.
```
jmikola: It looks like MongoDB 3.2 changed the default version to 3, so this is quite out of date :)

I think we can just remove this paragraph entirely and keep a single sentence above. The current language might encourage users to supply this option, so I'd suggest we change it to:

```
Overrides the server's default version for  a :manual:`2dsphere </core/2dsphere>` index.
```

I noticed the MongoDB manual was also outdated and reported that in [DOCS-11278](https://jira.mongodb.org/browse/DOCS-11278).

src/Model/TypeMapArrayIterator.php
```diff
@@ -23,10 +23,8 @@
 /**
  * Iterator for applying a type map to documents in inline command results.
  *
- * This iterator may be used to apply a type map to an array of documents
- * returned by a database command (e.g. aggregate on servers < 2.6) and allows
- * for functional equivalence with commands that return their results via a
- * cursor (e.g. aggregate on servers >= 2.6).
+ * This iterator  allows for functional equivalence with commands that return
+ * their results via a cursor (e.g. aggregate).
```
jmikola: I think we can drop this paragraph entirely.

src/Operation/CreateIndexes.php
```diff
@@ -180,22 +172,4 @@ private function executeCommand(Server $server)
...
-    /**
-     * Create one or more indexes for the collection by inserting into the
-     * "system.indexes" collection (MongoDB <2.6).
-     *
-     * @param Server $server
-     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
-     */
-    private function executeLegacy(Server $server)
-    {
-        $bulk = new Bulk(['ordered' => true]);
```
jmikola: This should allow removal of `use MongoDB\Driver\BulkWrite as Bulk;` up top.

PR #479, opened 2018-02-05
PHPLIB-322: Add resume logic to ChangeStream::rewind()

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-322

### DIFF
--- a/src/ChangeStream.php
+++ b/src/ChangeStream.php
@@ -128,7 +128,7 @@ class ChangeStream implements Iterator
                 $resumable = true;
             }
             if ($e->getCode() === self::CURSOR_NOT_FOUND) {
-               $resumable = true;
+                $resumable = true;
             }
             if ($e instanceof ConnectionTimeoutException) {
                 $resumable = true;
--- a/src/ChangeStream.php
+++ b/src/ChangeStream.php
@@ -117,7 +117,26 @@ class ChangeStream implements Iterator
      */
     public function rewind()
     {
-        $this->csIt->rewind();
+        $resumable = false;
+        try {
+            $this->csIt->rewind();
+            if ($this->valid()) {
+                $this->extractResumeToken($this->csIt->current());
+            }
+        } catch (RuntimeException $e) {
+            if (strpos($e->getMessage(), "not master") !== false) {
+                $resumable = true;
+            }
+            if ($e->getCode() === self::CURSOR_NOT_FOUND) {
+               $resumable = true;
+            }
+            if ($e instanceof ConnectionTimeoutException) {
+                $resumable = true;
+            }
+        }
+        if ($resumable) {
+            $this->resume();
+        }
     }
 
     /**
--- a/tests/Operation/WatchFunctionalTest.php
+++ b/tests/Operation/WatchFunctionalTest.php
@@ -73,9 +73,6 @@ class WatchFunctionalTest extends FunctionalTestCase
         $this->assertSameDocument($expectedResult, $changeStream->current());
     }
 
-    /**
-     * @todo test that rewind() also resumes once PHPLIB-322 is implemented
-     */
     public function testNextResumesAfterConnectionException()
     {
         /* In order to trigger a dropped connection, we'll use a new client with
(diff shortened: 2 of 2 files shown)


src/ChangeStream.php
```diff
@@ -117,7 +117,26 @@ public function next()
...
+            $this->csIt->rewind();
+            if ($this->valid()) {
+                $this->extractResumeToken($this->csIt->current());
+            }
+        } catch (RuntimeException $e) {
+            if (strpos($e->getMessage(), "not master") !== false) {
+                $resumable = true;
+            }
+            if ($e->getCode() === self::CURSOR_NOT_FOUND) {
+               $resumable = true;
```
jmikola: Extra space for indentation.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -269,13 +315,26 @@ public function testNextCannotExtractResumeToken()
         $operation = new Watch($this->manager, $this->getDatabaseName(), $this->getCollectionName(), $pipeline, ['maxAwaitTimeMS' => 100]);
         $changeStream = $operation->execute($this->getPrimaryServer());
 
-        $changeStream->rewind();
```
jmikola: I assume this was removed because `rewind()` now throws its own exception and would prevent us from testing `next()`

I'd suggest replacing with:

```
/* Note: we intentionally do not start iteration with rewind() to ensure
 * that we test extraction functionality within next(). */
```

Borrowed from the socket exception test.

jmikola: Side thought: does this actually throw or does it only introduce a delay with its own `getMore`? Since we've yet to insert a document, I'd expect `rewind()` to not find any result to attempt extraction.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -320,6 +379,54 @@ public function testMaxAwaitTimeMS()
...
+        $this->insertDocument(['_id' => 1, 'x' => 'foo']);
+
+        $this->killChangeStreamCursor($changeStream);
+
+        $changeStream->rewind();
+        $this->assertFalse($changeStream->valid());
+        $this->assertNull($changeStream->current());
+    }
+
+    public function testResumeAfterKillThenOperation()
```
jmikola: `testRewindExtractsResumeTokenAndNextResumes()`

tests/Operation/WatchFunctionalTest.php
```diff
@@ -320,6 +379,54 @@ public function testMaxAwaitTimeMS()
         $this->assertTrue($changeStream->valid());
     }
 
+    public function testResumeAfterKillThenNoOperations()
```
jmikola: `testRewindResumesAfterCursorNotFound()`

tests/Operation/WatchFunctionalTest.php
```diff
@@ -320,6 +379,54 @@ public function testMaxAwaitTimeMS()
...
+    {
+        $operation = new Watch($this->manager, $this->getDatabaseName(), $this->getCollectionName(), [], ['maxAwaitTimeMS' => 100]);
+        $changeStream = $operation->execute($this->getPrimaryServer());
+
+        $this->insertDocument(['_id' => 1, 'x' => 'foo']);
```
jmikola: I think we can simply omit this line. Asserting that `rewind()` doesn't throw an exception is enough to test that it can resume after not finding a cursor.

tests/Operation/WatchFunctionalTest.php
kvwalker: You're right, the rewind() doesn't throw because it doesn't attempt extraction.

jmikola: This now makes for very amusing `diff` output.

tests/Operation/WatchFunctionalTest.php
jmikola: This comment was resolved. No longer applicable to whatever line this points to.

PR #486, opened 2018-02-06
Add missing version info for 1.3.0 release

jmikola:
I took a look through docs changes in the [diff between `v1.2` and `master`](https://github.com/mongodb/mongo-php-library/compare/v1.2...master) and I believe this is everything.

Tested with a local docs build.


docs/reference/method/MongoDBClient-startSession.txt
```diff
@@ -2,10 +2,10 @@
 MongoDB\\Client::startSession()
 ===============================
 
-.. default-domain:: mongodb
-
 .. versionadded:: 1.3
 
+.. default-domain:: mongodb
```
jmikola: This rendered fine before. Just re-ordering it for consistency with other RST files.

docs/reference/method/MongoDBCollection-watch.txt
```diff
@@ -1,6 +1,8 @@
-================================
+============================
 MongoDB\\Collection::watch()
-================================
+============================
```
jmikola: I didn't notice build warnings for the previous header lines (not sure if strict warnings are being suppressed), but it is something RST can sometimes complain about so this was a quick fix.

docs/includes/apiargs-MongoDBCollection-method-dropIndex-option.yaml
```diff
@@ -5,6 +5,8 @@ source:
 source:
```
kvwalker: Should maxTimeMS also show `versionadded:: 1.3` here and in dropIndexes-option??

docs/includes/apiargs-MongoDBCollection-method-watch-option.yaml
```diff
@@ -70,6 +70,8 @@ optional: true
 source:
   file: apiargs-common-option.yaml
   ref: session
+post: |
+  .. versionadded:: 1.3
```
kvwalker: Do you need to specify this here even though every option here was added in version 1.3 since Watch is new?

docs/includes/apiargs-MongoDBCollection-method-dropIndex-option.yaml
jmikola: Ah, I missed https://github.com/mongodb/mongo-php-library/commit/3e83aa78ef5cf7cd52492e68899f3fe62e98c333. Thanks for catching that!

docs/includes/apiargs-MongoDBCollection-method-watch-option.yaml
jmikola: Definitely not. This was likely the result of an over aggressive search/replace.

PR #491, opened 2018-02-08
PHPLIB-252: Allow seeking to advance chunk iterator instead of always re-querying

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-252

### DIFF
--- a/src/GridFS/ReadableStream.php
+++ b/src/GridFS/ReadableStream.php
@@ -186,9 +186,26 @@ class ReadableStream
         $this->chunkOffset = (integer) floor($offset / $this->chunkSize);
         $this->bufferOffset = $offset % $this->chunkSize;
 
-        if ($lastChunkOffset !== $this->chunkOffset) {
+        /* If we are seeking to a previous chunk, we need to reinitialize the
+         * chunk iterator.
+         */
+        if ($lastChunkOffset > $this->chunkOffset) {
             $this->buffer = null;
             $this->chunksIterator = null;
+            return;
+        }
+
+        /* If we are seeking to a subsequent chunk, we do not need to
+         * reinitalize the chunk iterator. Instead, we can simply move forward
+         * to $this->chunkOffset.
+         */
+        $numChunks = $this->chunkOffset - $lastChunkOffset;
+        $i = 0;
+        if ($this->chunksIterator !== null) {
+            while ($i < $numChunks) {
+                $this->chunksIterator->next();
+                $i++;
+            }
         }
     }
 
--- a/src/GridFS/ReadableStream.php
+++ b/src/GridFS/ReadableStream.php
@@ -186,11 +186,21 @@ class ReadableStream
         $this->chunkOffset = (integer) floor($offset / $this->chunkSize);
         $this->bufferOffset = $offset % $this->chunkSize;
 
+        if ($lastChunkOffset === $this->chunkOffset) {
+            return;
+        }
+
+        if ($this->chunksIterator === null) {
+            return;
+        }
+
+        // Clear the buffer since the current chunk will be changed
+        $this->buffer = null;
+
         /* If we are seeking to a previous chunk, we need to reinitialize the
          * chunk iterator.
          */
         if ($lastChunkOffset > $this->chunkOffset) {
-            $this->buffer = null;
             $this->chunksIterator = null;
             return;
         }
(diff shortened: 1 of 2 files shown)


tests/GridFS/ReadableStreamFunctionalTest.php
```diff
@@ -189,4 +189,32 @@ public function testSeekOutOfRange()
 
         $stream->seek(11);
     }
+
+    public function testSeekPreviousChunk()
```
kvwalker: I'm not sure of the best way to show that `chunksIterator` was rewound.

src/GridFS/ReadableStream.php
```diff
@@ -186,9 +186,26 @@ public function seek($offset)
...
+
+        /* If we are seeking to a subsequent chunk, we do not need to
+         * reinitalize the chunk iterator. Instead, we can simply move forward
+         * to $this->chunkOffset.
+         */
+        $numChunks = $this->chunkOffset - $lastChunkOffset;
+        $i = 0;
+        if ($this->chunksIterator !== null) {
+            while ($i < $numChunks) {
+                $this->chunksIterator->next();
```
jmikola: If we look at the previous code, both `$buffer` and `$chunksIterator` were reset when the chunk offset changed. When `readBytes()` is called, it initializes those properties again if they are `null`.

The current change in this PR leaves `$buffer` as-is so long as the current chunk offset is >= the previous offset. That becomes problematic if you do end up advancing to the next chunk, as you do here.

The test case in https://github.com/kvwalker/mongo-php-library/pull/5/files#diff-40e5981925700af63fd6c7a3417a12cdR294 demonstrates a potential failure where `$buffer` retains the contents of the first chunk and we end up reading "abcd" instead of "efgh".

src/GridFS/ReadableStream.php
```diff
@@ -186,9 +186,26 @@ public function seek($offset)
...
+            return;
+        }
+
+        /* If we are seeking to a subsequent chunk, we do not need to
+         * reinitalize the chunk iterator. Instead, we can simply move forward
+         * to $this->chunkOffset.
+         */
+        $numChunks = $this->chunkOffset - $lastChunkOffset;
+        $i = 0;
+        if ($this->chunksIterator !== null) {
```
jmikola: This can be [refactored as a guard clause with an early return](https://refactoring.guru/replace-nested-conditional-with-guard-clauses) instead of nesting `while`.

src/GridFS/ReadableStream.php
```diff
@@ -186,9 +186,26 @@ public function seek($offset)
...
+            while ($i < $numChunks) {
```
jmikola: Given that `$i` is initialized a couple of lines up and the loop includes `$i++`, a `for` loop might be more concise.

src/GridFS/ReadableStream.php
```diff
@@ -186,9 +186,26 @@ public function seek($offset)
         $this->chunkOffset = (integer) floor($offset / $this->chunkSize);
         $this->bufferOffset = $offset % $this->chunkSize;
 
-        if ($lastChunkOffset !== $this->chunkOffset) {
+        /* If we are seeking to a previous chunk, we need to reinitialize the
```
jmikola: An idea for an early guard statement might be:

```php
if ($lastChunkOffset === $this->chunkOffset) {
    return;
}

// Clear the buffer as the current chunk will be changed
$this->buffer = null;

// If we are seeking to a previous chunk...
```

If the chunk isn't changing at all, we have nothing more to do after updating `$chunkOffset` and `$bufferOffset` a few lines up. Otherwise, we know the chunk will change, so the existing `$buffer` is no longer relevant and should be cleared (`readBytes()` will reinitialize it when needed). That should address the bug I mentioned below and in the related test PR.

From there, I think your existing logic is correct:

 * If we're seeking to a previous chunk, reset `$chunksIterator` and return (`readBytes()` will reinitialize it when needed)
 * If we're seeking to a later chunk, advance until we reach that chunk

This should also make it easier to add logic that prefers resetting `$chunksIterator` if we're seeking _very_ far ahead in the file stream. This is something I alluded to in the description of [PHPLIB-252](https://jira.mongodb.org/browse/PHPLIB-252), which we can revisit later on. The idea there is that if we're seeking forward, advancing chunks withing the same cursor batch is free (the driver already has the data client-side) and a `getMore` still preferable to a new query (the server already has the data and we avoid the overhead of a new query and `killCursors` command for discarding the old cursor). However, if we're dealing with a very large file and want to seek to the end, it may very well be cheaper to query anew instead of walk the cursor with potentially many `getMore` commands.

src/GridFS/ReadableStream.php
```diff
@@ -186,11 +186,21 @@ public function seek($offset)
         $this->chunkOffset = (integer) floor($offset / $this->chunkSize);
         $this->bufferOffset = $offset % $this->chunkSize;
 
+        if ($lastChunkOffset === $this->chunkOffset) {
+            return;
+        }
+
+        if ($this->chunksIterator === null) {
+            return;
```
jmikola: Good catch!

PR #493, opened 2018-02-09
PHPLIB-324: Cache extension installation for Travis CI jobs

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-324

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ cache:
   directories:
     - ${HOME}/.cache/pip
     - ${HOME}/.composer/cache/files
+    - ${HOME}/php-ext
 
 env:
   global:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -46,6 +46,22 @@ before_install:
   - mongod --version
   - mongo-orchestration --version
   - export MO_PATH=`python -c 'import mongo_orchestration; from os import path; print(path.dirname(mongo_orchestration.__file__));'`
+  - |
+    # tpecl is a helper to compile and cache php extensions
+      tpecl () {
+          local ext_name=ext-mongodb
+          local ext_so=mongodb.so
```
jmikola: I'd suggest we just copy the `tpecl()` definition from https://github.com/symfony/symfony/blob/32cc2e06a326b1dd888c02e1cc48988a14cb9a65/.travis.yml#L101 completely and parametrize the extension name and merge when invoking the `tpecl` function.

.travis.yml
```diff
@@ -46,6 +46,22 @@ before_install:
...
+          local ext_cache=~/php-ext/$(basename $ext_dir)/$ext_name
+          if [[ -e $ext_cache/$ext_so ]]; then
+              echo extension = $ext_cache/$ext_so >> $INI
+          else
+              mkdir -p $ext_cache
+              echo yes | pecl install -f $ext_name &&
+              cp $ext_dir/$ext_so $ext_cache
+          fi
+      }
+      export -f tpecl
```
jmikola: `before_script` below is still using `pecl install`, so that will need to be changed to `tpecl` (see: https://github.com/symfony/symfony/blob/32cc2e06a326b1dd888c02e1cc48988a14cb9a65/.travis.yml#L137). I don't think we need to bother with `tfold` right now to [collapse Travis CI log output](https://blog.travis-ci.com/2013-05-22-improving-build-visibility-log-folds/); that's something we can always follow up with in the future.

Calling `tpecl` with the extension name (usual argument to `pecl install`) and `.so` filename should suffice:

```
tpecl mongodb-${DRIVER_VERSION} mongodb.so
```

.travis.yml
```diff
@@ -46,11 +46,27 @@ before_install:
   - mongod --version
   - mongo-orchestration --version
   - export MO_PATH=`python -c 'import mongo_orchestration; from os import path; print(path.dirname(mongo_orchestration.__file__));'`
+  - |
+    # tpecl is a helper to compile and cache php extensions
+      tpecl () {
+          local ext_name=$1
+          local ext_so=$2
+          local ext_dir=$(php -r "echo ini_get('extension_dir');")
+          local ext_cache=~/php-ext/$(basename $ext_dir)/$ext_name
```
jmikola: Ah, I missed this earlier but we'll also need to add `php-ext` to the list of cached directories further up in the Travis config. See: https://github.com/symfony/symfony/blob/32cc2e06a326b1dd888c02e1cc48988a14cb9a65/.travis.yml#L35. The following should suffice:

```
- ${HOME}/php-ext
```

.travis.yml
```diff
@@ -46,11 +47,27 @@ before_install:
...
+          if [[ -e $ext_cache/$ext_so ]]; then
+              echo extension = $ext_cache/$ext_so >> $INI
```
jmikola: This depends on an `INI` variable being defined. Let's add this line before defining `tpecl()`: https://github.com/symfony/symfony/blob/32cc2e06a326b1dd888c02e1cc48988a14cb9a65/.travis.yml#L91

PR #497, opened 2018-02-14
PHPLIB-156: Allow operations to be explained

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-156


src/Operation/Explainable.php
```diff
@@ -0,0 +1,39 @@
...
+ * find, findAndModify, delete, and update).
+ *
+ * @internal
+ */
+interface Explainable extends Executable
+{
+    /**
+     * Explain the operation.
+     *
+     * @param Server $server
```
kvwalker: I added $server to explain() so explain can be executed inside of the operation classes.

src/Operation/Find.php
```diff
@@ -293,6 +294,38 @@ public function execute(Server $server)
...
+     * @return array|object
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function explain($server, $command, $options = [])
+    {
+        if ( ! isset($options['verbosity'])) {
+            $options['verbosity'] = 'allPlansExecution';
+        }
+
+        $cmd = new Command(['explain' => ['find' => $this->collectionName, 'query' => $command], 'verbosity' => $options['verbosity']]);
```
kvwalker: Is it correct to have `'query' => $command` here? I wasn't sure based on the [explain reference](https://docs.mongodb.com/manual/reference/command/explain/) since there are no examples for `find`.

src/Operation/Count.php
```diff
@@ -34,7 +34,7 @@
  * @see \MongoDB\Collection::count()
  * @see http://docs.mongodb.org/manual/reference/command/count/
  */
-class Count implements Executable
+class Count implements Explainable
```
jmikola: Extending multiple classes is not permitted, but we can implement multiple interfaces. I realize Executable is `@internal` and not used for anything, but we don't need to remove it in this PR.

src/Operation/Find.php
jmikola: This didn't occur to me yesterday, but it's the same issue we have with explaining `update` and `delete` commands. Since those don't route through `executeCommand()`, PHPLIB never creates the actual command documents for those. We'll need logic to create command documents manually, which will be used purely for passing into the `explain` command.

For the time being, let's just focus on things that use `executeCommand()` (and related methods), such as Count.

src/Operation/Explainable.php
jmikola: One issue I see with this approach is that each Explainable instance will now need to contain the same logic for executing the `explain` command on its own. That will entail quite a bit of duplicated code.

The get the most leverage out of code re-use, we can break down the many variations of `explain` into the greatest common denominator: the command document. Every variation of `explain` is going to have the same overhead for parsing the `verbosity` option and executing the explain command with `Server::executeCommand()` (`explain` doesn't take a read concern itself, so I don't think `executeReadCommand()` is needed). Then, we create an interface that can provide that common command document.

This is the design I suggested yesterday:

```php
interface Explainable {
    /* Renaming this from getCommand(), since that typically returns a
     * MongoDB\Driver\Command instance and we're more interested in
     * the document itself (Command::__construct()'s first argument). */
    function getCommandDocument();
}

class Explain implements Executable {
    /* The Explainable knows what collection it targets, so we only need
     * a database to run `explain` on. Alternatively, we might decide to also
     * pull the database from Explainable somehow, since all Operations we
     * might explain also require a database name.
     *
     * Options will at least be verbosity (the only documented explain option)
     * and typeMap, which we can apply to the cursor before the execute()
     * method returns current($cursor->toArray()) or $cursor->toArray()[0]
     * (both are equivalent). */
    public function __construct($databaseName, Explainable $explainable, array $options = []);

    public function execute(Server $server);
}
```

src/Operation/Count.php
```diff
@@ -162,6 +162,39 @@ public function execute(Server $server)
...
+     * @return array|object
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function explain($server, $command, $options = [])
+    {
+        if ( ! isset($options['verbosity'])) {
+            $options['verbosity'] = 'allPlansExecution';
+        }
+
+        $cmd = new Command(['explain' => ['count' => $this->collectionName, 'query' => $command], 'verbosity' => $options['verbosity']]);
```
jmikola: The value of `explain` here will need to be the entire document used to create the `MongoDB\Driver\Command` object in `createCommand()` below. Specifically, the [constructor's](http://php.net/manual/en/mongodb-driver-command.construct.php) first argument.

src/Operation/Count.php
```diff
@@ -162,6 +162,39 @@ public function execute(Server $server)
...
+     * @param Server $server
+     * @param $command
+     * @param array $options
+     * @return array|object
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function explain($server, $command, $options = [])
+    {
+        if ( ! isset($options['verbosity'])) {
+            $options['verbosity'] = 'allPlansExecution';
```
jmikola: While I doubt this will break sharding (like MapReduce recently), it may be safer to simply omit this option unless the user specifies a value. I believe the server will enforce this default if `verbosity` is absent in the `explain` command document.

Side note: we can also create constants for the three known modes on the Explain class, as you did with Watch and the lookup modes. Users don't _need_ to use those constants, but we can use them in tests and it's a nice reminder for those browsing the code.

src/Operation/Count.php
```diff
@@ -162,6 +162,39 @@ public function execute(Server $server)
...
+        if (empty($command)) {
+            $cmd = new Command(['explain' => ['count' => $this->collectionName], 'verbosity' => $options['verbosity']]);
+        }
+
+        $result = $server->executeCommand($this->databaseName, $cmd);
+
+        $resultArray = get_object_vars($result->toArray()[0]); // cast $result to array
+
+        if ($options['verbosity'] === 'queryPlanner') {
+            return ['queryPlanner' => $resultArray['queryPlanner']];
```
jmikola: Is there a reason for not returning the result document as-is? There are a few other operation classes where we do so (e.g. CreateCollection and drop operations). There may be some extra fields, such as `ok` and gossiped cluster time, but I don't think we need to insulate users from that. We can always look at crafting an ExplainResult class down the line if the raw result document is too noisy.

src/Operation/Count.php
```diff
@@ -162,6 +162,39 @@ public function execute(Server $server)
...
+
+        $cmd = new Command(['explain' => ['count' => $this->collectionName, 'query' => $command], 'verbosity' => $options['verbosity']]);
+
+        if (empty($command)) {
+            $cmd = new Command(['explain' => ['count' => $this->collectionName], 'verbosity' => $options['verbosity']]);
+        }
+
+        $result = $server->executeCommand($this->databaseName, $cmd);
+
+        $resultArray = get_object_vars($result->toArray()[0]); // cast $result to array
```
jmikola: Applying a type map of `['root' => 'array']` to the cursor (`$result` in this case) would be the more conventional way to work with an array here, as that allows the driver to do the conversion on its own and forego this object-to-array cast.

That said, if we want to give users control over the returned document, we should solicit a `typeMap` option and use that to apply to the array. If we think of the Explain operation as returning only a single document, it's quite similar to FindOne in that respect. With FindOne, the `typeMap` option gets applied to the Find operation's cursor, and FindOne just unpacks and returns the first and only result document in that cursor.

Also a reminder that PHPLIB has its own default type maps (BSONDocument and BSONArray classes), so even though it may be uncommon for users to specify their own type maps, we use this option to keep the various operations consistent. That's why operations such as CreateCollection take a typeMap option, even though they return a raw command result.

src/Operation/Count.php
```diff
@@ -162,12 +162,27 @@ public function execute(Server $server)
         return (integer) $result->n;
     }
 
+    public function getCommandDocument()
+    {
+        return $this->createCommandDocument();
```
jmikola: I was originally going to suggest factoring out the private `createCommandDocument()` method and just turning it into this public method, but I think it's reasonable to keep them separate for now. That will allow the interface to be tweaked without disturbing the innards of each operation class.

I'm contemplating whether it may make sense to have some form of "explain" in the interface method name, but we'll have plenty of time to debate names before releasing 1.4.0. This is totally fine as-is :)

src/Operation/Explain.php
```diff
@@ -0,0 +1,73 @@
...
+ * @api
+ * @see \MongoDB\Collection::explain()
+ * @see http://docs.mongodb.org/manual/reference/command/explain/
+ */
+class Explain implements Executable
+{
+
+    const VERBOSITY_ALL_PLANS = 'allPlansExecution';
+    const VERBOSITY_EXEC_STATS = 'executionStats';
+    const VERBOSITY_QUERY = 'queryPlanner';
```
jmikola: Nothing to change now, but I'd like to think over whether we should use unabridged names for constants, as we do in other operations.

kvwalker:
I've handled the command document creation for 'find', 'findOne', 'findAndModify', and 'distinct'. I still need to check on the edge cases for the 'find' commands.

src/Operation/Explain.php
```diff
@@ -29,33 +31,66 @@
...
         $this->options = $options;
     }
 
     public function execute(Server $server)
     {
+        if (! \MongoDB\server_supports_feature($server, self::$wireVersionForExplain)) {
+            throw UnsupportedException::explainNotSupported();
+        }
+
+        if ($this->explainable instanceOf \MongoDB\Operation\Distinct) {
```
jmikola: PHP's lexer doesn't seem to mind, but [`instanceof`](http://php.net/instanceof) is the canonical form.

src/Operation/Explain.php
```diff
@@ -29,33 +31,66 @@
...
+            if (! \MongoDB\server_supports_feature($server, self::$wireVersionForDistinct)) {
+                throw UnsupportedException::explainNotSupported();
+            }
+        }
+
+        if ($this->explainable instanceOf \MongoDB\Operation\FindAndModify) {
+            if (! \MongoDB\server_supports_feature($server, self::$wireVersionForFindAndModify)) {
```
jmikola: These can easily be combined into a single condition with `&&`. It'd only make sense to keep them separate if we expected to do other things for FindAndModify instances apart from the wire version check.

tests/Operation/ExplainFunctionalTest.php
```diff
@@ -0,0 +1,222 @@
...
+            ['x' => 2],
+            ['y' => 3]
+        ]);
+        $insertMany->execute($this->getPrimaryServer());
+
+        $operation = new Count($this->getDatabaseName(), $this->getCollectionName(), ['x' => ['$gte' => 1]], []);
+        $explainOperation = new Explain($this->getDatabaseName(), $operation, ['verbosity' => Explain::VERBOSITY_ALL_PLANS, 'typeMap' => ['root' => 'array']]);
+        $result = $explainOperation->execute($this->getPrimaryServer());
+
+        $this->assertTrue(array_key_exists('queryPlanner', $result));
```
jmikola: [`assertArrayHasKey()`](https://phpunit.de/manual/current/en/appendixes.assertions.html#appendixes.assertions.assertArrayHasKey) can be used here. In addition to being more concise, it also also PHPUnit to use a more intelligent error message if the assertion fails (e.g. "array didn't have the key we expected" instead of "expected true and got false").

tests/Operation/ExplainFunctionalTest.php
```diff
@@ -0,0 +1,222 @@
...
+        ]);
+        $insertMany->execute($this->getPrimaryServer());
+
+        $operation = new Count($this->getDatabaseName(), $this->getCollectionName(), ['x' => ['$gte' => 1]], []);
+        $explainOperation = new Explain($this->getDatabaseName(), $operation, ['typeMap' => ['root' => 'array']]);
+        $result = $explainOperation->execute($this->getPrimaryServer());
+
+        $this->assertTrue(array_key_exists('queryPlanner', $result));
+        $this->assertTrue(array_key_exists('executionStats', $result));
+        $this->assertTrue(array_key_exists('allPlansExecution', $result['executionStats']));
```
jmikola: According to the server docs, the default `verbosity` is `allPlansExecution`. I believe that makes this test very similar to `testCountAllPlansExecution()`.

In the interest of reducing code duplication throughout ExplainFunctionalTest, perhaps you can use a data provider here. I think there are a few variables:

 * `verbosity` string (if `null` is provided, your test can interpret that as an absent option)
 * Whether an `executionStats` key is expected in the result
 * Whether an `allPlansExecution` key is expected in `executionStats`

Assuming the verbosity behavior holds true for all commands, you could then write a single test method per command. To further cut down on duplication, you can also create a private `assertExplainResult()` helper that takes booleans for whether `executionStats` and `allPlansExecution` are expected.

tests/Operation/ExplainTest.php
```diff
@@ -0,0 +1,45 @@
...
+
+class ExplainTest extends TestCase
+{
+    /**
+     * @expectedException MongoDB\Exception\InvalidArgumentException
+     * @dataProvider provideInvalidConstructorOptions
+     */
+    public function testConstructorOptionTypeChecksForCount(array $options)
+    {
+        $explainable = new Count($this->getDatabaseName(), $this->getCollectionName(),[]);
```
jmikola: Since you know that constructor option validation doesn't care about the Explainable's concrete type, you shouldn't need separate tests for each possibility. If the constructor _did_ care for some reason, we'd certainly want to test that.

We can use PHPUnit to create a mock object from the interface itself:

```php
$explainable = $this->createMock('MongoDB\Operation\Explainable');
```

For more info on this, see [Test Doubles](https://phpunit.de/manual/current/en/test-doubles.html) in the PHPUnit docs.

tests/Operation/ExplainFunctionalTest.php
```diff
@@ -0,0 +1,195 @@
...
+     * @param integer $n
+     */
+    private function createFixtures($n)
+    {
+        $bulkWrite = new BulkWrite(['ordered' => true]);
+
+        for ($i = 1; $i <= $n; $i++) {
+            $bulkWrite->insert([
+                '_id' => $i,
+                'x' => (object) ['foo' => 'bar'],
```
jmikola: If none of the tests need `x` to be an embedded document, can we change this to `'x' => (integer) ($i . $i)`? Then `testCount()` could easily use `createFixtures()` (like `testFind()`) and still have it's condition (`x >= 1`) make sense.

src/Operation/Find.php
```diff
@@ -293,33 +293,49 @@ public function execute(Server $server)
...
-     *
-     * @see http://php.net/manual/en/mongodb-driver-server.executequery.php
-     * @return array
+     * Construct a command document for Find
      */
-    private function createOptions()
+    private function createCommandDocument()
     {
-        $options = [];
+        $cmd = ['find' => $this->collectionName, 'filter' => new BSONDocument($this->filter)];
```
jmikola: I assume the intention here is to ensure that an array like `[]` still encodes as a BSON document. We can shortcut this with an `(object)` cast, which is exactly what [`BSONDocument::bsonSerialize()`](https://github.com/mongodb/mongo-php-library/blob/1.3.0/src/Model/BSONDocument.php#L72) would do (after copying `$this->filter` a couple of times and creating an ArrayObject instance).

src/Operation/Find.php
```diff
@@ -293,33 +293,49 @@ public function execute(Server $server)
...
+                $cmd['tailable'] = true;
+            }
+            if ($options['cursorType'] === self::TAILABLE_AWAIT) {
+                $cmd['tailable'] = true;
+                $cmd['awaitData'] = true;
+            }
         }
 
-        return $options;
+        return $cmd + $options;
```
jmikola: `modifiers` may need special handling here, since that isn't a `find` option at all.

It may help to look at what Query constructor does:

 * [`hint` option falls back to `$hint` modifier](https://github.com/mongodb/mongo-php-driver/blob/1.4.1/src/MongoDB/Query.c#L138)
 * [Other options fall back to similarly named modifiers](https://github.com/mongodb/mongo-php-driver/blob/1.4.1/src/MongoDB/Query.c#L270) (see macros linked with `else`)

One approach here might be to create a local array of pairs, each containing the original option and modifier fallback. Iterating on this array, you could then check if the modern option is not set and then assign the modifier's value. That is essentially what the `PHONGO_QUERY_OPT_X()` macros in PHPC do.

Normally, I'd suggest using `array_key_exists()` to be correct, but `isset()` should work fine in this case based on the logic in `createQuery()` (now `createFindOptions()`). This operation doesn't assign any options with `null` values. Differentiating `null` values from missing keys is the main reason we'd want to use the more-expensive `array_key_exists()` function.

src/Operation/Find.php
```diff
@@ -293,33 +293,49 @@ public function execute(Server $server)
...
+        if (empty($options)) {
+            return $cmd;
         }
 
-        if (isset($this->options['session'])) {
-            $options['session'] = $this->options['session'];
+        // maxAwaitTimeMS is a Query level option so should not be considered here
+        unset($options['maxAwaitTimeMS']);
+
+        if (isset($options['cursorType'])) {
```
jmikola: I believe `createFindOptions()` already handles this.

src/Operation/Find.php
```diff
@@ -293,33 +293,49 @@ public function execute(Server $server)
...
+        $options = $this->createFindOptions();
+
+        if (empty($options)) {
+            return $cmd;
         }
 
-        if (isset($this->options['session'])) {
-            $options['session'] = $this->options['session'];
+        // maxAwaitTimeMS is a Query level option so should not be considered here
+        unset($options['maxAwaitTimeMS']);
```
jmikola: Good catch.

src/Operation/Find.php
jmikola: Second thought: since you've already called `createFindOptions()` before this point, any modifiers will be under a `modifiers` option, which will have an array type. Therefore, you can pluck the relevant fallback values from there and finally `unset($options['modifiers'])` after the fact.

src/Operation/Find.php
```diff
@@ -293,33 +293,49 @@ public function execute(Server $server)
...
+     * Create options for the find query.
      *
-     * @return Query
+     * Note that these are separate from the options for executing the command,
+     * which are created in createOptions().
+     *
+     * @return array
      */
-    private function createQuery()
+    private function createFindOptions()
```
jmikola: I'd suggest `createQueryOptions()`.

src/Operation/Find.php
```diff
@@ -351,6 +367,27 @@ private function createQuery()
...
+        return $options;
+    }
+
+    /**
+     * Create options for executing the command.
+     *
+     * @see http://php.net/manual/en/mongodb-driver-server.executequery.php
+     * @return array
+     */
+    private function createOptions()
```
jmikola: `createExecuteOptions()` might be clearer. I might suggest this for any files where we'll now have multiple functions creating "options". I realize the doc block explains this, but a rename will ensure it's perfectly clear what the method does from the places it is being called.

tests/Operation/ExplainFunctionalTest.php
```diff
@@ -0,0 +1,195 @@
...
+use MongoDB\Operation\FindOne;
+use MongoDB\Operation\InsertMany;
+
+class ExplainFunctionalTest extends FunctionalTestCase
+{
+    public function setUp()
+    {
+        parent::setUp();
+        if (version_compare($this->getServerVersion(), '3.0.0', '<')) {
+            $this->markTestSkipped('Explain is not supported');
```
jmikola: "Explain command is not supported"

tests/Operation/ExplainFunctionalTest.php
```diff
@@ -0,0 +1,195 @@
...
+     */
+    public function testCount($verbosity, $executionStatsExpected, $allPlansExecutionExpected)
+    {
+        $insertMany = new InsertMany($this->getDatabaseName(), $this->getCollectionName(), [
+            ['x' => 0],
+            ['x' => 1],
+            ['x' => 2],
+            ['y' => 3]
+        ]);
+        $insertMany->execute($this->getPrimaryServer());
```
jmikola: This can probably use `createFixtures()` (see my related comment below).

tests/Operation/ExplainFunctionalTest.php
```diff
@@ -0,0 +1,195 @@
...
+        $this->assertExplainResult($result, $executionStatsExpected, $allPlansExecutionExpected);
+    }
+
+    /**
+     * @dataProvider provideVerbosityInformation
+     */
+    public function testDistinct($verbosity, $executionStatsExpected, $allPlansExecutionExpected)
+    {
+        if (version_compare($this->getServerVersion(), '3.2.0', '<')) {
+            $this->markTestSkipped('Distinct is not supported on servers with version < 3.2');
```
jmikola: "Explain command does not support distinct"

As-is, this reads as if Distinct itself isn't supported. Most of our skip messages don't repeat the version requirement, but you're welcome to add that in. Based on your change stream skip message, the following could work: "Explaining distinct command requires server version >= 3.2"

tests/Operation/ExplainFunctionalTest.php
```diff
@@ -0,0 +1,195 @@
...
+        $this->assertExplainResult($result, $executionStatsExpected, $allPlansExecutionExpected);
+    }
+
+    /**
+     * @dataProvider provideVerbosityInformation
+     */
+    public function testFindAndModify($verbosity, $executionStatsExpected, $allPlansExecutionExpected)
+    {
+        if (version_compare($this->getServerVersion(), '3.2.0', '<')) {
+            $this->markTestSkipped('FindAndModify is not supported on servers with version < 3.2');
```
jmikola: Same comment re: Distinct's skip message.

tests/Operation/ExplainFunctionalTest.php
```diff
@@ -0,0 +1,195 @@
...
+        $explainOperation = new Explain($this->getDatabaseName(), $operation, ['verbosity' => $verbosity, 'typeMap' => ['root' => 'array', 'document' => 'array']]);
+        $result = $explainOperation->execute($this->getPrimaryServer());
+
+        $this->assertExplainResult($result, $executionStatsExpected, $allPlansExecutionExpected);
+    }
+
+    /**
+     * @dataProvider provideVerbosityInformation
+     */
+    public function testFindMaxAwait($verbosity, $executionStatsExpected, $allPlansExecutionExpected)
```
jmikola: Is this just here to test how the `cursorType` option sets `tailable` and `awaitData` and ensure that `maxAwaitTimeMS` is stripped? This might make more sense single test (no data provider) that uses CommandObserver test to assert the structure of the document that ends up in the `explain.find` property of the command document.

tests/Operation/ExplainFunctionalTest.php
```diff
@@ -0,0 +1,195 @@
...
+        $databaseName = $this->getDatabaseName();
+        $cappedCollectionName = $this->getCollectionName();
+        $cappedCollectionOptions = [
+            'capped' => true,
+            'max' => 100,
+            'size' => 1048576,
+        ];
+
+        $operation = new CreateCollection($databaseName, $cappedCollectionName, $cappedCollectionOptions);
+        $operation->execute($this->getPrimaryServer());
```
jmikola: I'd also check if creating a capped collection is actually necessary. I expect it may be, since `explain` does run the query to gather its results.

tests/Operation/ExplainFunctionalTest.php
```diff
@@ -0,0 +1,195 @@
...
+
+        // Insert documents into the capped collection.
+        $bulkWrite = new BulkWrite(['ordered' => true]);
+        $bulkWrite->insert(['_id' => 1]);
+        $bulkWrite->insert(['_id' => 2]);
+        $result = $this->manager->executeBulkWrite($this->getNamespace(), $bulkWrite);
```
jmikola: Would `createFixtures()` work here as well? We don't actually care about the structure of the documents.

tests/Operation/ExplainFunctionalTest.php
```diff
@@ -0,0 +1,195 @@
...
+
+        $this->assertExplainResult($result, $executionStatsExpected, $allPlansExecutionExpected);
+    }
+
+    public function provideVerbosityInformation()
+    {
+        return [
+            [Explain::VERBOSITY_ALL_PLANS, true, true],
+            [Explain::VERBOSITY_EXEC_STATS, true, false],
+            [Explain::VERBOSITY_QUERY, false, false]
```
jmikola: Trailing comma on the last array element is optional in PHP, but I typically include it to make the line consistent with those above. It also allows lines to be reordered easily and makes the diff more concise if we add more elements later (probably not relevant in this case).

tests/Operation/ExplainFunctionalTest.php
kvwalker: It is necessary as you expected.

src/Operation/Update.php
```diff
@@ -167,40 +167,32 @@ public function execute(Server $server)
...
 
-        $writeResult = $server->executeBulkWrite($this->databaseName . '.' . $this->collectionName, $bulk, $this->createOptions());
+        $writeResult = $server->executeBulkWrite($this->databaseName . '.' . $this->collectionName, $bulk, $this->createExecuteOptions());
 
         return new UpdateResult($writeResult);
     }
 
+    public function getCommandDocument()
+    {
+        return ['update' => $this->collectionName, 'updates' => [['q' => $this->filter] + ['u' => $this->update] + $this->createUpdateOptions()]];
```
jmikola: `['q' => $this->filter] + ['u' => $this->update]` can be reduced to a single array.

src/Operation/Delete.php
```diff
@@ -117,18 +117,38 @@ public function execute(Server $server)
             throw UnsupportedException::collationNotSupported();
         }
 
+        $deleteOptions = $this->createDeleteOptions();
```
jmikola: This doesn't appear to be used, since `$bulk->delete()` calls `$this->createDeleteOptions()` directly.

tests/Operation/ExplainFunctionalTest.php
```diff
@@ -0,0 +1,255 @@
...
+
+    public function testFindModifiers()
+    {
+        $this->createFixtures(3);
+
+        $operation = new Find(
+            $this->getDatabaseName(),
+            $this->getCollectionName(),
+            [],
+            ['readConcern' => $this->createDefaultReadConcern(), 'modifiers' => ['$max' => ['_id' => 2.2]]]
```
jmikola: Is `readConcern` needed here? This may be a copy/paste from `testFind()`.

tests/Operation/FindAndModifyFunctionalTest.php
```diff
@@ -4,6 +4,7 @@
 
 use MongoDB\Driver\BulkWrite;
 use MongoDB\Model\BSONDocument;
+use MongoDB\Operation\Explain;
```
jmikola: This may be a leftover from when you were testing explain functionality in various operation tests.

tests/Operation/ExplainTest.php
```diff
@@ -0,0 +1,36 @@
...
+namespace MongoDB\Tests\Operation;
+
+use MongoDB\Operation\Count;
+use MongoDB\Operation\Distinct;
+use MongoDB\Operation\Explain;
+
+class ExplainTest extends TestCase
+{
+    /**
+     * @requires PHPUnit 5.4.0
```
jmikola: Neat. I wasn't aware this existed.

tests/Operation/ExplainTest.php
```diff
@@ -0,0 +1,36 @@
...
+     * @expectedException MongoDB\Exception\InvalidArgumentException
+     * @dataProvider provideInvalidConstructorOptions
+     */
+    public function testConstructorOptionTypeChecks(array $options)
+    {
+        $explainable = $this->createMock('MongoDB\Operation\Explainable');
```
jmikola: I didn't realize that `createMock()` was only added in later PHPUnit versions. You should be able to replace this with `$this->getMockBuilder('MongoDB\Operation\Explainable')->getMock();` for compatibility across all versions. That would allow this to be run on our PHP 5.x environments.

src/Operation/FindOne.php
```diff
@@ -126,4 +126,9 @@ public function execute(Server $server)
 
         return ($document === false) ? null : $document;
     }
+
+    public function getCommandDocument()
+    {
+        return $this->find->getCommandDocument();
```
jmikola: One other thing that just occurred to me. Unlike, `Find`, `FindAndModify` is an internal operation class (ideally, we don't want users to create it directly). I think the classes that compose `FindAndModify` (e.g. `FindOneAndUpdate`) should also implement Explainable and proxy the `getCommandDocument()` call, as you're doing here for `Find`.


 Similar to `Find` and `FindOne`, those sub-classes  various sub-classes of

src/Operation/Explain.php
```diff
@@ -0,0 +1,105 @@
...
+        $this->options = $options;
+    }
+
+    public function execute(Server $server)
+    {
+        if (! \MongoDB\server_supports_feature($server, self::$wireVersionForExplain)) {
+            throw UnsupportedException::explainNotSupported();
+        }
+
+        if ($this->explainable instanceof \MongoDB\Operation\Distinct && ! \MongoDB\server_supports_feature($server, self::$wireVersionForDistinct)) {
```
jmikola: Just noticed: since we're already in the `MongoDB\Operation` namespace, we can simply refer to this as `Distinct`. There's no need to `use` it, either.

src/Operation/Explain.php
```diff
@@ -0,0 +1,105 @@
...
+            throw UnsupportedException::explainNotSupported();
+        }
+
+        if ($this->explainable instanceof \MongoDB\Operation\FindAndModify && ! \MongoDB\server_supports_feature($server, self::$wireVersionForFindAndModify)) {
```
jmikola: Once the other three `findAndModify` operations are supported, the first half of this conditional should be changed to consider all four classes. If that makes the conditional a bit unmanageable, a private `isFindAndModify()` may help.

jmikola:
Failed Travis build appears to be due to this exception:

```
1) MongoDB\Tests\Operation\ExplainFunctionalTest::testFindModifiers
MongoDB\Driver\Exception\ConnectionException: Failed to parse: { find: "ExplainFunctionalTest.b7c5adbe", filter: {}, max: { _id: 2.2 } }. Unrecognized field 'max'.
/home/travis/build/mongodb/mongo-php-library/src/Operation/Explain.php:97
/home/travis/build/mongodb/mongo-php-library/tests/Operation/ExplainFunctionalTest.php:168
/home/travis/build/mongodb/mongo-php-library/tests/CommandObserver.php:25
/home/travis/build/mongodb/mongo-php-library/tests/Operation/ExplainFunctionalTest.php:173
```

According to the [`find` command documentation](https://docs.mongodb.com/manual/reference/command/find/), `max` is a supported option. This also doesn't appear to raise an error in 3.2, which makes me think it might have been a bug in server 3.0.x only. It may be worth pinging someone on the query team to ask about that.

In the meantime, perhaps you can use another modifier in the test to circumvent the issue.

tests/Operation/ExplainTest.php
jmikola: This can still be removed.

src/Operation/Delete.php
```diff
@@ -117,18 +117,36 @@ public function execute(Server $server)
...
+        $bulk->delete($this->filter, $this->createDeleteOptions());
+
+        $writeResult = $server->executeBulkWrite($this->databaseName . '.' . $this->collectionName, $bulk, $this->createExecuteOptions());
+
+        return new DeleteResult($writeResult);
+    }
+
+    public function getCommandDocument()
+    {
+        return ['delete' => $this->collectionName, 'deletes' => [['q' => $this->filter] + $this->createDeleteOptions()]];
```
jmikola: While I was doing a sanity check of this method against the [`delete` command reference](https://docs.mongodb.com/manual/reference/command/delete/), I noticed that it's technically missing top-level `ordered` and `writeConcern` keys.

I'm not too worried about `ordered`. We don't pass it to `BulkWrite` above, although I believe libmongoc will default it to `true` explicitly. The server default is also `true`, and it really has no effect for a single-statement write operation anyway.

I also doubt that `writeConcern` is relevant to an explain, but I do think it's worth including since it is an option on this class.

I'd suggest creating the command document as you do in the return statement and then conditionally setting `writeConcern` if it's set.

src/Operation/Explainable.php
```diff
@@ -0,0 +1,31 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace MongoDB\Operation;
+
+use MongoDB\Driver\Server;
```
jmikola: This looks like dead code.

src/Operation/Update.php
```diff
@@ -167,40 +167,32 @@ public function execute(Server $server)
...
 
-        $writeResult = $server->executeBulkWrite($this->databaseName . '.' . $this->collectionName, $bulk, $this->createOptions());
+        $writeResult = $server->executeBulkWrite($this->databaseName . '.' . $this->collectionName, $bulk, $this->createExecuteOptions());
 
         return new UpdateResult($writeResult);
     }
 
+    public function getCommandDocument()
+    {
+        return ['update' => $this->collectionName, 'updates' => [['q' => $this->filter, 'u' => $this->update] + $this->createUpdateOptions()]];
```
jmikola: Similar to my comment on `Delete::getCommandDocument()`, I think you should add top-level `writeConcern` and `bypassDocumentValidation` fields here if they're set as class options. See: [`update` command reference](https://docs.mongodb.com/manual/reference/command/update/).

src/Operation/Update.php
```diff
@@ -167,40 +167,42 @@ public function execute(Server $server)
...
 
+    public function getCommandDocument()
+    {
+        $cmd = ['update' => $this->collectionName, 'updates' => [['q' => $this->filter, 'u' => $this->update] + $this->createUpdateOptions()]];
+
+        if (isset($this->options['writeConcern'])) {
+            $cmd['writeConcern'] = $this->options['writeConcern'];
+        }
+
+        if (isset($this->options['bypassDocumentValidation']) && \MongoDB\server_supports_feature($server, self::$wireVersionForDocumentLevelValidation)) {
```
jmikola: Adding another thing to the list of late-breaking discoveries, I think `getCommandDocument()` will also need to take a `Server` instance. Without it, we'll get an undefined variable error on the second half of this condition.

The various FindAndModify operations appear to have the same issue.

I suppose the Explainable interface will need that `use` statement after all :stuck_out_tongue:

src/Operation/FindAndModify.php
```diff
@@ -239,13 +239,17 @@ public function execute(Server $server)
...
     /**
-     * Create the findAndModify command.
+     * Create the findAndModify command document.
      *
-     * @param Server $server
-     * @return Command
+     * @return array
      */
-    private function createCommand(Server $server)
+    private function createCommandDocument()
```
jmikola: This definitely needs a Server passed in, both for Explainable and creating the actual command for `execute()`.

tests/Operation/ExplainFunctionalTest.php
```diff
@@ -63,6 +67,44 @@ public function testDelete($verbosity, $executionStatsExpected, $allPlansExecuti
         $this->assertExplainResult($result, $executionStatsExpected, $allPlansExecutionExpected);
     }
 
+    /**
+     * @dataProvider provideVerbosityInformation
+     */
+    public function testDeleteMany($verbosity, $executionStatsExpected, $allPlansExecutionExpected)
+    {
+        $this->collection = new Collection($this->manager, $this->getDatabaseName(), $this->getCollectionName());
```
jmikola: Also not used (copied from `testDelete()`).

tests/Operation/ExplainFunctionalTest.php
```diff
@@ -63,6 +67,44 @@ public function testDelete($verbosity, $executionStatsExpected, $allPlansExecuti
...
+
+        $this->assertExplainResult($result, $executionStatsExpected, $allPlansExecutionExpected);
+    }
+
+    /**
+     * @dataProvider provideVerbosityInformation
+     */
+    public function testDeleteOne($verbosity, $executionStatsExpected, $allPlansExecutionExpected)
+    {
+        $this->collection = new Collection($this->manager, $this->getDatabaseName(), $this->getCollectionName());
```
jmikola: Also not used (copied from `testDelete()`).

tests/Operation/ExplainFunctionalTest.php
```diff
@@ -0,0 +1,387 @@
...
+
+        $this->assertExplainResult($result, $executionStatsExpected, $allPlansExecutionExpected);
+    }
+
+    /**
+     * @dataProvider provideVerbosityInformation
+     */
+    public function testDelete($verbosity, $executionStatsExpected, $allPlansExecutionExpected)
+    {
+        $this->collection = new Collection($this->manager, $this->getDatabaseName(), $this->getCollectionName());
```
jmikola: I don't believe this is ever used.

PR #498, opened 2018-02-20
Fix ReadPreference

bsormagec:
executeQuery function's third parameter must be instance of ReadPreference but this function giving array and this PR fixing that.
Thanks..

### DIFF
--- a/src/Operation/Find.php
+++ b/src/Operation/Find.php
@@ -284,7 +284,7 @@ class Find implements Executable
             throw UnsupportedException::readConcernNotSupported();
         }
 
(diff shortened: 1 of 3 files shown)


jmikola:
I expect this is because you've not updated the `mongodb` extension to 1.4.0, which is [required](https://github.com/mongodb/mongo-php-library/blob/1.3.0/composer.json#L16) with library version 1.3.0. See https://github.com/mongodb/mongo-php-driver/issues/757#issuecomment-365690509 for more information. 

The extension dependency bump was mentioned in the [release notes](https://github.com/mongodb/mongo-php-library/releases/tag/1.3.0). If you use Composer to install/upgrade the library, it also should have alerted you to the incompatibility unless you were ignoring platform dependencies (i.e. `--ignore-platform-reqs`), which is not advisable.

PR #500, opened 2018-02-23
PHPLIB-172: Support explain option for aggregate command

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-172

### DIFF
--- a/src/Operation/Aggregate.php
+++ b/src/Operation/Aggregate.php
@@ -77,6 +77,9 @@ class Aggregate implements Executable
      *  * comment (string): An arbitrary string to help trace the operation
      *    through the database profiler, currentOp, and logs.
      *
+     *  * explain (boolean): Specifies whether or not to return the information
+     *    on the processing of the pipeline.
+     *
      *  * hint (string|document): The index to use. Specify either the index
      *    name as a string or the index key pattern as a document. If specified,
      *    then the query system will only consider plans using the hinted index.
@@ -160,6 +163,10 @@ class Aggregate implements Executable
             throw InvalidArgumentException::invalidType('"comment" option', $options['comment'], 'string');
         }
 
+        if (isset($options['explain']) && ! is_bool($options['explain'])) {
+            throw InvalidArgumentException::invalidType('"explain" option', $options['explain'], 'boolean');
+        }
+
         if (isset($options['hint']) && ! is_string($options['hint']) && ! is_array($options['hint']) && ! is_object($options['hint'])) {
             throw InvalidArgumentException::invalidType('"hint" option', $options['hint'], 'string or array or object');
         }
@@ -208,6 +215,10 @@ class Aggregate implements Executable
             unset($options['writeConcern']);
         }
 
+        if ( ! empty($options['explain'])) {
+            $options['useCursor'] = false;
+        }
+
         $this->databaseName = (string) $databaseName;
         $this->collectionName = (string) $collectionName;
         $this->pipeline = $pipeline;
@@ -238,16 +249,17 @@ class Aggregate implements Executable
             throw UnsupportedException::writeConcernNotSupported();
         }
 
+        $hasExplain = ! empty($this->options['explain']);
         $hasOutStage = \MongoDB\is_last_pipeline_operator_out($this->pipeline);
 
         $command = $this->createCommand($server);
-        $options = $this->createOptions($hasOutStage);
+        $options = $this->createOptions($hasOutStage, $hasExplain);
 
-        $cursor = $hasOutStage
+        $cursor = ($hasOutStage && ! $hasExplain)
             ? $server->executeReadWriteCommand($this->databaseName, $command, $options)
             : $server->executeReadCommand($this->databaseName, $command, $options);
 
-        if ($this->options['useCursor']) {
+        if ($this->options['useCursor'] || $hasExplain) {
             if (isset($this->options['typeMap'])) {
                 $cursor->setTypeMap($this->options['typeMap']);
             }
@@ -288,7 +300,7 @@ class Aggregate implements Executable
             $cmd['bypassDocumentValidation'] = $this->options['bypassDocumentValidation'];
         }
 
-        foreach (['comment', 'maxTimeMS'] as $option) {
+        foreach (['comment', 'explain', 'maxTimeMS'] as $option) {
             if (isset($this->options[$option])) {
                 $cmd[$option] = $this->options[$option];
             }
@@ -323,7 +335,7 @@ class Aggregate implements Executable
      * @param boolean $hasOutStage
      * @return array
      */
-    private function createOptions($hasOutStage)
+    private function createOptions($hasOutStage, $hasExplain)
     {
         $options = [];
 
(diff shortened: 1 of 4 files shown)


tests/Operation/AggregateFunctionalTest.php
```diff
@@ -109,6 +109,17 @@ public function testTypeMapOption(array $typeMap = null, array $expectedDocument
...
 
+    public function testExplainOption()
+    {
+        $this->createFixtures(3);
+
+        $pipeline = [['$match' => ['_id' => ['$ne' => 2]]]];
+        $operation = new Aggregate($this->getDatabaseName(), $this->getCollectionName(), $pipeline, ['explain' => true, 'typeMap' => ['root' => 'array', 'document' => 'array']]);
+        $results = iterator_to_array($operation->execute($this->getPrimaryServer()));
+
+        $this->assertArrayHasKey('queryPlanner', $results[0]['stages'][0]['$cursor']);
```
jmikola: I think we can simply assert that the count of `$results` is 1 and that a top-level `stages` field exists in the result document.

That should make this fairly resilient to the note in [Return Information on the Aggregation Operation](https://docs.mongodb.com/manual/reference/command/aggregate/#return-information-on-the-aggregation-operation):

> The explain output is subject to change between releases.

src/Operation/Aggregate.php
```diff
@@ -77,6 +77,9 @@ class Aggregate implements Executable
      *  * comment (string): An arbitrary string to help trace the operation
      *    through the database profiler, currentOp, and logs.
      *
+     *  * explain (boolean): Specifies whether or not to return the information
+     *    on the processing of the pipeline.
+     *
```
jmikola: This should be added to the RST docs as well.

tests/Operation/AggregateFunctionalTest.php
jmikola: Also, I think we can remove the `typeMap` option above. The `ArrayHasKey` constraint [does support the ArrayAccess interface](https://github.com/sebastianbergmann/phpunit/blob/7.0.1/src/Framework/Constraint/ArrayHasKey.php#L65), which our default BSONDocument class implements. Looking at the git history of the constraint class, that functionality goes back to PHPUnit 3.x, so we can rely on it for all variants in the test matrix.

src/Operation/Aggregate.php
```diff
@@ -302,6 +309,10 @@ private function createCommand(Server $server)
             $cmd['collation'] = (object) $this->options['collation'];
         }
 
+        if (isset($this->options['explain'])) {
```
jmikola: This can be rolled into the `foreach (['comment', 'maxTimeMS'] as $option)` loop above, since it uses the basic isset/assign behavior with no casting or other logic.

tests/Operation/AggregateFunctionalTest.php
kvwalker: If I remove the `typeMap` option, the result returned is an array wrapped in a `stdClass` object so I still need `['typeMap' => ['root' => 'array']]`.

src/Operation/Aggregate.php
```diff
@@ -251,7 +262,7 @@ public function execute(Server $server)
             ? $server->executeReadWriteCommand($this->databaseName, $command, $options)
             : $server->executeReadCommand($this->databaseName, $command, $options);
 
-        if ($this->options['useCursor']) {
+        if ($this->options['useCursor'] || ( ! $this->options['useCursor'] && ! empty($this->options['explain']))) {
```
jmikola: `! $this->options['useCursor']` can be factored out here, since we'd only reach the right-hand side of `||` if `useCursor` was not true.

tests/Operation/AggregateFunctionalTest.php
```diff
@@ -109,6 +109,18 @@ public function testTypeMapOption(array $typeMap = null, array $expectedDocument
...
 
+    public function testExplainOption()
+    {
+        $this->createFixtures(3);
+
+        $pipeline = [['$match' => ['_id' => ['$ne' => 2]]]];
+        $operation = new Aggregate($this->getDatabaseName(), $this->getCollectionName(), $pipeline, ['explain' => true, 'typeMap' => ['root' => 'array']]);
+        $results = iterator_to_array($operation->execute($this->getPrimaryServer()));
+
+        $this->assertSame(1, count($results));
```
jmikola: `assertCount()` should work here.

tests/Operation/AggregateFunctionalTest.php
```diff
@@ -109,6 +110,46 @@ public function testTypeMapOption(array $typeMap = null, array $expectedDocument
...
+
+        $this->assertCount(1, $results);
+        $this->assertArrayHasKey('stages', $results[0]);
+    }
+
+    public function testExplainOptionWithWriteConcern()
+    {
+        if (version_compare($this->getServerVersion(), '3.4.0', '<')) {
+            $this->markTestSkipped('Sessions are not supported');
+        }
```
jmikola: This skip doesn't look relevant. Missed that from an earlier version of the PR.

tests/Operation/AggregateFunctionalTest.php
```diff
@@ -109,6 +110,46 @@ public function testTypeMapOption(array $typeMap = null, array $expectedDocument
...
+        $options = ['explain' => true, 'writeConcern' => new WriteConcern(1)];
+
+        (new CommandObserver)->observe(
+            function() use ($pipeline, $options) {
+                $operation = new Aggregate($this->getDatabaseName(), $this->getCollectionName(), $pipeline, $options);
+
+                $results = iterator_to_array($operation->execute($this->getPrimaryServer()));
+
+                $this->assertCount(1, $results);
+                $this->assertObjectHasAttribute('stages', current($results));
```
jmikola: Did `$this->assertArrayHasKey('stages', $results[0]);` not work here?

I'm actually curious how this works given the implementation of [`ObjectHasAttribute::matches()`](https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/Constraint/ObjectHasAttribute.php#L30):

```php
$a = new ArrayObject(['foo'=>1]);
$ro = new ReflectionClass($a);
$ro->hasProperty('foo'); // false
```

jmikola: Nevermind, I'm an idiot.

You're using the Aggregation operation class directly, so there is no type map and this is a `stdClass` instance.

tests/Operation/AggregateFunctionalTest.php
kvwalker: This skip is relevant because, according to the [documentation for aggregate](https://docs.mongodb.com/php-library/master/reference/method/MongoDBCollection-aggregate/index.html), writeConcern "is not supported for server versions prior to 3.4 and will result in an exception at execution time if used." I forgot to change the message, which I have now corrected.

tests/Operation/AggregateFunctionalTest.php
jmikola: Aye, I forgot we're not inheriting the Collection's type map here.

PR #502, opened 2018-02-23
PHPLIB-333: Allow typeMap option when useCursor is false for Aggregate

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-333

### DIFF
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -205,7 +205,7 @@ class Collection
             $options['readConcern'] = $this->readConcern;
         }
 
-        if ( ! isset($options['typeMap']) && ( ! isset($options['useCursor']) || $options['useCursor'])) {
+        if ( ! isset($options['typeMap'])) {
             $options['typeMap'] = $this->typeMap;
         }
 
--- a/tests/Operation/AggregateFunctionalTest.php
+++ b/tests/Operation/AggregateFunctionalTest.php
@@ -5,6 +5,7 @@ namespace MongoDB\Tests\Operation;
 use MongoDB\Driver\BulkWrite;
 use MongoDB\Operation\Aggregate;
 use MongoDB\Tests\CommandObserver;
+use ArrayIterator;
 use stdClass;
 
 class AggregateFunctionalTest extends FunctionalTestCase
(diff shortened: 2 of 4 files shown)


src/Collection.php
```diff
@@ -205,10 +205,6 @@ public function aggregate(array $pipeline, array $options = [])
             $options['readConcern'] = $this->readConcern;
         }
 
-        if ( ! isset($options['typeMap']) && ( ! isset($options['useCursor']) || $options['useCursor'])) {
-            $options['typeMap'] = $this->typeMap;
-        }
-
```
jmikola: We only need to get ride of the `useCursor` logic here, as we sill want to inherit the type map if it's not set in `$options`.

tests/Operation/AggregateFunctionalTest.php
```diff
@@ -103,12 +103,29 @@ public function testTypeMapOption(array $typeMap = null, array $expectedDocument
...
 
+    public function testTypeMapInline()
+    {
+        if (version_compare($this->getServerVersion(), '3.6.0', '>=')) {
+            $this->markTestSkipped('Inline aggregations are not supported');
+        }
+
+        $this->createFixtures(3);
+
+        $pipeline = [['$match' => ['x' => 1]], ['$out' => $this->getCollectionName() . '.output']];
```
jmikola: The `$out` operator implies that no results will be returned. I think it'd make more sense to rewrite this test case to copy the logic in `testTypeMapOption()` and add the `instanceof` check. Its data provider can also be re-used as-is.

This test can then be renamed `testTypeMapOptionWithoutCursor()` or something to that effect.

tests/Operation/AggregateFunctionalTest.php
```diff
@@ -103,12 +103,29 @@ public function testTypeMapOption(array $typeMap = null, array $expectedDocument
...
+
+        $operation = new Aggregate($this->getDatabaseName(), $this->getCollectionName(), $pipeline, ['typeMap' => ['root' => 'array'], 'useCursor' => false]);
+        $results = $operation->execute($this->getPrimaryServer());
+
+        $this->assertTrue($results instanceof \MongoDB\Model\TypeMapArrayIterator);
```
jmikola: `assertInstanceOf()` can be used here. Reminder that its class name string won't need the leading `\`, since it assumes the global namespace (like `@expectedException`).

tests/Operation/AggregateFunctionalTest.php
```diff
@@ -110,20 +110,23 @@ public function testTypeMapOption(array $typeMap = null, array $expectedDocument
...
 
-        $pipeline = [['$match' => ['x' => 1]], ['$out' => $this->getCollectionName() . '.output']];
+        $pipeline = [['$match' => ['_id' => ['$ne' => 2]]]];
 
-        $operation = new Aggregate($this->getDatabaseName(), $this->getCollectionName(), $pipeline, ['typeMap' => ['root' => 'array'], 'useCursor' => false]);
+        $operation = new Aggregate($this->getDatabaseName(), $this->getCollectionName(), $pipeline, ['typeMap' => $typeMap, 'useCursor' => false]);
         $results = $operation->execute($this->getPrimaryServer());
 
-        $this->assertTrue($results instanceof \MongoDB\Model\TypeMapArrayIterator);
+        $this->assertInstanceOf(\ArrayIterator::class, $results);
```
jmikola: Ah, good idea to use [`::class`](http://php.net/manual/en/language.oop5.basic.php#language.oop5.basic.class.class) here. We couldn't use it before since we were supporting 5.4, but that's no longer a concern.

I would suggest `use ArrayIterator` up top and removing the `\` here, just so we're consistent about listing all classes at the top of each file (e.g. `stdClass`).

tests/Operation/AggregateFunctionalTest.php
```diff
@@ -110,20 +110,23 @@ public function testTypeMapOption(array $typeMap = null, array $expectedDocument
         $this->assertEquals($expectedDocuments, $results);
     }
 
-    public function testTypeMapInline()
+    /**
+     * @dataProvider provideTypeMapOptionsAndExpectedDocuments
+     */
+    public function testTypeMapOptionWithoutCursor(array $typeMap = null, array $expectedDocuments)
```
jmikola: `$expectedDocuments` isn't being used for anything. Can you add `$this->assertEquals($expectedDocuments, $results);` (as `testTypeMapOption()` does) down below?

PR #501, opened 2018-02-23
PHPLIB-332: Create explain helper on Collection class

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-332

PR #503, opened 2018-02-26
PHPLIB-156: Add readPreference and session options for Explain

kvwalker:


src/Operation/Explain.php
```diff
@@ -49,26 +51,40 @@ class Explain implements Executable
      *
      * Supported options:
      *
-     *  * verbosity (string): The mode in which the explain command will be run.
+     *  * readPreference (MongoDB\Driver\ReadPreference): Read preference. This
+     *    will be used to select a new server when resuming. Defaults to a
+     *    "primary" read preference.
```
jmikola: Let's chop this at "Read preference." The rest only applies to the Watch operation. Sorry for not noticing this in #501.

PR #507, opened 2018-02-28
PHPLIB-216: Use expectException() method instead of annotation in tests

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-216

### DIFF
--- a/tests/Operation/BulkWriteFunctionalTest.php
+++ b/tests/Operation/BulkWriteFunctionalTest.php
@@ -6,6 +6,8 @@ use MongoDB\BulkWriteResult;
 use MongoDB\Collection;
 use MongoDB\Driver\BulkWrite as Bulk;
 use MongoDB\Driver\WriteConcern;
+use MongoDB\Exception\BadMethodCallException;
+use MongoDB\Exception\InvalidArgumentException;
 use MongoDB\Model\BSONDocument;
 use MongoDB\Operation\BulkWrite;
 use MongoDB\Tests\CommandObserver;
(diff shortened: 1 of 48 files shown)


kvwalker:
expectException() was introduced in PHPUnit 5.2 so I'm not sure of a good way to make this work on PHPUnit 4.8. One [recommendation](https://github.com/sebastianbergmann/phpunit/issues/2074) was to use setExpectedException, but that was removed in PHPUnit 6.

tests/Operation/BulkWriteFunctionalTest.php
```diff
@@ -161,82 +163,80 @@ public function testUnacknowledgedWriteConcern()
 
     /**
      * @depends testUnacknowledgedWriteConcern
-     * @expectedException MongoDB\Exception\BadMethodCallException
-     * @expectedExceptionMessageRegExp /[\w:\\]+ should not be called for an unacknowledged write result/
      */
     public function testUnacknowledgedWriteConcernAccessesDeletedCount(BulkWriteResult $result)
     {
+        $this->expectException(BadMethodCallException::class);
+        $this->expectExceptionMessageRegExp('/[\w] should not be called for an unacknowledged write result/');
```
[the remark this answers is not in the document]
kvwalker: I realize that `'/[\w:\\]+ should not...` is what was originally in the `@expectedExceptionMessageRegExp`, but when I use that regular expression in `$this->expectExceptionMessageRegExp()`, I get a `PHPUnit\Framework\Exception: Invalid expected exception message regex given`.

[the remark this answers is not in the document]
jmikola: Ah! I think that's because `\\` in the PHP string translates to a single, escaped backslash. Regex then thinks we're escaping the follow `]`. Since we need to provide `\\` to the regex parser, we may need to use `\\\\` in the PHP string.

tests/TestCase.php
```diff
@@ -11,6 +11,30 @@
 
 abstract class TestCase extends BaseTestCase
 {
+    public function expectException($exception)
+    {
+        if (method_exists('PHPUnit\Framework\TestCase', 'expectException')) {
+            return parent::expectException($exception);
```
jmikola: Since `expectException()` has a `void` return type, this will likely cause a notice with code inspection tools like PHPStan. I'd break this up and `return;` on the next line to steer clear of that.

Likewise for the methods below.

tests/TestCase.php
```diff
@@ -11,6 +11,30 @@
...
+        }
+        return parent::setExpectedException($exception);
```
jmikola: We can simply remove the `return` here. Likewise for the methods below.

tests/TestCase.php
```diff
@@ -11,6 +11,30 @@
 
 abstract class TestCase extends BaseTestCase
 {
+    public function expectException($exception)
+    {
+        if (method_exists('PHPUnit\Framework\TestCase', 'expectException')) {
```
jmikola: This can be `BaseTestCase::class` in all three cases. Alternatively, `get_parent_class()` could also be used (https://3v4l.org/Wh5YO) to keep things flexible, but I can't imagine we're going to change the parent class.

PR #509, opened 2018-03-07
removed collectionName attribute in GridFSbucket construct signature

maxime-alves:
I saw on the mongodb docs that this was false, you can verify in the code.


jmikola:
Indeed, it looks like bad copypasta from the Collection documentation. The constructor has never had a `$collectionName` argument: https://github.com/mongodb/mongo-php-library/blob/1.1.0/src/GridFS/Bucket.php#L68

Thanks for the correction. I'll manually merge this to ensure it makes its way through all versions of the documentation.

PR #510, opened 2018-03-12
PHPLIB-327, PHPLIB-335: ChangeStream::next() key increment fixes

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-327
https://jira.mongodb.org/browse/PHPLIB-335

### DIFF
--- a/src/ChangeStream.php
+++ b/src/ChangeStream.php
@@ -95,11 +95,11 @@ class ChangeStream implements Iterator
         try {
             $this->csIt->next();
             if ($this->valid()) {
-                $this->resumeToken = $this->extractResumeToken($this->csIt->current());
                 if ($this->hasAdvanced) {
                     $this->key++;
                 }
                 $this->hasAdvanced = true;
+                $this->resumeToken = $this->extractResumeToken($this->csIt->current());
             }
         } catch (RuntimeException $e) {
             if (strpos($e->getMessage(), "not master") !== false) {
@@ -127,8 +127,8 @@ class ChangeStream implements Iterator
         try {
             $this->csIt->rewind();
             if ($this->valid()) {
-                $this->resumeToken = $this->extractResumeToken($this->csIt->current());
                 $this->hasAdvanced = true;
+                $this->resumeToken = $this->extractResumeToken($this->csIt->current());
             }
         } catch (RuntimeException $e) {
             if (strpos($e->getMessage(), "not master") !== false) {
--- a/src/ChangeStream.php
+++ b/src/ChangeStream.php
@@ -38,7 +38,8 @@ class ChangeStream implements Iterator
     private $resumeToken;
     private $resumeCallable;
     private $csIt;
-    private $key;
+    private $key = 0;
+    private $hasAdvanced = false;
 
     const CURSOR_NOT_FOUND = 43;
 
@@ -53,8 +54,6 @@ class ChangeStream implements Iterator
     {
         $this->resumeCallable = $resumeCallable;
         $this->csIt = new IteratorIterator($cursor);
-
-        $this->key = 0;
     }
 
     /**
@@ -97,7 +96,10 @@ class ChangeStream implements Iterator
             $this->csIt->next();
             if ($this->valid()) {
                 $this->resumeToken = $this->extractResumeToken($this->csIt->current());
-                $this->key++;
+                if ($this->hasAdvanced) {
+                    $this->key++;
+                }
+                $this->hasAdvanced = true;
             }
         } catch (RuntimeException $e) {
             if (strpos($e->getMessage(), "not master") !== false) {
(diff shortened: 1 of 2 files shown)


jmikola:
At a quick glance, this PR should target the v1.3 branch. Rebasing may introduce extra commits, so it may be easier to run `git reset --hard v1.3` on the branch, cherry-pick this commit into it, and force push before changing the target in the PR.

In the future, I find it helps to prefix my branches with the target name (e.g. "1.3-phplib-327"). That helps me catch this before opening a PR against master out of habit.

src/ChangeStream.php
```diff
@@ -40,6 +40,8 @@ class ChangeStream implements Iterator
     private $csIt;
     private $key;
 
+    private $previousValid;
```
jmikola: Blank line above this can be removed so all `private` declarations are in the same paragraph.

src/ChangeStream.php
```diff
@@ -55,6 +57,8 @@ public function __construct(Cursor $cursor, callable $resumeCallable)
         $this->csIt = new IteratorIterator($cursor);
 
         $this->key = 0;
+
+        $this->previousValid = false;
```
jmikola: Can we initialize `$key` and `$previousValid` in the `private` declarations above? That would reduce a few lines in the constructor.

src/ChangeStream.php
```diff
@@ -94,10 +98,15 @@ public function next()
     {
         $resumable = false;
         try {
+            if ($this->valid()) {
+                $this->previousValid = true;
```
jmikola: I'd propose `$hasAdvanced` as a property name. Like "is", the prefix implies that it's a boolean type, and I think it better matches the purpose of the variable as a flag that is initially `false` and is switched once to `true` (where it remains thereafter).

Also, I think we can move this to the existing `if ($this->valid())` branch after `$this->csIt->next();`.

Assuming you implement [PHPLIB-335](https://jira.mongodb.org/browse/PHPLIB-335) in the same PR, I think it'd make sense for `$this->hasAdvanced = true` to occur between the conditional key increment and resume token extraction. For example:

```php
$this->csIt->next();
if ($this->valid()) {
    if ($this->hasAdvanced) {
        $this->key++;
    }
    $this->hasAdvanced = true;
    $this->resumeToken = $this->extractResumeToken($this->csIt->current());
}
```

In the spirit of PHPLIB-335, this will ensure that `$hasAdvanced` flag is toggled even if resume token extraction fails.

jmikola:
Additionally, can you add a regression test for PHPLIB-327? I think you can model it after the test I attached to [PHPLIB-335](https://jira.mongodb.org/browse/PHPLIB-335) and simply remove the projection and ResumeTokenException expectations. In effect:

 * Create a change stream
 * Insert two documents
 * Advance with `next()` and assert that the key is zero
 * Advance with `next()` and assert that the key is one

You should also be able to verify locally that the test fails without the fix in this PR, which will demonstrate that it's a suitable regression test.

src/ChangeStream.php
```diff
@@ -39,8 +39,10 @@ class ChangeStream implements Iterator
     private $resumeCallable;
     private $csIt;
     private $key;
+    private $hasAdvanced;
 
-    private $previousValid;
+    $this->key = 0;
+    $this->hasAdvanced = false;
```
jmikola: Declarations should be:

```php
private $key = 0;
private $hasAdvanced = false;
```

I assume this was a typo copying content from the constructor, but if not: http://php.net/manual/en/language.oop5.properties.php#example-180

src/ChangeStream.php
```diff
@@ -98,15 +96,13 @@ public function next()
...
-            }
             $this->csIt->next();
             if ($this->valid()) {
-                $this->resumeToken = $this->extractResumeToken($this->csIt->current());
-                if ($this->previousValid) {
+                if ($this->hasAdvanced) {
                     $this->key++;
                 }
+                $this->hasAdvanced = true;
+                $this->resumeToken = $this->extractResumeToken($this->csIt->current());
```
jmikola: I'd propose keeping this line at the top of the conditional scope for the PHPLIB-327 commit. A second commit for PHPLIB-335 can then move it to the end of the conditional scope.

If you're adding regression tests for each issue in the respective commits, the second commit for PHPLIB-335 may also need to adjust PHPLIB-327's test.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -536,6 +535,25 @@ public function provideTypeMapOptionsAndExpectedChangeDocument()
         ];
     }
 
+    public function testNextAdvancesKey()
+    {
+        $operation = new Watch($this->manager, $this->getDatabaseName(), $this->getCollectionName(), [], $this->defaultOptions);
+        $changeStream = $operation->execute($this->getPrimaryServer());
+
+        /* Note: we intentionally do not start iteration with rewind() to ensure
+         * that we test extraction functionality within next(). */
```
jmikola: I'd suggest replacing "extraction functionality within next()" with "incrementing when first advancing within next()" or simply deleting this comment entirely.

PR #513, opened 2018-03-15
PHPLIB-66: Wrapper method for collMod command

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-66

### DIFF
--- a/docs/includes/apiargs-MongoDBDatabase-method-modifyCollection-param.yaml
+++ b/docs/includes/apiargs-MongoDBDatabase-method-modifyCollection-param.yaml
@@ -1,0 +1,20 @@
+source:
+  file: apiargs-common-param.yaml
+  ref: $collectionName
+replacement:
+  subject: "collection or view"
+  action: " to modify"
+---
+arg_name: param
+name: $collectionOptions
+type: array
+description: |
+  Collection or view options to assign.
+interface: phpmethod
+operation: ~
+optional: false
+---
+source:
+  file: apiargs-common-param.yaml
+  ref: $options
+...
--- a/docs/reference/method/MongoDBDatabase-modifyCollection.txt
+++ b/docs/reference/method/MongoDBDatabase-modifyCollection.txt
@@ -1,0 +1,81 @@
+=====================================
+MongoDB\\Database::modifyCollection()
+=====================================
+
+.. versionadded:: 1.4
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Definition
+----------
+
+.. phpmethod:: MongoDB\\Database::modifyCollection()
+
+   Modifies a collection or view according to the specified
+   ``$collectionOptions``.
+
+   .. code-block:: php
+
+      function modifyCollection($collectionName, array $collectionOptions, array $options = []): array|object
+
+   This method has the following parameters:
+
+   .. include:: /includes/apiargs/MongoDBDatabase-method-modifyCollection-param.rst
+
+   The ``$options`` parameter supports the following options:
+
+   .. include:: /includes/apiargs/MongoDBDatabase-method-modifyCollection-option.rst
+
+Return Values
+-------------
+
+An array or object with the result document of the :manual:`collMod
+</reference/command/collMod>` command.
+
+Errors/Exceptions
+-----------------
+
+.. include:: /includes/extracts/error-invalidargumentexception.rst
+.. include:: /includes/extracts/error-driver-runtimeexception.rst
+
+Example
+-------
+
+The following example changes the expiration time of a TTL collection in the
+``test`` database:
+
+.. code-block:: php
+
+   <?php
+
+   $db = (new MongoDB\Client)->test;
+
+   $result = $db->modifyCollection('users', [
+       'keyPattern' => ['lastAccess' => 1],
+       'expireAfterSeconds' => 1000
+   ]);
+
+   var_dump($result);
+
+The output would then resemble::
+
+   object(stdClass)#2779 {
+     ["expireAfterSeconds_old"]=>
+     int(3)
+     ["expireAfterSeconds_new"]=>
+     int(1000)
+     ["ok"]=>
+     float(1)
+   }
+
+See Also
+--------
+
+- :manual:`collMod </reference/command/collMod>` command reference in the MongoDB
+  manual
--- a/src/Database.php
+++ b/src/Database.php
@@ -33,6 +33,7 @@ use MongoDB\Operation\DatabaseCommand;
 use MongoDB\Operation\DropCollection;
 use MongoDB\Operation\DropDatabase;
 use MongoDB\Operation\ListCollections;
+use MongoDB\Operation\ModifyCollection;
 
 class Database
 {
@@ -336,6 +337,33 @@ class Database
     {
         $operation = new ListCollections($this->databaseName, $options);
         $server = $this->manager->selectServer(new ReadPreference(ReadPreference::RP_PRIMARY));
+
+        return $operation->execute($server);
+    }
+
+    /**
+     * Modifies a collection or view.
+     *
+     * @see ModifyCollection::__construct() for supported options
+     * @param string $collectionName    Collection or view to modify
+     * @param array  $collectionOptions Collection or view options to assign
+     * @param array  $options           Command options
+     * @throws InvalidArgumentException for parameter/option parsing errors
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function modifyCollection($collectionName, array $collectionOptions, array $options = [])
+    {
+        if ( ! isset($options['typeMap'])) {
+            $options['typeMap'] = $this->typeMap;
+        }
+
+        $server = $this->manager->selectServer(new ReadPreference(ReadPreference::RP_PRIMARY));
+
+        if ( ! isset($options['writeConcern']) && \MongoDB\server_supports_feature($server, self::$wireVersionForWritableCommandWriteConcern)) {
+            $options['writeConcern'] = $this->writeConcern;
+        }
+
+        $operation = new ModifyCollection($this->databaseName, $collectionName, $collectionOptions, $options);
 
         return $operation->execute($server);
     }
--- a/src/Operation/ModifyCollection.php
+++ b/src/Operation/ModifyCollection.php
@@ -1,0 +1,135 @@
+<?php
+/*
+ * Copyright 2018 MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace MongoDB\Operation;
+
+use MongoDB\Driver\Command;
+use MongoDB\Driver\Server;
+use MongoDB\Driver\Session;
+use MongoDB\Driver\WriteConcern;
+use MongoDB\Driver\Exception\RuntimeException as DriverRuntimeException;
+use MongoDB\Exception\InvalidArgumentException;
+
+/**
+ * Operation for the collMod command.
+ *
+ * @api
+ * @see \MongoDB\Database::modifyCollection()
+ * @see http://docs.mongodb.org/manual/reference/command/collMod/
+ */
+class ModifyCollection implements Executable
+{
+    private $databaseName;
+    private $collectionName;
+    private $collectionOptions;
+    private $options;
+
+    /**
+     * Constructs a collMod command.
+     *
+     * Supported options:
+     *
+     *  * session (MongoDB\Driver\Session): Client session.
+     *
+     *    Sessions are not supported for server versions < 3.6.
+     *
+     *  * typeMap (array): Type map for BSON deserialization. This will only be
+     *    used for the returned command result document.
+     *
+     *  * writeConcern (MongoDB\Driver\WriteConcern): Write concern.
+     *
+     *    This is not supported for server versions < 3.2 and will result in an
+     *    exception at execution time if used.
+     *
+     * @param string       $databaseName      Database name
+     * @param string       $collectionName    Collection or view to modify
+     * @param string       $collectionOptions Collection or view options to assign
+     * @param array        $options           Command options
+     * @throws InvalidArgumentException for parameter/option parsing errors
+     */
+    public function __construct($databaseName, $collectionName, array $collectionOptions, array $options = [])
+    {
+        if (empty($collectionOptions)) {
+            throw new InvalidArgumentException('$collectionOptions is empty');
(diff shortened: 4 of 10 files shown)


src/Operation/CollMod.php
```diff
@@ -0,0 +1,100 @@
...
+use MongoDB\Exception\UnexpectedValueException;
+
+/**
+ * Operation for the collMod command.
+ *
+ * @api
+ * @see \MongoDB\Collection::collMod()
+ * @see http://docs.mongodb.org/manual/reference/command/collMod/
+ */
+class CollMod implements Executable
```
jmikola: Can we call this "ModifyCollection" instead?

src/Operation/CollMod.php
```diff
@@ -0,0 +1,100 @@
...
+     * Constructs a collMod command.
+     *
+     * Supported options:
+     *
+     *  * typeMap (array): Type map for BSON deserialization. This will only be
+     *    used for the returned command result document.
+     *
+     * @param string       $databaseName     Database name
+     * @param string       $collectionOrView Collection or view to modify
+     * @param array|object $collModOption    Option for collMod
```
jmikola: Shouldn't this always be a string?

src/Operation/CollMod.php
```diff
@@ -0,0 +1,100 @@
...
+     * @param array        $collModValue     Value for collModOption
```
jmikola: I think this type can be `mixed`. Looking at the [`collMod` examples](https://docs.mongodb.com/manual/reference/command/collMod/), option values may be booleans, strings, arrays, and documents (what I noticed at a glance). I'd suggest we accept anything and allow the server to raise an error if the type is incorrect.

src/Operation/CollMod.php
```diff
@@ -0,0 +1,100 @@
...
+        $this->options = $options;
+    }
+
+    /**
+     * Execute the operation.
+     *
+     * @see Executable::execute()
+     * @param Server $server
+     * @return array|object Command result document
+     * @throws UnexpectedValueException if the command response was malformed
```
jmikola: This looks like copypasta from FindAndModify (or another operation class).

src/Database.php
```diff
@@ -152,6 +153,18 @@ public function __toString()
         return $this->databaseName;
     }
 
+    public function collMod($collectionOrView, $collModOption, $collModValue, array $options = [])
```
jmikola: Can we call this `modifyCollection()` as well, for consistency with `createCollection()` and `dropCollection()`?

src/Operation/CollMod.php
```diff
@@ -0,0 +1,100 @@
...
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function execute(Server $server)
+    {
+        $cursor = $server->executeCommand($this->databaseName, $this->createCommand());
```
jmikola: I think we can use `executeWriteCommand()` here.

Note that CreateCollection also supports a `writeConcern` option. Both `create` and `collMod` are listed in [commands that write](https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#other-commands-that-write) in the write concern spec.

Note: that option isn't mentioned in the MongoDB manual but there's an open ticket to do so ([DOCS-9822](https://jira.mongodb.org/browse/DOCS-9822)).

src/Operation/CollMod.php
```diff
@@ -0,0 +1,100 @@
...
+    private $collModValue;
+    private $options;
+
+    /**
+     * Constructs a collMod command.
+     *
+     * Supported options:
+     *
+     *  * typeMap (array): Type map for BSON deserialization. This will only be
+     *    used for the returned command result document.
```
jmikola: `writeConcern` and `session` options should also be supported here after the switch to `executeWriteCommand()`.

src/Operation/CollMod.php
```diff
@@ -0,0 +1,100 @@
...
+     *
+     * @param string       $databaseName     Database name
+     * @param string       $collectionOrView Collection or view to modify
```
jmikola: This reminds me that we never added support for view-related options to CreateCollection. We can chat later about that, but it should likely be a feature to include in this next release.

src/Operation/ModifyCollection.php
```diff
@@ -0,0 +1,146 @@
...
+            throw InvalidArgumentException::invalidType('"writeConcern" option', $options['writeConcern'], 'MongoDB\Driver\WriteConcern');
+        }
+
+        if (isset($options['writeConcern']) && $options['writeConcern']->isDefault()) {
+            unset($options['writeConcern']);
+        }
+
+        $this->databaseName = (string) $databaseName;
+        $this->collectionOrView = (string) $collectionOrView;
+        $this->collModOptionValue = $collModOptionValue;
```
jmikola: Can be renamed to `$this->collectionOptions`.

src/Operation/ModifyCollection.php
```diff
@@ -0,0 +1,146 @@
...
+ * @api
+ * @see \MongoDB\Database::modifyCollection()
+ * @see http://docs.mongodb.org/manual/reference/command/collMod/
+ */
+class ModifyCollection implements Executable
+{
+    private $databaseName;
+    private $collectionName;
+    private $collModOption;
+    private $collModValue;
```
jmikola: These properties can be consolidated to `private $collectionOptions`.

src/Operation/ModifyCollection.php
```diff
@@ -0,0 +1,146 @@
...
+        if (isset($options['writeConcern']) && ! $options['writeConcern'] instanceof WriteConcern) {
+            throw InvalidArgumentException::invalidType('"writeConcern" option', $options['writeConcern'], 'MongoDB\Driver\WriteConcern');
+        }
+
+        if (isset($options['writeConcern']) && $options['writeConcern']->isDefault()) {
+            unset($options['writeConcern']);
+        }
+
+        $this->databaseName = (string) $databaseName;
+        $this->collectionOrView = (string) $collectionOrView;
```
jmikola: The property name defined above is `private $collectionName`. Let's use that here, too.

src/Operation/ModifyCollection.php
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    /**
+     * Create the collMod command.
+     *
+     * @return Command
+     */
+    private function createCommand()
+    {
+        $cmd = ['collMod' => $this->collectionOrView];
```
jmikola: `$this->collectionName`

src/Database.php
```diff
@@ -340,6 +341,33 @@ public function listCollections(array $options = [])
...
+     * Modifies a collection or view.
+     *
+     * @see ModifyCollection::__construct() for supported options
+     * @param string $collectionOrView   Collection or view to modify
+     * @param string $collModOptionValue Options and values for collMod
+     * @param array  $options            Command options
+     * @throws InvalidArgumentException for parameter/option parsing errors
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function modifyCollection($collectionOrView, $collModOptionValue, array $options = [])
```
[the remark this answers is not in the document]
jmikola: Also, let's type-hint `$collectionOptions` as an array. Earlier I mentioned that it could be an array or object, since it's a document, but it's easier if we just stick to an array. It's unlikely that any users are going going to pass in an object here.

Expecting an array is also consistent with the CreateIndexes command, as we expect each index specification (technically a document) to be a PHP array. CreateCollection also takes collection options as an array, but that's somewhat special since it's mixed in with the operation options (e.g. writeConcern, typeMap).

src/Operation/ModifyCollection.php
```diff
@@ -0,0 +1,146 @@
...
+        foreach($this->collModOptionValue as $key => $value) {
+            $cmd[$key] = $value;
+        }
```
jmikola: If `$this->collectionOptions` is going to be an array, we can simplify this entire method as:

```php
return new Command(['collMod' => $this->collectionName] + $this->collectionOptions);
```

That effectively merges the arrays with a union, ensuring the `collMod` in the left-hand argument remains as-is.

You may decide to just do this in the `execute()` method for forgo `createCommand()` entirely.

src/Operation/ModifyCollection.php
```diff
@@ -0,0 +1,146 @@
...
+     *    This is not supported for server versions < 3.2 and will result in an
+     *    exception at execution time if used.
+     *
+     * @param string       $databaseName       Database name
+     * @param string       $collectionOrView   Collection or view to modify
+     * @param string       $collModOptionValue Options and values for collMod
+     * @param array        $options            Command options
+     * @throws InvalidArgumentException for parameter/option parsing errors
+     */
+    public function __construct($databaseName, $collectionOrView, $collModOptionValue, array $options = [])
```
jmikola: ... `, $collectionName, array $collectionOptions, ` ...

Let's not default `$collectionOptions` to anything, since I think it makes sense to require some value (related to my next comment).

src/Operation/ModifyCollection.php
```diff
@@ -0,0 +1,146 @@
...
+    {
```
jmikola: Before we validate options, I think we can add a small check to ensure that the `$collectionOptions` array is not empty, as is done for `$operations` in BulkWrite.

In my tests, MongoDB 3.6 lets us run the command `{ "collMod": "foo" }` with no additional options successfully, but I have to assume that's a no-op. You can double-check with @ksuarz and/or @visemet, but think it'd make sense to yell at our users for trying to do that. If anything, it'd help them realize they're inadvertently running a no-op.

tests/Operation/ModifyCollectionFunctionalTest.php
```diff
@@ -0,0 +1,32 @@
...
+        $indexes = [['key' => ['lastAccess' => 1], 'expireAfterSeconds' => 3]];
+        $createIndexes = new CreateIndexes($this->getDatabaseName(), $this->getCollectionName(), $indexes);
+        $createIndexes->execute($this->getPrimaryServer());
+
+        $keyPattern = (object) ['lastAccess' => 1];
+        $modifyCollection = new ModifyCollection(
+                                $this->getDatabaseName(),
+                                $this->getCollectionName(),
+                                ['index' => ['keyPattern' => $keyPattern, 'expireAfterSeconds' => 1000]],
+                                ['typeMap' => ['root' => 'array']]
```
jmikola: Just confirming: this is needed because we the operation doesn't inherit a typeMap by default, so the driver returns the root document as a `stdClass`.

tests/Operation/ModifyCollectionFunctionalTest.php
```diff
@@ -0,0 +1,32 @@
...
+                            );
+        $result = ($modifyCollection->execute($this->getPrimaryServer()));
```
jmikola: Extra parens around the return value can be removed.

tests/Database/DatabaseFunctionalTest.php
```diff
@@ -152,6 +153,23 @@ public function testGetSelectsCollectionAndInheritsOptions()
...
 
+    public function testModifyCollection()
+    {
+        $this->database->createCollection($this->getCollectionName());
+
+        $indexes = [['key' => ['lastAccess' => 1], 'expireAfterSeconds' => 3]];
+        $createIndexes = new CreateIndexes($this->getDatabaseName(), $this->getCollectionName(), $indexes);
+        $createIndexes->execute($this->getPrimaryServer());
+
+        $keyPattern = (object) ['lastAccess' => 1];
```
jmikola: Is the `(object)` cast necessary? `['lastAccess' => 1]` should already serialize as a BSON document.

If it's not needed, can we just use a `['lastAccess' => 1]` literal below and skip this variable?

tests/Operation/ModifyCollectionFunctionalTest.php
```diff
@@ -0,0 +1,32 @@
...
+    public function testCollMod()
+    {
+        $operation = new CreateCollection($this->getDatabaseName(), $this->getCollectionName());
+        $operation->execute($this->getPrimaryServer());
+
+        $indexes = [['key' => ['lastAccess' => 1], 'expireAfterSeconds' => 3]];
+        $createIndexes = new CreateIndexes($this->getDatabaseName(), $this->getCollectionName(), $indexes);
+        $createIndexes->execute($this->getPrimaryServer());
+
+        $keyPattern = (object) ['lastAccess' => 1];
```
jmikola: Likewise, not sure if `(object)` cast is needed. If not, can we just use the `['lastAccess' => 1]` literal below and do without this variable?

tests/Operation/ModifyCollectionFunctionalTest.php
```diff
@@ -0,0 +1,32 @@
...
+        $modifyCollection = new ModifyCollection(
+                                $this->getDatabaseName(),
```
jmikola: I realize we don't have a code formatting profile yet, but if you'd like to indent a constructor/function call, let's use 4 spaces to indent the arguments and then have `);` on its own line without indentation. For example:

```php
$foo = new Foo(
    'foo',
    'bar'
);
```

tests/Operation/ModifyCollectionTest.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace MongoDB\Tests\Operation;
+
+use MongoDB\Exception\InvalidArgumentException;
+use MongoDB\Operation\ModifyCollection;
+
+class ModifyCollectionTest extends TestCase
+{
```
jmikola: If you end up adding a non-empty check for `$collectionOptions`, we can have another test here to demonstrate that `[]` throws an exception.

src/Operation/ModifyCollection.php
ksuarz: I looked at the {{collMod}} code in the server; it does what you'd expect in that the options for the collection or view are not changed. But it doesn't look like we optimize it to be a "no-op"; my reading of the code is that it will still commit a metadata write to the storage layer.

I don't think it really matters either way; a {{collMod}} with no options is still a valid command and the server will just do what you tell it to do.

jmikola: Understandable that it's not technically a no-op on the server if it commits. Thanks for looking into it.

@kvwalker: Let's still throw here.

docs/includes/apiargs-MongoDBDatabase-method-modifyCollection-param.yaml
```diff
@@ -0,0 +1,19 @@
+source:
+  file: apiargs-common-param.yaml
+  ref: $collectionName
+replacement:
+  action: " or view to modify"
```
jmikola: In the interest of not overloading the purpose of `{{action}}`, can you see if we can change the `$collectionName` definition in `apiargs-common-param.yaml` to:

```yaml
arg_name: param
name: $collectionName
type: string
description: |
  The name of the {{subject}}{{action}}.
interface: phpmethod
operation: ~
optional: false
replacement:
  subject: "collection"
  action: ""
```

And then this can use the following replacement:

```yaml
replacement:
  subject: "collection or view"
  action: " to modify"
```

docs/reference/method/MongoDBDatabase-modifyCollection.txt
```diff
@@ -0,0 +1,110 @@
...
+The output would then resemble::
+
+   object(stdClass)#2779 {
+     ["expireAfterSeconds_old"]=>
+     int(3)
+     ["expireAfterSeconds_new"]=>
+     int(1000)
+     ["ok"]=>
+     float(1)
+     ["operationTime"]=>
```
jmikola: I think we can remove `["operationTime"]` and `["$clusterTime"]` from the output. "would then resemble" above implies that we may not be presenting the exact response from the server.

PR #519, opened 2018-03-30
PHPLIB-84: Allow Collection::dropIndex() to accept an IndexInfo object

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-84

### DIFF
--- a/docs/includes/apiargs-MongoDBCollection-method-dropIndex-param.yaml
+++ b/docs/includes/apiargs-MongoDBCollection-method-dropIndex-param.yaml
@@ -1,10 +1,10 @@
 arg_name: param
 name: $indexName
-type: string
+type: string| :phpclass:`MongoDB\\Model\\IndexInfo`
 description: |
-  The name of the index to drop. View the existing indexes on the collection
-  using the :phpmethod:`listIndexes() <MongoDB\\Collection::listIndexes>`
-  method.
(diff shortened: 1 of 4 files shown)


src/Collection.php
```diff
@@ -443,7 +443,7 @@ public function drop(array $options = [])
      * Drop a single index in the collection.
      *
      * @see DropIndexes::__construct() for supported options
-     * @param string $indexName Index name
+     * @param string|IndexInfo $indexName Index name or model object
```
jmikola: Please add `use MongoDB\Model\IndexInfo;` up top.

docs/includes/apiargs-MongoDBCollection-method-dropIndex-param.yaml
```diff
@@ -1,10 +1,10 @@
 arg_name: param
 name: $indexName
-type: string
+type: string| :phpclass:`MongoDB\\Model\\IndexInfo`
```
kvwalker: Note: I added a space after the `|` so the `:phpclass:` renders properly in the docs.

jmikola: Interesting. Can you ask someone in #docs about this later? Just curious if they have a work-around for it.

PR #520, opened 2018-04-02
PHPLIB-226: Implement ArrayAccess for CollectionInfo and DatabaseInfo

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-226

### DIFF
--- a/src/Model/DatabaseInfo.php
+++ b/src/Model/DatabaseInfo.php
@@ -17,6 +17,8 @@
 
 namespace MongoDB\Model;
 
+use MongoDB\Exception\BadMethodCallException;
+use ArrayAccess;
 /**
  * Database information model class.
  *
@@ -27,7 +29,7 @@ namespace MongoDB\Model;
  * @see \MongoDB\Client::listDatabases()
  * @see http://docs.mongodb.org/manual/reference/command/listDatabases/
  */
-class DatabaseInfo
+class DatabaseInfo implements ArrayAccess
 {
     private $info;
 
@@ -80,5 +82,51 @@ class DatabaseInfo
     public function isEmpty()
     {
         return (boolean) $this->info['empty'];
+    }
+
+    /**
+     * Check whether a field exists in the database information.
+     *
+     * @see http://php.net/arrayaccess.offsetexists
+     * @param mixed $key
+     * @return boolean
+     */
+    public function offsetExists($key)
+    {
+        return array_key_exists($key, $this->info);
+    }
+
+    /**
+     * Return the field's value from the database information.
+     *
+     * @see http://php.net/arrayaccess.offsetget
+     * @param mixed $key
+     * @return mixed
+     */
+    public function offsetGet($key)
+    {
+        return $this->info[$key];
+    }
+
+    /**
+     * Not supported.
+     *
+     * @see http://php.net/arrayaccess.offsetset
+     * @throws BadMethodCallException
+     */
+    public function offsetSet($key, $value)
+    {
+        throw BadMethodCallException::classIsImmutable(__CLASS__);
+    }
+
+    /**
+     * Not supported.
+     *
+     * @see http://php.net/arrayaccess.offsetunset
+     * @throws BadMethodCallException
+     */
+    public function offsetUnset($key)
+    {
+        throw BadMethodCallException::classIsImmutable(__CLASS__);
     }
 }
--- a/tests/Model/DatabaseInfoTest.php
+++ b/tests/Model/DatabaseInfoTest.php
@@ -2,6 +2,7 @@
 
 namespace MongoDB\Tests\Model;
 
+use MongoDB\Exception\BadMethodCallException;
 use MongoDB\Model\DatabaseInfo;
 use MongoDB\Tests\TestCase;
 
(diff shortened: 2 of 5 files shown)


tests/Operation/ListCollectionsFunctionalTest.php
```diff
@@ -34,6 +34,34 @@ public function testListCollectionsForNewlyCreatedDatabase()
...
+        $writeResult = $insertOne->execute($server);
+        $this->assertEquals(1, $writeResult->getInsertedCount());
+
+        $operation = new ListCollections($this->getDatabaseName(), ['filter' => ['name' => $this->getCollectionName()]]);
+        $collections = $operation->execute($server);
+
+        $this->assertInstanceOf('MongoDB\Model\CollectionInfoIterator', $collections);
+
+        foreach ($collections as $collection) {
+            $this->assertInstanceOf('MongoDB\Model\CollectionInfo', $collection);
```
derickr: Would it make sense to assert the type of the return of ``getInfo()`` and ``getIdIndex()`` values to the documented "document" too, if they are not ``null``?

kvwalker: Good point. The return type should actually be `array|object|null`, since that is used in place of `document`. I fixed that in the most recent commit. I don't think the tests usually check the type of the return if it's just `array` or `object`.

derickr: It's possible that ``document`` was used to refer to the ``array|object`` combination before, so perhaps ``document|null`` was "expected". In any case, your [change](https://github.com/mongodb/mongo-php-library/pull/520/files#diff-3b734177ac5b3b9a25cc0a63acfa4989R89) now left out the ``|null`` part, which I'm not sure whether that's what you wanted to do.

kvwalker: My bad, thanks for catching that!

tests/Operation/ListCollectionsFunctionalTest.php
```diff
@@ -34,6 +34,33 @@ public function testListCollectionsForNewlyCreatedDatabase()
...
+            $this->assertArrayHasKey('readOnly', $collection['info']);
+            $this->assertEquals($collection['idIndex'], ['v' => 2, 'key' => ['_id' => 1], 'name' => '_id_', 'ns' => $this->getNamespace()]);
```
jmikola: The expected value comes before the actual value in [`assertEquals()`](https://phpunit.de/manual/current/en/appendixes.assertions.html#appendixes.assertions.assertEquals) (and other assertions). This makes no difference for a passing test but becomes important when PHPUnit renders an error message for a test failure.

tests/Operation/ListCollectionsFunctionalTest.php
```diff
@@ -34,6 +34,33 @@ public function testListCollectionsForNewlyCreatedDatabase()
...
+    public function testOffsetGet()
+    {
+        if (version_compare($this->getServerVersion(), '3.4.0', '<')) {
+            $this->markTestSkipped('idIndex and info are not supported');
+        }
+
+        $server = $this->getPrimaryServer();
+
+        $operation = new DropDatabase($this->getDatabaseName());
+        $operation->execute($server);
```
jmikola: Is this necessary? The base FunctionalTestCase class for Operations already drops the collection during setup.

Assuming this was copied from `testListCollectionsForNewlyCreatedDatabase()`, I think that was a unique case where we needed to ensure that the database was just created during the test.

tests/Operation/ListDatabasesFunctionalTest.php
```diff
@@ -52,6 +52,25 @@ public function testFilterOption()
         }
     }
 
+    public function testOffsetGet()
```
jmikola: These would be more suited as unit tests within the existing model test classes (DatabaseInfoTest and CollectionInfoTest). The following IndexInfoTest should work as prior art:

 * `testImplementsArrayAccess()`
 * `testOffsetSetCannotBeCalled()`
 * `testOffsetUnsetCannotBeCalled()`

tests/Operation/ListCollectionsFunctionalTest.php
```diff
@@ -34,6 +34,30 @@ public function testListCollectionsForNewlyCreatedDatabase()
         }
     }
 
+    public function testOffsetGet()
+    {
+        if (version_compare($this->getServerVersion(), '3.4.0', '<')) {
+            $this->markTestSkipped('idIndex and info are not supported');
+        }
```
jmikola: I don't see any harm in keeping this for testing 3.4+ servers, but I'd suggest renaming it from `testOffsetGet()` to something like `testIdIndexAndInfo()` instead.

tests/Operation/ListDatabasesFunctionalTest.php
jmikola: Unlike the CollectionInfo functional test (for [SERVER-26513](https://jira.mongodb.org/browse/SERVER-26513)), I don't think we need a functional test for extra fields in DatabaseInfo.

src/Model/DatabaseInfo.php
```diff
@@ -81,4 +83,51 @@ public function isEmpty()
...
+     * Not supported.
+     *
+     * @see http://php.net/arrayaccess.offsetunset
+     * @throws BadMethodCallException
+     */
+    public function offsetUnset($key)
+    {
+        throw BadMethodCallException::classIsImmutable(__CLASS__);
+    }
+
```
jmikola: Extra newline here can be removed.

tests/Model/DatabaseInfoTest.php
```diff
@@ -39,4 +40,29 @@ public function testDebugInfo()
...
+    }
+
+    public function testOffsetUnsetCannotBeCalled()
+    {
+        $info = new DatabaseInfo(['name' => 'foo', 'sizeOnDisk' => 1048576, 'empty' => false]);
+        $this->expectException(BadMethodCallException::class);
+        $this->expectExceptionMessage('MongoDB\Model\DatabaseInfo is immutable');
+        unset($info['empty']);
+    }
+
```
jmikola: Extra newline here as well.

PR #521, opened 2018-04-03
PHPLIB-80: Add IndexInfo helper methods for special index type options

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-80

### DIFF
--- a/docs/reference/method/MongoDBModelIndexInfo-is2dSphere.txt
+++ b/docs/reference/method/MongoDBModelIndexInfo-is2dSphere.txt
@@ -1,0 +1,59 @@
+=======================================
+MongoDB\\Model\\IndexInfo::is2dSphere()
+=======================================
+
+.. versionadded:: 1.4
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Definition
+----------
+
+.. phpmethod:: MongoDB\\Model\\IndexInfo::is2dSphere()
+
+   Return whether the index is a :manual:`2dsphere </core/2dsphere>`
+   index.
+
+   .. code-block:: php
+
+      function is2dSphere(): boolean
+
+Return Values
+-------------
+
+A boolean indicating whether the index is a 2dsphere index.
+
+Examples
+--------
+
+.. code-block:: php
+
+   <?php
+
+   $collection = (new MongoDB\Client)->selectCollection('test', 'places');
+
+   $collection->createIndex(['pos' => '2dsphere']);
+
+   foreach ($collection->listIndexes() as $index) {
+       if ($index->is2dSphere()) {
+           printf("%s has 2dsphereIndexVersion: %d\n", $index->getName(), $index['2dsphereIndexVersion']);
+       }
+   }
+
+The output would then resemble::
+
+   pos_2dsphere has 2dsphereIndexVersion: 3
+
+See Also
+--------
+
+- :phpmethod:`MongoDB\\Collection::createIndex()`
+- :phpmethod:`MongoDB\\Collection::listIndexes()`
+- :manual:`2dsphere Indexes </core/2dsphere>` reference in the MongoDB
+  manual
--- a/docs/reference/method/MongoDBModelIndexInfo-isGeoHaystack.txt
+++ b/docs/reference/method/MongoDBModelIndexInfo-isGeoHaystack.txt
@@ -1,0 +1,59 @@
+==========================================
+MongoDB\\Model\\IndexInfo::isGeoHaystack()
+==========================================
+
+.. versionadded:: 1.4
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Definition
+----------
+
+.. phpmethod:: MongoDB\\Model\\IndexInfo::isGeoHaystack()
+
+   Return whether the index is a :manual:`geoHaystack
+   </core/geohaystack>` index.
+
+   .. code-block:: php
+
+      function isGeoHaystack(): boolean
+
+Return Values
+-------------
+
+A boolean indicating whether the index is a geoHaystack index.
+
+Examples
+--------
+
+.. code-block:: php
+
+   <?php
+
+   $collection = (new MongoDB\Client)->selectCollection('test', 'places');
+
+   $collection->createIndex(['pos' => 'geoHaystack', 'x' => 1], ['bucketSize' => 5]);
+
+   foreach ($collection->listIndexes() as $index) {
+       if ($index->isGeoHaystack()) {
+           printf("%s has bucketSize: %d\n", $index->getName(), $index['bucketSize']);
+       }
+   }
+
+The output would then resemble::
+
+   pos_geoHaystack_x_1 has bucketSize: 5
+
+See Also
+--------
+
+- :phpmethod:`MongoDB\\Collection::createIndex()`
+- :phpmethod:`MongoDB\\Collection::listIndexes()`
+- :manual:`geoHaystack Indexes </core/geohaystack>` reference in the MongoDB
+  manual
--- a/docs/reference/method/MongoDBModelIndexInfo-isText.txt
+++ b/docs/reference/method/MongoDBModelIndexInfo-isText.txt
@@ -1,0 +1,58 @@
+===================================
+MongoDB\\Model\\IndexInfo::isText()
+===================================
+
+.. versionadded:: 1.4
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Definition
+----------
+
+.. phpmethod:: MongoDB\\Model\\IndexInfo::isText()
+
+   Return whether the index is a :manual:`text </core/index-text>` index.
+
+   .. code-block:: php
+
+      function isText(): boolean
+
+Return Values
+-------------
+
+A boolean indicating whether the index is a text index.
+
+Examples
+--------
+
+.. code-block:: php
+
+   <?php
+
+   $collection = (new MongoDB\Client)->selectCollection('test', 'restaurants');
+
+   $collection->createIndex(['name' => 'text']);
+
+   foreach ($collection->listIndexes() as $index) {
+       if ($index->isText()) {
+           printf("%s has default language: %d\n", $index->getName(), $index['default_language']);
+       }
+   }
+
+The output would then resemble::
+
+   name_text has default language: english
+
+See Also
+--------
+
+- :phpmethod:`MongoDB\\Collection::createIndex()`
+- :phpmethod:`MongoDB\\Collection::listIndexes()`
+- :manual:`Text Indexes </core/index-text>` reference in the MongoDB
+  manual
--- a/src/Model/IndexInfo.php
+++ b/src/Model/IndexInfo.php
@@ -111,6 +111,26 @@ class IndexInfo implements ArrayAccess
     }
 
     /**
+     * Return whether or not this index is of type 2dsphere.
+     *
+     * @return boolean
+     */
+    public function is2dSphere()
+    {
+        return array_search('2dsphere', $this->getKey(), true) !== false;
+    }
+
+    /**
+     * Return whether or not this index is of type geoHaystack.
+     *
+     * @return boolean
+     */
+    public function isGeoHaystack()
+    {
+        return array_search('geoHaystack', $this->getKey(), true) !== false;
+    }
+
+    /**
      * Return whether this is a sparse index.
      *
      * @see http://docs.mongodb.org/manual/core/index-sparse/
@@ -119,6 +139,16 @@ class IndexInfo implements ArrayAccess
     public function isSparse()
     {
         return ! empty($this->info['sparse']);
+    }
+
+    /**
+     * Return whether or not this index is of type text.
+     *
+     * @return boolean
+     */
+    public function isText()
+    {
+        return array_search('text', $this->getKey(), true) !== false;
     }
 
     /**
--- a/tests/Model/IndexInfoFunctionalTest.php
+++ b/tests/Model/IndexInfoFunctionalTest.php
@@ -1,0 +1,78 @@
+<?php
+
+namespace MongoDB\Tests\Model;
+
+use MongoDB\Collection;
+use MongoDB\Tests\FunctionalTestCase;
+
+class IndexInfoFunctionalTest extends FunctionalTestCase
+{
+    private $collection;
+
+    public function setUp()
+    {
+        parent::setUp();
+
+        $this->collection = new Collection($this->manager, $this->getDatabaseName(), $this->getCollectionName());
+        $this->collection->drop();
+    }
+
+    public function tearDown()
+    {
+        if ($this->hasFailed()) {
+            return;
+        }
+
+        $this->collection->drop();
+    }
+
+    public function testIs2dSphere()
+    {
+        $indexName = $this->collection->createIndex(['pos' => '2dsphere']);
+        $result = $this->collection->listIndexes();
+
+        $result->rewind();
+        $result->next();
+        $index = $result->current();
+
+        $this->assertEquals($indexName, $index->getName());
+        $this->assertTrue($index->is2dSphere());
+
+        $expectedVersion = version_compare($this->getServerVersion(), '3.2.0', '<') ? 2 : 3;
+        $this->assertEquals($expectedVersion, $index['2dsphereIndexVersion']);
+    }
+
+    public function testIsGeoHaystack()
+    {
+        $indexName = $this->collection->createIndex(['pos' => 'geoHaystack', 'x' => 1], ['bucketSize' => 5]);
+        $result = $this->collection->listIndexes();
+
+        $result->rewind();
+        $result->next();
+        $index = $result->current();
+
+        $this->assertEquals($indexName, $index->getName());
+        $this->assertTrue($index->isGeoHaystack());
+        $this->assertEquals(5, $index['bucketSize']);
+    }
+
+    public function testIsText()
+    {
+        $indexName = $this->collection->createIndex(['x' => 'text']);
+        $result = $this->collection->listIndexes();
+
+        $result->rewind();
+        $result->next();
+        $index = $result->current();
+
+        $this->assertEquals($indexName, $index->getName());
+        $this->assertTrue($index->isText());
+        $this->assertEquals('english', $index['default_language']);
+        $this->assertEquals('language', $index['language_override']);
+
+        $expectedVersion = version_compare($this->getServerVersion(), '3.2.0', '<') ? 2 : 3;
+        $this->assertEquals($expectedVersion, $index['textIndexVersion']);
+
+        $this->assertSameDocument(['x' => 1], $index['weights']);
+    }
+}
--- a/tests/Model/IndexInfoTest.php
+++ b/tests/Model/IndexInfoTest.php
@@ -21,7 +21,10 @@ class IndexInfoTest extends TestCase
         $this->assertSame(['x' => 1], $info->getKey());
         $this->assertSame('x_1', $info->getName());
         $this->assertSame('foo.bar', $info->getNamespace());
+        $this->assertFalse($info->is2dSphere());
+        $this->assertFalse($info->isGeoHaystack());
         $this->assertFalse($info->isSparse());
+        $this->assertFalse($info->isText());
         $this->assertFalse($info->isTtl());
         $this->assertFalse($info->isUnique());
     }
@@ -40,7 +43,10 @@ class IndexInfoTest extends TestCase
         $this->assertSame(['y' => 1], $info->getKey());
         $this->assertSame('y_sparse', $info->getName());
         $this->assertSame('foo.bar', $info->getNamespace());
+        $this->assertFalse($info->is2dSphere());
+        $this->assertFalse($info->isGeoHaystack());
         $this->assertTrue($info->isSparse());
+        $this->assertFalse($info->isText());
         $this->assertFalse($info->isTtl());
         $this->assertFalse($info->isUnique());
     }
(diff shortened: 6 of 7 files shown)


src/Model/IndexInfo.php
```diff
@@ -100,6 +154,42 @@ public function getNamespace()
...
+
+    /**
+     * Return the text index version.
+     *
+     * @return integer|null
+     */
+    public function getTextIndexVersion()
+    {
+        if ( ! empty($this->info['textIndexVersion'])) {
+            return $this->info['textIndexVersion'];
```
derickr: In line 200 (just below), the ``getVersion()`` method does a cast to an ``(integer)`` for the version number. Perhaps you want to do that here, as well as for ``get2dsphereIndexVersion`` too?

src/Model/IndexInfo.php
```diff
@@ -110,6 +200,29 @@ public function getVersion()
...
 
+    /**
+     * Return the weights document for text indexes.
+     *
+     * @return document
+     */
+    public function getWeights()
+    {
+        if (empty($this->info['weights'])) {
+            return self::DEFAULT_WEIGHT;
```
derickr: ``self::DEFAULT_WEIGHT`` is not declared as a ``document`` (array) in the class definition, but as a simple number ``1``. I think it should return ``null`` if ``weights`` is empty, but happy to hear with @jmikola has to say on this too.

src/Model/IndexInfo.php
```diff
@@ -110,6 +200,29 @@ public function getVersion()
...
+    }
+
+    /**
+     * Return whether background is true.
+     *
+     * @return boolean
+     */
+    public function isBackground()
+    {
+        return ! empty($this->info['background']);
```
derickr: As you can set the ``background`` option to ``false``, this line is not right, as it would return ``true`` even if ``background`` is set to false, as in:

```
> db.foo1.createIndex( { test2: 'text' }, { background: false } );
{
	"createdCollectionAutomatically" : true,
	"numIndexesBefore" : 1,
	"numIndexesAfter" : 2,
	"ok" : 1
}
> db.foo1.getIndexes();
[
…
	{
		"v" : 2,
		"key" : {
			"_fts" : "text",
			"_ftsx" : 1
		},
		"name" : "test2_text",
		"ns" : "test.foo1",
		"background" : false,
		"weights" : {
			"test2" : 1
		},
		"default_language" : "english",
		"language_override" : "language",
		"textIndexVersion" : 3
	}
]
```

src/Model/IndexInfo.php
derickr: The next comment shows how ``weights`` looks like, when a text index exists:

tests/Model/IndexInfoTest.php
```diff
@@ -139,4 +139,24 @@ public function testOffsetUnsetCannotBeCalled()
...
+        $this->assertNull($info->get2dSphereIndexVersion());
+        $this->assertNull($info->getBucketSize());
+        $this->assertEquals(IndexInfo::DEFAULT_LANGUAGE, $info->getDefaultLanguage());
+        $this->assertEquals(IndexInfo::DEFAULT_LANGUAGE_OVERRIDE, $info->getLanguageOverride());
+        $this->assertNull($info->getPartialFilterExpression());
+        $this->assertNull($info->getStorageEngineName());
+        $this->assertNull($info->getTextIndexVersion());
+        $this->assertEquals(IndexInfo::DEFAULT_WEIGHT, $info->getWeights());
+        $this->assertFalse($info->isBackground());
+    }
```
derickr: I think it's good that you test for default values in case elements in the IndexInfo struct are missing, but I also think the tests should cover the cases where the elements are present. In other comments I've shown a few possible cases (for simple ``text``, and ``2dsphere`` indexes), and I think it would be good to add tests cases for these as well.

src/Model/IndexInfo.php
```diff
@@ -70,6 +74,43 @@ public function __toString()
...
+
+    /**
+     * Return the default language of a text index.
+     *
+     * @return string
+     */
+    public function getDefaultLanguage()
+    {
+        if (empty($this->info['default_language'])) {
+            return self::DEFAULT_LANGUAGE;
```
derickr: I'm quite quite sure whether we should return the defaults ``self::DEFAULT_LANGUAGE_OVERRIDE`` and ``self::DEFAULT_LANGUAGE`` that you have defined, or just ``null``, as all non-text indexes will not have this information, and returning ``english``/``language`` might not be expected here.

src/Model/IndexInfo.php
derickr: Teaching me PHP now huh ;-) You're absolutely right. My mistake.

tests/Model/IndexInfoFunctionalTest.php
```diff
@@ -0,0 +1,153 @@
...
+    public function testGetWeights()
+    {
+        $this->collection->createIndex(['x' => 'text']);
+        $result = $this->collection->listIndexes();
+
+        $result->rewind();
+        $this->assertInstanceOf('MongoDB\Model\IndexInfo', $result->current());
+        $result->next();
+
+        $this->assertSameDocument(['x' => 1], $result->current()->getWeights());
```
derickr: I would probably add one more test here that actually sets up the weights for different text fields, as is described [here](https://docs.mongodb.com/manual/tutorial/control-results-of-text-search/).

docs/reference/method/MongoDBModelIndexInfo-getBucketSize.txt
```diff
@@ -0,0 +1,55 @@
...
+.. code-block:: php
+
+   <?php
+   $collection->createIndex(['x' => 'geoHaystack', 'y' => 1], ['bucketSize' => 5]);
+
+   $resultIterator = $collection->listIndexes();
+   $resultIterator->rewind();
+   $resultIterator->next();
+
+   var_dump($resultIterator->current()->getBucketSize());
```
jmikola: If we change this documentation to demonstrate `isGeoHaystack()`:

```php
$collection = (new MongoDB\Client)->selectCollection('test', 'places');

$collection->createIndex(['pos' => 'geoHaystack', 'x' => 1], ['bucketSize' => 5]);

foreach ($collection->listIndexes() as $index) {
    if ($index->isGeoHaystack()) {
        printf("%s has bucketSize: %d\n", $index->getName(), $index['bucketSize']);
    }
}
```

This should output:

> pos_geoHaystack_x_1 has bucketSize: 5

src/Model/IndexInfo.php
```diff
@@ -70,6 +70,42 @@ public function __toString()
         return $this->getName();
     }
 
+    /**
```
jmikola: Looking at the methods introduced in this PR, I think it'd be more worthwhile to add "is" methods for the various index types. That would be both more maintainable _and_ easier to document.

 * `is2d()`
 * `is2dSphere()`
 * `isGeoHaystack()`
 * `isText()`

Consider the following shell session:

```
> db.foo.createIndex({title:"text",description:"text"})
{
	"createdCollectionAutomatically" : true,
	"numIndexesBefore" : 1,
	"numIndexesAfter" : 2,
	"ok" : 1
}
> db.foo.createIndex({pos1:"2d"})
{
	"createdCollectionAutomatically" : false,
	"numIndexesBefore" : 2,
	"numIndexesAfter" : 3,
	"ok" : 1
}
> db.foo.createIndex({pos2:"2dsphere"})
{
	"createdCollectionAutomatically" : false,
	"numIndexesBefore" : 3,
	"numIndexesAfter" : 4,
	"ok" : 1
}
> db.foo.createIndex({pos3:"geoHaystack",x:1},{bucketSize:1})
{
	"createdCollectionAutomatically" : false,
	"numIndexesBefore" : 4,
	"numIndexesAfter" : 5,
	"ok" : 1
}
> db.foo.getIndexSpecs()
[
	{
		"v" : 2,
		"key" : {
			"_id" : 1
		},
		"name" : "_id_",
		"ns" : "test.foo"
	},
	{
		"v" : 2,
		"key" : {
			"_fts" : "text",
			"_ftsx" : 1
		},
		"name" : "title_text_description_text",
		"ns" : "test.foo",
		"weights" : {
			"description" : 1,
			"title" : 1
		},
		"default_language" : "english",
		"language_override" : "language",
		"textIndexVersion" : 3
	},
	{
		"v" : 2,
		"key" : {
			"pos1" : "2d"
		},
		"name" : "pos1_2d",
		"ns" : "test.foo"
	},
	{
		"v" : 2,
		"key" : {
			"pos2" : "2dsphere"
		},
		"name" : "pos2_2dsphere",
		"ns" : "test.foo",
		"2dsphereIndexVersion" : 3
	},
	{
		"v" : 2,
		"key" : {
			"pos3" : "geoHaystack",
			"x" : 1
		},
		"name" : "pos3_geoHaystack_x_1",
		"ns" : "test.foo",
		"bucketSize" : 1
	}
]
>
```

A robust method to implement `is<type>()` for each of these four types might be:

```php
return array_search('typeName', $this->getKey(), true) !== false;
```

This relies on `getKey()` to cast the "key" field to an array, since it is technically a document and could be a stdClass, and then uses [`array_search()`](http://php.net/array_search) to check if any field (i.e. key) has the type we're looking for. Roughly equivalent to:

```php
foreach ($this->info['key'] as $type) {
    if ($type === 'typeName') {
        return true;
    }
}

return false;
```

docs/reference/method/MongoDBModelIndexInfo-get2dSphereIndexVersion.txt
```diff
@@ -0,0 +1,54 @@
...
+--------
+
+.. code-block:: php
+
+   <?php
+   $collection->createIndex(['x' => '2dsphere']);
+
+   $resultIterator = $collection->listIndexes();
+   $resultIterator->rewind();
+   $resultIterator->next();
```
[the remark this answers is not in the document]
jmikola: If we change this documentation to demonstrate `is2dSphere()`:

```php
$collection = (new MongoDB\Client)->selectCollection('test', 'places');

$collection->createIndex(['pos' => '2dsphere']);

foreach ($collection->listIndexes() as $index) {
    if ($index->is2dSphere()) {
        printf("%s has 2dsphereIndexVersion: %d\n", $index->getName(), $index['2dsphereIndexVersion']);
    }
}
```

This should output:

> pos_2dsphere has 2dsphereIndexVersion: 3

Side benefit here would be that we're also demonstrating how users can rely on `offsetGet()` to access type-specific index fields on their own.

docs/reference/method/MongoDBModelIndexInfo-is2dSphere.txt
```diff
@@ -0,0 +1,58 @@
+==========================================
+MongoDB\\Model\\IndexInfo::is2dSphere()
+==========================================
```
jmikola: IIRC, RST emits a warning if the character length of headings and lines are not equal. Three characters should be removed on this and the first line.

docs/reference/method/MongoDBModelIndexInfo-is2dSphere.txt
```diff
@@ -0,0 +1,58 @@
...
+
+A boolean indicating whether the index is a 2dsphere index.
+
+Examples
+--------
+
+.. code-block:: php
+
+   <?php
+   $collection = (new MongoDB\Client)->selectCollection('test', 'places');
```
jmikola: Newline after `<?php` for consistency with other examples.

docs/reference/method/MongoDBModelIndexInfo-is2dSphere.txt
```diff
@@ -0,0 +1,58 @@
...
+
+   $collection->createIndex(['pos' => '2dsphere']);
+
+   foreach ($collection->listIndexes() as $index) {
+   if ($index->is2dSphere()) {
+      printf("%s has 2dsphereIndexVersion: %d\n", $index->getName(), $index['2dsphereIndexVersion']);
+      }
```
jmikola: Indentation is a bit off here. Should be:

```
foreach ($collection->listIndexes() as $index) {
    if ($index->is2dSphere()) {
        printf("%s has 2dsphereIndexVersion: %d\n", $index->getName(), $index['2dsphereIndexVersion']);
    }
}
```

docs/reference/method/MongoDBModelIndexInfo-is2dSphere.txt
```diff
@@ -0,0 +1,58 @@
...
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Definition
+----------
+
+.. phpmethod:: MongoDB\\Model\\IndexInfo::is2dSphere()
+
+   Return whether or not the index is a :manual:`2dsphere </core/2dsphere>`
```
jmikola: I think we can remove "or not" to be consistent with other documentation for "is" methods.

docs/reference/method/MongoDBModelIndexInfo-isBackground.txt
```diff
@@ -0,0 +1,53 @@
+=========================================
+MongoDB\\Model\\IndexInfo::isBackground()
```
jmikola: I'd propose deleting this method. I'm not sure it's relevant to users once the index has been created, and it's trivial to determine this via `offsetGet()`.

docs/reference/method/MongoDBModelIndexInfo-isGeoHaystack.txt
```diff
@@ -0,0 +1,58 @@
...
+
+   <?php
+   $collection = (new MongoDB\Client)->selectCollection('test', 'places');
+
+   $collection->createIndex(['pos' => 'geoHaystack', 'x' => 1], ['bucketSize' => 5]);
+
+   foreach ($collection->listIndexes() as $index) {
+      if ($index->isGeoHaystack()) {
+         printf("%s has bucketSize: %d\n", $index->getName(), $index['bucketSize']);
+      }
```
jmikola: Looks like this is using 3-space instead of 4 for indents.

docs/reference/method/MongoDBModelIndexInfo-isGeoHaystack.txt
```diff
@@ -0,0 +1,58 @@
...
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Definition
+----------
+
+.. phpmethod:: MongoDB\\Model\\IndexInfo::isGeoHaystack()
+
+   Return whether or not the index is a :manual:`geoHaystack
```
jmikola: ~~or not~~

docs/reference/method/MongoDBModelIndexInfo-isGeoHaystack.txt
```diff
@@ -0,0 +1,58 @@
...
+
+A boolean indicating whether the index is a geoHaystack index.
+
+Examples
+--------
+
+.. code-block:: php
+
+   <?php
+   $collection = (new MongoDB\Client)->selectCollection('test', 'places');
```
jmikola: Newline after `<?php` for consistency.

docs/reference/method/MongoDBModelIndexInfo-isText.txt
```diff
@@ -0,0 +1,57 @@
+==========================================
+MongoDB\\Model\\IndexInfo::isText()
+==========================================
```
jmikola: Title length.

docs/reference/method/MongoDBModelIndexInfo-isText.txt
```diff
@@ -0,0 +1,57 @@
...
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Definition
+----------
+
+.. phpmethod:: MongoDB\\Model\\IndexInfo::isText()
+
+   Return whether or not the index is a :manual:`text </core/index-text>` index.
```
jmikola: ~~or not~~

docs/reference/method/MongoDBModelIndexInfo-isText.txt
```diff
@@ -0,0 +1,57 @@
...
+-------------
+
+A boolean indicating whether the index is a text index.
+
+Examples
+--------
+
+.. code-block:: php
+
+   <?php
```
jmikola: Newline after.

docs/reference/method/MongoDBModelIndexInfo-isText.txt
```diff
@@ -0,0 +1,57 @@
...
+   $collection = (new MongoDB\Client)->selectCollection('test', 'restaurants');
+
+   $collection->createIndex(['name' => 'text']);
+
+   foreach ($collection->listIndexes() as $index) {
+   if ($index->isText()) {
+      printf("%s has default language: %d\n", $index->getName(), $index['default_language']);
+      }
+   }
```
jmikola: 4-space indents.

tests/Model/IndexInfoFunctionalTest.php
```diff
@@ -0,0 +1,102 @@
...
+        $this->collection = new Collection($this->manager, $this->getDatabaseName(), $this->getCollectionName());
+    }
+
+    public function testIs2dSphere()
+    {
+        $this->collection->createIndex(['pos' => '2dsphere']);
+        $result = $this->collection->listIndexes();
+
+        $result->rewind();
+        $this->assertInstanceOf('MongoDB\Model\IndexInfo', $result->current());
```
jmikola: This is asserting the first IndexInfo for the `_id` index, which we then ignore when advancing. Perhaps we can just remove this, as `testListIndexesForNewlyCreatedCollection()` in ListIndexesFunctionalTest already tests IndexInfoIterator in this regard. I'd propose we instead capture the name returned by `createIndex()` and then assert that it's the same as `$index->getName()` below to ensure we're dealing with the expected index record.

Alternatively, we can just remove this and go right to asserting `is2dSphere()` on the second element.

tests/Model/IndexInfoFunctionalTest.php
```diff
@@ -0,0 +1,102 @@
...
+        $result->next();
+        $index = $result->current();
+
+        $this->assertTrue($index->is2dSphere());
+
+        if (version_compare($this->getServerVersion(), '3.2.0', '<')) {
+            $this->assertEquals(2, $index['2dsphereIndexVersion']);
+            return;
```
jmikola: Consider:

```
$expectedVersion = version_compare($this->getServerVersion(), '3.2.0', '<') ? 2 : 3;
$this->assertEquals($expectedVersion, $index['2dsphereIndexVersion']);
```

This ensures that if more assertions are later appended to this test, we limit the chances of a bug missing these for < 3.2.

tests/Model/IndexInfoFunctionalTest.php
```diff
@@ -0,0 +1,102 @@
...
+
+        $this->assertTrue($index->isText());
+        $this->assertEquals('english', $index['default_language']);
+        $this->assertEquals('language', $index['language_override']);
+
+        if (version_compare($this->getServerVersion(), '3.2.0', '<')) {
+            $this->assertEquals(2, $index['textIndexVersion']);
+            return;
+        }
+        $this->assertEquals(3, $index['textIndexVersion']);
```
jmikola: Same idea as 2dsphere.

tests/Model/IndexInfoFunctionalTest.php
```diff
@@ -0,0 +1,102 @@
...
+
+    public function testIsText()
+    {
+        $this->collection->createIndex(['x' => 'text']);
+        $result = $this->collection->listIndexes();
+
+        $result->rewind();
+        $this->assertInstanceOf('MongoDB\Model\IndexInfo', $result->current());
+        $result->next();
+        $index = $result->current();
```
jmikola: Same as 2dsphere.

tests/Model/IndexInfoFunctionalTest.php
```diff
@@ -0,0 +1,102 @@
...
+
+    public function testIsGeoHaystack()
+    {
+        $this->collection->createIndex(['pos' => 'geoHaystack', 'x' => 1], ['bucketSize' => 5]);
+        $result = $this->collection->listIndexes();
+
+        $result->rewind();
+        $this->assertInstanceOf('MongoDB\Model\IndexInfo', $result->current());
+        $result->next();
+        $index = $result->current();
```
jmikola: Same as 2dsphere.

tests/Model/IndexInfoTest.php
```diff
@@ -139,4 +139,19 @@ public function testOffsetUnsetCannotBeCalled()
...
+            'key' => ['x' => 1],
+            'name' => 'x_1',
+            'ns' => 'foo.bar',
+        ]);
+
+        $this->assertFalse($info->is2dSphere());
+        $this->assertFalse($info->isGeoHaystack());
+        $this->assertFalse($info->isText());
+        $this->assertFalse($info->isBackground());
+    }
```
jmikola: Using the example specs I shared in https://github.com/mongodb/mongo-php-library/pull/521#discussion_r180154102, can you create tests based on the `testTtlIndex()` (and friends) for truthy return values? I think you can then also just add `$this->assertFalse()` for the new "is" methods to existing tests (see: `$this->assertFalse($info->isSparse());` in `testTtlIndex()`.

Since you already created functional tests, I think it's worth keeping those since we didn't have functional tests for creating all of these index types previously.

tests/Model/IndexInfoFunctionalTest.php
```diff
@@ -0,0 +1,70 @@
+<?php
+
+namespace MongoDB\Tests\Model;
+
+use MongoDB\Collection;
+use MongoDB\Operation\CreateIndexes;
+use MongoDB\Operation\ListIndexes;
```
jmikola: These operations aren't used, so we can remove these `use` statements.

tests/Model/IndexInfoFunctionalTest.php
```diff
@@ -0,0 +1,70 @@
...
+use MongoDB\Tests\Operation\FunctionalTestCase;
```
jmikola: I'm not keen on using a base test class from another sibling namespace. That said, I changed this locally to extend `MongoDB\Tests\FunctionalTestCase` instead and it passed without issue.

tests/Model/IndexInfoFunctionalTest.php
```diff
@@ -0,0 +1,70 @@
...
+
+class IndexInfoFunctionalTest extends FunctionalTestCase
+{
+    private $collection;
+
+    public function setUp()
+    {
+        parent::setUp();
+
+        $this->collection = new Collection($this->manager, $this->getDatabaseName(), $this->getCollectionName());
```
jmikola: Since we're creating a collection in each test, let's drop it upon `setup()` and tearDown()`. Consider `MongoDB\Tests\Collection\FunctionalTestCase` for prior art. In particular, we can leave the collection as-is in the event of a failure.

I see no reason we can't use the Collection methods, so `$this->collection->drop();` is fine and we can skip the DropCollection operation class.

PR #522, opened 2018-04-04
PHPLIB-278: Add examples for calling methods on database, collection, and index models

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-278

### DIFF
--- a/docs/reference/method/MongoDBModelIndexInfo-isTtl.txt
+++ b/docs/reference/method/MongoDBModelIndexInfo-isTtl.txt
@@ -28,6 +28,23 @@ Return Values
 
 A boolean indicating whether the index is a TTL index.
 
+Examples
+--------
+
+.. code-block:: php
+
+   <?php
+
+   $info = new IndexInfo([
+       'expireAfterSeconds' => 100,
+   ]);
+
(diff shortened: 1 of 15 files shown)


docs/reference/method/MongoDBModelIndexInfo-isTtl.txt
```diff
@@ -28,6 +28,23 @@ Return Values
...
+
+.. code-block:: php
+
+   <?php
+
+   $info = new IndexInfo([
+       'expireAfterSeconds' => 100,
+   ]);
+
+   var_dump($info->isTtl());
```
derickr: Do we really use ``isTtl`` and not ``isTTL``?

kvwalker: [Yes](https://github.com/mongodb/mongo-php-library/blob/master/src/Model/IndexInfo.php#L130), we do!

kvwalker:
This doesn't depend on #521 because this is just in response to a user asking for more examples for `getName()`. I decided to go ahead and add examples for the other methods in the models classes. I can add examples to the documentation as part of #521.

PR #524, opened 2018-04-09
PHPLIB-190: Support object cloning for BSONArray and BSONDocument

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-190


src/Model/BSONDocument.php
```diff
@@ -32,6 +32,18 @@
  */
 class BSONDocument extends ArrayObject implements JsonSerializable, Serializable, Unserializable
 {
+    /**
+     * Deep clone this BSONDocument.
+     */
+    public function __clone()
+    {
+        foreach ($this as $key => $value) {
```
jmikola: This approach works if all arrays are BSONArray, but we previously talked about needing to copy elements within a basic array as well. That is possible if the user is only returning root and embedded documents as BSONDocument and requesting BSON arrays as basic PHP arrays.

Additionally, I believe this approach can segfault:

```php

class Foo {
    function __clone() {
        foreach ($this as $key => $value) {
            if (is_object($this->$key)) {
                $this->$key = clone $value;
            }
        }
    }
}

$a = new Foo;
$a->a = $a;
$b = clone $a;
```

It may be worth looking into how [DeepCopy](https://github.com/myclabs/DeepCopy) handles this by maintaining a hash of clones created during the top-most clone operation. This allows it to avoid a recursive rabbit hole. The readme explains some of the challenges with naive cloning approaches.

Looping in @derickr for advise here. Maybe we don't care about segfaulting on a recursive loop (PHP does that by design). I don't believe it's possible with a document originating directly from the server. As in my example, it requires purposefully creating a cycle in the object hierarchy.

derickr: As @jmikola says, this can't happen with documents coming form the server, so I'm fine if we don't mitigate this issue.

tests/Model/BSONArrayTest.php
```diff
@@ -4,9 +4,9 @@
 
 use MongoDB\Model\BSONArray;
 use MongoDB\Model\BSONDocument;
-use MongoDB\Tests\TestCase;
+use MongoDB\Tests\FunctionalTestCase;
```
kvwalker: I changed this from `TestCase` to `FunctionalTestCase` to access `assertSameDocument()`. Alternatively, I could add `assertSameDocument()` and `normalizeBSON()` to  `TestCase`.

tests/Model/BSONArrayTest.php
```diff
@@ -17,6 +17,24 @@ public function testBsonSerializeReindexesKeys()
         $this->assertSame(['foo', 'bar'], $array->bsonSerialize());
     }
 
+    public function testClone()
+    {
+        $arr1 = new BSONArray(['x' => 1]);
+        $arr2 = new BSONArray(['a', 'b', 'c']);
+        $arr3 = new BSONArray(['foo' => 'bar']);
```
jmikola: `$arr1` and `$arr3` are technically documents, since you're using associative arrays. Perhaps we can use the following as our object:

```
new BSONArray([
    // array of multiple types (scalar, object, array)
    [
        'foo',
        new stdClass,
        ['bar', new stdClass],
    ],
    // BSONArray of multiple types (scalar, object, array)
    new BSONArray([
        'foo',
        new stdClass,
        ['bar', new stdClass],
    ]),
]);
```

The second element should cover nested BSONArray objects.

While `assertSameDocument()` should work just fine on a BSONArray and its clone, you can optionally wrap it in an associative array first (e.g. `['array' => $bsonArrayBeingTested]`).

After the cloning, you'd then want to assert objects are different in the following indices:

 * `$document['array'][0][1]`
 * `$document['array'][0][2][1]`
 * `$document['array'][1]`
 * `$document['array'][1][1]`
 * `$document['array'][1][2][1]`

tests/TestCase.php
```diff
@@ -5,12 +5,44 @@
...
+        $normalizeRootDocuments = function($document) {
+            return \MongoDB\BSON\toJSON(\MongoDB\BSON\fromPHP($this->normalizeBSON($document)));
+        };
+
+        $this->assertEquals(
+            array_map($normalizeRootDocuments, $expectedDocuments),
+            array_map($normalizeRootDocuments, $actualDocuments)
+        );
+    }
+
```
jmikola: Small nit, but these `protected` methods should go after the `public` method below.

tests/Model/BSONDocumentTest.php
```diff
@@ -4,10 +4,10 @@
 
 use MongoDB\Model\BSONArray;
 use MongoDB\Model\BSONDocument;
-use MongoDB\Tests\FunctionalTestCase;
+use MongoDB\Tests\TestCase;
 use ArrayObject;
 
-class BSONDocumentTest extends FunctionalTestCase
+class BSONDocumentTest extends TestCase
```
jmikola: The `testClone()` method from BSONArrayTest can be ported over to this class as well:

```php
$document = new BSONDocument([
    'a' => [
        'a' => 'foo',
        'b' => new stdClass,
        'c' => ['bar', new stdClass],
    ],
    'b' => new BSONDocument([
        'a' => 'foo',
        'b' => new stdClass,
        'c' => ['bar', new stdClass],
    ]),
]);
```

tests/TestCase.php
```diff
@@ -48,6 +43,44 @@ public function provideInvalidDocumentValues()
...
+    public function expectException($exception)
```
jmikola: I think this was moved by mistake.

src/functions.php
```diff
@@ -194,3 +194,17 @@ function is_string_array($input) {
     return true;
 }
 
+function recursive_copy($element) {
```
jmikola: Let's preface this with a doc block, and mark it `@internal` to advise users not to call it directly.

```
/**
 * Performs a deep copy of a value.
 *
 * This function will clone objects and recursively copy values within arrays.
 *
 * @internal
 * @see https://bugs.php.net/bug.php?id=49664
 * @param mixed $element Value to be copied
 * @return mixed
 */
```

jmikola: Related reading: http://phpsadness.com/sad/50

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -9,6 +9,7 @@
 use MongoDB\Driver\ReadPreference;
 use MongoDB\Driver\WriteConcern;
 use MongoDB\Exception\InvalidArgumentException;
+use MongoDB\Model\BSONDocument;
```
jmikola: Is this needed?

PR #526, opened 2018-04-10
PHPLIB-262: Implement iterator for BSON files

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-262

### DIFF
--- a/src/Model/BSONIterator.php
+++ b/src/Model/BSONIterator.php
@@ -1,0 +1,135 @@
+<?php
+/*
+ * Copyright 2018 MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace MongoDB\Model;
+
+use MongoDB\Exception\UnexpectedValueException;
+use MongoDB\Model\BSONDocument;
+use Iterator;
+
+/**
+ * Iterator for BSON documents.
+ */
+class BSONIterator implements Iterator
+{
+    private $buffer;
+    private $bufferLength;
+    private $current;
+    private $key = 0;
+    private $position = 0;
+    private $options;
+
+    const BSON_SIZE = 4;
+
+    /**
+     * Constructs a BSON Iterator.
+     *
+     * Supported options:
+     *
+     *  * typeMap (array): Type map for BSON deserialization.
+     *
+     * @internal
+     * @see http://php.net/manual/en/function.mongodb.bson-tophp.php
+     * @param string $data    Concatenated, valid, BSON-encoded documents
+     * @param array  $options Iterator options
+     * @throws InvalidArgumentException for parameter/option parsing errors
+     */
+    public function __construct($data, array $options = [])
+    {
+        if (isset($options['typeMap']) && ! is_array($options['typeMap'])) {
+            throw InvalidArgumentException::invalidType('"typeMap" option', $options['typeMap'], 'array');
+        }
+
+        if ( ! isset($options['typeMap'])) {
+            $options['typeMap'] = [];
+        }
+
+        $this->buffer = $data;
+        $this->bufferLength = strlen($data);
+        $this->options = $options;
+    }
+
+    /**
+     * @see http://php.net/iterator.current
+     * @return mixed
+     */
+    public function current()
+    {
+        return $this->current;
+    }
+
+    /**
+     * @see http://php.net/iterator.key
+     * @return mixed
+     */
+    public function key()
+    {
+        return $this->key;
+    }
+
+    /**
+     * @see http://php.net/iterator.next
+     * @return void
+     */
+    public function next()
+    {
+        $this->key++;
+        $this->current = null;
+        $this->advance();
+    }
+
+    /**
+     * @see http://php.net/iterator.rewind
+     * @return void
+     */
+    public function rewind()
+    {
+        $this->key = 0;
+        $this->position = 0;
+        $this->current = null;
+        $this->advance();
+    }
+
+    /**
+     * @see http://php.net/iterator.valid
+     * @return boolean
+     */
+    public function valid()
+    {
+        return $this->current !== null;
+    }
+
+    private function advance()
+    {
+        if ($this->position === $this->bufferLength) {
+            return;
+        }
+
+        if (($this->bufferLength - $this->position) < self::BSON_SIZE) {
+            throw new UnexpectedValueException(sprintf('Expected at least %d bytes; %d remaining', self::BSON_SIZE, $this->bufferLength - $this->position));
+        }
+
+        list(,$documentLength) = unpack('V', substr($this->buffer, $this->position, self::BSON_SIZE));
+
+        if (($this->bufferLength - $this->position) < $documentLength) {
+            throw new UnexpectedValueException(sprintf('Expected %d bytes; %d remaining', $documentLength, $this->bufferLength - $this->position));
+        }
+
+        $this->current = \MongoDB\BSON\toPHP(substr($this->buffer, $this->position, $documentLength), $this->options['typeMap']);
+        $this->position += $documentLength;
+    }
+}
--- a/tests/Model/BSONIteratorTest.php
+++ b/tests/Model/BSONIteratorTest.php
@@ -1,0 +1,133 @@
+<?php
+
+namespace MongoDB\Tests\Model;
+
+use MongoDB\Exception\UnexpectedValueException;
+use MongoDB\Model\BSONArray;
+use MongoDB\Model\BSONDocument;
+use MongoDB\Model\BSONIterator;
+use MongoDB\Tests\TestCase;
+use stdClass;
+
+class BSONIteratorTest extends TestCase
+{
+    /**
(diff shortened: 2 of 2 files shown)


kvwalker:
I ended up pretty much using @jmikola's [BsonIterator](https://github.com/jmikola/react-mongodb/blob/master/src/BsonIterator.php). Locally, I tested this using `mongodump` on the `phplib_test` collection, but I'm not sure of the best way to write a consistent test for this.

src/Model/BSONIterator.php
```diff
@@ -0,0 +1,117 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace MongoDB\Model;
+
+use MongoDB\BSON;
```
jmikola: This looks like a dead import, since `MongoDB\BSON` isn't used below. Was this done in anticipation of callingthe [BSON functions](http://php.net/manual/en/ref.bson.functions.php)?

PHP 5.6 does support [`use function`](https://wiki.php.net/rfc/use_function), although that won't work on 5.5, which we're still supporting. For now, I'd stick to calling `\MongoDB\BSON\toPHP()` and the like.

src/Model/BSONIterator.php
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    private function getNextBSON()
+    {
+        if ($this->position === $this->bufferLength) {
+            return;
+        }
+
+        if (($this->bufferLength - $this->position) < 5) {
+            throw new UnexpectedValueException(sprintf('Expected at least %d bytes; %d remaining', self::BSON_SIZE, $this->bufferLength - $this->position));
```
jmikola: I think this error message is actually incorrect. We're expecting at least 5 bytes (i.e. smallest possible document), which would include the 32-bit integer denoting the root document size and a trailing null byte.

PHP 5.5 won't allow us to define a constant relative to `self::BSON_SIZE` ([constant expressions](http://php.net/manual/en/language.oop5.constants.php) were introduced in 5.6), but I'd suggest a new `MIN_DOCUMENT_LENGTH` constant defined to 5, which we can reference in this error message and the conditional above (instead of using a [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming))).

src/Model/BSONIterator.php
```diff
@@ -0,0 +1,117 @@
...
+     * @param string $data Concatenated, valid, BSON-encoded documents
+     */
+    public function __construct($data)
+    {
+        $this->buffer = $data;
+        $this->bufferLength = strlen($data);
+    }
+
+    /**
+     */
```
jmikola:
```
* @see http://php.net/iterator.current
* @return mixed
```

src/Model/BSONIterator.php
```diff
@@ -0,0 +1,117 @@
...
+            throw new UnexpectedValueException(sprintf('Expected at least %d bytes; %d remaining', self::BSON_SIZE, $this->bufferLength - $this->position));
+        }
+
+        list(,$documentLength) = unpack('V', substr($this->buffer, $this->position, self::BSON_SIZE));
+
+        if (($this->bufferLength - $this->position) < $documentLength) {
+            throw new UnexpectedValueException(sprintf('Expected %d bytes; %d remaining', $documentLength, $this->bufferLength - $this->position));
+        }
+
+        $this->current = new BSONDocument((array)(substr($this->buffer, $this->position, self::BSON_SIZE + $documentLength)));
```
jmikola: This doesn't look correct, as raw bytes from BSON data are being cast to an array and fed to the BSONDocument constructor. Try to `var_dump()` the elements of this iterator to see what this produces.

You're going to need to use the [BSON functions](http://php.net/manual/en/ref.bson.functions.php) to convert these bytes into a PHP value to assign to `$this->current`. Additionally, since `MongoDB\BSON\toPHP()` supports a type map it'd be reasonable if the constructor accepted an optional type map. In the interest of consistency, it should probably be through a new `array $options = []` argument and parsed similar type maps for operation classes.

If there is a type map available, we can use it when calling `toPHP()`; otherwise, we can default to an empty array, which is equivalent to no type map for `toPHP()`. I'm not sure if `null` is accepted and ignored in the same fashion, but you could check the extension source to confirm.

kvwalker:
I added tests for the first two cases you mentioned, but I'm confused by the last one. Maybe we can go over it tomorrow morning and then I can add that test case.

tests/Model/BSONIteratorTest.php
```diff
@@ -0,0 +1,85 @@
...
+            /*
+                Exception for remaining buffer after 32-bit length field being too small, for
+                both the first and second positions in the sequence.
+            */
+        ];
+    }
+
+    public function testRemainingBufferLength()
+    {
+        $binaryString = '1234';
```
jmikola: For the exception tests, I'd suggest using the base document `fromPHP([])`, which should represent the smallest valid BSON document (in binary form). It contains a 32-bit integer with a value of `5`, followed by a null byte (per the `document` definition in the [spec](http://bsonspec.org/spec.html) where `e_list` is empty).



The first exception is when we don't have sufficient data in the buffer to read the 4-byte length header. For this, I'd like a test for both the very first document and a subsequent document:

 * `testCannotReadLengthFromFirstDocument()`: start with one document as our input, and trim it to the first three bytes (anything less than four will suffice).
 * `testCannotReadLengthFromSubsequentDocument()`: start with two documents as our input, but trim the second document to only three bytes

The second exception is when we _were_ able to read the 4-byte length header, but there is not sufficient data in the buffer to read the remainder.

 * `testCannotReadFirstDocument()`: start with one document as our input and trim to the first four bytes
 * `testCannotReadSecondDocument()`: start with two documents as our input, but trim the second document to only four bytes

In PHPC, I use `pack()` to create these BSON documents. You may decide that's preferable (see: [bson-toJSON_error-002.phpt](https://github.com/mongodb/mongo-php-driver/blob/master/tests/bson/bson-toJSON_error-002.phpt)).

tests/Model/BSONIteratorTest.php
```diff
@@ -0,0 +1,85 @@
...
+class BSONIteratorTest extends TestCase
+{
+    /**
+     * @dataProvider provideTypeMapOptionsAndExpectedDocuments
+     */
+    public function testValidValues(array $typeMap, $binaryString, array $expectedDocuments)
+    {
+        $bsonIt = new BSONIterator($binaryString, $typeMap);
+
+        $this->assertSameDocuments($expectedDocuments, $bsonIt);
```
jmikola: I just remembered that `assertSameDocuments()` converts documents to JSON before it compares, so it's not useful for testing that type maps have been applied. See `AggregateFunctionalTest::testTypeMapOption()` and `provideTypeMapOptionsAndExpectedDocuments()` for a good approach.

tests/Model/BSONIteratorTest.php
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    public function provideTypeMapOptionsAndExpectedDocuments()
+    {
+        return [
+            [
+                [],
```
jmikola: Let's make sure that one test passes `null` as the `typeMap` option, which will ensure code coverage of the constructor assigning `[]` in that case.

tests/Model/BSONIteratorTest.php
```diff
@@ -0,0 +1,133 @@
...
+    public function provideTypeMapOptionsAndExpectedDocuments()
+    {
+        return [
+            [
+                null,
+                implode(array_map(
+                    'MongoDB\BSON\fromPHP',
+                    [
+                        (object) ['_id' => 1, 'x' => (object) ['foo' => 'bar']],
+                        (object) ['_id' => 3, 'x' => (object) ['foo' => 'bar']],
```
jmikola: We can drop the object-casting from all of the `fromPHP()` inputs, unless you actually need an array like `[ 0 => 'foo' ]` to serialize as `{ "0": "foo" }` instead of `[ "foo" ]`.

tests/Model/BSONIteratorTest.php
```diff
@@ -0,0 +1,133 @@
...
+    /**
+     * @dataProvider provideTypeMapOptionsAndExpectedDocuments
+     */
+    public function testValidValues(array $typeMap = null, $binaryString, array $expectedDocuments)
+    {
+        $bsonIt = new BSONIterator($binaryString, ['typeMap' => $typeMap]);
+
+        $results = iterator_to_array($bsonIt);
+
+        $this->assertSameDocuments($expectedDocuments, $results);
```
jmikola: This should use `assertEquals()` to ensure we're checking expected types after type map conversion (as is done in AggregateFunctionalTest).

tests/Model/BSONIteratorTest.php
```diff
@@ -0,0 +1,133 @@
...
+                    ['_id' => 3, 'x' => ['foo' => 'bar']],
+                ]
+            ],
+            [
+                ['root' => 'object', 'document' => 'array'],
+                implode(array_map(
+                    'MongoDB\BSON\fromPHP',
+                    [
+                        (object) ['_id' => 1, 'x' => ['foo' => 'bar']],
+                        (object) ['_id' => 3, 'x' => ['foo' => 'bar']],
```
jmikola: Drop `(object)`.

tests/Model/BSONIteratorTest.php
```diff
@@ -0,0 +1,133 @@
...
+                    (object) ['_id' => 3, 'x' => ['foo' => 'bar']],
+                ]
+            ],
+            [
+                ['root' => 'array', 'document' => 'stdClass'],
+                implode(array_map(
+                    'MongoDB\BSON\fromPHP',
+                    [
+                        ['_id' => 1, 'x' => (object) ['foo' => 'bar']],
+                        ['_id' => 3, 'x' => (object) ['foo' => 'bar']],
```
jmikola: Drop `(object)`.

PR #528, opened 2018-04-16
PHPLIB-342: Change streams should use the same session when resuming

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-342

I marked this [WIP] because I'll update `WatchFunctionalTest` to verify that the `lsid`s of all aggregates and getMores are the same after [PHPC-1152](https://jira.mongodb.org/browse/PHPC-1152) is merged.

### DIFF
--- a/src/ChangeStream.php
+++ b/src/ChangeStream.php
@@ -101,6 +101,15 @@ class ChangeStream implements Iterator
                 $this->hasAdvanced = true;
                 $this->resumeToken = $this->extractResumeToken($this->csIt->current());
             }
+            /* If the cursorId is 0, the server has invalidated the cursor so we
+             * will never perform another getMore. This means that we cannot
+             * resume and we can therefore unset the resumeCallable, which will
+             * free any reference to Watch. This will also free the only
+             * reference to an implicit session, since any such reference
+             * belongs to Watch. */
+            if ((string)$this->getCursorId() === '0') {
+                $this->resumeCallable = null;
+            }
         } catch (RuntimeException $e) {
             if (strpos($e->getMessage(), "not master") !== false) {
                 $resumable = true;
@@ -129,6 +138,10 @@ class ChangeStream implements Iterator
             if ($this->valid()) {
                 $this->hasAdvanced = true;
                 $this->resumeToken = $this->extractResumeToken($this->csIt->current());
+            }
+            // As with next(), free the callable once we know it will never be used.
+            if ((string)$this->getCursorId() === '0') {
+                $this->resumeCallable = null;
             }
         } catch (RuntimeException $e) {
             if (strpos($e->getMessage(), "not master") !== false) {
--- a/tests/Operation/WatchFunctionalTest.php
+++ b/tests/Operation/WatchFunctionalTest.php
@@ -8,7 +8,9 @@ use MongoDB\Driver\ReadPreference;
 use MongoDB\Driver\Server;
 use MongoDB\Driver\Exception\ConnectionTimeoutException;
 use MongoDB\Exception\ResumeTokenException;
+use MongoDB\Operation\CreateCollection;
 use MongoDB\Operation\DatabaseCommand;
+use MongoDB\Operation\DropCollection;
 use MongoDB\Operation\InsertOne;
 use MongoDB\Operation\Watch;
 use MongoDB\Tests\CommandObserver;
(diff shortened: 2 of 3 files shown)


src/ChangeStream.php
```diff
@@ -101,6 +101,15 @@ public function next()
                 $this->hasAdvanced = true;
                 $this->resumeToken = $this->extractResumeToken($this->csIt->current());
             }
+            // If the cursorId is 0, the server has invalidated the cursor so we
+            // will never perform another getMore. This means that we cannot
+            // resume and we can therefore unset the resumeCallable, which will
+            // free any reference to Watch. This will also free the only
+            // reference to an implicit session, since any such reference
+            // belongs to Watch.
```
jmikola:
```
/* Feel free to
 * use multi-line
 * comments. */
```

src/ChangeStream.php
```diff
@@ -130,6 +139,15 @@ public function rewind()
...
```
jmikola: I propose pointing to the earlier extended comment for `next()` rather than reproducing the full explanation:

```
// As with next(), free the callable once we know it will never be used
```

tests/Operation/WatchFunctionalTest.php
```diff
@@ -588,6 +588,46 @@ public function testResumeTokenNotFoundAdvancesKey()
...
+            },
+            function($changeStream) use (&$originalSession) {
+                if (isset($changeStream->aggregate)) {
+                    $originalSession = $changeStream->lsid;
+                }
+            }
+        );
+
+        $changeStream->rewind();
+        $this->killChangeStreamCursor($changeStream);
```
jmikola: Please test this against a replica set with auth if possible. That should highlight whether `killChangeStreamCursor()` will need to also take the same implicit session created within the Watch class.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -588,6 +588,46 @@ public function testResumeTokenNotFoundAdvancesKey()
...
+        // lsid of any aggregates after the change stream resumes. After
+        // PHPC-1152 is complete, we will ensure that the lsid of the initial
+        // aggregate matches the lsid of any subsequent aggregates and getMores.
+        (new CommandObserver)->observe(
+            function() use ($operation, &$changeStream) {
+                $changeStream = $operation->execute($this->getPrimaryServer());
+            },
+            function($changeStream) use (&$originalSession) {
+                if (isset($changeStream->aggregate)) {
+                    $originalSession = $changeStream->lsid;
```
jmikola: AFAIK `lsid` may include other fields, so you may want to only capture `lsid->id` (and/or possibly cast it to a hex string as I'm doing in the PHPC tests:

```
bin2hex((string) $command->lsid->id);
```

This makes output easier to digest if comparisons fail, since you're no longer dealing with raw binary strings and `MongoDB\BSON\Binary` objects.

tests/Operation/WatchFunctionalTest.php
kvwalker: After testing with auth, it looks like `killChangeStreamCursor()` does not need to take the same implicit session created within the Watch class.

jmikola: Excellent. That answers my question in https://github.com/mongodb/mongo-php-driver/pull/803#discussion_r182100188.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -588,6 +588,74 @@ public function testResumeTokenNotFoundAdvancesKey()
...
+             * driver to kill it (via mongoc_cursor_destroy()). */
+            'killCursors',
+            /* Finally, ChangeStream will rewind the new cursor as the last step
+             * of the resume process. This results in one last getMore. */
+            'getMore',
+        ];
+
+        $this->assertSame($expectedCommands, $commands);
+
+        foreach ($sessionAfterResume as $session) {
```
jmikola: Very nice :)

src/ChangeStream.php
```diff
@@ -101,6 +101,15 @@ public function next()
                 $this->hasAdvanced = true;
                 $this->resumeToken = $this->extractResumeToken($this->csIt->current());
             }
+            /* If the cursorId is 0, the server has invalidated the cursor so we
+             * will never perform another getMore. This means that we cannot
+             * resume and we can therefore unset the resumeCallable, which will
+             * free any reference to Watch. This will also free the only
+             * reference to an implicit session, since any such reference
+             * belongs to Watch. */
+            if ($this->getCursorId() === 0) {
```
jmikola: `getCursorId()` returns a `MongoDB\Driver\CursorId`, so this will never evaluate to true. You may need to cast the CursorId to a string and compare with `'0'`.

jmikola:
Let's rebase on v1.3 tomorrow before you merge this.

PR #531, opened 2018-04-20
PHPLIB-339: Deprecate snapshot option

kvwalker:
https://jira.mongodb.org/browse/PHPLIB-339

### DIFF
--- a/tests/Operation/FindTest.php
+++ b/tests/Operation/FindTest.php
@@ -128,6 +128,17 @@ class FindTest extends TestCase
         return $options;
     }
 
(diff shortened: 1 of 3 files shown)


tests/Operation/FindTest.php
```diff
@@ -128,6 +128,17 @@ public function provideInvalidConstructorOptions()
         return $options;
     }
 
+    public function testAutoIndexIdOptionIsDeprecated()
```
jmikola: `testSnapshotOptionIsDeprecated()`

PR #533, opened 2018-04-23
PHPLIB-346: Deprecate autoIndexId option for CreateCollection

jmikola:
https://jira.mongodb.org/browse/PHPLIB-346

### DIFF
--- a/src/Operation/CreateCollection.php
+++ b/src/Operation/CreateCollection.php
@@ -53,6 +53,10 @@ class CreateCollection implements Executable
      *    of an index on the _id field. For replica sets, this option cannot be
      *    false. The default is true.
      *
+     *    This option has been deprecated since MongoDB 3.2. As of MongoDB 4.0,
+     *    this option cannot be false when creating a replicated collection
+     *    (i.e. a collection outside of the local database in any mongod mode).
+     *
      *  * capped (boolean): Specify true to create a capped collection. If set,
      *    the size option must also be specified. The default is false.
      *
@@ -168,6 +172,10 @@ class CreateCollection implements Executable
 
         if (isset($options['writeConcern']) && $options['writeConcern']->isDefault()) {
             unset($options['writeConcern']);
+        }
+
+        if (isset($options['autoIndexId'])) {
+            trigger_error('The "autoIndexId" option is deprecated and will be removed in a future release', E_USER_DEPRECATED);
         }
 
         $this->databaseName = (string) $databaseName;
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -52,6 +52,23 @@ abstract class TestCase extends BaseTestCase
         return $this->wrapValuesForDataProvider($this->getInvalidDocumentValues());
     }
 
+    protected function assertDeprecated(callable $execution)
+    {
+        $errors = [];
+
+        set_error_handler(function($errno, $errstr) use (&$errors) {
+            $errors[] = $errstr;
+        }, E_USER_DEPRECATED);
+
+        try {
+            call_user_func($execution);
+        } finally {
+            restore_error_handler();
+        }
+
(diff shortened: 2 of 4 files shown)


tests/TestCase.php
```diff
@@ -52,6 +52,23 @@ public function provideInvalidDocumentValues()
         return $this->wrapValuesForDataProvider($this->getInvalidDocumentValues());
     }
 
+    protected function assertDeprecated(callable $execution)
```
jmikola: If this looks OK, I think we can use it for #531 and #532.

src/Operation/CreateCollection.php
```diff
@@ -170,6 +174,10 @@ public function __construct($databaseName, $collectionName, array $options = [])
             unset($options['writeConcern']);
         }
 
+        if (isset($options['autoIndexId'])) {
```
kvwalker: Should this also include a version check? As is, won't this trigger an error even if a user is using version 2.6 or 3.0?

jmikola: I considered that, and/or making an exception for the `local` database, which MongoDB 4.0 should [still allow](https://jira.mongodb.org/browse/SERVER-8237?focusedCommentId=1808020&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-1808020).

I'll ask this question in [DRIVERS-473](https://jira.mongodb.org/browse/DRIVERS-473).

Making an exception for the `local` database would be easy, although restrictions based on server version would require us to move this notice to `execute()`, which is a bit lousy.

jmikola: > As is, won't this trigger an error even if a user is using version 2.6 or 3.0?

Thinking about this some more, I think we're right to raise this for all server versions. The documentation for this option has always stated:

> For replica sets, do not set ``autoIndexId`` to ``false``.

The logic behind the server not permitting this at all (outside of `local`) regardless of mongod type is that a standalone could later be promoted to a replica set. This is irrespective of server version, so I don't see the harm in annoying users with a notice in this case.

----

On a separate note, it may not even make sense to allow an exemption for `local`. The point of deprecating this is that we'd ultimately like to remove the option altogether in our next major version. Keeping it around for `local` would complicate that goal.

If we deprecate it for later removal, users that really want to use this option down the line can likely just run the `create` command manually.

kvwalker: I think it makes sense that we don't allow an exemption for `local` and raise the error for all server versions.

src/Operation/CreateCollection.php
```diff
@@ -170,6 +174,10 @@ public function __construct($databaseName, $collectionName, array $options = [])
...
+            trigger_error('The "autoIndexId" option is deprecated and will be removed in a future release', E_USER_DEPRECATED);
```
alcaeus: Do you want to show deprecations by default? For reference, Symfony silences these `trigger_error` calls to make deprecation warnings opt-in instead of opt-out: without silencing, the default PHP error handler will start printing notices about deprecations, even if the user has no intention of fixing them or listening to them. Using the silencing operator would effectively require the user to use some kind of deprecation listener (such as the one provided by Symfony) to "listen" to these deprecation warnings.

alcaeus: > I assume it stops PHP's default handler from emitting noise but doesn't actually prevent a custom error handler from being notified?

Your assumption is correct. Symfony changed this behavior shortly after introducing deprecation warnings in Symfony 2.7 after complaints about large number of deprecation notices being found in logs. The discussion around the change can be found in https://github.com/symfony/symfony/pull/14900.

jmikola: Created [PHPLIB-347](https://jira.mongodb.org/browse/PHPLIB-347) to track this further. Thanks for the suggestion!

PR #541, opened 2018-06-21
 PHPLIB-319: Added 'disableMD5' option as GridFS MD5 digest must be optional

derickr:


src/GridFS/Bucket.php
```diff
@@ -120,6 +128,7 @@ public function __construct(Manager $manager, $databaseName, array $options = []
         $this->databaseName = (string) $databaseName;
         $this->bucketName = $options['bucketName'];
         $this->chunkSizeBytes = $options['chunkSizeBytes'];
+        $this->disableMD5 = isset($options['disableMD5']) ? $options['disableMD5'] : false;
```
jmikola: For consistency, you can add `'disableMD5' => false,` when `$options` is appended at the start of the constructor. This can then become `$this->disableMD5 = $options['disableMD5'];`. Prior art: https://github.com/mongodb/mongo-php-library/blob/baf85f46d578acf25ef11d65eb28ff5bd4b75503/src/Operation/FindAndModify.php#L114-L118

src/GridFS/Bucket.php
```diff
@@ -470,6 +479,9 @@ public function openDownloadStreamByName($filename, array $options = [])
      *  * chunkSizeBytes (integer): The chunk size in bytes. Defaults to the
      *    bucket's chunk size.
      *
+     *  * disableMD5 (boolean): When true, no MD5 sum will be generated for
+     *    the stored file. Defaults to "false".
```
jmikola: These two lines should like be added to the doc block for `uploadFromStream()` further down in the file.

src/GridFS/WritableStream.php
```diff
@@ -96,7 +104,11 @@ public function __construct(CollectionWrapper $collectionWrapper, $filename, arr
 
         $this->chunkSize = $options['chunkSizeBytes'];
         $this->collectionWrapper = $collectionWrapper;
-        $this->ctx = hash_init('md5');
+        $this->disableMD5 = isset($options['disableMD5']) ? $options['disableMD5'] : false;
```
jmikola: As with Bucket's constructor, `'disableMD5' => false,` when initially appending `$options` and then this can be a basic assignment.

src/GridFS/WritableStream.php
```diff
@@ -167,7 +179,7 @@ public function getSize()
      * written. Since seeking is not supported and writes are appended, this is
      * always the end of the stream.
      *
-     * @see WriteableStream::getSize()
+     * @see WritableStream::getSize()
```
jmikola: Thanks!

src/GridFS/WritableStream.php
```diff
@@ -219,12 +231,14 @@ private function abort()
...
     {
-        $md5 = hash_final($this->ctx);
-
         $this->file['length'] = $this->length;
-        $this->file['md5'] = $md5;
         $this->file['uploadDate'] = new UTCDateTime;
 
+        if ( ! $this->disableMD5) {
+            $md5 = hash_final($this->hashCtx);
+            $this->file['md5'] = $md5;
```
jmikola: This can probably be consolidated into one line. I likely only broke it up previously to finalize the hash context ASAP and then delayed `$this->file['md5']` assignment to keep it alphabetized with other fields.

src/GridFS/Bucket.php
```diff
@@ -100,6 +104,10 @@ public function __construct(Manager $manager, $databaseName, array $options = []
             throw new InvalidArgumentException(sprintf('Expected "chunkSizeBytes" option to be >= 1, %d given', $options['chunkSizeBytes']));
         }
 
+        if (isset($options['disableMD5']) && ! is_bool($options['disableMD5'])) {
+            throw InvalidArgumentException::invalidType('"disableMD5" option', $options['disableMD5'], 'boolean');
+        }
```
jmikola: Can we add some tests for this? Should just require adding the following to `BucketFunctionalTest::provideInvalidConstructorOptions()`:

```
foreach ($this->getInvalidBooleanValues() as $value) {
    $options[][] = ['disableMD5' => $value];
}
```

docs/includes/apiargs-MongoDBGridFSBucket-method-construct-option.yaml
```diff
@@ -17,6 +17,17 @@ interface: phpmethod
 operation: ~
 optional: true
 ---
+arg_name: option
+name: disableMD5
```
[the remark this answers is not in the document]
derickr: I've changed that, but why wasn't a similar thing done for chuckSizeBytes as well?

docs/includes/apiargs-MongoDBDatabase-method-selectGridFSBucket-option.yaml
```diff
@@ -17,6 +17,17 @@ interface: phpmethod
 operation: ~
 optional: true
 ---
+arg_name: option
+name: disableMD5
```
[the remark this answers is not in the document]
derickr: I just copied what was done for chuckSizeBytes again ;-) Updated.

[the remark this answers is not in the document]
jmikola: Ah, does `chunkSizeBytes` bytes duplicate text when it could otherwise share? If so, I'll look into that later.

PR #542, opened 2018-06-21
 PHPLIB-354: Implement new count API 

derickr:


src/Operation/CountDocuments.php
```diff
@@ -0,0 +1,240 @@
...
+
+        $cursor = $server->executeReadCommand($this->databaseName, new Command($this->createCommandDocument()), $this->createOptions());
+        $result = current($cursor->toArray());
+
+        // Older server versions may return a float
+        if ( ! isset($result->n) || ! (is_integer($result->n) || is_float($result->n))) {
+            throw new UnexpectedValueException('count command did not return a numeric "n" value');
+        }
+
+        return (integer) $result->n;
```
jmikola: It's an edge case, but I'm not sure if this will play nicely with the `MongoDB\BSON\Int64` wrapper ([PHPC-1211](https://jira.mongodb.org/browse/PHPC-1211)). Would it be better to just return the value as-is in that case?

I realize we document an `integer` return value, but we may want to change that to `integer|MongoDB\BSON\Int64`. I don't think this needs to be addressed in this PR specifically. I've opened [PHPLIB-356](https://jira.mongodb.org/browse/PHPLIB-356) to track this.

src/Operation/CountDocuments.php
```diff
@@ -0,0 +1,240 @@
...
+     *  * skip (integer): The number of documents to skip before returning the
+     *    documents.
+     *
+     * @param string       $databaseName   Database name
+     * @param string       $collectionName Collection name
+     * @param array|object $filter         Query by which to filter documents
+     * @param array        $options        Command options
+     * @throws InvalidArgumentException for parameter/option parsing errors
+     */
+    public function __construct($databaseName, $collectionName, $filter = [], array $options = [])
```
jmikola: In the interest of keeping these Operation's strict, let's not default `$filter` here (consistent with [Find](https://github.com/mongodb/mongo-php-library/blob/master/src/Operation/Find.php)). The convenience stuff can go in the Collection's helper method.

src/Operation/CountDocuments.php
```diff
@@ -0,0 +1,240 @@
...
+     * Create the count command document.
+     *
+     * @return array
+     */
+    private function createCommandDocument()
+    {
+        $pipeline = [];
+
+        if ( ! empty($this->filter)) {
+            $pipeline[] = ['$match' => (object) $this->filter];
```
jmikola: I don't think you need a conditional here. `$this->filter` can always be cast to an object. If it happens to be an empty array already, we end up with an empty `stdClass`.

src/Operation/CountDocuments.php
```diff
@@ -0,0 +1,240 @@
...
+use MongoDB\Exception\UnsupportedException;
+
+/**
+ * Operation for the count command.
+ *
+ * @api
+ * @see \MongoDB\Collection::count()
+ * @see http://docs.mongodb.org/manual/reference/command/count/
+ */
+class CountDocuments implements Executable, Explainable
```
jmikola: This should not be Explainable. The `explain` command doesn't handle aggregation pipelines, thanks to `aggregate` having its own unique API for explanation.

src/Operation/CountDocuments.php
```diff
@@ -0,0 +1,240 @@
...
+
+        // Older server versions may return a float
+        if ( ! isset($result->n) || ! (is_integer($result->n) || is_float($result->n))) {
+            throw new UnexpectedValueException('count command did not return a numeric "n" value');
+        }
+
+        return (integer) $result->n;
+    }
+
+    public function getCommandDocument(Server $server)
```
jmikola: This can be removed when you drop Explainable.

src/Operation/EstimatedDocumentCount.php
```diff
@@ -0,0 +1,173 @@
...
+     *  * session (MongoDB\Driver\Session): Client session.
+     *
+     *    Sessions are not supported for server versions < 3.6.
+     *
+     * @param string       $databaseName   Database name
+     * @param string       $collectionName Collection name
+     * @param array        $options        Command options
+     * @throws InvalidArgumentException for parameter/option parsing errors
+     */
+    public function __construct($databaseName, $collectionName, array $options = [])
```
jmikola: Similar to `CountDocuments` encapsulating an `Aggregate`, I think this class an encapsulate a `Count` operation. We'd then leverage it's existing implementation of Executable and Explainable.

Feel free to lump this in with the ticket for refactoring `CountDocuments`.

src/Operation/EstimatedDocumentCount.php
```diff
@@ -0,0 +1,173 @@
...
+
+        $cursor = $server->executeReadCommand($this->databaseName, new Command($this->createCommandDocument()), $this->createOptions());
+        $result = current($cursor->toArray());
+
+        // Older server versions may return a float
+        if ( ! isset($result->n) || ! (is_integer($result->n) || is_float($result->n))) {
+            throw new UnexpectedValueException('count command did not return a numeric "n" value');
+        }
+
+        return (integer) $result->n;
```
jmikola: Another candidate for [PHPLIB-356](https://jira.mongodb.org/browse/PHPLIB-356), along with the original `Count` operation of course.

tests/Collection/CrudSpecFunctionalTest.php
```diff
@@ -123,12 +123,18 @@ private function executeOperation(array $operation)
...
             case 'count':
+            case 'countDocuments':
             case 'find':
                 return $this->collection->{$operation['name']}(
                     isset($operation['arguments']['filter']) ? $operation['arguments']['filter'] : [],
                     array_diff_key($operation['arguments'], ['filter' => 1])
                 );
 
+            case 'estimatedDocumentCount':
+                return $this->collection->estimatedDocumentCount(
```
jmikola: This can probably fit on one line.

tests/Collection/CrudSpecFunctionalTest.php
```diff
@@ -352,7 +360,7 @@ private function executeAssertResult(array $operation, $expectedResult, $actualR
                 break;
 
             default:
-                throw new LogicException('Unsupported operation: ' . $operationName);
+                throw new LogicException('Unsupported operation: ' . $operation['name']);
```
jmikola: Was `$operationName` undefined? If so, good catch!

src/Collection.php
```diff
@@ -273,6 +277,35 @@ public function count($filter = [], array $options = [])
...
+     * @see Count::__construct() for supported options
+     * @param array|object $filter  Query by which to filter documents
+     * @param array        $options Command options
+     * @return integer
+     * @throws UnexpectedValueException if the command response was malformed
+     * @throws UnsupportedException if options are not supported by the selected server
+     * @throws InvalidArgumentException for parameter/option parsing errors
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function countDocuments($filter = [], array $options = [])
```
jmikola: As noted elsewhere, this default value on `$filter` is sufficient to provide a convenience to users. The Operation class can omit the default value.

docs/reference/class/MongoDBCollection.txt
```diff
@@ -71,6 +72,7 @@ Methods
    /reference/method/MongoDBCollection-drop
    /reference/method/MongoDBCollection-dropIndex
    /reference/method/MongoDBCollection-dropIndexes
+   /reference/method/MongoDBCollection-estimatedDocumentCount
```
jmikola: Building the docs produced the following warning:

> source/reference/class/MongoDBCollection.txt:59: WARNING: toctree contains reference to nonexisting document u'reference/method/MongoDBCollection-estimatedDocumentCount'

Looks like a typo in the filename for `MongoDBCollection-estimateDocumentCount.txt` (missing a "d" before "Document").

docs/reference/method/MongoDBCollection-estimateDocumentCount.txt
```diff
@@ -0,0 +1,56 @@
+============================================
+MongoDB\\Collection::estimateDocumentCount()
```
jmikola: "estimatedDocumentCount"

docs/reference/method/MongoDBCollection-estimateDocumentCount.txt
```diff
@@ -0,0 +1,56 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Definition
+----------
+
+.. phpmethod:: MongoDB\\Collection::estimateDocumentCount()
```
jmikola: "estimateDocumentCount"

PR #544, opened 2018-06-22
[WIP] PHPLIB-351: Support cluster and database-level change streams

jmikola:
https://jira.mongodb.org/browse/PHPLIB-351
https://jira.mongodb.org/browse/PHPLIB-358


src/ChangeStream.php
```diff
@@ -201,6 +191,30 @@ private function extractResumeToken($document)
...
     }
 
+    /**
+     * Determines if an exception is a resumable error.
+     *
+     * @see https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst#resumable-error
+     * @param RuntimeException $exception
+     * @return boolean
+     */
+    private function isResumableError(RuntimeException $exception)
```
jmikola: The criteria for a resumable error changed since the spec was first implemented in 1.3.0. See: [Resumable Error](https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst#resumable-error).

src/ChangeStream.php
```diff
@@ -201,6 +191,30 @@ private function extractResumeToken($document)
...
+        if ($exception instanceof ConnectionException) {
+            return true;
+        }
+
+        if ( ! $exception instanceof ServerException) {
+            return false;
+        }
+
+        if (in_array($exception->getCode(), [self::$errorCodeCappedPositionLost, self::$errorCodeCursorKilled, self::$errorCodeInterrupted])) {
+            return false;
```
jmikola: My read of the new resumable error criteria lead me to conclude that checking the error message string for "not master" or "node is recovering" was redundant. This logic takes the into account and I've opened https://github.com/mongodb/specifications/pull/365 to correct the spec.

src/Operation/Watch.php
```diff
@@ -79,22 +92,44 @@ class Watch implements Executable
      *  * resumeAfter (document): Specifies the logical starting point for the
      *    new change stream.
      *
+     *    Using this option in conjunction with "startAtOperationTime" will
+     *    result in a server error. The options are mutually exclusive.
```
jmikola: Per [`startatoperationtime`](https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst#startatoperationtime) in the spec, we're not supposed to raise a client-side error if both options are used.

src/Operation/Watch.php
```diff
@@ -104,79 +139,145 @@ public function __construct(Manager $manager, $databaseName, $collectionName, ar
...
+
+        // Null database name implies a cluster-wide change stream
+        if ($databaseName === null) {
+            $databaseName = 'admin';
+            $this->changeStreamOptions['allChangesForCluster'] = true;
         }
 
         $this->databaseName = (string) $databaseName;
-        $this->collectionName = (string) $collectionName;
+        $this->collectionName = isset($collectionName) ? (string) $collectionName : null;
```
jmikola: `Aggregate` takes care of handling the `null` Collection name (per [PHPLIB-358](https://jira.mongodb.org/browse/PHPLIB-358)).

src/Operation/Watch.php
```diff
@@ -104,79 +139,145 @@ public function __construct(Manager $manager, $databaseName, $collectionName, ar
...
+    /** @internal */
+    final public function commandStarted(CommandStartedEvent $event)
+    {
+    }
+
+    /** @internal */
+    final public function commandSucceeded(CommandSucceededEvent $event)
+    {
+        if ($event->getCommandName() !== 'aggregate') {
+            return;
```
jmikola: I considered throwing a LogicException here, as this class should never be registered as a CommandSubscriber except for the brief duration that we execute an `aggregate` command; however, I don't think that's necessary. Open to changing this if you'd prefer the extra sanity check, though.

src/Operation/Watch.php
```diff
@@ -104,79 +139,145 @@ public function __construct(Manager $manager, $databaseName, $collectionName, ar
...
+        }
+
+        $reply = $event->getReply();
+
+        if (isset($reply->operationTime) && $reply->operationTime instanceof TimestampInterface) {
```
jmikola: By this point, we've already verified that the server's wire version is >= 7. If `operationTime` isn't a Timestamp, something is quite wrong. I just wanted to be defensive.

Failing silently also makes sense as this resume behavior is internal behavior (akin to us creating an "implicit" session in the constructor and not bothering the user if that fails).

tests/Operation/AggregateFunctionalTest.php
```diff
@@ -12,6 +12,28 @@
...
+        if (version_compare($this->getServerVersion(), '3.6.0', '<')) {
+            $this->markTestSkipped('$currentOp is not supported');
+        }
+
+        (new CommandObserver)->observe(
+            function() {
+                $operation = new Aggregate(
+                    'admin',
+                    null,
+                    [['$currentOp' => (object) []]]
```
jmikola: I'm not specifying any options nor am I asserting the response as this is only testing that executing such a command with a `null` collection name results in `{ aggregate: 1 }` in the command document.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -130,6 +131,100 @@ function(stdClass $command) use (&$commands) {
         $this->assertSame($expectedCommands, $commands);
     }
 
+    public function testResumeBeforeReceivingAnyResultsIncludesStartAtOperationTime()
+    {
+        $operation = new Watch($this->manager, $this->getDatabaseName(), $this->getCollectionName(), [], $this->defaultOptions);
```
jmikola: I originally inserted a document before starting the change stream, but that lead to a strange behavior. The initial `aggregate` command would not find the document (of course), but it's response would include an `operationTime` time based on the document's insertion since it was the most recent entry in the oplog.

When we kill and resume the cursor in the subsequent test activity, that operation time would be used for `startAtOperationTime` in the resuming `aggregate` and the document would show up.

I was under the impression that using the operation time from `aggregate` was intended to prevent extra documents from showing up from before the launch point of the original change stream, but perhaps this is a known edge case.

Perhaps @daprahamian, @kevinAlbs, or @ShaneHarvey can chime in here.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -484,16 +579,8 @@ public function testTypeMapOption(array $typeMap, $expectedChangeDocument)
 
         $changeStream->next();
         $this->assertTrue($changeStream->valid());
-        $changeDocument = $changeStream->current();
-
-        // Unset the resume token and namespace, which are intentionally omitted
```
jmikola: Thanks to `assertMatchesDocument()`, we can ignore extra fields in the actual document that are not present in our expected value. This allows us to do without these manual unsets.

tests/TestCase.php
```diff
@@ -69,6 +69,49 @@ protected function assertDeprecated(callable $execution)
...
+     */
+    protected function assertMatchesDocument($expectedDocument, $actualDocument)
+    {
+        $normalizedExpectedDocument = $this->normalizeBSON($expectedDocument);
+        $normalizedActualDocument = $this->normalizeBSON($actualDocument);
+
+        $extraKeys = [];
+
+        /* Avoid unsetting fields while we're iterating on the ArrayObject to
+         * work around https://bugs.php.net/bug.php?id=70246 */
```
jmikola: I had a great time discovering this bug. :stuck_out_tongue:

src/Operation/Aggregate.php
```diff
@@ -289,7 +292,7 @@ public function execute(Server $server)
     private function createCommand(Server $server)
     {
         $cmd = [
-            'aggregate' => $this->collectionName,
+            'aggregate' => isset($this->collectionName) ? $this->collectionName : 1,
```
derickr: The use of "1" wasn't really documented: https://github.com/mongodb/docs/pull/3357

src/Operation/Watch.php
```diff
@@ -79,22 +92,44 @@ class Watch implements Executable
...
-     * @param string         $collectionName Collection name
+     * Note: A database-level change stream may be created by specifying null
+     * for the collection name. A cluster-level change stream may be created by
+     * specifying null for both the database and collection name.
+     *
+     * @param string|null    $databaseName   Database name
+     * @param string|null    $collectionName Collection name
      * @param array          $pipeline       List of pipeline operations
      * @param array          $options        Command options
      * @param Manager        $manager        Manager instance from the driver
```
derickr: Why is "Manager" the last option in this list, but the first argument in the method declaration?

tests/Operation/WatchFunctionalTest.php
derickr: Does that also address the extra "clusterTime" information that made tests fail?

tests/TestCase.php
derickr: For fun I would say: @sgolemon, fix yo shit ;-)

src/Operation/Watch.php
jmikola: Good catch! Looks like an old typo: https://github.com/mongodb/mongo-php-library/blame/master/src/Operation/Watch.php#L93

I'll fix it.

tests/Operation/WatchFunctionalTest.php
jmikola: @ShaneHarvey mentioned that he and @behackett talked about this some time ago and concluded that it was expected behavior (if no other write operation on the server occurs between the `insert` and `aggregate` nor does enough time elapse for a periodic no-op entry to be created).

They tossed around the idea of having the driver manually increment the operation time returned by `aggregate`, but nothing came of that.

src/Operation/Watch.php
```diff
@@ -104,79 +139,145 @@ public function __construct(Manager $manager, $databaseName, $collectionName, ar
...
+            throw InvalidArgumentException::invalidType('"resumeAfter" option', $options['resumeAfter'], 'array or object');
+        }
+
+        if (isset($options['startAtOperationTime']) && ! $options['startAtOperationTime'] instanceof TimestampInterface) {
+            throw InvalidArgumentException::invalidType('"startAtOperationTime" option', $options['startAtOperationTime'], TimestampInterface::class);
         }
 
+        /* In the absence of an explicit session, create one to ensure that the
+         * initial aggregation and any resume attempts can use the same session
+         * ("implicit from the user's perspective per PHPLIB-342). We can ignore
```
kvwalker: I think there's a " missing after 'perspective'

tests/Operation/WatchFunctionalTest.php
ShaneHarvey: Yeah I just talked with Bernie, incrementing the operation time does not really solve the problem. We can fix the issue in SPEC-1122.

jmikola: Aye, SPEC-1122 ("Update startAtOperationTime during change stream iteration" for those without access) would help provided we're not resuming because the first `getMore` has failed; however, I'll admit that this is a fairly contrived edge case more likely to show up in a test suite than real life.

PR #551, opened 2018-06-27
PHPLIB-350: Provide Transactions example for Docs

derickr:


tests/DocumentationExamplesTest.php
```diff
@@ -1180,6 +1184,223 @@ public function testIndex_example_2()
...
+        } catch (\MongoDB\Driver\Exception\Exception $error) {
+            echo "Caught exception during transaction, aborting.\n";
+            $session->abortTransaction();
+            throw $error;
+        }
+
+        while (true) {
+            try {
+                $session->commitTransaction();
+                echo "Transaction committed.\n";
```
derickr: When running the tests, this is printed out. Not sure what to do about this, as it's what the Doc examples want.

tests/DocumentationExamplesTest.php
```diff
@@ -1180,6 +1184,223 @@ public function testIndex_example_2()
...
+    }
+
+    // Start Transactions Retry Example 1
+    private function runTransactionWithRetry1(callable $txnFunc, Manager $manager, Session $session)
+    {
+        while (true) {
+            try {
+                $txnFunc($manager, $session);  // performs transaction
+                break;
+            } catch (\MongoDB\Driver\Exception\CommandException $error) {
```
derickr: I thought to be specific about the Exception classes here for now. We should update these of course when we have error labels added.

tests/DocumentationExamplesTest.php
```diff
@@ -1180,6 +1184,223 @@ public function testIndex_example_2()
         $this->assertEquals('cuisine_1_name_1', $indexName);
     }
 
+    // Start Transactions Intro Example 1
+    private function updateEmployeeInfo1(Manager $manager, Session $session)
```
jmikola: AFAIK, none of the other examples refer to class names. If we do, I think we should use the FQCN. It's quite verbose, but it ensures the examples can be easily copy-pasted. Users can always add `use` statements on their own.

I find it odd that Session gets passed in here instead of creating it within, but I suppose that's up to the docs team.

tests/DocumentationExamplesTest.php
```diff
@@ -1180,6 +1184,223 @@ public function testIndex_example_2()
...
+    {
+        $employees = (new Database($manager, "hr"))->employees;
+        $events = (new Database($manager, "reporting"))->events;
```
jmikola: I'm also not sure how I feel about revealing the Manager here. If we can avoid it, I'd rather not suggest that users work with Managers directly nor suggest that they manually construct Database and Collection.

Can you inject the Database object (or a Client) instead? From there, you can use `$db->employees` and `$db->events`, respectively.

tests/DocumentationExamplesTest.php
```diff
@@ -1180,6 +1184,223 @@ public function testIndex_example_2()
...
+            } catch (\MongoDB\Driver\Exception\Exception $error) {
+                echo "Error during commit ...\n";
+                throw $error;
+            }
+        }
+    }
+    // End Transactions Intro Example 1
+
+    public function testTransactions_intro_example_1()
+    {
```
jmikola: Let's add skips to all three test methods that utilize transactions:

```
if ($this->getPrimaryServer()->getType() === Server::TYPE_STANDALONE) {
    $this->markTestSkipped('Transactions are not supported on standalone servers');
}

if (version_compare($this->getFeatureCompatibilityVersion(), '4.0', '<')) {
    $this->markTestSkipped('Transactions are only supported on FCV 4.0 or higher');
}
```

tests/DocumentationExamplesTest.php
```diff
@@ -1180,6 +1184,223 @@ public function testIndex_example_2()
...
+        $hrDb = new Database($this->manager, "hr");
+        $reportingDb = new Database($this->manager, "reporting");
+
+        $hrDb->dropCollection('employees');
+        $reportingDb->dropCollection('events');
+
+        /* Collections need to be created before a transaction starts */
+        $hrDb->createCollection('employees');
+        $reportingDb->createCollection('events');
```
jmikola: Should the drop or create operations utilize a majority write concern? I seem to recall that coming up earlier in the PHPC tests. Is it not a problem because you're doing a create after the drop?

tests/DocumentationExamplesTest.php
```diff
@@ -1180,6 +1184,223 @@ public function testIndex_example_2()
...
+        }
+    }
+    // End Transactions Retry Example 2
+
+    // Start Transactions Retry Example 3
+    private function runTransactionWithRetry3(callable $txnFunc, Manager $manager, Session $session)
+    {
+        while (true) {
+            try {
+                $txnFunc($manager, $session);  // performs transaction
```
jmikola: First argument can be `$db`.

tests/DocumentationExamplesTest.php
```diff
@@ -1180,6 +1184,223 @@ public function testIndex_example_2()
...
+            } catch (\MongoDB\Driver\Exception\Exception $error) {
+                echo "Error during commit ...\n";
+                throw $error;
+            }
+        }
+    }
+    // End Transactions Retry Example 2
+
+    // Start Transactions Retry Example 3
+    private function runTransactionWithRetry3(callable $txnFunc, Manager $manager, Session $session)
```
jmikola: Second argument can be `\MongoDB\Database $db`.

tests/DocumentationExamplesTest.php
```diff
@@ -1180,6 +1184,223 @@ public function testIndex_example_2()
...
+                }
+            } catch (\MongoDB\Driver\Exception\Exception $error) {
+                throw $error;
+            }
+        }
+    }
+    // End Transactions Retry Example 1
+
+    // Start Transactions Retry Example 2
+    private function commitWithRetry2(Manager $manager, Session $session)
```
jmikola: First argument can be `\MongoDB\Database $db`.

tests/DocumentationExamplesTest.php
```diff
@@ -1180,6 +1184,223 @@ public function testIndex_example_2()
...
+        $hrDb->createCollection('employees');
+        $reportingDb->createCollection('events');
+
+        $session = $this->manager->startSession();
+
+        $this->updateEmployeeInfo1($this->manager, $session);
+    }
+
+    // Start Transactions Retry Example 1
+    private function runTransactionWithRetry1(callable $txnFunc, Manager $manager, Session $session)
```
jmikola: Second argument can be `\MongoDB\Database $db`.

tests/DocumentationExamplesTest.php
```diff
@@ -1180,6 +1184,223 @@ public function testIndex_example_2()
...
+    {
+        while (true) {
+            try {
+                $txnFunc($manager, $session);  // performs transaction
```
jmikola: First argument can be `$db`.

tests/DocumentationExamplesTest.php
```diff
@@ -1180,6 +1184,223 @@ public function testIndex_example_2()
...
+                } else {
+                    throw $error;
+                }
+            } catch (\MongoDB\Driver\Exception\Exception $error) {
+                throw $error;
+            }
+        }
+    }
+
+    private function commitWithRetry3(Manager $manager, Session $session)
```
jmikola: First argument can be `\MongoDB\Database $db`.

jmikola: On second thought, why does this method take anything more than `$session`? The other parameter isn't used at all.

tests/DocumentationExamplesTest.php
```diff
@@ -1180,6 +1184,223 @@ public function testIndex_example_2()
...
+                ['employee' => 3, 'status' => [ 'new' => 'Inactive', 'old' => 'Active']],
+                ['session' => $session]
+            );
+        } catch (\MongoDB\Driver\Exception\Exception $error) {
+            echo "Caught exception during transaction, aborting.\n";
+            $session->abortTransaction();
+            throw $error;
+        }
+
+        $this->commitWithRetry3($manager, $session);
```
jmikola: `commitWithRetry3()` shouldn't require the `$manager` (or `$db` if you make other suggested changes).

tests/DocumentationExamplesTest.php
```diff
@@ -1180,6 +1184,223 @@ public function testIndex_example_2()
...
+                    throw $error;
+                }
+            } catch (\MongoDB\Driver\Exception\Exception $error) {
+                echo "Error during commit ...\n";
+                throw $error;
+            }
+        }
+    }
+
+    private function updateEmployeeInfo3(Manager $manager, Session $session)
```
jmikola: First argument can be `\MongoDB\Database $db`.

tests/DocumentationExamplesTest.php
```diff
@@ -1180,6 +1184,223 @@ public function testIndex_example_2()
...
+    {
+        $employees = (new Database($manager, "hr"))->employees;
+        $events = (new Database($manager, "reporting"))->events;
```
jmikola: `$db->employees` and `$db->events`.

tests/DocumentationExamplesTest.php
```diff
@@ -1180,6 +1184,223 @@ public function testIndex_example_2()
...
+
+        $session->startTransaction([
+            'readConcern' => new ReadConcern("snapshot"),
+            'writeConcern' => new WriteConcern(WriteConcern::MAJORITY)
```
jmikola: FQCN for both of these. I realize that makes the WriteConcern example a bit long :D

tests/DocumentationExamplesTest.php
```diff
@@ -1180,6 +1184,223 @@ public function testIndex_example_2()
...
+        }
+    }
+
+    private function updateEmployeeInfo3(Manager $manager, Session $session)
+    {
+        $employees = (new Database($manager, "hr"))->employees;
+        $events = (new Database($manager, "reporting"))->events;
+
+        $session->startTransaction([
+            'readConcern' => new ReadConcern("snapshot"),
```
jmikola: Single-quoted strings where possible for consistency.

tests/DocumentationExamplesTest.php
```diff
@@ -1180,6 +1184,223 @@ public function testIndex_example_2()
...
+        /* Prepare databases for transaction */
+		$hrDb = new Database($this->manager, "hr");
+        $reportingDb = new Database($this->manager, "reporting");
+
+		$hrDb->dropCollection('employees');
+		$reportingDb->dropCollection('events');
+
+        /* Collections need to be created before a transaction starts */
+		$hrDb->createCollection('employees');
+        $reportingDb->createCollection('events');
```
jmikola: You can left-align everything here. No need for fancy C-style method alignments :)

tests/DocumentationExamplesTest.php
```diff
@@ -1180,6 +1184,223 @@ public function testIndex_example_2()
...
+
+        // Start a session.
+        $session = $this->manager->startSession();
+
+        try {
+            $this->runTransactionWithRetry3(array($this, 'updateEmployeeInfo3'), $this->manager, $session);
```
jmikola: We can use short array syntax everywhere. Also consider replacing `$manager` with `$db`.

tests/DocumentationExamplesTest.php
```diff
@@ -1180,6 +1184,223 @@ public function testIndex_example_2()
...
+        } catch (\MongoDB\Driver\Exception\Exception $error) {
+            echo "Caught exception during transaction, aborting.\n";
+            $session->abortTransaction();
+            throw $error;
+        }
+
+        $this->commitWithRetry3($manager, $session);
+    }
+
+    public function testTransactions_retry_example_3()
```
jmikola: Rather than include this in "Transactions Retry Example 3", I think we should create another function and keep the "test" function internal. That will also allow you to add the necessary skip logic and his that from the documentation examples.

The example wants to demonstrate creation of a session, so you'll need to pass our new function a single `\MongoDB\Client` object. `client = new Client($this->getUri());` should get the job done. From there, you can use its `startSession()` method and also select the necessary databases with `->`.

tests/DocumentationExamplesTest.php
```diff
@@ -1180,6 +1184,223 @@ public function testIndex_example_2()
...
+        $reportingDb = new Database($this->manager, "reporting");
+
+        $hrDb->dropCollection('employees');
+        $reportingDb->dropCollection('events');
+
+        /* Collections need to be created before a transaction starts */
+        $hrDb->createCollection('employees');
+        $reportingDb->createCollection('events');
+
+        $session = $this->manager->startSession();
```
jmikola: Be sure to use the Client object here. Although the Client and Manager share the same sockets (same URI), they are different Manager instances (Client creates its own internally) and you'd get a driver exception if a Session from one Manager is used with another.

tests/DocumentationExamplesTest.php
derickr: Done

tests/DocumentationExamplesTest.php
derickr: Done

tests/DocumentationExamplesTest.php
derickr: Done

tests/DocumentationExamplesTest.php
derickr: Done

tests/DocumentationExamplesTest.php
derickr: Done

tests/DocumentationExamplesTest.php
derickr: Done, I think.

tests/DocumentationExamplesTest.php
derickr: Done

tests/DocumentationExamplesTest.php
derickr: Done

tests/DocumentationExamplesTest.php
derickr: Done

tests/DocumentationExamplesTest.php
derickr: Done

tests/DocumentationExamplesTest.php
derickr: Done

tests/DocumentationExamplesTest.php
derickr: Done

tests/DocumentationExamplesTest.php
derickr: Done

tests/DocumentationExamplesTest.php
derickr: Done

tests/DocumentationExamplesTest.php
```diff
@@ -1180,6 +1184,223 @@ public function testIndex_example_2()
...
+
+        $this->updateEmployeeInfo1($this->manager, $session);
```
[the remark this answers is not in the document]
derickr: Done

tests/DocumentationExamplesTest.php
derickr: Yes --- done

tests/DocumentationExamplesTest.php
derickr: Done

tests/DocumentationExamplesTest.php
derickr: Done

tests/DocumentationExamplesTest.php
derickr: Used output buffering

tests/DocumentationExamplesTest.php
derickr: Done, using client

tests/DocumentationExamplesTest.php
```diff
@@ -1180,6 +1184,223 @@ public function testIndex_example_2()
         $this->assertEquals('cuisine_1_name_1', $indexName);
     }
 
+    // Start Transactions Intro Example 1
+    private function updateEmployeeInfo1(Manager $manager, Session $session)
+    {
+        $employees = (new Database($manager, "hr"))->employees;
```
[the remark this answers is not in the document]
derickr: DELETED.

tests/DocumentationExamplesTest.php
derickr: Done

jmikola:
LGTM. Two warnings from PHPUnit, but we can address those later:

```
There were 2 risky tests:

1) MongoDB\Tests\DocumentationExamplesTest::testTransactions_intro_example_1
This test did not perform any assertions

2) MongoDB\Tests\DocumentationExamplesTest::testTransactions_retry_example_3
This test did not perform any assertions
```

PR #553, opened 2018-06-27
PHPLIB-363: Require ext-mongodb 1.5.0 and drop support for server versions < 3.0

jmikola:
https://jira.mongodb.org/browse/PHPLIB-363


src/Operation/ListCollections.php
```diff
@@ -143,35 +134,4 @@ private function executeCommand(Server $server)
 
         return new CollectionInfoCommandIterator(new CachingIterator($cursor));
     }
-
-    /**
-     * Returns information for all collections in this database by querying the
-     * "system.namespaces" collection (MongoDB <3.0).
```
derickr: I found it highly annoying there was no space after the ``<`` :-D

jmikola: You should have spoken up three years ago! https://github.com/mongodb/mongo-php-library/commit/99650bb320b79c9bc5e94dcfc4b2106a814ca33f#diff-6561571108dcf5d57f7d6fbee2beff25R95

Issue #559, opened 2018-07-05
Trying to clone an uncloneable object of class MongoDB\BSON\ObjectId


jrslv:
Is an issue in 1.40
Is not an issue in 1.3.2

https://github.com/mongodb/mongo-php-library/blob/1.4.0/src/Model/BSONArray.php#L41
throws an error "Trying to clone an uncloneable object of class MongoDB\BSON\ObjectId"
if passed an array of `ObjectID` objects

jmikola:
Thanks for bringing this to our attention. [PHPLIB-190](https://jira.mongodb.org/browse/PHPLIB-190) implemented `__clone()` handlers for the BSONArray and BSONDocument classes in PHPLIB 1.4.0; however, it looks like the BSON classes in the driver were never clone-able. I've opened [PHPC-1230](https://jira.mongodb.org/browse/PHPC-1230) to track the necessary driver improvement.

I've also opened [PHPLIB-367](https://jira.mongodb.org/browse/PHPLIB-367) to develop a quick workaround in this library, which we should be able to publish much sooner in a patch release.

PR #560, opened 2018-07-05
PHPLIB-367: BSONArray/Document cloning should respect uncloneable objects

jmikola:
https://jira.mongodb.org/browse/PHPLIB-367

This should address the issue reported in https://github.com/mongodb/mongo-php-library/issues/559.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -23,6 +23,7 @@ use MongoDB\Driver\Server;
 use MongoDB\Driver\WriteConcern;
 use MongoDB\Exception\InvalidArgumentException;
 use stdClass;
+use ReflectionClass;
 
 /**
  * Applies a type map to a document.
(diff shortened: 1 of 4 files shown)


src/functions.php
```diff
@@ -216,5 +217,9 @@ function recursive_copy($element) {
         return $element;
     }
 
+    if ( ! (new ReflectionClass($element))->isCloneable()) {
```
derickr: Is cloneable is reported to not work with internal classes (in a note on http://php.net/manual/en/reflectionclass.iscloneable.php) — could that be a problem for us?

jmikola: That note was the reason I added `testCloneSupportsBSONTypes()`. I didn't look into `isCloneable()`, but it certainly appears to do the job for our BSON classes in PHP 5.5+.

Issue #562, opened 2018-07-12
Error: Cannot set read preferences after starting transaction


sandrokeil:
I don't explicitly set a read preference but it looks that this library sets always read preference `primary` for each query if not specified. This results in the error `Cannot set read preferences after starting transaction`.

Here is an example which fails:

```php
$client = new Client();
$collection = $client->selectCollection('mydb', 'test');
$session = $client->startSession();
$session->startTransaction(
    [
        'readConcern' => new \MongoDB\Driver\ReadConcern('snapshot'),
        'writeConcern' => new \MongoDB\Driver\WriteConcern(\MongoDB\Driver\WriteConcern::MAJORITY),
    ]
);

$options = [
    'session' => $session
];

$collection->findOne([], $options); // -> Exception
```

How can I read data from a collection inside a transaction?

jmikola:
Thanks for the report. This looks to be a libmongoc bug, which I've reported in [CDRIVER-2744](https://jira.mongodb.org/browse/CDRIVER-2744) and submitted a PR for in https://github.com/mongodb/mongo-c-driver/pull/517. It looks like this can be addressed in libmongoc's upcoming 1.11.1 release, for which we will likely publish a new PHPC release for those using the bundled version of libmongoc instead of system libraries. I've created [PHPC-1236](https://jira.mongodb.org/browse/PHPC-1236) to track that.

Additionally, I believe we can make an improvement to PHPLIB by not sending an explicit read preference option on commands/queries unless we're communicating with a sharded cluster (where transactions are not even supported in 4.0). I've opened [PHPLIB-375](https://jira.mongodb.org/browse/PHPLIB-375) to track that. While this won't fix the libmongoc bug, it would allow PHPLIB to work around it.

Please watch both [PHPC-1236](https://jira.mongodb.org/browse/PHPC-1236) and [PHPLIB-375](https://jira.mongodb.org/browse/PHPLIB-375) for updates (or [CDRIVER-2744](https://jira.mongodb.org/browse/CDRIVER-2744) if you've configured the PHP extension to use libmongoc and libbson as system libraries). I expect that any one of those issues will resolve this for you.

----

[...]

PR #564, opened 2018-07-12
Regression tests for find() and aggregate() using primary RP within transaction

jmikola:
See: https://github.com/mongodb/mongo-php-library/issues/562

This is a regression test for [CDRIVER-2744](https://jira.mongodb.org/browse/CDRIVER-2744). It builds upon https://github.com/mongodb/mongo-php-library/pull/563 (for [PHPLIB-373](https://jira.mongodb.org/browse/PHPLIB-373)) and may be fixed by either:

 * [PHPLIB-375](https://jira.mongodb.org/browse/PHPLIB-375): omitting read preference option for commands and queries sent in non-sharded topologies
 * [PHPC-1236](https://jira.mongodb.org/browse/PHPC-1236): Bumping libmongoc to 1.11.1 (once released)

I don't think we should merge this while failing and would suggest leaving it open until one of the solutions can be implemented.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
         "php": ">=5.5",
         "ext-hash": "*",
         "ext-json": "*",
-        "ext-mongodb": "^1.5.0"
+        "ext-mongodb": "^1.5.2"
     },
     "require-dev": {
         "phpunit/phpunit": "^4.8.36 || ^6.4"
--- a/tests/Collection/CollectionFunctionalTest.php
+++ b/tests/Collection/CollectionFunctionalTest.php
@@ -10,8 +10,10 @@ use MongoDB\Driver\ReadPreference;
 use MongoDB\Driver\WriteConcern;
 use MongoDB\Exception\InvalidArgumentException;
 use MongoDB\Operation\Count;
+use MongoDB\Operation\CreateCollection;
 use MongoDB\Operation\MapReduce;
 use MongoDB\Tests\CommandObserver;
+use Exception;
 use stdClass;
 
 /**
@@ -104,6 +106,39 @@ class CollectionFunctionalTest extends FunctionalTestCase
         $this->assertEquals($this->getNamespace(), $this->collection->getNamespace());
     }
 
+    public function testAggregateWithinTransaction()
+    {
+        $this->skipIfTransactionsAreNotSupported();
+
+        // Collection must be created before the transaction starts
+        $options = ['writeConcern' => new WriteConcern(WriteConcern::MAJORITY)];
+        $operation = new CreateCollection($this->getDatabaseName(), $this->getCollectionName(), $options);
+        $operation->execute($this->getPrimaryServer());
+
+        $session = $this->manager->startSession();
+        $session->startTransaction();
+
+        try {
+            $this->createFixtures(3, ['session' => $session]);
+
+            $cursor = $this->collection->aggregate(
+                [['$match' => ['_id' => ['$lt' => 3]]]],
+                ['session' => $session]
+            );
+
+            $expected = [
+                ['_id' => 1, 'x' => 11],
+                ['_id' => 2, 'x' => 22],
+            ];
+
+            $this->assertSameDocuments($expected, $cursor);
+
+            $session->commitTransaction();
+        } finally {
+            $session->endSession();
+        }
+    }
+
     public function testCreateIndexSplitsCommandOptions()
     {
         if (version_compare($this->getServerVersion(), '3.6.0', '<')) {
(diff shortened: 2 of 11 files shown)


tests/Collection/CollectionFunctionalTest.php
```diff
@@ -104,6 +106,40 @@ public function testGetNamespace()
         $this->assertEquals($this->getNamespace(), $this->collection->getNamespace());
     }
 
+    public function testAggregateWithinTransaction()
```
jmikola: I decided to add tests for `aggregate()` in response to https://github.com/mongodb/mongo-php-library/issues/562#issuecomment-404792939. Best I can tell, there are no problems with `aggregate()`, which jives with the underlying libmongoc bug.

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -179,6 +215,40 @@ public function testFindOne()
...
+            $expected = [
+                ['_id' => 1, 'x' => 11],
+                ['_id' => 2, 'x' => 22],
+            ];
+
+            $this->assertSameDocuments($expected, $cursor);
+
+            $this->commitTransaction();
+        } catch (Exception $e) {
+            $session->abortTransaction();
```
jmikola: Since the `find()` tests throw an exception, we must be careful to abort the transaction before allowing PHPUnit to process the test failure. Without doing so, we leave the transaction open on the server and a subsequent run of this test will hang when attempting to drop the collection during `setUp()`. Consider the following mongod log:

```
2018-07-13T11:40:18.069-0400 I COMMAND  [conn80] CMD: drop phplib_test.CollectionFunctionalTest.1ab43437

*** more than 60 seconds elapse ***

2018-07-13T11:41:20.741-0400 I STORAGE  [conn80] dropCollection: phplib_test.CollectionFunctionalTest.1ab43437 (88ff4330-4ee4-44b2-afcd-933f32b5d741) - renaming to drop-pending collection: phplib_test.system.drop.1531496480i1t3.CollectionFunctionalTest.1ab43437 with drop optime { ts: Timestamp(1531496480, 1), t: 3 }
2018-07-13T11:41:20.741-0400 I STORAGE  [conn80] renameCollection: renaming collection 88ff4330-4ee4-44b2-afcd-933f32b5d741 from phplib_test.CollectionFunctionalTest.1ab43437 to phplib_test.system.drop.1531496480i1t3.CollectionFunctionalTest.1ab43437
2018-07-13T11:41:20.744-0400 I COMMAND  [conn80] command phplib_test.CollectionFunctionalTest.1ab43437 command: drop { drop: "CollectionFunctionalTest.1ab43437", $db: "phplib_test", lsid: { id: UUID("d3fa5cf7-d8a2-4e9f-b764-c09c7dbaa401") }, $clusterTime: { clusterTime: Timestamp(1531496418, 4), signature: { hash: BinData(0, 0000000000000000000000000000000000000000), keyId: 0 } } } numYields:0 reslen:234 locks:{ Global: { acquireCount: { r: 2, w: 2 } }, Database: { acquireCount: { w: 1, W: 1 }, acquireWaitCount: { W: 1 }, timeAcquiringMicros: { W: 62671648 } }, oplog: { acquireCount: { w: 1 } } } protocol:op_msg 62675ms
2018-07-13T11:41:20.744-0400 I REPL     [replication-1] Completing collection drop for phplib_test.system.drop.1531496480i1t3.CollectionFunctionalTest.1ab43437 with drop optime { ts: Timestamp(1531496480, 1), t: 3 } (notification optime: { ts: Timestamp(1531496480, 1), t: 3 })
2018-07-13T11:41:20.745-0400 I STORAGE  [replication-1] Finishing collection drop for phplib_test.system.drop.1531496480i1t3.CollectionFunctionalTest.1ab43437 (88ff4330-4ee4-44b2-afcd-933f32b5d741).
```

jmikola: It looks like this is actually the result of failing to use a majority write concern during the initial drop/create of the test collection. See [PHPLIB-377](https://jira.mongodb.org/browse/PHPLIB-377).

jmikola: On second thought, it may actually be a problem with an in-progress transaction not being aborted when the `MongoDB\Driver\Session` is destroyed. This may be a libmongoc bug.

jmikola: Fourth time's the charm. After some more testing, I can attest that this is not a libmongoc bug. Rather, the exception being thrown includes a reference to the `$session` and keeps it and the transaction open beyond the scope of the test method itself.

I'll see if I can work out another approach that doesn't require explicit aborting in the meantime.

jmikola: Note: once [PHPC-1243](https://jira.mongodb.org/browse/PHPC-1243) is resolved and released, we can change this to `finally { $session->endSession(); }` (with a bump to our PHPC dependency).

composer.json
```diff
@@ -13,7 +13,7 @@
         "php": ">=5.5",
         "ext-hash": "*",
         "ext-json": "*",
-        "ext-mongodb": "^1.5.0"
+        "ext-mongodb": "^1.5.2"
```
jmikola: This ensures that [CDRIVER-2744](https://jira.mongodb.org/browse/CDRIVER-2744) is in place via libmongoc 1.12.

jmikola:
@derickr: This has been rebased on master, so please review at your leisure. I don't expect the `ext-mongodb` bump is controversial, since the master branch will likely become PHPLIB 1.5.0 and depend on PHPC 1.6.0.

tests/Operation/AggregateFunctionalTest.php
```diff
@@ -274,12 +277,50 @@ public function provideTypeMapOptionsAndExpectedDocuments()
...
     }
 
+    public function testReadPreferenceWithinTransaction()
+    {
+        $this->skipIfTransactionsAreNotSupported();
+
+        // Collection must be created before the transaction starts
+        $options = ['writeConcern' => new WriteConcern(WriteConcern::MAJORITY)];
+        $operation = new CreateCollection($this->getDatabaseName(), $this->getCollectionName(), $options);
+        $operation->execute($this->getPrimaryServer());
```
derickr: These three lines are copied many times, perhaps they should go into a helper function in FunctionalTestCase?

jmikola: Good idea. I'll make the change and update the PR.

jmikola:
@derickr: I created generic create/drop collection helpers that apply a majority WC when supported by the server. Testing locally, there was no problem using a majority WC with a standalone, but we'll see what Travis says.

I left some explicit uses of CreateCollection and DropCollection operations as-is, particularly when they were using custom collection names.

PR #566, opened 2018-07-16
PHPLIB-376: countDocuments() throws an exception if no documents are found

derickr:


src/Operation/CountDocuments.php
```diff
@@ -152,8 +152,15 @@ public function execute(Server $server)
         }
 
         $cursor = $server->executeReadCommand($this->databaseName, new Command($this->createCommandDocument()), $this->createOptions());
-        $result = current($cursor->toArray());
+        $allResults = $cursor->toArray();
 
+        /* If there are no documents to count, the aggregation pipeline has no items to group, and
+         * hence the result is an empty array (PHPLIB-376) */
+        if (is_array($allResults) && count($allResults) == 0) {
```
jmikola: `is_array()` seems redundant here, as `Cursor::toArray()` will never not return an array.

I think we can simplify this by leaving `$result = current($cursor->toArray());` as-is and just checking for the appropriate return value from [`current()`](http://php.net/current):

> If the internal pointer points beyond the end of the elements list or _the array is empty_, current() returns FALSE.

If you're adamant about removing ` current($cursor->toArray());`, I think we can at least get rid of `is_array()`.

tests/Operation/CountDocumentsFunctionalTest.php
```diff
@@ -0,0 +1,30 @@
...
+use MongoDB\Operation\CountDocuments;
+use MongoDB\Operation\InsertMany;
+use stdClass;
+
+class CountDocumentsFunctionalTest extends FunctionalTestCase
+{
+    public function testEmptyCollection()
+    {
+        $operation = new CountDocuments($this->getDatabaseName(), $this->getCollectionName(), []);
+        $this->assertEquals(0, $operation->execute($this->getPrimaryServer()));
```
jmikola: When possible, `assertSame()` is helpful to ensure we return the correct types. This is also nice because we can't yet you return type hints.

derickr: I've added a separate commit to update the other test too

jmikola: Did you forget to push changes?

PR #571, opened 2018-07-31
PHPLIB-341: Add LocalDateTime example to tutorial

derickr:


docs/tutorial/custom-types.txt
```diff
@@ -0,0 +1,174 @@
...
+        private $tz;
+        public function __construct($milliseconds = null, \DateTimeZone $timezone = null)
+        {
+            $this->utc = new \MongoDB\BSON\UTCDateTime($milliseconds);
+            if ($timezone === null) {
+                $timezone = new \DateTimeZone(date_default_timezone_get());
+            }
+            $this->tz = $timezone;
+        }
+    ?>
```
derickr: If you want to say that we should remove the ``?>`` than I will say no. Without it, the code highlighting in VIM is royally screwed up.

docs/tutorial/custom-types.txt
```diff
@@ -0,0 +1,174 @@
...
+
+The driver serializes PHP variables, including objects, into BSON when it
+communicates to the server, and deserializes BSON back into PHP variables when
+it receives data from the server.
+
+It is possible to influence the behaviour by implementing the
+:php:`MongoDB\\BSON\\Persistable <class.mongodb-bson-persistable>` interface.
+If a class implements this interface, then upon serialization the
+:php:`bsonSerialize <mongodb-bson-serializable.bsonserialize>` method is
+called. This method is responsible for returning an array with the fields that
```
jmikola: This can return an an array or object so we can properly differentiate between empty arrays and documents. I'd suggest:

> This method is responsible for returning an array or stdClass object to convert to BSON and store in the database. That data will later be used to reconstruct the object upon reading from the database.

docs/tutorial/custom-types.txt
```diff
@@ -0,0 +1,174 @@
...
+        {
+            return [
+                'utc' => $this->utc,
+                'tz' => $this->tz->getName(),
+            ];
+        }
+    ?>
+
+The driver will additionally add a ``__pclass`` field to the document, and
+store that in the database too. This field contains the PHP class name so that
```
jmikola: "in the database, too" (comma before "too").

docs/tutorial/custom-types.txt
```diff
@@ -0,0 +1,174 @@
...
+upon deserialization the driver knows which class to use for recreating the
+stored object.
+
+When the document is read from the database, the driver detects whether a
+``__pclass`` field is present and then executes
+:php:`MongoDB\BSON\Persistable::bsonUnserialize
+<mongodb-bson-unserializable.bsonunserialize>` method which is
+responsible for restoring the object's original status.
```
jmikola: "original state" seems more correct.

docs/tutorial/custom-types.txt
```diff
@@ -0,0 +1,174 @@
...
+
+In the code below, we make sure that the data in the ``utc`` and ``tz`` fields
+are of the right time, and then assigns their values to the two private
```
jmikola: "and then assign their values". "assign" is agreeing with the subject "we".

docs/tutorial/custom-types.txt
```diff
@@ -0,0 +1,174 @@
...
+        }
+    ?>
+
+You may have noticed that the class also implements the
+:php:`MongoDB\\BSON\\UTCDateTimeInterface
+<class.mongodb-bson-utcdatetimeinterface>` interface. This interface defines
+the two non-constructor methods of the :php:`MongoDB\\BSON\\UTCDateTime
+<class.mongodb-bson-utcdatetime>` class. It is recommended that wrappers
+around existing BSON classes implement their respective interface so that
+their objects can be used in the same context as their original unwrapped
```
jmikola: "so that the wrapper objects" avoids ambiguity with "their objects".

docs/tutorial/custom-types.txt
```diff
@@ -0,0 +1,174 @@
...
+version.
```
jmikola: I might add a suggestion that users shouldn't also type-hint against these interfaces instead of our BSON type classes.

docs/tutorial/custom-types.txt
```diff
@@ -0,0 +1,174 @@
...
+    }
+    object(DateTime)#5 (3) {
+      ["date"]=>
+      string(26) "2018-07-31 14:07:23.716000"
+      ["timezone_type"]=>
+      int(3)
+      ["timezone"]=>
+      string(13) "Europe/London"
+    }
+
```
jmikola: Final thought, but would it be worth mentioning that this approach plays nicely with the aggregation framework, which allows you to perform operations using Olson time zone identifier strings? I don't think we need a code example, but a final blurb that links out to the aggregation manual entries may form a good conclusion.

docs/tutorial/custom-types.txt
derickr: No need for the grammar lesson :-þ I make typoes too.

docs/tutorial/custom-types.txt
jmikola: To clarify, there is no benefit to implementing this interfaces if other code is going to expect an actual UTCDateTime instance. "Can be used in the same context" really only works if folks commit to using the interfaces everywhere.

docs/tutorial/custom-types.txt
```diff
@@ -0,0 +1,182 @@
...
+You may have noticed that the class also implements the
+:php:`MongoDB\\BSON\\UTCDateTimeInterface
+<class.mongodb-bson-utcdatetimeinterface>` interface. This interface defines
+the two non-constructor methods of the :php:`MongoDB\\BSON\\UTCDateTime
+<class.mongodb-bson-utcdatetime>` class.
+
+It is recommended that wrappers around existing BSON classes implement their
+respective interface (i.e. :php:`MongoDB\\BSON\\UTCDateTimeInterface
+<class.mongodb-bson-utcdatetimeinterface>` so that the wrapper objects can be
```
jmikola: Missing a closing parent after the class link between the backtick and "so"

docs/tutorial/custom-types.txt
```diff
@@ -0,0 +1,182 @@
...
+used in the same context as their original unwrapped version. It is also
+recommended that you always type-hint against the interface i.e.
```
jmikola: Missing an opening paren before "i.e."

docs/tutorial/custom-types.txt
```diff
@@ -0,0 +1,182 @@
...
+:php:`MongoDB\\BSON\\UTCDateTimeInterface
+<class.mongodb-bson-utcdatetimeinterface>`), and never against the concrete
```
jmikola: no need for a comma between closing paren and "and".

docs/tutorial/custom-types.txt
```diff
@@ -0,0 +1,182 @@
...
+class (i.e. :php:`MongoDB\\BSON\\UTCDateTime
+<class.mongodb-bson-utcdatetime>`), as this would prevent wrapped objects from
+being accepted into methods.
```
jmikola: As-is, this sentence reads:

> It is also recommended that you always type-hint against the interface and never against the concrete, as this would prevent wrapped objects from being accepted into methods.

I understand the last clause applies to type-hinting "against the concrete" clause, but I think we can phrase this better:

> It is also recommended that you always type-hint against the interface instead of the concrete class. This will ensure that methods can accept wrapped objects and unwrapped objects interchangeably.

PR #572, opened 2018-08-03
PHPLIB-382: Causal Consistency Examples for the Manual

derickr:


tests/DocumentationExamplesTest.php
```diff
@@ -1408,6 +1410,57 @@ public function testTransactions_retry_example_3()
...
 
+    function testCausalConsistency()
+    {
+        $this->skipIfCausalConsistencyIsNotSupported();
+
+        // Prep
+        $client = new Client($this->getUri());
+        $test = $client->selectDatabase(
+            'test',
+            [ 'writeConcern' => new \MongoDB\Driver\WriteConcern(\MongoDB\Driver\WriteConcern::MAJORITY) ]
```
jmikola: This is not within the example code, so we can use `WriteConcern` here (it's already `use`-ed up top).

tests/DocumentationExamplesTest.php
```diff
@@ -1408,6 +1410,57 @@ public function testTransactions_retry_example_3()
...
+        );
+        $items = $client->test->items;
+
+        $items->drop();
+        $items->insertOne(
+            [ 'sku' => '111', 'name' => 'Peanuts', 'start' => new \MongoDB\BSON\UTCDateTime() ]
```
jmikola: Likewise, let's `use` this so we can use the short name here.

tests/DocumentationExamplesTest.php
```diff
@@ -1408,6 +1410,57 @@ public function testTransactions_retry_example_3()
...
+        $s2 = $client->startSession([ 'causalConsistency' => true ]);
+        $s2->advanceClusterTime($s1->getClusterTime());
+        $s2->advanceOperationTime($s1->getOperationTime());
+
+        $items = $client->selectDatabase(
+            'test',
+            [ 'readPreference' => new \MongoDB\Driver\ReadPreference(\MongoDB\Driver\ReadPreference::RP_SECONDARY) ]
+        )->items;
+
+        foreach ($items->find( [ 'end' => [ '$exists' => false ] ], [ 'session' => $s2 ] ) as $item) {
```
jmikola: It may be more readable to do `find()` on its own (splitting the arguments on their own line) and change this to `foreach ($cursor as $item)`.

tests/DocumentationExamplesTest.php
```diff
@@ -1408,6 +1410,57 @@ public function testTransactions_retry_example_3()
...
+            var_dump( $item );
```
jmikola: I realize this is your preferred coding style, but keep in mind that this may trigger warnings if we add code inspection to Travis down the line. I've no objections to spaces between array brackets and elements, as you've done above.

Side note: is this inconsistent with methods that take array arguments? For instance you have `startSession([ ...` above. Although you also have `$items->find( [ ...`, which aren't consistent with one another.

tests/FunctionalTestCase.php
```diff
@@ -114,6 +114,26 @@ protected function getServerStorageEngine(ReadPreference $readPreference = null)
...
+    {
+        if ($this->getPrimaryServer()->getType() === Server::TYPE_STANDALONE) {
+            $this->markTestSkipped('Causal Consistency is not supported on standalone servers');
+        }
+
+        if ($this->getPrimaryServer()->getType() === Server::TYPE_MONGOS) {
+            if (version_compare($this->getServerVersion(), '3.6.0', '<')) {
+                $this->markTestSkipped('Causal Consistency is only supported on MongoDB 3.6 or higher');
+            }
+        } else {
```
jmikola: Rather than nest conditionals, I propose we use an early return after the version compare conditional.

Alternatively, you can switch on the server type and handle mongos and a replica set explicitly. Then, `default:` can always `markTestSkipped()`.

tests/DocumentationExamplesTest.php
derickr: I did remove the spaces inside the parenthesis here, and fixed the ``startSession`` case by having the arguments on their own line.

tests/FunctionalTestCase.php
```diff
@@ -114,6 +115,74 @@ protected function getServerStorageEngine(ReadPreference $readPreference = null)
...
+        );
+
+        $cursor->setTypeMap(['root' => 'array', 'document' => 'array']);
+        $document = current($cursor->toArray());
+
+        if (! $document ) {
+            return false;
+        }
+
+        return preg_match('@^.*/.*:\d+@', $document['host']);
```
jmikola: Is a `preg_match` actually necessary? This looks to just be testing that the URI starts with any scheme followed by a forward slash and then anything before a `:\d+` (port specifier). What distinguishes this as a replica set?

tests/FunctionalTestCase.php
```diff
@@ -114,6 +115,74 @@ protected function getServerStorageEngine(ReadPreference $readPreference = null)
...
+        switch ( $this->getPrimaryServer()->getType() )
+        {
+            case Server::TYPE_MONGOS:
+                if (version_compare($this->getServerVersion(), '3.6.0', '<')) {
+                    $this->markTestSkipped('$changeStream is only supported on MongoDB 3.6 or higher');
+                }
+                if (!$this->isShardedClusterUsingReplicasets()) {
+                    $this->markTestSkipped('$changeStream is only supported with replicasets');
+                }
+                break;
```
jmikola: Can also just `return` here, but no objections to `break`.

tests/FunctionalTestCase.php
```diff
@@ -114,6 +115,74 @@ protected function getServerStorageEngine(ReadPreference $readPreference = null)
         throw new UnexpectedValueException('Could not determine server storage engine');
     }
 
+    protected function isShardedClusterUsingReplicasets()
```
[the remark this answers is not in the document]
derickr: It doesn't depend on it, but as PHPLIB doesn't test anything besides standalone, it *is* going to be useful once we move the topologies over from PHPC. Right now, I'm using the Travis topologies from PHPC to test PHPLIB too locally so that we can find these "bugs" in the test skips.

tests/FunctionalTestCase.php
derickr: It's necessary. With a replicaset: ``"host" : "dec6d8a7-9bc1-4c0e-960c-615f860b956f/localhost:4400,localhost:4401"``
Without a replicaset: ``"host" : "localhost:4100"``

I'll add a comment.

PR #573, opened 2018-08-14
PHPLIB-382: Updated Causal Consistency example for docs

derickr:


tests/DocumentationExamplesTest.php
```diff
@@ -1411,29 +1411,38 @@ function testCausalConsistency()
...
-        $items = $client->test->items;
+        )->items;
 
         $items->drop();
         $items->insertOne(
             [ 'sku' => '111', 'name' => 'Peanuts', 'start' => new UTCDateTime() ]
         );
 
         // Start Causal Consistency Example 1
+        $items = $client->selectDatabase(
```
derickr: I know we have just done this in ``:1414``, but it now needs to be part of the example, and we needed ``$items`` in 1414 so we can make the ``drop``/``insertOne`` to make the test execute.

tests/DocumentationExamplesTest.php
```diff
@@ -1411,29 +1411,38 @@ function testCausalConsistency()
...
         $s1 = $client->startSession(
             [ 'causalConsistency' => true ]
         );
 
+        $currentDate = new \MongoDB\BSON\UTCDateTime();
+
         $items->updateOne(
             [ 'sku' => '111', 'end' => [ '$exists' => false ] ],
-            [ '$currentDate' => [ 'end' => true ] ],
+            [ '$set' => [ 'end' => $currentDate ] ],
```
jmikola: Is there a particular reason the examples no longer want to use `$currentDate`?

PR #574, opened 2018-08-21
PHPLIB-380: Add alternative topologies to Travis CI

derickr:


tests/Operation/ModifyCollectionFunctionalTest.php
```diff
@@ -25,7 +25,21 @@ public function testCollMod()
...
         $result = $modifyCollection->execute($this->getPrimaryServer());
 
-        $this->assertSame(3, $result['expireAfterSeconds_old']);
-        $this->assertSame(1000, $result['expireAfterSeconds_new']);
+        if (array_key_exists('raw', $result)) {
+            /* Sharded environment, where we only assert if a shard had a successful update. For
+             * non-primary shards that don't have chunks for the collection, the result contains a
+             * "ns does not exist" error. */
+            foreach ($result['raw'] as $shard) {
+                $shard = (array) $shard;
```
jmikola: Why not add `'document' => 'array'` to the ModifyCollection `'typeMap` option to avoid this cast? We're already doing it for "root" to allow assertions on `$result['expireAfterSeconds_old']` and the like.

tests/FunctionalTestCase.php
```diff
@@ -4,6 +4,7 @@
 
 use MongoDB\Driver\Command;
 use MongoDB\Driver\Cursor;
+use MongoDB\Driver\Exception\CommandException;
```
jmikola: Small nit, but I usually alphabetize all classes under `MongoDB\Driver` together and then list sub-namespaces afterwards (akin to a directory listing showing all files and then directories at the end). I'm not too worried if this stays as is. We might catch it down the line with code inspection and can always clean things up then.

tests/FunctionalTestCase.php
```diff
@@ -53,23 +54,32 @@ protected function getFeatureCompatibilityVersion(ReadPreference $readPreference
...
+            if (isset($document['featureCompatibilityVersion']['version']) && is_string($document['featureCompatibilityVersion']['version'])) {
+                return $document['featureCompatibilityVersion']['version'];
+            }
+
+            // MongoDB 3.4: featureCompatibilityVersion is a string
+            if (isset($document['featureCompatibilityVersion']) && is_string($document['featureCompatibilityVersion'])) {
+                return $document['featureCompatibilityVersion'];
+            }
+        } catch (CommandException $e) {
+            /* It is not possible to get the featureCompatibilityVersion from mongos, so if the
```
jmikola: Did you write this with the expectation that a future version of mongos might support querying for FCV?

Would it make sense to leave most of the original method as-is and introduce an early conditional that instead selects a server and defers to `$this->getServerVersion($readPreference)` if its type is a mongos node?

That would also avoid using a magic constant (`17`) below, and allow us to get rid of the entire try/catch here.

tests/FunctionalTestCase.php
```diff
@@ -53,23 +54,32 @@ protected function getFeatureCompatibilityVersion(ReadPreference $readPreference
...
+             * command fails, fall back to the server version if the exception's code signals it was
+             * a "no option found to get" error. */
+            if ($e->getCode() == 17) {
+                return $this->getServerVersion($readPreference);
+            }
```
jmikola: As-is, I think this `catch` should re-throw the exception if we don't return the server version. Otherwise, there is no return value and things might fail unexpectedly. Rethrowing would lead to a much clearer test failure. That said, I think my suggestion above (i.e. use an early conditional) may make this a moot point.

tests/Database/DatabaseFunctionalTest.php
```diff
@@ -162,10 +162,26 @@ public function testModifyCollection()
...
-        $this->assertSame(1000, $commandResult['expireAfterSeconds_new']);
+
+        $commandResult = (array) $commandResult;
+
+        if (array_key_exists('raw', $commandResult)) {
+            /* Sharded environment, where we only assert if a shard had a successful update. For
+             * non-primary shards that don't have chunks for the collection, the result contains a
+             * "ns does not exist" error. */
+            foreach ($commandResult['raw'] as $shard) {
+                $shard = (array) $shard;
```
jmikola: Can we make this consistent with ModifyCollectionFunctionalTest by using the following options when calling `modifyCollection()` above?

```
['typeMap' => ['root' => 'array', 'document' => 'array']]
```

derickr: Done

PR #582, opened 2018-10-02
PHPLIB-387: Support errors and unordered writes in CRUD spec tests

jmikola:
https://jira.mongodb.org/browse/PHPLIB-387

These improvements complement test revisions in https://github.com/mongodb/specifications/pull/399.

We can wait until the new spec tests are merged so that they can be incorporated into this PR; however, the test runner changes can be reviewed immediately.

### DIFF
--- a/tests/Collection/CrudSpecFunctionalTest.php
+++ b/tests/Collection/CrudSpecFunctionalTest.php
@@ -2,7 +2,11 @@
 
 namespace MongoDB\Tests\Collection;
 
+use MongoDB\BulkWriteResult;
 use MongoDB\Collection;
+use MongoDB\InsertManyResult;
+use MongoDB\Driver\Exception\BulkWriteException;
+use MongoDB\Driver\Exception\RuntimeException;
 use MongoDB\Operation\FindOneAndReplace;
 use IteratorIterator;
 use LogicException;
@@ -39,8 +43,16 @@ class CrudSpecFunctionalTest extends FunctionalTestCase
         $expectedData = isset($test['outcome']['collection']['data']) ? $test['outcome']['collection']['data'] : null;
         $this->initializeData($initialData, $expectedData);
 
-        $result = $this->executeOperation($test['operation']);
-        $this->executeOutcome($test['operation'], $test['outcome'], $result);
+        $result = null;
+        $exception = null;
+
+        try {
+            $result = $this->executeOperation($test['operation']);
+        } catch (RuntimeException $e) {
+            $exception = $e;
+        }
+
+        $this->executeOutcome($test['operation'], $test['outcome'], $result, $exception);
     }
 
     public function provideSpecificationTests()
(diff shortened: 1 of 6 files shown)


tests/Collection/CrudSpecFunctionalTest.php
```diff
@@ -116,10 +128,9 @@ private function executeOperation(array $operation)
                 );
 
             case 'bulkWrite':
-                $results = $this->prepareBulkWriteArguments($operation['arguments']);
                 return $this->collection->bulkWrite(
-                    array_diff_key($results, ['options' => 1]),
-                    $results['options']
+                    array_map([$this, 'prepareBulkWriteRequest'], $operation['arguments']['requests']),
```
jmikola: The original `prepareBulkWriteArguments()` method returned an array consisting of both requests (under numeric indices) and options (under a single associative index). I refactored that method to instead format each request element independently (hence the use of `array_map`).

tests/Collection/CrudSpecFunctionalTest.php
```diff
@@ -212,6 +233,42 @@ private function executeOutcome(array $operation, array $outcome, $actualResult)
...
+            case 'bulkWrite':
+                $insertedIds = isset($outcome['result']['insertedIds']) ? $outcome['result']['insertedIds'] : [];
+
+                if ($exception instanceof BulkWriteException) {
+                    return new BulkWriteResult($exception->getWriteResult(), $insertedIds);
+                }
+                break;
+
+            case 'insertMany':
+                $insertedIds = isset($outcome['result']['insertedIds']) ? $outcome['result']['insertedIds'] : [];
```
jmikola: Per https://github.com/mongodb/specifications/pull/399#discussion_r222099362, the spec tests will not actually contain `insertedIds` in the expected result document; however, I would like our test runner to respect `insertedIds` if it's present.

In practice, a spec test with `insertedIds` would cause our driver to fail as our BulkWrite and InsertMany operations have no way of injecting inserted IDs into the thrown BulkWriteException. Doing so would require a custom PHPLIB exception class that contained a PHPLIB write result (which does include inserted IDs).

tests/Collection/CrudSpecFunctionalTest.php
```diff
@@ -290,6 +350,7 @@ private function executeAssertResult(array $operation, $expectedResult, $actualR
 
             case 'deleteMany':
             case 'deleteOne':
+                $this->assertInternalType('array', $expectedResult);
```
jmikola: These are extra sanity checks, which appear in their own commit.

jmikola:
@derickr: This is ready for a final review.

tests/Collection/CrudSpecFunctionalTest.php
```diff
@@ -39,8 +43,16 @@ public function testSpecification(array $initialData, array $test, $minServerVer
...
 
-        $result = $this->executeOperation($test['operation']);
-        $this->executeOutcome($test['operation'], $test['outcome'], $result);
+        $result = null;
+        $exception = null;
+
+        try {
+            $result = $this->executeOperation($test['operation']);
+        } catch (RuntimeException $e) {
+            $exception = $e;
```
derickr: This assignment isn't strictly necessary. The variable will still exist outside the ``catch`` scope. You can just do ``catch (RuntimeException $exception)``.

jmikola: Noted. Just wanted to err on the side of clarity here, which is why I also initialized `$exception` above as `null`.

PR #586, opened 2018-10-25
PHPLIB-352: Only send bypassDocumentValidation if it's true

derickr:


src/Operation/Aggregate.php
```diff
@@ -299,7 +299,11 @@ private function createCommand(Server $server)
 
         $cmd['allowDiskUse'] = $this->options['allowDiskUse'];
 
-        if (isset($this->options['bypassDocumentValidation']) && \MongoDB\server_supports_feature($server, self::$wireVersionForDocumentLevelValidation)) {
+        if (
+            isset($this->options['bypassDocumentValidation']) &&
+            $this->options['bypassDocumentValidation'] &&
+            \MongoDB\server_supports_feature($server, self::$wireVersionForDocumentLevelValidation)
+        ) {
```
jmikola: Optional: refactor `isset($x) && $x` as `!empty($x)`. There's some prior art for this in the library (and in `Aggregate.php` in fact).

src/Operation/BulkWrite.php
```diff
@@ -324,7 +324,11 @@ public function execute(Server $server)
 
         $options = ['ordered' => $this->options['ordered']];
 
-        if (isset($this->options['bypassDocumentValidation']) && \MongoDB\server_supports_feature($server, self::$wireVersionForDocumentLevelValidation)) {
+        if (
+            isset($this->options['bypassDocumentValidation']) &&
+            $this->options['bypassDocumentValidation'] &&
+            \MongoDB\server_supports_feature($server, self::$wireVersionForDocumentLevelValidation)
+        ) {
```
jmikola: [CDRIVER-2685](https://jira.mongodb.org/browse/CDRIVER-2685) was addressed in 1.13, so I assume that applies to all `executeBulkWrite()` code paths (and is the reason we didn't have a corresponding PHPC ticket). That said, there's probably no harm in keeping this check here for consistency. :+1:

PR #588, opened 2018-10-30
PHPLIB-356: Audit documentation for possible BSON 64-bit integer return values

derickr:


src/MapReduceResult.php
```diff
@@ -71,7 +71,7 @@ public function getCounts()
      */
     public function getExecutionTimeMS()
     {
-        return $this->executionTimeMS;
+        return (integer) $this->executionTimeMS;
```
jmikola: Is this redundant in light of `$this->executionTimeMS = (integer) $result->timeMillis;` in the constructor? No objections if you want to keep this as a sanity check.

I realize we're at a disadvantage since we can't yet use type hints for class properties.

src/Model/CollectionInfo.php
```diff
@@ -63,6 +63,7 @@ public function __debugInfo()
      */
     public function getCappedMax()
     {
+        /* The MongoDB server might return this number as an integer or float */
```
jmikola: Can use `//` for single-line comments -- unless we are using `/*` in this fashion elsewhere.

derickr: I had used ``/*`` ... ``*/`` for PHPLIB-376 in src/Operation/CountDocuments.php too — couldn't really see other inline comments.

src/MapReduceResult.php
derickr: It's redundant, but makes it more the same with other places. I'm just going to keep it.

PR #591, opened 2018-11-13
PHPLIB-368: Operations executed within transaction should not inherit read/write concern

derickr:
https://jira.mongodb.org/browse/PHPLIB-368

### DIFF
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -551,7 +551,7 @@ class Collection
      * @throws InvalidArgumentException for parameter/option parsing errors
      * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
      */
-    public function EstimatedDocumentCount(array $options = [])
+    public function estimatedDocumentCount(array $options = [])
     {
         if ( ! isset($options['readPreference'])) {
             $options['readPreference'] = $this->readPreference;
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -203,10 +203,13 @@ class Collection
 
         /* A "majority" read concern is not compatible with the $out stage, so
          * avoid providing the Collection's read concern if it would conflict.
+         *
+         * A read concern is also not compatible with transactions.
          */
         if ( ! isset($options['readConcern']) &&
-             ! ($hasOutStage && $this->readConcern->getLevel() === ReadConcern::MAJORITY) &&
-            \MongoDB\server_supports_feature($server, self::$wireVersionForReadConcern)) {
+            ! ($hasOutStage && $this->readConcern->getLevel() === ReadConcern::MAJORITY) &&
+            \MongoDB\server_supports_feature($server, self::$wireVersionForReadConcern) &&
+            ! $this->isInTransaction($options)) {
             $options['readConcern'] = $this->readConcern;
         }
 
@@ -214,7 +217,10 @@ class Collection
             $options['typeMap'] = $this->typeMap;
         }
 
-        if ($hasOutStage && ! isset($options['writeConcern']) && \MongoDB\server_supports_feature($server, self::$wireVersionForWritableCommandWriteConcern)) {
+        if ($hasOutStage &&
+            ! isset($options['writeConcern']) &&
+            \MongoDB\server_supports_feature($server, self::$wireVersionForWritableCommandWriteConcern) &&
+            ! $this->isInTransaction($options)) {
             $options['writeConcern'] = $this->writeConcern;
         }
 
@@ -236,7 +242,7 @@ class Collection
      */
     public function bulkWrite(array $operations, array $options = [])
     {
-        if ( ! isset($options['writeConcern'])) {
+        if ( ! isset($options['writeConcern']) && ! $this->isInTransaction($options)) {
             $options['writeConcern'] = $this->writeConcern;
         }
 
@@ -268,7 +274,7 @@ class Collection
 
         $server = $this->manager->selectServer($options['readPreference']);
 
-        if ( ! isset($options['readConcern']) && \MongoDB\server_supports_feature($server, self::$wireVersionForReadConcern)) {
+        if ( ! isset($options['readConcern']) && \MongoDB\server_supports_feature($server, self::$wireVersionForReadConcern) && ! $this->isInTransaction($options)) {
             $options['readConcern'] = $this->readConcern;
         }
 
@@ -297,7 +303,7 @@ class Collection
 
         $server = $this->manager->selectServer($options['readPreference']);
 
-        if ( ! isset($options['readConcern']) && \MongoDB\server_supports_feature($server, self::$wireVersionForReadConcern)) {
+        if ( ! isset($options['readConcern']) && \MongoDB\server_supports_feature($server, self::$wireVersionForReadConcern) && ! $this->isInTransaction($options)) {
             $options['readConcern'] = $this->readConcern;
         }
 
@@ -359,7 +365,7 @@ class Collection
     {
         $server = $this->manager->selectServer(new ReadPreference(ReadPreference::RP_PRIMARY));
 
-        if ( ! isset($options['writeConcern']) && \MongoDB\server_supports_feature($server, self::$wireVersionForWritableCommandWriteConcern)) {
+        if ( ! isset($options['writeConcern']) && \MongoDB\server_supports_feature($server, self::$wireVersionForWritableCommandWriteConcern) && ! $this->isInTransaction($options)) {
             $options['writeConcern'] = $this->writeConcern;
         }
 
@@ -382,7 +388,7 @@ class Collection
      */
     public function deleteMany($filter, array $options = [])
     {
-        if ( ! isset($options['writeConcern'])) {
+        if ( ! isset($options['writeConcern']) && ! $this->isInTransaction($options)) {
             $options['writeConcern'] = $this->writeConcern;
         }
 
@@ -406,7 +412,7 @@ class Collection
      */
     public function deleteOne($filter, array $options = [])
     {
-        if ( ! isset($options['writeConcern'])) {
+        if ( ! isset($options['writeConcern']) && ! $this->isInTransaction($options)) {
             $options['writeConcern'] = $this->writeConcern;
         }
 
@@ -437,7 +443,7 @@ class Collection
 
         $server = $this->manager->selectServer($options['readPreference']);
 
-        if ( ! isset($options['readConcern']) && \MongoDB\server_supports_feature($server, self::$wireVersionForReadConcern)) {
+        if ( ! isset($options['readConcern']) && \MongoDB\server_supports_feature($server, self::$wireVersionForReadConcern) && ! $this->isInTransaction($options)) {
             $options['readConcern'] = $this->readConcern;
         }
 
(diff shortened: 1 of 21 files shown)


src/Collection.php
```diff
@@ -882,6 +882,15 @@ public function insertMany(array $documents, array $options = [])
      */
     public function insertOne($document, array $options = [])
     {
+        if (isset($options['session']) && $options['session']->isInTransaction()) {
```
jmikola: This may need an `instanceof` check between `isset()` and calling `isInTransaction()`.

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -311,6 +311,68 @@ public function testMapReduce()
         $this->assertNotEmpty($result->getCounts());
     }
 
+    /**
+     * @expectedException MongoDB\Driver\Exception\InvalidArgumentException
+     * @expectedExceptionMessage inherited
+     */
+    public function testInsertOneInTransactionWithInheritedWriteConcern()
```
jmikola: This test can change if the Collection avoids inheriting WC if it knows it would conflict with the current state of the session when `insertOne()` is called.

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -311,6 +311,68 @@ public function testMapReduce()
...
+            ['test' => 'foo'],
+            ['session' => $session]
+        );
+
+        $this->dropCollection();
+    }
+
+    /**
+     * @expectedException MongoDB\Driver\Exception\InvalidArgumentException
+     * @expectedExceptionMessage option
```
jmikola: We can still use a substring here, but it should probably be something more specific in the final test. For instance, if the final message is `The "readConcern" option cannot be specified within a transaction. Instead, specify it when starting the transaction.` we might just go with `"readConcern" option cannot be specified within a transaction`.

tests/Collection/CollectionFunctionalTest.php
derickr: Instead of changing it, I've removed it. I suppose I could replace it with a test that uses Command Observer? ``libmongoc`` doesn't seem to throw an error when we do this thing.

jmikola: I don't believe libmongoc enforces this. It may just be done server side.

An APM test to ensure we don't inherit a non-default write concern would probably be prudent.

src/Collection.php
```diff
@@ -882,6 +882,15 @@ public function insertMany(array $documents, array $options = [])
...
+            if ($this->writeConcern instanceof WriteConcern && !$this->writeConcern->isDefault()) {
+                throw new InvalidArgumentException("Cannot use an inherited writeConcern during a transaction");
+            }
+            if (isset($options['writeConcern']) && !$options['writeConcern']->isDefault()) {
+                throw new InvalidArgumentException("Cannot specify 'writeConcern' option during a transaction");
```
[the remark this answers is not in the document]
jmikola: My understanding is that only `insertOne()` needs the "silently ignore" logic. If it sees a session option and that session is in a transaction, it should not inherit the Collection's RC/WC. Suppose we have the following in the Collection helper:

```php
$inTransaction = isset($options['session']) && $options['session'] instanceof Session && $options['session']->isInTransaction();

if ( ! isset($options['writeConcern']) && ! $inTransaction) {
    $options['writeConcern'] = $this->writeConcern;
}
```

Later, in the Operation's `execute()` method:

```php
// Constructor would already have validated the session's type (if set) by this point
$inTransaction = isset($options['session']) && $options['session']->isInTransaction();

// Constructor would already have unset a default WC by this point
if (isset($this->options['writeConcern']) && $inTransaction) {
    throw UnsupportedException::writeConcernNotSupportedInTransaction();
}
```

Under normal operation, the session's state should not change between the Collection helper deciding it's safe to inherit and `execute()` being called, since that all happens within the helper method.

I can think of two edge cases that would still allow for an error. Once SDAM monitoring is implemented, it's possible that a listener method could be invoked when the helper calls `Manager::selectServer()` (between inheritance logic and `execute()`). The session could also be altered during the helper method's execution via a thread. These cases would either require to be doing something quite strange or acting in bad faith, so I don't think we need to worry about it.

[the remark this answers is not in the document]
derickr: Changed.

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -311,6 +311,71 @@ public function testMapReduce()
         $this->assertNotEmpty($result->getCounts());
     }
 
+    public function testDontInheritWriteConcernInTransaction()
```
jmikola:
```suggestion
    public function testInsertOneDoesNotInheritWriteConcernInTransaction()
```

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -311,6 +311,71 @@ public function testMapReduce()
...
+                        'session' => $session,
+                    ]
+                );
+            },
+            function(array $event) {
+                $this->assertObjectNotHasAttribute('writeConcern', $event['started']->getCommand());
+            }
+        );
+
+        $this->dropCollection();
```
jmikola: IIRC, we only need to explicitly _create_ collections for transaction tests. `MongoDB\Tests\Collection::tearDown()` already drops the collection under test iff the test has not failed (otherwise it leaves it in place for inspection), so this can be removed.

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -311,6 +311,71 @@ public function testMapReduce()
...
+            [ 'writeConcern' => new \MongoDB\Driver\WriteConcern(1) ]
+        );
+
+        (new CommandObserver)->observe(
+            function() use ($collection, $session) {
+                $collection->insertOne(
+                    ['test' => 'foo'],
+                    [
+                        'session' => $session,
+                    ]
```
jmikola: The options seem small enough to rewrite concisely as `['session' => $session]`.

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -311,6 +311,71 @@ public function testMapReduce()
...
+
+        $this->createCollection();
+
+        $session = $this->manager->startSession();
+        $session->startTransaction([
+            'readConcern' => new \MongoDB\Driver\ReadConcern("snapshot"),
+            'writeConcern' => new \MongoDB\Driver\WriteConcern(\MongoDB\Driver\WriteConcern::MAJORITY)
+        ]);
+
+        $collection = new Collection(
```
jmikola: Assuming you're going to do this frequently in these tests, this may be more concise:

```
$collection = $this->collection->withOptions([
    'writeConcern' => new \MongoDB\Driver\WriteConcern(1),
]);
```

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -311,6 +311,71 @@ public function testMapReduce()
...
+    {
+        $this->skipIfTransactionsAreNotSupported();
+
+        $this->createCollection();
+
+        $session = $this->manager->startSession();
+        $session->startTransaction([
+            'readConcern' => new \MongoDB\Driver\ReadConcern("snapshot"),
+            'writeConcern' => new \MongoDB\Driver\WriteConcern(\MongoDB\Driver\WriteConcern::MAJORITY)
+        ]);
```
jmikola: Are either of these options relevant? AFAICT, the logic prohibits per-operation RC/WC regardless of whether the transaction specified those options on its own, so `$session->startTransaction()` would be more concise.

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -311,6 +311,71 @@ public function testMapReduce()
...
+
+        $collection = new Collection($this->manager, $this->getDatabaseName(), $this->getCollectionName());
```
jmikola: If Collection options aren't changing, we can just use `$this->collection` (as in other tests within this file).

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -311,6 +311,71 @@ public function testMapReduce()
...
+
+        $collection->insertOne(
+            ['test' => 'foo'],
+            [
+                'session' => $session,
+                'writeConcern' => new \MongoDB\Driver\WriteConcern(1)
+            ]
+        );
+
+        $this->dropCollection();
```
jmikola: Can also be removed thanks to `tearDown()`.

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -311,6 +311,71 @@ public function testMapReduce()
...
+                    ]
+                );
+            },
+            function(array $event) {
+                $this->assertObjectNotHasAttribute('writeConcern', $event['started']->getCommand());
+            }
+        );
+
+        $this->dropCollection();
+    }
```
jmikola: Would it be prudent to call `$session->endSession();` at the end of these functions? I recall that coming up when I first added the aggregate/find transaction tests. See https://github.com/mongodb/mongo-php-library/pull/564#discussion_r202395078 for more context.

This is particularly necessary for tests that throw, since the exception's stack trace ends up holding a reference to the session variable, which prevents it from being closed and terminating the transaction. If we don't clean up the session, I believe future test methods could block unnecessarily.

A `try { ... } finally { $session->endSession(); }` should suffice for now. Long term, I think we can come up with a better solution for this. I'm experimenting with something in my work for SPEC-1133 to track _all_ collections used during a test (e.g. output collections) for cleanup during `tearDown()`, instead of only having `tearDown()` drop the single collection under test. If that works out, I can see us doing something similar to track sessions used during a test to ensure that `tearDown()` calls `endSession()` for us.

src/Collection.php
```diff
@@ -78,6 +78,20 @@ class Collection
     private $typeMap;
     private $writeConcern;
 
+    /**
+     * Returns whether we are currently in a transaction
+     *
+     * @param array  $options  Command options
+     * @return bool
+     */
+    private function isInTransaction(array $options)
```
jmikola: PedantryTest may complain about this, since it checks that private methods appear after all public and protected methods. Consider moving below `public function withOptions()`.

tests/Collection/CollectionFunctionalTest.php
jmikola: Created [PHPLIB-399](https://jira.mongodb.org/browse/PHPLIB-399) to record the idea about tracking sessions.

src/Collection.php
```diff
@@ -203,18 +203,25 @@ public function aggregate(array $pipeline, array $options = [])
...
 
         if ( ! isset($options['typeMap'])) {
             $options['typeMap'] = $this->typeMap;
         }
 
-        if ($hasOutStage && ! isset($options['writeConcern']) && \MongoDB\server_supports_feature($server, self::$wireVersionForWritableCommandWriteConcern)) {
+        if ($hasOutStage &&
+            ! isset($options['writeConcern']) &&
+            \MongoDB\server_supports_feature($server,
+                self::$wireVersionForWritableCommandWriteConcern) &&
```
jmikola: Does this function call need to be split across two lines? It should fall within 120 characters easily.

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -311,6 +311,353 @@ public function testMapReduce()
...
+            [
+                function($collection, $session, $options = []) {
+                    $collection->dropIndexes(
+                        ['session' => $session] + $options
+                    );
+                }, 'w'
+            ],
+            */
+
+            /* Disabled, as count() can't be used in transactions
```
jmikola: I assume this will ultimately be deleted, but if not the comment should probably refer to `count` command for `estimatedDocumentCount()` method.

PR #592, opened 2018-11-15
PHPLIB-398: Aggregate with $out should ignore batchSize option

jmikola:
https://jira.mongodb.org/browse/PHPLIB-398

### DIFF
--- a/src/Operation/Aggregate.php
+++ b/src/Operation/Aggregate.php
@@ -255,7 +255,7 @@ class Aggregate implements Executable
         $hasExplain = ! empty($this->options['explain']);
         $hasOutStage = \MongoDB\is_last_pipeline_operator_out($this->pipeline);
 
-        $command = $this->createCommand($server);
+        $command = $this->createCommand($server, $hasOutStage);
         $options = $this->createOptions($hasOutStage, $hasExplain);
 
         $cursor = ($hasOutStage && ! $hasExplain)
@@ -287,9 +287,10 @@ class Aggregate implements Executable
      * Create the aggregate command.
      *
      * @param Server  $server
+     * @param boolean $hasOutStage
      * @return Command
      */
-    private function createCommand(Server $server)
+    private function createCommand(Server $server, $hasOutStage)
     {
         $cmd = [
             'aggregate' => isset($this->collectionName) ? $this->collectionName : 1,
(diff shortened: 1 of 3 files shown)


tests/Collection/CrudSpecFunctionalTest.php
```diff
@@ -81,7 +86,7 @@ public function provideSpecificationTests()
      */
     private function assertEquivalentCollections($expectedCollection, $actualCollection)
     {
-        $mi = new MultipleIterator;
+        $mi = new MultipleIterator(MultipleIterator::MIT_NEED_ANY);
```
jmikola: I didn't realize when originally writing the test that [MultipleIterator](http://php.net/MultipleIterator)'s default behavior is to stop when any iterator stops yielding results. This caused us to not actually compare the actual and expected output collections because the actual collection had no documents (due to the pipeline never executing).

Note: the GridFS spec tests also use MultipleIterator for a similar assertion and would benefit from using `MIT_NEED_ANY`, but I'll take care of that in a separate PR.

tests/Operation/AggregateFunctionalTest.php
```diff
@@ -14,6 +14,25 @@
...
 class AggregateFunctionalTest extends FunctionalTestCase
 {
+    public function testBatchSizeIsIgnoredIfPipelineIncludesOutStage()
+    {
+        (new CommandObserver)->observe(
+            function() {
+                $operation = new Aggregate(
+                    $this->getDatabaseName(),
+                    $this->getCollectionName(),
+                    [['$out' => $this->getCollectionName() . '.output']],
```
derickr: Shouldn't this new collection be drop'ed in the tearDown?

jmikola: Added a drop later in the test, along with another test method that was leaving behind an output collection.

src/Operation/Aggregate.php
```diff
@@ -287,9 +287,10 @@ public function execute(Server $server)
      * Create the aggregate command.
      *
      * @param Server  $server
+     * @param boolean $hasOutStage
      * @return Command
      */
-    private function createCommand(Server $server)
+    private function createCommand(Server $server, $hasOutStage)
```
derickr: Instead of passing in ``$hasOutStage`` as an argument, would not be nicer to have this method run ``$hasOutStage = \MongoDB\is_last_pipeline_operator_out($this->pipeline);`` itself? I realise it's a ``private`` function, but IMO passing around unnecessary arguments is bad form.

PR #596, opened 2018-11-30
PHPLIB-364: Document that transactions do not allow per-op read/write concerns

derickr:


docs/includes/apiargs-MongoDBCollection-common-option.yaml
```diff
@@ -42,6 +42,12 @@ description: |
 
    This is not supported for server versions prior to 3.2 and will result in an
    exception at execution time if used.
+
+   It is not possible to specify a :manual:`read concern
+   </reference/read-concern>` for individual operations as part of a
+   transaction. Instead, set the :manual:`read concern
```
jmikola: We've already linked to the read concern manual entry, so I think this bit can read "set the ``readConcern`` option" so that we refer to the actual option name for `startTransaction()`.

docs/includes/apiargs-MongoDBCollection-common-option.yaml
```diff
@@ -68,6 +74,12 @@ type: :php:`MongoDB\\Driver\\WriteConcern <class.mongodb-driver-writeconcern>`
 description: |
    :manual:`Write concern </reference/write-concern>` to use for the operation.
    Defaults to the collection's write concern.
+
+   It is not possible to specify a :manual:`write concern
+   </reference/write-concern>` for individual operations as part of a
+   transaction. Instead, set the :manual:`write concern
```
jmikola: Likewise, ``writeConcern`` to refer to option name.

PR #599, opened 2018-12-06
[WIP] PHPLIB-389: Add withOptions helper to Client

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-389

Allows fetching a new client instance with changed `readConcern`, `readPreference`, `writeConcern` and `typeMap` options. If only the `typeMap` changes, the original manager is reused. Changing any of the `readConcern`, `readPreference`, `writeConcern` options causes instantiation of a new manager to ensure consistency between the managers' options and those stored in the client.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -42,10 +42,12 @@ class Client
     private static $wireVersionForReadConcern = 4;
     private static $wireVersionForWritableCommandWriteConcern = 5;
 
+    private $driverOptions;
     private $manager;
     private $readConcern;
     private $readPreference;
     private $uri;
+    private $uriOptions;
     private $typeMap;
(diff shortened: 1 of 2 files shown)


jmikola:
Will get to this closer to the 1.5.0 release. As previously discussed off-issue, this needs some improvements to ensure we can recreate a URI from RC, RP, and WC options.

alcaeus:
Closing the PR as the implementation is insufficient. This requires additional logic to make sure the driver does not unnecessarily create new internal connections, which IMO is not justified considering that the original use-case (changing the typemap on an already created client) can easily be worked around by passing a typemap when fetching a `Collection` or `Database` instance.

PR #610, opened 2019-05-09
PHPLIB-307: Retryable writes test runner

jmikola:
https://jira.mongodb.org/browse/PHPLIB-307

This depends on WriteResult accessibility for bulk write errors ([PHPC-1373](https://jira.mongodb.org/browse/PHPC-1373)) and `retryWrites=true` by default ([PHPC-1324](https://jira.mongodb.org/browse/PHPC-1324)); however, the test runner can work around that for the time being and we can revisit those with TODO items after bumping to `ext-mongodb` 1.6.0.

### DIFF
--- a/tests/SpecTests/RetryableWritesSpecTest.php
+++ b/tests/SpecTests/RetryableWritesSpecTest.php
@@ -1,0 +1,65 @@
+<?php
+
+namespace MongoDB\Tests\SpecTests;
+
+/**
+ * Retryable writes spec tests.
+ *
+ * @see https://github.com/mongodb/specifications/tree/master/source/retryable-writes
+ */
+class RetryableWritesSpecTest extends FunctionalTestCase
+{
+    /**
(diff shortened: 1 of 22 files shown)


tests/SpecTests/RetryableWritesSpecTest.php
```diff
@@ -0,0 +1,69 @@
...
+     * @param array  $test  Individual "tests[]" document
+     * @param array  $runOn Top-level "runOn" document
+     * @param array  $data  Top-level "data" array to initialize collection
+     */
+    public function testRetryableWrites($name, array $test, array $runOn = null, array $data)
+    {
+        $this->setName($name);
+
+        if (isset($runOn)) {
+            $this->checkServerRequirements(
```
jmikola: This is incorrectly implemented, since `runOn` is an array of requirements and we should only skip if none are satisfied. Will revise in a subsequent commit.

PR #614, opened 2019-05-29
PHPLIB-365: Transactions test runner

jmikola:
https://jira.mongodb.org/browse/PHPLIB-365

### DIFF
--- a/tests/SpecTests/TransactionsSpecTest.php
+++ b/tests/SpecTests/TransactionsSpecTest.php
@@ -1,0 +1,264 @@
+<?php
+
+namespace MongoDB\Tests\SpecTests;
+
+use MongoDB\BSON\Int64;
+use MongoDB\BSON\Timestamp;
+use MongoDB\Driver\Command;
+use MongoDB\Driver\Manager;
+use MongoDB\Driver\ReadConcern;
+use MongoDB\Driver\ReadPreference;
+use MongoDB\Driver\Server;
+use MongoDB\Driver\WriteConcern;
+use MongoDB\Driver\Exception\ServerException;
+use stdClass;
+
+/**
+ * Transactions spec tests.
+ *
+ * @see https://github.com/mongodb/specifications/tree/master/source/transactions
+ */
+class TransactionsSpecTest extends FunctionalTestCase
+{
+    const INTERRUPTED = 11601;
+
+    /* In addition to the useMultipleMongoses tests, these should all pass
+     * before the driver can be considered compatible with MongoDB 4.2. */
+    private static $incompleteTests = [
+        'error-labels: add unknown commit label to MaxTimeMSExpired' => 'PHPC-1382',
+        'error-labels: add unknown commit label to writeConcernError MaxTimeMSExpired' => 'PHPC-1382',
+        'read-concern: only first countDocuments includes readConcern' => 'PHPLIB-417',
+        'read-concern: countDocuments ignores collection readConcern' => 'PHPLIB-417',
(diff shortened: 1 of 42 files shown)


jmikola:
@kvwalker: Tests are not all passing, but feel free to take a look at the test runner implementation if you have the time. In particular, I'd appreciate a sanity check that I'm covering the steps in https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst#use-as-integration-tests.

tests/SpecTests/TransactionsSpecTest.php
```diff
@@ -0,0 +1,236 @@
...
+            $this->markTestSkipped($test['skipReason']);
+        }
+
+        $databaseName = isset($databaseName) ? $databaseName : $this->getDatabaseName();
+        $collectionName = isset($collectionName) ? $collectionName : $this->getCollectionName();
+
+        $context = Context::fromTransactions($test, $databaseName, $collectionName);
+        $this->setContext($context);
+
+        $this->dropTestAndOutcomeCollections();
```
kvwalker: According to the spec, it looks like the following should occur:
```
(3) Create a collection object from the MongoClient,
using the database_name and collection_name fields of the YAML file.

(4) Drop the test collection, using writeConcern "majority".

(5) Execute the "create" command to recreate the collection, using writeConcern "majority".
(Creating the collection inside a transaction is prohibited, so create it explicitly.)
```

It looks like this covers steps 4 and 5, but where is step 3 done?

tests/SpecTests/TransactionsSpecTest.php
```diff
@@ -0,0 +1,236 @@
...
+        $context->session0->endSession();
+        $context->session1->endSession();
+
+        if (isset($commandExpectations)) {
+            $commandExpectations->stopMonitoring();
+            $commandExpectations->assert($this, $context);
+        }
+
+        if (isset($test['outcome']['collection']['data'])) {
+            $this->assertOutcomeCollectionData($test['outcome']['collection']['data']);
```
kvwalker: Does this fulfill "Ensure this find reads the latest data by using primary read preference with local read concern even when the MongoClient is configured with another read preference or read concern."?

tests/SpecTests/TransactionsSpecTest.php
jmikola: `dropTestAndOutcomeCollections()` internally creates a collection by calling `Context::getCollection()`. This is identical to how we create a collection for each operation, with the notable difference that the operation may have options to provide.

Since collection _and_ database options vary by each operation, I saw no reason to create the objects up front as we do with the Client and both Session objects.

Later on in the transactions spec README, we have:

> **11.4**: Create a Collection object from the Database, using the collection_name field at the top level of the test file. If collectionOptions is present create the Collection object with the provided options. Otherwise create the object with the default options.

jmikola:
Revised the test suite to ensure we preserve array/object types when parsing JSON spec test files. This ensures that our assertions (e.g. APM command documents) are more accurate.

Also added a list of tests we expect to fail due to unimplemented functionality in the PHP driver and/or libmongoc. This will at least ensure we can merge this without introducing failures to the overall test suite.

jmikola:
Note: this may require a new alpha release of PHPC to ensure that error labels are available on cursor exceptions ([PHPC-1387](https://jira.mongodb.org/browse/PHPC-1387)). Alternatively, I can skip the affected tests as was done in https://github.com/mongodb/mongo-php-library/pull/614/commits/5d06103314d477ca4b1b9ffa6a7028b439fc00fc.

PR #615, opened 2019-06-05
PHPLIB-410: Command Monitoring test runner

jmikola:
https://jira.mongodb.org/browse/PHPLIB-410

Spec tests synced with mongodb/specifications@4820c7bad9b06b71f169c0d4b99aba6fbb319243

### DIFF
--- a/tests/SpecTests/CommandExpectations.php
+++ b/tests/SpecTests/CommandExpectations.php
@@ -2,7 +2,6 @@
 
 namespace MongoDB\Tests\SpecTests;
 
-use MongoDB\BSON\Timestamp;
 use MongoDB\Driver\Monitoring\CommandFailedEvent;
 use MongoDB\Driver\Monitoring\CommandStartedEvent;
 use MongoDB\Driver\Monitoring\CommandSucceededEvent;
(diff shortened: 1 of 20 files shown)


tests/SpecTests/CommandExpectations.php
```diff
@@ -2,7 +2,6 @@
 
 namespace MongoDB\Tests\SpecTests;
 
-use MongoDB\BSON\Timestamp;
```
jmikola: This was leftover from some earlier wip commit in a previous PR when the command document assertion was done within this class. That logic was ultimately moved to `TransactionsSpecTest::assertCommandMatches()`.

tests/SpecTests/TransactionsSpecTest.php
```diff
@@ -56,89 +56,62 @@ public function tearDown()
...
     }
 
-    /**
-     * Kill all sessions on the cluster.
-     *
-     * This will clean up any open transactions that may remain from a
-     * previously failed test. For sharded clusters, this command will be run
-     * on all mongos nodes.
-     */
-    private static function killAllSessions()
```
jmikola: This was moved to the bottom of the file, to be consistent with ordering of public, protected, and private members.

tests/SpecTests/TransactionsSpecTest.php
```diff
@@ -56,89 +56,62 @@ public function tearDown()
...
-     * Note: this method may modify the $expectedCommand object.
+     * Note: this method may modify the $expected object.
      *
-     * @param stdClass $expectedCommand Expected command document
-     * @param stdClass $actualCommand   Actual command document
+     * @param stdClass $expected Expected command document
+     * @param stdClass $actual   Actual command document
      */
-    public function assertSameCommand(stdClass $expectedCommand, stdClass $actualCommand)
+    public static function assertCommandMatches(stdClass $expected, stdClass $actual)
```
jmikola: PHPUnit tends to prefer `static` methods for assertions when possible. This wasn't necessary for our own needs, but I didn't see the harm in following convention. I haven't given this much thought in the past because I was seldom digging through PHPUnit's source; however, this came up while I was crafting our custom Constraint class.

tests/SpecTests/DocumentsMatchConstraint.php
```diff
@@ -0,0 +1,164 @@
...
+     * to make document comparisons more efficient, we may consider supporting
+     * this option. */
+    private $sortKeys = false;
+    private $value;
+
+    /**
+     * Creates a new constraint.
+     *
+     * @param array|object $value
+     * @param boolean      $ignoreExtraKeys If true, ignore extra keys within actual value's document(s)
```
jmikola: This may need to be split into separate flags for ignoring top-level keys alone and extra keys throughout all sub-documents. Per my comment in https://github.com/mongodb/specifications/pull/560#pullrequestreview-246810295:

> In the Command Monitoring spec test README, [Additional Values](https://github.com/mongodb/specifications/blob/master/source/command-monitoring/tests/README.rst#additional-values) explicitly states that the allowance for extra fields only pertains to top-level fields, and an example is provided for a failed comparison due to an extra field within a sub-document.
>
> This is in contrast to the [Spec Test Match Function](https://github.com/mongodb/specifications/blob/master/source/change-streams/tests/README.rst#spec-test-match-function) described in the Change Streams spec test README, which I believe some drivers use as a general matching algorithm for documents in other spec's tests.

jmikola: Done in subsequent commit.

tests/SpecTests/Operation.php
```diff
@@ -43,6 +44,17 @@ private function __construct(stdClass $operation)
         }
     }
 
+    public static function fromCommandMonitoring(stdClass $operation)
+    {
+        $o = new self($operation);
+
+        if (isset($operation->collectionOptions)) {
+            $o->collectionOptions = (array) $operation->collectionOptions;
+        }
```
jmikola: This was originally missing. After fixing the extra keys logic in DocumentsMatchConstraint, the test in `unacknowledgedBulkWrite.json` began failing because the test runner never applied the write concern in `collectionOptions`. It's now handled properly.

PR #616, opened 2019-06-11
PHPLIB-366: Change streams test runner

jmikola:
Depends on https://github.com/mongodb/mongo-php-library/pull/615. Can be rebased after that is merged.

https://jira.mongodb.org/browse/PHPLIB-366

### DIFF
--- a/tests/SpecTests/ChangeStreamsSpecTest.php
+++ b/tests/SpecTests/ChangeStreamsSpecTest.php
@@ -1,0 +1,266 @@
+<?php
+
+namespace MongoDB\Tests\SpecTests;
+
+use MongoDB\ChangeStream;
+use MongoDB\Client;
+use MongoDB\Driver\Exception\Exception;
+use MongoDB\Model\BSONDocument;
+use ArrayIterator;
+use LogicException;
+use MultipleIterator;
+use stdClass;
+
+/**
+ * Change Streams spec tests.
+ *
+ * @see https://github.com/mongodb/specifications/tree/master/source/change-streams
+ */
+class ChangeStreamsSpecTest extends FunctionalTestCase
+{
+    /* These should all pass before the driver can be considered compatible with
+     * MongoDB 4.2. */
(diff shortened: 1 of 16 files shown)


jmikola:
Rebased on master after merging #615. This should still wait until https://github.com/mongodb/specifications/pull/567 is merged, since it requires a test fix.

tests/SpecTests/ChangeStreamsSpecTest.php
```diff
@@ -0,0 +1,240 @@
...
+ * Change Streams spec tests.
+ *
+ * @see https://github.com/mongodb/specifications/tree/master/source/change-streams
+ */
+class ChangeStreamsSpecTest extends FunctionalTestCase
+{
+    /* These should all pass before the driver can be considered compatible with
+     * MongoDB 4.2. */
+    private static $incompleteTests = [
+        'change-streams: Test consecutive resume' => 'PHPLIB-416',
```
jmikola: There is actually no reason for this to fail. I believe the problem is that I'm using `iterator_to_array()` to capture the change stream result rather than the [documented iteration procedure](https://docs.mongodb.com/php-library/current/reference/method/MongoDBCollection-watch/#examples).

Will attempt to get this fixed and passing before we merge.

jmikola:
@kvwalker: This is ready for another look. While implementing this, I came across two bugs that appear to date back to the original implementation in 1.3.0:

 * https://jira.mongodb.org/browse/PHPLIB-443
 * https://jira.mongodb.org/browse/PHPLIB-442

I'd like to fix these in the active release branch (i.e. 1.4.x) and then merge up to master.

tests/SpecTests/ChangeStreamsSpecTest.php
jmikola: Just to confirm, this was actually one of the two bugs cited below: [PHPLIB-442](https://jira.mongodb.org/browse/PHPLIB-442)

PR #617, opened 2019-06-12
PHPLIB-429: Remove support for PHP 5.5

jmikola:
https://jira.mongodb.org/browse/PHPLIB-429

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,13 +10,13 @@
         { "name": "Katherine Walker", "email": "katherine.walker@mongodb.com" }
     ],
     "require": {
-        "php": ">=5.5",
+        "php": "^5.6 || ^7.0",
         "ext-hash": "*",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -10,13 +10,13 @@
         { "name": "Katherine Walker", "email": "katherine.walker@mongodb.com" }
     ],
     "require": {
-        "php": ">=5.5",
+        "php": ">=5.6",
```
alcaeus: Would it make sense to use ^5.6 || ^7.0?

jmikola: I didn't want to discriminate against any PHP 6 users.

My only concern is that if PHP 8.0 releases start popping up (e.g. alphas), we'd need to publish a patch release to add `|| ^8.0`. I suppose that's easily fixed when the issue arises, though.

PR #621, opened 2019-06-19
 PHPLIB-442: Ensure change stream resume token is updated after resume

jmikola:
https://jira.mongodb.org/browse/PHPLIB-442

This extracts common logic in next() and rewind() to a new method, which is now also called by resume() after it rewinds the internal iterator.

PR #622, opened 2019-06-19
PHPLIB-445: Leave key as-is if ChangeStream::next() fails to extract resume token

jmikola:
https://jira.mongodb.org/browse/PHPLIB-445

Note: this will conflict with https://github.com/mongodb/mongo-php-library/pull/621, so it'd be preferable to resolve that first and then rebase this after merging up to master.

### DIFF
--- a/src/ChangeStream.php
+++ b/src/ChangeStream.php
@@ -212,6 +212,8 @@ class ChangeStream implements Iterator
             return;
         }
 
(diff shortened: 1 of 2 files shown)


jmikola:
Rebased after merging https://github.com/mongodb/mongo-php-library/pull/621 into `v1.4` and up to `master`.

PR #626, opened 2019-06-21
PHPLIB-441: Increment ChangeStream key consistently when resuming

jmikola:
https://jira.mongodb.org/browse/PHPLIB-441

### DIFF
--- a/src/ChangeStream.php
+++ b/src/ChangeStream.php
@@ -50,6 +50,11 @@ class ChangeStream implements Iterator
     private $resumeCallable;
     private $csIt;
     private $key = 0;
+
+    /**
+     * Whether the change stream has advanced to its first result. This is used
+     * to determine whether $key should be incremented after an iteration event.
+     */
     private $hasAdvanced = false;
 
     /**
@@ -102,7 +107,7 @@ class ChangeStream implements Iterator
     {
         try {
             $this->csIt->next();
-            $this->onIteration(true);
+            $this->onIteration($this->hasAdvanced);
         } catch (RuntimeException $e) {
             if ($this->isResumableError($e)) {
                 $this->resume();
@@ -118,6 +123,9 @@ class ChangeStream implements Iterator
     {
         try {
             $this->csIt->rewind();
+            /* Unlike next() and resume(), the decision to increment the key
+             * does not depend on whether the change stream has advanced. This
+             * ensures that multiple calls to rewind() do not alter state. */
             $this->onIteration(false);
         } catch (RuntimeException $e) {
             if ($this->isResumableError($e)) {
@@ -193,12 +201,12 @@ class ChangeStream implements Iterator
     }
 
     /**
-     * Perform housekeeping after an iteration event (i.e. next or rewind).
+     * Perform housekeeping after an iteration event.
      *
-     * @param boolean $isNext Whether the iteration event was a call to next()
+     * @param boolean $incrementKey Increment $key if there is a current result
      * @throws ResumeTokenException
      */
-    private function onIteration($isNext)
+    private function onIteration($incrementKey)
     {
         /* If the cursorId is 0, the server has invalidated the cursor and we
          * will never perform another getMore nor need to resume since any
@@ -210,13 +218,13 @@ class ChangeStream implements Iterator
             $this->resumeCallable = null;
         }
 
+        /* Return early if there is not a current result. Avoid any attempt to
+         * increment the iterator's key or extract a resume token */
         if (!$this->valid()) {
             return;
         }
 
-        /* Increment the key if the iteration event was a call to next() and we
-         * have already advanced past the first result. */
-        if ($isNext && $this->hasAdvanced) {
+        if ($incrementKey) {
             $this->key++;
         }
 
@@ -234,6 +242,14 @@ class ChangeStream implements Iterator
         $newChangeStream = call_user_func($this->resumeCallable, $this->resumeToken);
         $this->csIt = $newChangeStream->csIt;
         $this->csIt->rewind();
-        $this->onIteration(false);
+        /* Note: if we are resuming after a call to ChangeStream::rewind(),
+         * $hasAdvanced will always be false. For it to be true, rewind() would
+         * need to have thrown a RuntimeException with a resumable error, which
+         * can only happen during the first call to IteratorIterator::rewind()
+         * before onIteration() has a chance to set $hasAdvanced to true.
+         * Otherwise, IteratorIterator::rewind() would either NOP (consecutive
+         * rewinds) or throw a LogicException (rewind after next), neither of
+         * which would result in a call to resume(). */
+        $this->onIteration($this->hasAdvanced);
     }
 }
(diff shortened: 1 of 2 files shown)


src/ChangeStream.php
```diff
@@ -214,9 +222,7 @@ private function onIteration($isNext)
             return;
         }
 
-        /* Increment the key if the iteration event was a call to next() and we
-         * have already advanced past the first result. */
-        if ($isNext && $this->hasAdvanced) {
+        if ($incrementKey) {
```
jmikola: I contemplated renaming this to `$incrementKeyIfValid`, but adding a comment above the early return for `!$this->valid()` seemed preferable.

src/ChangeStream.php
```diff
@@ -234,6 +242,13 @@ private function resume()
...
         $this->csIt->rewind();
-        $this->onIteration(false);
+        /* Note: if we are resuming after a call to rewind(), $hasAdvanced will
+         * always be false. If it were true, then the previous call to
+         * IteratorIterator::rewind() would either have been a NOP and succeeded
+         * because the cursor was still at its first position or have thrown a
+         * LogicException for attempting to rewind an advanced cursor. In the
+         * latter case, the exception would have bypassed the catch block for
+         * resuming. */
+        $this->onIteration($this->hasAdvanced);
```
jmikola: I'd appreciate a sanity check on this. While coming up with the current implementation in this PR, I ran the test suite with a hard assertion of `$this->hasAdvanced === false` placed before the call to `resume()` in `rewind()`'s `catch` block, and it was never triggered.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -225,6 +226,62 @@ private function assertStartAtOperationTime(TimestampInterface $expectedOperatio
         $this->assertEquals($expectedOperationTime, $command->pipeline[0]->{'$changeStream'}->startAtOperationTime);
     }
 
+    public function testRewindMultipleTimesWithResults()
```
jmikola: These two methods aren't related to resuming, so I included them in their own commit. This is the `rewind()` counterpart to the existing `testNextAdvancesKey` test method.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -322,20 +379,67 @@ public function testNoChangeAfterResumeBeforeInsert()
         $this->assertMatchesDocument($expectedResult, $changeStream->current());
     }
 
-    public function testResumeTokenIsUpdatedAfterResuming()
+    public function testResumeMultipleTimesInSuccession()
```
jmikola: This method was introduced in #621 for [PHPLIB-442](https://jira.mongodb.org/browse/PHPLIB-442) earlier this week, so I didn't feel too bad about renaming it. Refactoring it to _also_ check for key increments (in addition to the existing result assertions) seemed preferable to duplicating the entire test to only check key increments.

I also added some additional resume cases, which weren't in the original test.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -689,6 +795,8 @@ public function testNextAdvancesKey()
         $this->insertDocument(['x' => 1]);
         $this->insertDocument(['x' => 2]);
 
+        /* Note: we intentionally do not start iteration with rewind() to ensure
+         * that next() behaves identically when called without rewind(). */
```
jmikola: I happened to read through this test while writing the multiple rewind tests and thought this could use some clarity. This change is also in its own commit, since it doesn't pertain to resuming.

src/ChangeStream.php
kvwalker: I think the comment makes this clear.

src/ChangeStream.php
kvwalker: I'm not sure I follow this part:
```
If it were true, then the previous call to IteratorIterator::rewind() would either have been
a NOP and succeeded because the cursor was still at its first position
```
How can `$hasAdvanced` be true if the cursor was still at its first position?

tests/Operation/WatchFunctionalTest.php
```diff
@@ -322,20 +379,67 @@ public function testNoChangeAfterResumeBeforeInsert()
...
+    public function testResumeMultipleTimesInSuccession()
     {
-        $this->insertDocument(['_id' => 1]);
+        $operation = new CreateCollection($this->getDatabaseName(), $this->getCollectionName());
+        $operation->execute($this->getPrimaryServer());
 
         $operation = new Watch($this->manager, $this->getDatabaseName(), $this->getCollectionName(), [], $this->defaultOptions);
         $changeStream = $operation->execute($this->getPrimaryServer());
 
+        /* Killing the cursor when there are no results will tests that neither
```
kvwalker: nit: `tests` should be `test

src/ChangeStream.php
jmikola: My point here was that "if it were true" cannot happen in practice. Either the `rewind()` succeeds as a NOP and doesn't throw, so we never call `resume()` in the first place, or it throws a LogicException, which we don't catch (and also don't call `resume()`). I'll reword this a bit to better explain that, as I feel I only conveyed that the LogicException case is when we never reach this point.

> How can `$hasAdvanced` be true if the cursor was still at its first position?

If this was a more general question (apart from `resume()` being invoked), I can address that. When a cursor is first created (basic or tailable alike), it starts before the first element. `valid()` is `false` and there is no current element. The initial call to `rewind()` or `next()` is responsible for advancing to the first position (if any), updating `valid()`, and optionally populating the current element.

We use `$hasAdvanced` here to track if we've advanced before the initial pre-iteration state. If it's more helpful, I suppose we can rename this to `$hasIterationStarted`.

Another edge case is that it's kosher to call `rewind()` multiple times in sequence. And since `$key` is initialized to `0`, we should never increment the key after a successful call to `rewind()`, since, unlike `next()` or resuming, `rewind()` itself never moves past the first element.

kvwalker: Oh I see, thanks for the explanation! All of this makes sense.

jmikola: > I suppose we can rename this to `$hasIterationStarted`.

Ah, I forgot that I actually added a doc block to `$hasAdvanced` to explain exactly what it implies ([here](https://github.com/mongodb/mongo-php-library/pull/626/files#diff-d407b16b9acca873bbbaf0198a74fe57R54)). I think that's sufficient, as `$hasIterated` might also sound confusing since calls to `rewind()` and `next()` can still leave it `false` if the change stream doesn't come across a change event. I'll stick with the shorter name and just revise this comment block to be more clear.

jmikola: Note: this is a slight edge case where `rewind()` _could_ be called immediately after resuming without encountering a LogicException from the driver. Since resuming creates a fresh cursor and rewinds it, a user could technically call `ChangeStream::rewind()` after a resume and not see an error. This is very trivial and may not be worth fixing, but I opened [PHPLIB-448](https://jira.mongodb.org/browse/PHPLIB-448) to track it.

jmikola:
Test failure fixed in https://github.com/mongodb/mongo-php-library/pull/628.

PR #636, opened 2019-07-05
PHPLIB-451: ChangeStream::rewind() should never execute getMore

jmikola:
https://jira.mongodb.org/browse/PHPLIB-451
https://jira.mongodb.org/browse/PHPLIB-416

A few highlights from the commit messages:

 * Write ops in change stream tests now use `w:majority` to avoid spurious test failures due to timing.
 * Introduced an internal TailableCursorIterator class, which can be used to ensure that rewinding is a NOP (i.e. `getMore` will not be executed) for tailable cursors with an empty `firstBatch`.
 * Watch tests were revised to no longer expect `rewind()` to encounter an error and resume. In most tests, `rewind()` is a NOP because the change stream has no events to return upon creation.
 * Since Watch now always executes aggregate commands with APM, the check for `startAtOperationTime` support was moved to `createResumeCallable()`.
 * Various assertions for `current()` returning null were also changed to instead check `valid()`. Both forms are equivalent, but checking `valid()` is more consistent with our iteration examples.

### DIFF
--- a/src/Model/TailableCursorIterator.php
+++ b/src/Model/TailableCursorIterator.php
@@ -38,12 +38,12 @@ class TailableCursorIterator extends IteratorIterator
      *
      * @internal
      * @param Cursor  $cursor
-     * @param boolean $isFirstBatchIsEmpty
+     * @param boolean $isFirstBatchEmpty
      */
-    public function __construct(Cursor $cursor, $isFirstBatchIsEmpty)
+    public function __construct(Cursor $cursor, $isFirstBatchEmpty)
     {
         parent::__construct($cursor);
-        $this->isRewindNop = $isFirstBatchIsEmpty;
+        $this->isRewindNop = $isFirstBatchEmpty;
     }
 
     /**
--- a/src/Model/TailableCursorIterator.php
+++ b/src/Model/TailableCursorIterator.php
@@ -1,0 +1,80 @@
+<?php
+/*
+ * Copyright 2019 MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace MongoDB\Model;
+
+use MongoDB\Driver\Cursor;
+use IteratorIterator;
+
+/**
+ * Iterator for tailable cursors.
+ *
+ * This iterator may be used to wrap a tailable cursor. By indicating whether
+ * the cursor's first batch of results is empty, this iterator can NOP initial
+ * calls to rewind() and prevent it from executing a getMore command.
+ *
+ * @internal
+ */
+class TailableCursorIterator extends IteratorIterator
+{
+    private $isRewindNop;
+
+    /**
+     * Constructor.
+     *
+     * @internal
+     * @param Cursor  $cursor
+     * @param boolean $isFirstBatchIsEmpty
+     */
+    public function __construct(Cursor $cursor, $isFirstBatchIsEmpty)
+    {
+        parent::__construct($cursor);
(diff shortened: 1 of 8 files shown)


src/Model/TailableCursorIterator.php
```diff
@@ -0,0 +1,80 @@
...
+    private $isRewindNop;
+
+    /**
+     * Constructor.
+     *
+     * @internal
+     * @param Cursor  $cursor
+     * @param boolean $isFirstBatchIsEmpty
+     */
+    public function __construct(Cursor $cursor, $isFirstBatchIsEmpty)
```
kvwalker: `$isFirstBatchIsEmpty` is a little oddly phrased to me. Would `$isFirstBatchEmpty` make more sense?

src/Operation/Watch.php
```diff
@@ -186,6 +187,12 @@ public function __construct(Manager $manager, $databaseName, $collectionName, ar
     /** @internal */
     final public function commandStarted(CommandStartedEvent $event)
     {
+        if ($event->getCommandName() !== 'aggregate') {
+            return;
+        }
+
+        $this->operationTime = null;
```
kvwalker: Why do we need to set `operationTime` here?

src/Operation/Watch.php
```diff
@@ -241,40 +254,36 @@ private function createResumeCallable(Manager $manager)
...
      *
      * @param Server $server
      * @return Cursor
      */
     private function executeAggregate(Server $server)
     {
-        /* If we've already captured an operation time or the server does not
-         * support resuming from an operation time (e.g. MongoDB 3.6), execute
-         * the aggregation directly and return its cursor. */
-        if ($this->operationTime !== null || ! \MongoDB\server_supports_feature($server, self::$wireVersionForStartAtOperationTime)) {
```
kvwalker: Why was this removed?

src/Model/TailableCursorIterator.php
jmikola: Total brain fart. This was originally `firstBatchIsEmpty` and I must have forgotten to rephrase that when I switched to an `is` prefix. Will fix.

src/Operation/Watch.php
jmikola: The original logic was actually a bit backwards, but it worked in practice.

The server has reported operation times for as long as change streams have been supported (since 3.6); however, the `startAtOperationTime` option has only been supported since 4.0. Previously, `createResumeCallable` would decide to use `startAtOperationTime` purely by whether `$this->operationTime` was set, as we would only ever set that if this condition was true and we ran the `aggregate` command through APM.

With this PR, we need to always check whether `aggregate`'s `firstBatch` is empty and therefore always run through APM. That means that `$this->operationTime` was always captured and could result in the driver attempting to use `startAtOperationTime` to resume against a 3.6 server (in the absence of a resume token).

To address this, I moved the `server_supports_feature` check to `createResumeCallable` before we decide to use that option. That fixes the issue with resuming against a 3.6 server when no changes have been returned. This also means that the APM callback can continue collecting `$this->operationTime` all the time. Alternatively, I suppose we could have added a `server_supports_feature` within the APM callback and decide whether to ignore the operation time, but `createResumeCallable` seems more straightforward since that's when we actually attempt to specify the option.

src/Operation/Watch.php
jmikola: Just being a bit defensive here. If `aggregate` fails, I didn't want any stale values to remain in place. Granted, if it _does_ fail then the change stream will expose the exception and won't bother to resume anyway.

Now that you bring this up, I do think there's a subtle behavioral difference with this PR in that we now _always_ collect `operationTime` from _every_ `aggregate` response. Previously, we would only collect `operationTime` (and use APM) for the very first `aggregate` command executed by a Watch instance, and skip APM if `$this->operationTime` was already set going into a second `aggregate` execution (i.e. resume attempt). [Resume Process](https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst#resume-process) in the spec does refer to "original aggregation", but I wasn't sure if that was simply to differentiate from a `getMore` command or it was actually specifying _not_ to capture an operation time from a resuming `aggregate`.

I don't see why a resuming `aggregate`'s operation time would be undesirable, but it would probably be best to revert this to the original behavior and only capture `operationTime` on the very first `aggregate`.

Note: this will be a moot point for 4.2+, since `postBatchOperationTime` will be available for empty batches. That said, it's still very relevant for a 4.0 server.

src/Operation/Watch.php
kvwalker: That makes sense, thanks!

PR #637, opened 2019-07-09
PHPLIB-407: Support startAfter change stream option

jmikola:
https://jira.mongodb.org/browse/PHPLIB-407

Note: this does not address using the startAfter option for resuming. That will be implemented in PHPLIB-411 with the introduction of a resume token accessor method and support for postBatchResumeToken.

### DIFF
--- a/docs/includes/apiargs-MongoDBClient-method-watch-option.yaml
+++ b/docs/includes/apiargs-MongoDBClient-method-watch-option.yaml
@@ -36,6 +36,12 @@ source:
 ---
 source:
   file: apiargs-method-watch-option.yaml
+  ref: startAfter
+post: |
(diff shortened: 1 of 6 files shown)


xyxsns:
PHP Fatal error:  Uncaught MongoDB\Driver\Exception\CommandException: BSON field '$changeStream.startAfter' is an unknown field. in /private/var/www/mongo-changestream-php-master/vendor/mongodb/mongodb/src/Operation/Aggregate.php:295

jmikola:
@xyxsns: This is not the appropriate place to follow-up with that question. Please report a [new issue](https://github.com/mongodb/mongo-php-library/issues/new/choose), provide as much as the requested information as possible from the template, and we'll look into it when we have a chance.

PR #638, opened 2019-07-09
[WIP] Investigating freeing of invalid pointer on shutdown

jmikola:
https://jira.mongodb.org/browse/PHPLIB-432


alcaeus:
Closing here: all evidence points to generators being the culprit, which we have removed in #694.

PR #642, opened 2019-07-23
PHPLIB-402: Add aggregate helper to Database class

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-402

I've extracted the `is_in_transaction` helper from `MongoDB\Collection` in a separate to functions since we need it more often. I've also decided to extract the options for the aggregate helper to a common file to avoid unnecessary duplication of option descriptions.

### DIFF
--- a/docs/includes/apiargs-aggregate-option.yaml
+++ b/docs/includes/apiargs-aggregate-option.yaml
@@ -1,0 +1,62 @@
+arg_name: option
+name: allowDiskUse
+type: boolean
+description: |
+  Enables writing to temporary files. When set to ``true``, aggregation stages
+  can write data to the ``_tmp`` sub-directory in the ``dbPath`` directory. The
+  default is ``false``.
+interface: phpmethod
+operation: ~
+optional: true
+---
+arg_name: option
+name: batchSize
+type: integer
+description: |
+  Specifies the initial batch size for the cursor. A batchSize of ``0`` means an
+  empty first batch and is useful for quickly returning a cursor or failure
+  message without doing significant server-side work.
+interface: phpmethod
+operation: ~
+optional: true
+---
+source:
+  file: apiargs-MongoDBCollection-common-option.yaml
+  ref: bypassDocumentValidation
+post: |
+  This only applies when using the :ref:`$out <agg-out>` stage.
+
+  Document validation requires MongoDB 3.2 or later: if you are using an earlier
+  version of MongoDB, this option will be ignored.
+---
+arg_name: option
+name: comment
+type: string
+description: |
+  Users can specify an arbitrary string to help trace the operation through the
+  database profiler, currentOp, and logs.
+interface: phpmethod
+operation: ~
+optional: true
+---
+arg_name: option
+name: explain
+type: boolean
+description: |
+  Specifies whether or not to return the information on the processing of the
+  pipeline.
+interface: phpmethod
+operation: ~
+optional: true
+---
+arg_name: option
+name: hint
+type: string|array|object
+description: |
+  The index to use. Specify either the index name as a string or the index key
+  pattern as a document. If specified, then the query system will only consider
+  plans using the hinted index.
+interface: phpmethod
+operation: ~
+optional: true
+...
--- a/docs/reference/method/MongoDBDatabase-aggregate.txt
+++ b/docs/reference/method/MongoDBDatabase-aggregate.txt
@@ -1,6 +1,8 @@
-================================
-MongoDB\\Collection::aggregate()
-================================
+==============================
+MongoDB\\Database::aggregate()
+==============================
+
+.. versionadded:: 1.5
 
 .. default-domain:: mongodb
 
@@ -13,10 +15,12 @@ MongoDB\\Collection::aggregate()
 Definition
 ----------
 
-.. phpmethod:: MongoDB\\Collection::aggregate()
+.. phpmethod:: MongoDB\\Database::aggregate()
 
-   Executes an :manual:`aggregation framework pipeline
-   </core/aggregation-pipeline>` operation on the collection.
+   Runs a specified :manual:`admin/diagnostic pipeline
+   </reference/operator/aggregation-pipeline/#db-aggregate-stages>` which does
+   not require an underlying collection. For aggregations on collection data,
+   see :phpmethod:`MongoDB\\Collection::aggregate()`.
 
    .. code-block:: php
 
@@ -24,11 +28,11 @@ Definition
 
    This method has the following parameters:
 
-   .. include:: /includes/apiargs/MongoDBCollection-method-aggregate-param.rst
+   .. include:: /includes/apiargs/MongoDBDatabase-method-aggregate-param.rst
 
    The ``$options`` parameter supports the following options:
 
-   .. include:: /includes/apiargs/MongoDBCollection-method-aggregate-option.rst
+   .. include:: /includes/apiargs/MongoDBDatabase-method-aggregate-option.rst
 
 Return Values
 -------------
@@ -45,25 +49,14 @@ Errors/Exceptions
 .. include:: /includes/extracts/error-invalidargumentexception.rst
 .. include:: /includes/extracts/error-driver-runtimeexception.rst
 
-.. _php-coll-agg-method-behavior:
-
-Behavior
---------
-
-:phpmethod:`MongoDB\\Collection::aggregate()`'s return value depends on the
-MongoDB server version and whether the ``useCursor`` option is specified. If
-``useCursor`` is ``true``, a :php:`MongoDB\\Driver\\Cursor
-<class.mongodb-driver-cursor>` object is returned. If ``useCursor`` is
-``false``, an :php:`ArrayIterator <arrayiterator>` is returned that wraps the
-``result`` array from the command response document. In both cases, the return
-value will be :php:`Traversable <traversable>`.
+.. _php-db-agg-method-behavior:
 
 .. todo: add examples
 
 See Also
 --------
 
-- :phpmethod:`MongoDB\\Database::aggregate()`
+- :phpmethod:`MongoDB\\Collection::aggregate()`
 - :manual:`aggregate </reference/command/aggregate>` command reference in the
   MongoDB manual
 - :manual:`Aggregation Pipeline </core/aggregation-pipeline>` documentation in
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -209,7 +209,7 @@ class Collection
         if ( ! isset($options['readConcern']) &&
             ! ($hasOutStage && $this->readConcern->getLevel() === ReadConcern::MAJORITY) &&
             \MongoDB\server_supports_feature($server, self::$wireVersionForReadConcern) &&
-            ! $this->isInTransaction($options)) {
+            ! \MongoDB\is_in_transaction($options)) {
             $options['readConcern'] = $this->readConcern;
         }
 
@@ -220,7 +220,7 @@ class Collection
         if ($hasOutStage &&
             ! isset($options['writeConcern']) &&
             \MongoDB\server_supports_feature($server, self::$wireVersionForWritableCommandWriteConcern) &&
-            ! $this->isInTransaction($options)) {
+            ! \MongoDB\is_in_transaction($options)) {
             $options['writeConcern'] = $this->writeConcern;
         }
 
@@ -242,7 +242,7 @@ class Collection
      */
     public function bulkWrite(array $operations, array $options = [])
     {
-        if ( ! isset($options['writeConcern']) && ! $this->isInTransaction($options)) {
+        if ( ! isset($options['writeConcern']) && ! \MongoDB\is_in_transaction($options)) {
             $options['writeConcern'] = $this->writeConcern;
         }
 
@@ -274,7 +274,7 @@ class Collection
 
         $server = $this->manager->selectServer($options['readPreference']);
 
-        if ( ! isset($options['readConcern']) && \MongoDB\server_supports_feature($server, self::$wireVersionForReadConcern) && ! $this->isInTransaction($options)) {
+        if ( ! isset($options['readConcern']) && \MongoDB\server_supports_feature($server, self::$wireVersionForReadConcern) && ! \MongoDB\is_in_transaction($options)) {
             $options['readConcern'] = $this->readConcern;
         }
 
@@ -303,7 +303,7 @@ class Collection
 
         $server = $this->manager->selectServer($options['readPreference']);
 
-        if ( ! isset($options['readConcern']) && \MongoDB\server_supports_feature($server, self::$wireVersionForReadConcern) && ! $this->isInTransaction($options)) {
+        if ( ! isset($options['readConcern']) && \MongoDB\server_supports_feature($server, self::$wireVersionForReadConcern) && ! \MongoDB\is_in_transaction($options)) {
             $options['readConcern'] = $this->readConcern;
         }
 
@@ -365,7 +365,7 @@ class Collection
     {
         $server = $this->manager->selectServer(new ReadPreference(ReadPreference::RP_PRIMARY));
 
-        if ( ! isset($options['writeConcern']) && \MongoDB\server_supports_feature($server, self::$wireVersionForWritableCommandWriteConcern) && ! $this->isInTransaction($options)) {
+        if ( ! isset($options['writeConcern']) && \MongoDB\server_supports_feature($server, self::$wireVersionForWritableCommandWriteConcern) && ! \MongoDB\is_in_transaction($options)) {
             $options['writeConcern'] = $this->writeConcern;
         }
 
@@ -388,7 +388,7 @@ class Collection
      */
     public function deleteMany($filter, array $options = [])
     {
-        if ( ! isset($options['writeConcern']) && ! $this->isInTransaction($options)) {
+        if ( ! isset($options['writeConcern']) && ! \MongoDB\is_in_transaction($options)) {
             $options['writeConcern'] = $this->writeConcern;
         }
 
@@ -412,7 +412,7 @@ class Collection
      */
     public function deleteOne($filter, array $options = [])
     {
-        if ( ! isset($options['writeConcern']) && ! $this->isInTransaction($options)) {
+        if ( ! isset($options['writeConcern']) && ! \MongoDB\is_in_transaction($options)) {
             $options['writeConcern'] = $this->writeConcern;
         }
 
@@ -443,7 +443,7 @@ class Collection
 
         $server = $this->manager->selectServer($options['readPreference']);
 
-        if ( ! isset($options['readConcern']) && \MongoDB\server_supports_feature($server, self::$wireVersionForReadConcern) && ! $this->isInTransaction($options)) {
+        if ( ! isset($options['readConcern']) && \MongoDB\server_supports_feature($server, self::$wireVersionForReadConcern) && ! \MongoDB\is_in_transaction($options)) {
             $options['readConcern'] = $this->readConcern;
         }
 
@@ -470,7 +470,7 @@ class Collection
 
         $server = $this->manager->selectServer(new ReadPreference(ReadPreference::RP_PRIMARY));
 
-        if ( ! isset($options['writeConcern']) && \MongoDB\server_supports_feature($server, self::$wireVersionForWritableCommandWriteConcern) && ! $this->isInTransaction($options)) {
+        if ( ! isset($options['writeConcern']) && \MongoDB\server_supports_feature($server, self::$wireVersionForWritableCommandWriteConcern) && ! \MongoDB\is_in_transaction($options)) {
             $options['writeConcern'] = $this->writeConcern;
         }
 
@@ -504,7 +504,7 @@ class Collection
 
         $server = $this->manager->selectServer(new ReadPreference(ReadPreference::RP_PRIMARY));
 
-        if ( ! isset($options['writeConcern']) && \MongoDB\server_supports_feature($server, self::$wireVersionForWritableCommandWriteConcern) && ! $this->isInTransaction($options)) {
+        if ( ! isset($options['writeConcern']) && \MongoDB\server_supports_feature($server, self::$wireVersionForWritableCommandWriteConcern) && ! \MongoDB\is_in_transaction($options)) {
             $options['writeConcern'] = $this->writeConcern;
         }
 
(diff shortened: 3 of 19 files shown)


docs/includes/apiargs-aggregate-option.yaml
```diff
@@ -0,0 +1,71 @@
...
+description: |
+  The index to use. Specify either the index name as a string or the index key
+  pattern as a document. If specified, then the query system will only consider
+  plans using the hinted index.
+interface: phpmethod
+operation: ~
+optional: true
+---
+arg_name: option
+name: useCursor
```
jmikola: Per [`aggregate`](https://docs.mongodb.com/manual/reference/command/aggregate/#command-fields) command docs, non-cursor aggregations are no longer supported in 3.6+. That's the same version where collection-agnostic aggregations were introduced (e.g. [`$currentOp`](https://docs.mongodb.com/manual/reference/operator/aggregation/currentOp/)), so I can't think of a case where an application could legitimately use `useCursor: false` with a DB-level aggregation.

In 3.6+, the only time a cursor won't be returned is if `explain` is used, and that option is mutually exclusive with `useCursor`. IIRC, the operation class actually checks for that.

I realize `useCursor` is a common option on the `Aggregate` operation, but I think we should leave this on the collection method.

docs/includes/apiargs-aggregate-option.yaml
```diff
@@ -0,0 +1,71 @@
...
+type: boolean
+description: |
+  Indicates whether the command will request that the server provide results
+  using a cursor. The default is ``true``.
```
jmikola: Whether or not this is moved back to the collection options file, I think we should add the following note:

> MongoDB 3.6+ no longer supports returning results _without_ a cursor (excluding when the `explain` option is used) and specifying `false` for this option will result in a server error.

docs/reference/method/MongoDBDatabase-aggregate.txt
```diff
@@ -0,0 +1,71 @@
...
+   :depth: 1
+   :class: singlecol
+
+Definition
+----------
+
+.. phpmethod:: MongoDB\\Database::aggregate()
+
+   Executes an :manual:`aggregation framework pipeline
+   </core/aggregation-pipeline>` operation on the collection.
```
jmikola: "on the collection" is copypasta, but we don't want to say "on the database" as that also isn't accurate. Let's borrow language from [`db.aggregate()`](https://docs.mongodb.com/manual/reference/method/db.aggregate/).

```
Runs a specified :manual:`admin/diagnostic pipeline </reference/operator/aggregation-pipeline/#db-aggregate-stages>` which does not require an underlying collection. For aggregations on collection data, see :phpmethod:`MongoDB\\Collection::aggregate()`.
```

The manual doesn't do so, but I think we can link "admin/diagnostic pipeline" to https://docs.mongodb.com/manual/reference/operator/aggregation-pipeline/#db-aggregate-stages.

docs/reference/method/MongoDBDatabase-aggregate.txt
```diff
@@ -0,0 +1,71 @@
...
+``false``, an :php:`ArrayIterator <arrayiterator>` is returned that wraps the
+``result`` array from the command response document. In both cases, the return
+value will be :php:`Traversable <traversable>`.
+
+.. todo: add examples
+
+See Also
+--------
+
+- :manual:`aggregate </reference/command/aggregate>` command reference in the
```
jmikola: Let's lead with:

```
- :phpmethod:`MongoDB\\Collection::aggregate()`
```

Likewise, `Collection::aggregate()`'s "See Also" can include a link back to `Database::aggregate()`.

docs/reference/method/MongoDBDatabase-aggregate.txt
```diff
@@ -0,0 +1,71 @@
...
+-----------------
+
+.. include:: /includes/extracts/error-unexpectedvalueexception.rst
+.. include:: /includes/extracts/error-unsupportedexception.rst
+.. include:: /includes/extracts/error-invalidargumentexception.rst
+.. include:: /includes/extracts/error-driver-runtimeexception.rst
+
+.. _php-agg-method-behavior:
+
+Behavior
```
jmikola: I think we can drop this section entirely, since we're also pulling `useCursor` documentation from this method. I'm not even sure if the admin/diagnostic pipeline operators are compatible with `explain`, which this text would technically explain, but it's such an edge case I think we'd do better to remove this entirely lest we risk confusing readers.

docs/reference/method/MongoDBDatabase-aggregate.txt
```diff
@@ -0,0 +1,71 @@
...
+   The ``$options`` parameter supports the following options:
+
+   .. include:: /includes/apiargs/MongoDBDatabase-method-aggregate-option.rst
+
+Return Values
+-------------
+
+A :php:`MongoDB\\Driver\\Cursor <class.mongodb-driver-cursor>` or
+:php:`ArrayIterator <arrayiterator>` object. In both cases, the return value
+will be :php:`Traversable <traversable>`.
```
jmikola: I was going to suggest double-checking to see if `explain` is compatible with any of the [DB-level stages](https://docs.mongodb.com/manual/reference/operator/aggregation-pipeline/#db-aggregate-stages), but even if none of those are supported it's possible something is introduced in the future that does work with `explain` and might result in an ArrayIterator returned. I think we can leave this as-is. It's still accurate, even without the verbose "Behavior" section left in.

src/Collection.php
```diff
@@ -209,7 +209,7 @@ public function aggregate(array $pipeline, array $options = [])
         if ( ! isset($options['readConcern']) &&
             ! ($hasOutStage && $this->readConcern->getLevel() === ReadConcern::MAJORITY) &&
             \MongoDB\server_supports_feature($server, self::$wireVersionForReadConcern) &&
-            ! $this->isInTransaction($options)) {
+            ! \MongoDB\is_in_transaction($options)) {
```
jmikola: [PHPLIB-429](https://jira.mongodb.org/browse/PHPLIB-429) dropped support for 5.5, so if you'd like to import the function you can. Alternatively, we can wait to do this across the board in a later PR.

src/functions.php
```diff
@@ -110,6 +110,22 @@ function is_first_key_operator($document)
     return (isset($firstKey[0]) && $firstKey[0] === '$');
 }
 
+/**
+ * Returns whether we are currently in a transaction
+ *
+ * @internal
+ * @param array  $options  Command options
+ * @return bool
```
jmikola: I actually think that might have been the only place `bool` was used instead of `boolean`. The pedant in me would see this changed for consistency (noting that we might go in an entirely different direction after implementing code standard checks down the line).

src/functions.php
```diff
@@ -110,6 +110,22 @@ function is_first_key_operator($document)
     return (isset($firstKey[0]) && $firstKey[0] === '$');
 }
 
+/**
+ * Returns whether we are currently in a transaction
+ *
+ * @internal
+ * @param array  $options  Command options
```
jmikola: Extra whitespace can be trimmed here since there's no other `@param` elements to align with. #pedantic

src/functions.php
```diff
@@ -110,6 +110,22 @@ function is_first_key_operator($document)
     return (isset($firstKey[0]) && $firstKey[0] === '$');
 }
 
+/**
+ * Returns whether we are currently in a transaction
```
jmikola: While we're moving this around, can we add a trailing period? #pedantic

tests/SpecTests/CommandExpectations.php
```diff
@@ -64,6 +64,18 @@ public static function fromCommandMonitoring(array $expectedEvents)
...
     }
 
+    public static function fromCrud(array $expectedEvents)
+    {
+        $o = new self($expectedEvents);
+
+        /* The bulkWrite-arrayFilters test only includes the commandStarted
+         * event without asserting against the result of the command, so ignore
+         * any other events. */
+        $o->ignoreExtraEvents = true;
```
jmikola: I think you want:

```
$o->ignoreCommandFailed = true;
$o->ignoreCommandSucceeded = true;
```

Instead of this. Basically, `fromCommandMonitoring()` is currently the only spec that has assertions for commands failing/succeeding. Every other spec only asserts the command started event.

 `ignoreExtraEvents` is currently only used in the change stream spec, because its runner purposefully tells us to only check as many events are found in the `expectations` array. This is mainly so that it doesn't need to deal with variation around `getMore` commands across drivers; however, it's also a problem I'm trying to address with [SPEC-1341](https://jira.mongodb.org/browse/SPEC-1341) since we _would_ like to assert `getMore` in _some_ tests.

But I digress. For the CRUD tests, I don't think any `expectations` arrays are incomplete, so we should be able to leave `$ignoreExtraEvents` alone here so we can catch failures down the line (if we actually do end up missing an event).

tests/SpecTests/CrudSpecTest.php
```diff
@@ -0,0 +1,162 @@
+<?php
+
+namespace MongoDB\Tests\SpecTests;
+
+use stdClass;
+
+/**
+ * Transactions spec tests.
+ *
+ * @see https://github.com/mongodb/specifications/tree/master/source/crid
```
jmikola: `crud` :)

tests/SpecTests/CrudSpecTest.php
```diff
@@ -0,0 +1,162 @@
+<?php
+
+namespace MongoDB\Tests\SpecTests;
+
+use stdClass;
+
+/**
+ * Transactions spec tests.
```
jmikola: "CRUD spec tests"

tests/SpecTests/CommandExpectations.php
```diff
@@ -64,6 +64,18 @@ public static function fromCommandMonitoring(array $expectedEvents)
         return new self($expectedEvents);
     }
 
+    public static function fromCrud(array $expectedEvents)
+    {
+        $o = new self($expectedEvents);
+
+        /* The bulkWrite-arrayFilters test only includes the commandStarted
+         * event without asserting against the result of the command, so ignore
+         * any other events. */
```
jmikola: Curious about this, but I commented below in CrudSpecTest.

tests/SpecTests/CrudSpecTest.php
```diff
@@ -0,0 +1,162 @@
...
+        'aggregate-merge: Aggregate with $merge and batch size of 0' => 'PHPLIB-438',
+        'aggregate-merge: Aggregate with $merge and majority readConcern' => 'PHPLIB-438',
+        'aggregate-merge: Aggregate with $merge and local readConcern' => 'PHPLIB-438',
+        'aggregate-merge: Aggregate with $merge and available readConcern' => 'PHPLIB-438',
+        'aggregate-out-readConcern: readConcern majority with out stage' => 'PHPLIB-431',
+        'aggregate-out-readConcern: readConcern local with out stage' => 'PHPLIB-431',
+        'aggregate-out-readConcern: readConcern available with out stage' => 'PHPLIB-431',
+        'aggregate-out-readConcern: readConcern linearizable with out stage' => 'PHPLIB-431',
+        'aggregate-out-readConcern: invalid readConcern with out stage' => 'PHPLIB-431',
+        'bulkWrite-arrayFilters: BulkWrite with arrayFilters' => 'Fails due to command assertions',
```
jmikola: Is it possible this only failed because you _weren't_ ignoring command failed/succeeded events? If so, see if my earlier suggested change allows this to pass.

The actual spec test for this PR (`db-aggregate.json`) happens to not include any APM expectations. By coincidence, you're already skipping the other tests since related tickets aren't done. That leaves `bulkWrite-arrayFilters` as the only test that was actually using APM, so it would make sense if the earlier issue was causing this.

tests/SpecTests/CrudSpecTest.php
```diff
@@ -0,0 +1,162 @@
...
+     * @param stdClass $expected Expected command document
+     * @param stdClass $actual   Actual command document
+     */
+    public static function assertCommandMatches(stdClass $expected, stdClass $actual)
+    {
+        /* The command may contain a session id, even though we don't start one.
+         * In this case, ignore any session identifier by removing it. */
+        if (isset($actual->lsid)) {
+            unset($actual->lsid);
+        }
```
jmikola: Can you confirm if this is necessary? `assertDocumentsMatch()` should ignore extra fields in the actual document (see doc block in FunctionalTestCase as well as DocumentsMatchConstraint for more info).

I'll note that the change stream spec, which doesn't assert `lsid` at all, didn't need this.

tests/SpecTests/CrudSpecTest.php
```diff
@@ -0,0 +1,162 @@
...
+    /**
+     * Execute an individual test case from the specification.
+     *
+     * @dataProvider provideTests
+     * @param string    $name           Test name
+     * @param stdClass  $test           Individual "tests[]" document
+     * @param array     $runOn          Top-level "runOn" array with server requirements
+     * @param array     $data           Top-level "data" array to initialize collection
+     * @param string    $databaseName   Name of database under test
+     * @param string    $collectionName Name of collection under test
```
jmikola: I realize this was copy-pasted from another file, but you can remove a space between the type and variable names. Looks like a missed that in the other files, but we can address it here.

tests/SpecTests/CrudSpecTest.php
```diff
@@ -0,0 +1,162 @@
...
+        $databaseName = isset($databaseName) ? $databaseName : $this->getDatabaseName();
+        $collectionName = isset($collectionName) ? $collectionName : $this->getCollectionName();
+
+        $context = Context::fromCrud($test, $databaseName, $collectionName);
+        $this->setContext($context);
+
+        $this->dropTestAndOutcomeCollections();
+        $this->createTestCollection();
+        $this->insertDataFixtures($data);
+        $this->preventStaleDbVersionError($test->operations);
```
jmikola: IIRC, this was only needed for the transaction spec tests. I think you can remove this.

tests/SpecTests/CrudSpecTest.php
```diff
@@ -0,0 +1,162 @@
...
+        }
+
+        $databaseName = isset($databaseName) ? $databaseName : $this->getDatabaseName();
+        $collectionName = isset($collectionName) ? $collectionName : $this->getCollectionName();
+
+        $context = Context::fromCrud($test, $databaseName, $collectionName);
+        $this->setContext($context);
+
+        $this->dropTestAndOutcomeCollections();
+        $this->createTestCollection();
```
jmikola: I think this also shouldn't be needed. We only did this in the transactions spec because the server does not allow creating a collection (or database AFAIK) within a multi-document transaction.

tests/SpecTests/Operation.php
```diff
@@ -101,6 +101,23 @@ public static function fromCommandMonitoring(stdClass $operation)
...
+
+        $o->resultExpectation = ResultExpectation::fromCrud($operation, $o->getResultAssertionType());
+
+        if (isset($operation->collectionOptions)) {
+            $o->collectionOptions = (array) $operation->collectionOptions;
+        }
+
+        if (isset($operation->databaseOptions)) {
+            $o->databaseOptions = (array) $operation->databaseOptions;
+        }
```
jmikola: `databaseOptions` technically isn't defined within the CRUD spec. It's possible this could show up in the future, so I'll defer to you if you'd like to keep it. Elsewhere, I was being very pedantic and only parsing fields in the documented test format (hoping to invite a failure if something new was introduced unknowingly).

tests/SpecTests/ResultExpectation.php
```diff
@@ -72,6 +72,19 @@ public static function fromChangeStreams(stdClass $result, callable $assertionCa
...
+        if (property_exists($operation, 'result') && !self::isErrorResult($operation->result)) {
+            $assertionType = $operation->result === null ? self::ASSERT_NULL : $defaultAssertionType;
+            $expectedValue = $operation->result;
+        } else {
+            $assertionType = self::ASSERT_NOTHING;
+            $expectedValue = null;
+        }
+
+        return new self($assertionType, $expectedValue);
+    }
```
jmikola: I believe this is identical to `fromTransactions()`. I think both methods should stay, but maybe we should extra the body into a new function so we can do something like:

```
list($assertionType, $expectedValue) = getAssertionTypeAndExpectedValue($operation);
```

Ideally, retryable writes will ultimately be migrated to use consistent syntax. I'm not sure if retryable reads (coming soon) does, too, so this refactoring could certainly wait until we have 3+ duplications.

src/Database.php
```diff
@@ -155,6 +158,67 @@ public function __toString()
...
 
+    /**
+     * Runs an aggregation framework pipeline on the database for pipeline
+     * stages that do not require an underlying collection, such as $currentOp
+     * and $listLocalSessions. Requires MongoDB >= 3.6
+     *
+     * Note: this method's return value depends on the MongoDB server version
+     * and the "useCursor" option. If "useCursor" is true, a Cursor will be
+     * returned; otherwise, an ArrayIterator is returned, which wraps the
+     * "result" array from the command response document.
```
jmikola: Per earlier discussions in the doc files, I think we can remove this or replace it with a note that the `useCursor` option should just be left alone.

docs/includes/apiargs-aggregate-option.yaml
```diff
@@ -0,0 +1,71 @@
...
+interface: phpmethod
+operation: ~
+optional: true
```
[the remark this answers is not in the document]
alcaeus: Didn't get an error when I tried building it, but I'll add it nonetheless since all other files have it.

docs/includes/apiargs-aggregate-option.yaml
alcaeus: Yes, that makes sense - I didn't realise the versions lined up like that. I moved the option back to the collection docs and added a note about servers > 3.6 throwing an error.

docs/reference/method/MongoDBDatabase-aggregate.txt
alcaeus: > I'm not even sure if the admin/diagnostic pipeline operators are compatible with `explain`

I had to check myself, but it's perfectly fine to use `explain` with diagnostic pipeline operators.
```
> db.aggregate([{$currentOp: {}}], {explain: true})
{ "stages" : [ { "$currentOp" : { } } ], "ok" : 1 }
```

Removing the block sounds good for now. It might be worth explaining the different behaviour when using `explain`, but mentioning `useCursor` is definitely confusing.

src/Collection.php
alcaeus: I'd like to do this across the board and bring in a slightly softer variant of https://github.com/doctrine/coding-standard that works with PHP 5.

src/functions.php
alcaeus: That is correct. I've changed it to `boolean` for the time being. The coding standard mentioned above includes a sniff (https://github.com/slevomat/coding-standard/blob/master/SlevomatCodingStandard/Sniffs/TypeHints/LongTypeHintsSniff.php) that forces short type names to be used (`int` and `bool` vs. `integer` and `boolean`, respectively). I suppose we'll discuss that when the time comes.

src/functions.php
alcaeus: I was wondering if you'd notice 😂

tests/SpecTests/CrudSpecTest.php
alcaeus: This might have been left over from me trying to get the bulkWrite test to pass. `assertDocumentsMatch` decided that those documents didn't match, but I'll have to re-check that anyways. I'll remove this piece of specific code and fix the bulkWrite spec test in a separate PR.

tests/SpecTests/CrudSpecTest.php
alcaeus: Will do. I'll let phpcs clean that up in other files once we add it to the build pipeline.

tests/SpecTests/Operation.php
alcaeus: This was shamelessly copy-pasted from `fromTransactions`. Since it's not an option of the spec as you mention, I'd suggest removing it from both methods until the time comes to add it.

tests/SpecTests/ResultExpectation.php
alcaeus: I'll refactor the `fromCrud`, `fromRetryableWrites` and `fromTransactions` methods to add a `getAssertion...` method like you suggested. We can then always keep more significant exceptions to the norm like `fromChangeStreams` while avoiding duplicating the usual checks. 👍

tests/SpecTests/CommandExpectations.php
alcaeus: I've replaced `ignoreExtraEvents` with `ignoreCommandFailed` and `ignoreCommandSucceeded`. `ignoreExtraEvents` is a bit blunt since it slices off the list of commands that were stored. I think it's better to not store commands we don't care about in the first place.

tests/SpecTests/CrudSpecTest.php
alcaeus: I've taken the time to step-debug this and the issue is that while we expect a `BSONArray` instance for `arrayFilters` in the bulkWrite update document, we seem to send a `BSONDocument` instead, which causes the constraint failure. I've created https://jira.mongodb.org/browse/PHPLIB-458 to investigate the issue at hand. Along the way, I've also created https://jira.mongodb.org/browse/PHPLIB-459 to update the `DocumentsMatchConstraint` in question to better explained which part of the documents don't match, as I was only able to figure this out by manually checking the objects in question.

tests/SpecTests/ResultExpectation.php
alcaeus: So, after starting to refactor I decided to put this off for now. The `fromRetryableWrites` helper is different, which means that I'll most likely do once we add retryable reads.

tests/SpecTests/CommandExpectations.php
jmikola: Looking at CommandExpectation with fresh eyes, the `array_slice()` is definitely odd. We do know the size of `expected` at the time the class is constructed, so it seems feasible that we could decide to ignore an event rather than continue appending to `$actualEvents`. Feel free to just refactor that in a later commit if you like (I don't think it warrants a review).

tests/SpecTests/Operation.php
```diff
@@ -101,6 +100,20 @@ public static function fromCommandMonitoring(stdClass $operation)
...
+    {
+        $o = new self($operation);
+
+        $o->resultExpectation = ResultExpectation::fromCrud($operation, $o->getResultAssertionType());
+
+        if (isset($operation->collectionOptions)) {
+            $o->collectionOptions = (array) $operation->collectionOptions;
+        }
+
+
```
jmikola: Extra newline from deleting the database options `if` block.

tests/SpecTests/crud/aggregate-merge.json
```diff
@@ -0,0 +1,415 @@
+{
+  "runOn": [
+    {
+      "minServerVersion": "4.2.0"
```
jmikola: https://github.com/mongodb/specifications/commit/efe9319a2b8b046d44fcea9661b9e18fdb109031 was merged, so you can update this here or in #644 as you wish.

jmikola: Just saw your comment in https://github.com/mongodb/mongo-php-library/pull/644#discussion_r306689346. I suppose this is the PR to update it then.

PR #643, opened 2019-07-23
PHPLIB-413: Add support for read concerns in aggregations with an $out stage

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-413


src/Collection.php
```diff
@@ -201,15 +202,18 @@ public function aggregate(array $pipeline, array $options = [])
...
+         * level will cause a server error.
          *
          * A read concern is also not compatible with transactions.
          */
         if ( ! isset($options['readConcern']) &&
-            ! ($hasOutStage && $this->readConcern->getLevel() === ReadConcern::MAJORITY) &&
             \MongoDB\server_supports_feature($server, self::$wireVersionForReadConcern) &&
-            ! \MongoDB\is_in_transaction($options)) {
+            ! \MongoDB\is_in_transaction($options) &&
+            ( ! $hasOutStage || \MongoDB\server_supports_feature($server, self::$wireVersionForReadConcernWithOutStage))
```
jmikola: I think this is also an improvement. IIRC, there was some discussion about venturing a bit too far outside the spec with previous behavior of silently deciding whether or not it was safe to inherit a particular read concern. I'd now say it's preferable to inherit as-is (if supported) and allow the server to throw if it needs to.

src/Collection.php
```diff
@@ -201,15 +202,18 @@ public function aggregate(array $pipeline, array $options = [])
 
         $server = $this->manager->selectServer($options['readPreference']);
 
-        /* A "majority" read concern is not compatible with the $out stage, so
-         * avoid providing the Collection's read concern if it would conflict.
+        /* Servers before 4.2 don't support a readConcern when an $out stage is
+         * being used. Starting with 4.2, a read concern can be used; however it
+         * does not support the "linearizable" level. Using the "linearizable"
+         * level will cause a server error.
```
jmikola: We're not going to check for particular RC values here (like the old code did) so I think we can remove the text about "linearizable". I think this can just be:

> MongoDB 4.2 and later supports a read concern when an $out stage is being used, but earlier versions do not.

src/Database.php
```diff
@@ -191,15 +192,18 @@ public function aggregate(array $pipeline, array $options = [])
 
         $server = $this->manager->selectServer($options['readPreference']);
 
-        /* A "majority" read concern is not compatible with the $out stage, so
-         * avoid providing the Collection's read concern if it would conflict.
+        /* Servers before 4.2 don't support a readConcern when an $out stage is
+         * being used. Starting with 4.2, a read concern can be used; however it
+         * does not support the "linearizable" level. Using the "linearizable"
+         * level will cause a server error.
```
jmikola: This can be synced with whatever is done for Collection.

tests/SpecTests/CrudSpecTest.php
```diff
@@ -19,11 +19,6 @@ class CrudSpecTest extends FunctionalTestCase
         'aggregate-merge: Aggregate with $merge and majority readConcern' => 'PHPLIB-438',
         'aggregate-merge: Aggregate with $merge and local readConcern' => 'PHPLIB-438',
         'aggregate-merge: Aggregate with $merge and available readConcern' => 'PHPLIB-438',
-        'aggregate-out-readConcern: readConcern majority with out stage' => 'PHPLIB-431',
-        'aggregate-out-readConcern: readConcern local with out stage' => 'PHPLIB-431',
-        'aggregate-out-readConcern: readConcern available with out stage' => 'PHPLIB-431',
-        'aggregate-out-readConcern: readConcern linearizable with out stage' => 'PHPLIB-431',
-        'aggregate-out-readConcern: invalid readConcern with out stage' => 'PHPLIB-431',
```
jmikola: Related to APM discussions in #402, do you know if these failed after removing the skips? Since these tests _do_ include `expectations`.

src/Collection.php
alcaeus: Ah, yes. This was left over from a previous iteration where I assumed that we needed to avoid adding a `linearisable` level to the command. I realised later the spec actually expects us to pass it and let the server error. I'll update the comment as you suggested 👍

tests/SpecTests/CrudSpecTest.php
alcaeus: Yes, these tests failed as they expect the `readConcern` option to be present in the command.

alcaeus:
Review comments taken care of, waiting for #642 to be merged.

tests/SpecTests/CrudSpecTest.php
```diff
@@ -19,11 +19,6 @@ class CrudSpecTest extends FunctionalTestCase
         'aggregate-merge: Aggregate with $merge and majority readConcern' => 'PHPLIB-438',
         'aggregate-merge: Aggregate with $merge and local readConcern' => 'PHPLIB-438',
         'aggregate-merge: Aggregate with $merge and available readConcern' => 'PHPLIB-438',
-        'aggregate-out-readConcern: readConcern majority with out stage' => 'PHPLIB-431',
```
alcaeus: While working on these tests I realised that neither the `aggregate-out-readConcern` nor `aggregate-merge` spec tests cover running database-level aggregations with these options. I was wondering if it would make sense to add such tests to the spec to ensure consistent behaviour of the `db.aggregate` and `collection.aggregate` helpers?

jmikola: Since neither of the [DB-level pipeline operations](https://docs.mongodb.com/manual/reference/operator/aggregation-pipeline/#db-aggregate-stages) receive data from a collection, I'm not actually sure how read concern would affect the command. I could imagine it being ignored or prompting an server error. This would be worth asking in `#driver-devs` or, if necessary, `#server-query`. So long as it doesn't produce a server error, I don't see the harm in reproducing spec tests for it, since they could at least use APM.

I see no reason we cannot test `$merge`, though. And since it was introduced _after_ DB-level aggregations, I think you could just add a `db.aggregate()` case directly to the existing `$merge` spec test file since the `runOn` clause will apply.

Feel free to log a SPEC ticket (or two) for these and take it on. It would be a good learning experience for submitting a spec test PR.

alcaeus: A quick test showed that running a DB-level pipeline (e.g. `$listLocalSessions`) with a read concern other than `local` will result in an error from the database. I'll put the information into a spec ticket once I've got the necessary permissions.

PR #644, opened 2019-07-23
PHPLIB-438: Unify handling for write stages in aggregation pipelines

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-438

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -127,7 +127,7 @@ function is_in_transaction(array $options)
 }
 
 /**
- * Return whether the aggregation pipeline ends with an $out operator.
+ * Return whether the aggregation pipeline ends with an $out or $merge operator.
  *
  * This is used for determining whether the aggregation pipeline must be
  * executed against a primary server.
@@ -136,7 +136,7 @@ function is_in_transaction(array $options)
  * @param array $pipeline List of pipeline operations
  * @return boolean
  */
-function is_last_pipeline_operator_out(array $pipeline)
+function is_last_pipeline_operator_write(array $pipeline)
 {
     $lastOp = end($pipeline);
 
@@ -146,7 +146,7 @@ function is_last_pipeline_operator_out(array $pipeline)
 
     $lastOp = (array) $lastOp;
 
-    return key($lastOp) === '$out';
+    return in_array(key($lastOp), ['$out', '$merge'], true);
 }
 
 /**
--- a/tests/SpecTests/Context.php
+++ b/tests/SpecTests/Context.php
@@ -63,6 +63,15 @@ final class Context
             $o->outcomeCollectionName = $test->outcome->collection->name;
         }
 
+        $o->defaultWriteOptions = [
+            'writeConcern' => new WriteConcern(WriteConcern::MAJORITY),
+        ];
+
+        $o->outcomeFindOptions = [
+            'readConcern' => new ReadConcern('local'),
+            'readPreference' => new ReadPreference('primary'),
+        ];
+
         $o->client = new Client(FunctionalTestCase::getUri(), $clientOptions);
 
         return $o;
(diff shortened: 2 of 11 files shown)


tests/SpecTests/crud/aggregate-merge.json
```diff
@@ -1,7 +1,7 @@
 {
   "runOn": [
     {
-      "minServerVersion": "4.2.0"
+      "minServerVersion": "4.1.11"
```
jmikola: Our spec tests typically do use the exact version when a feature was introduced. This is totally my fault dating back to https://github.com/mongodb/specifications/commit/b10b400be768f6d5df95eddf90fdb710df1975bd#diff-05d2232459eeabeaeec04d301113e3b3R6.

I may as well submit a PR to go back and change this in the main repo.

src/functions.php
```diff
@@ -136,7 +136,7 @@ function is_in_transaction(array $options)
  * @param array $pipeline List of pipeline operations
  * @return boolean
  */
-function is_last_pipeline_operator_out(array $pipeline)
+function is_last_pipeline_operator_write(array $pipeline)
```
jmikola: Doc block still refers to `$out` only.

src/functions.php
```diff
@@ -146,7 +146,7 @@ function is_last_pipeline_operator_out(array $pipeline)
 
     $lastOp = (array) $lastOp;
 
-    return key($lastOp) === '$out';
+    return in_array(key($lastOp), ['$out', '$merge']);
```
jmikola: It may not make a practical difference, but preserving `===` warrants setting `$strict` to `true`. I bring this up in case it's relevant for some future sniff (e.g. prohibiting loose equality checks across the board).

tests/SpecTests/crud/aggregate-merge.json
alcaeus: See https://github.com/mongodb/specifications/pull/614, changes there will be brought in with #642 with the temporary commit in this branch being dropped.

src/functions.php
alcaeus: Good point. I've added a `true` as third argument to enable strict checks.

tests/SpecTests/crud/aggregate-merge.json
alcaeus: See https://github.com/mongodb/specifications/pull/614, changes there will be brought in with #642 with the temporary commit in this branch being dropped.

alcaeus:
Review items applied, documentation updated. I also added read and write concerns to the `Context::fromCrud` helper to fix intermittent test failures on replica sets.

tests/SpecTests/crud/aggregate-merge.json
jmikola: Thanks for linking the PR. I forgot to come back and drop a comment here.

Since merged: https://github.com/mongodb/specifications/commit/efe9319a2b8b046d44fcea9661b9e18fdb109031

tests/SpecTests/Context.php
```diff
@@ -53,6 +53,30 @@ public static function fromCommandMonitoring(stdClass $test, $databaseName, $col
...
+
+        $clientOptions = isset($test->clientOptions) ? (array) $test->clientOptions : [];
+
+        if (isset($test->outcome->collection->name)) {
+            $o->outcomeCollectionName = $test->outcome->collection->name;
+        }
+
+        $o->defaultWriteOptions = [
+            'writeConcern' => new WriteConcern(WriteConcern::MAJORITY),
+        ];
```
jmikola: Was this (and `outcomeFindOptions`) actually needed? AFAIK, all of the CRUD tests read/write to the primary and never specified a custom read concern or read preference.

AFAIK, `defaultWriteOptions` was only used for transaction tests (per [implementation instructions](https://github.com/mongodb/specifications/tree/master/source/transactions/tests#use-as-integration-tests)).

alcaeus: I added this after a test kept failing on a replicaset setup. I got suspicious after I couldn’t replicate the failures when step-debugging through the test and thought that the problem might be with the secondaries being read before they have the data.

PR #652, opened 2019-07-29
PHPLIB-417: countDocuments should use group with _id: 1

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-417

### DIFF
--- a/src/Operation/CountDocuments.php
+++ b/src/Operation/CountDocuments.php
@@ -192,7 +192,7 @@ class CountDocuments implements Executable
             $pipeline[] = ['$limit' => $this->options['limit']];
         }
 
-        $pipeline[] = ['$group' => ['_id' => null, 'n' => ['$sum' => 1]]];
+        $pipeline[] = ['$group' => ['_id' => 1, 'n' => ['$sum' => 1]]];
 
         $cmd = [
             'aggregate' => $this->collectionName,


lindelius:
Isn't this just going to return either 1 or 0, though?

alcaeus:
> Isn't this just going to return either 1 or 0, though?

The `countDocuments` helper doesn't rely on collection metadata to return a document count, but rather runs an aggregation pipeline containing a `$group` pipeline stage to count the documents. In this case, we're grouping by the value `1`, which will cause the `$group` stage to emit a single document with the `n` field containing the number of documents that were emitted from the previous `$match`, `$skip`, and `$limit` stages.

lindelius:
Is the ”output a single document” functionality being changed to `1` from `null`? The shell documentation is still using `null` in the examples

alcaeus:
The motivation for the change is explained in https://github.com/mongodb/specifications/pull/504:
> The $group stage of an aggregation pipeline can use any constant to achieve the group-for-all effect. Switching to using 1 as a constant, to avoid test runner issues associated with null values

This has no implications on the result or on whether you should use `null` or `1` when doing such a `$group` aggregation yourself. In our case the change is necessary to pass the spec tests once adding support for retryable reads.

The "output a single document" functionality is not tied to anything: this is a side effect of using `$group` with a constant value for `_id`, no matter what this constant value is. You might as well use `_id: 'The quick brown fox jumps over the lazy dog'`.

lindelius:
Ok! Thanks for clarifying. I was under the impression that this was only for when setting the `_id` field to `null`.

alcaeus:
> It’s propbably not your table, but it would be useful to update the documentations to read ”constant value” rather than ”null” for the examples etc.

I don't know which table you are referring to. A quick search led me to the [documentation for `$group`](https://docs.mongodb.com/manual/reference/operator/aggregation/group/), which states:
> The _id field is mandatory; however, you can specify an _id value of null, or any other constant value, to calculate accumulated values for all the input documents as a whole.

jmikola:
@lindelius: Nothing to worry about on your end (or any other application). For purposes of `$group`, _any_ constant value (provided it's not some special operator/expression) should provide the same effect of grouping into a single accumulation bucket.

The impetus for this change was that several internal [specification](https://github.com/mongodb/specifications) tests had trouble asserting outgoing commands because those same tests often use `null` to signify that a field should _not_ exist in the document, rather than asserting that it contains a BSON null value.

PR #658, opened 2019-08-01
PHPLIB-242: Refactor command operations to use field path syntax for type maps 

alcaeus:
This adds a new internal function to modify a type map given to a command and add a `fieldPath` type for the given path. This is used in the `Aggregate`, `Distinct`, `FindAndModify`, and `MapReduce` operations to avoid manually applying the type map to the results. Any pre-existing field paths are prefixed with the internal field path to ensure consistent results. If a user already passes a type for the field path for the given operation, it is not overwritten as I assume the user "knows what they are doing". The root type is always set to `object`; otherwise the user would be able to break internal functionality by changing the root type. I'm not sure whether the name of the function is appropriate to convey what the function is doing; I'm open to suggestions in this regard.


src/functions.php
```diff
@@ -241,3 +242,41 @@ function recursive_copy($element) {
...
+ * An existing type map for the given field path will not be overwritten
+ *
+ * @internal
+ * @param array  $typeMap   The existing typeMap
+ * @param string $fieldPath The field path to apply the root type to
+ * @return array
+ */
+function create_field_path_type_map(array $typeMap, $fieldPath)
+{
+    if (isset($typeMap['root']) && ! isset($typeMap['fieldPaths'][$fieldPath])) {
```
jmikola: Should we skip this just because the user has specified both a custom `root` type _and_ happens to have specified our internal field path in their own `fieldPaths` option?

For instance, if we're using this with FindAndModify and the user specified a custom type for the `value` field path, I imagine we should just move that to the `value.value` field path. And independent of that, whatever the user specifies for `root` can always can always be assigned to `$typeMap['fieldPaths'][$fieldPath]` at the end (as you do later in the function).

I wonder if it'd make sense to just handle these independently.

tests/FunctionsTest.php
```diff
@@ -135,4 +169,53 @@ public function provideMapReduceOutValues()
...
+                ['document' => 'array', 'root' => 'object'],
+                ['document' => 'array'],
+            ],
+            'No field path' => [
+                ['root' => 'object', 'fieldPaths' => ['field' => 'array']],
+                ['root' => 'array'],
+            ],
+            'Field path exists' => [
+                ['root' => 'object', 'fieldPaths' => ['field' => 'object']],
+                ['root' => 'array', 'fieldPaths' => ['field' => 'object']],
```
jmikola: Related to my earlier comments on `create_field_path_type_map()`, I think this is where we'd expect:

```
[
    'root' => 'object',
    'fieldPaths' => [
        'field' => 'array',
        'field.field' => 'object',
    ],
]
```

At least I believe that's how this would have worked previously with TypeMapArrayIterator. It's quite possible we never had tests combining `fieldPaths` with TypeMapArrayIterator, since the class predated that extension feature.

src/functions.php
alcaeus: You are correct. The helper function now always nests all pre-existing field paths to prefix them with the given field path. It also applies an `object` type for the `root` key, and an `array` type for the `array` key if the given field path ends in `.$`. This was added because the Aggregate, Distinct, and MapReduce operations expect an array of results instead of an iterator. We may always optimise this further down the road by returning an iterator directly instead of wrapping the results after the call. For now, this works just fine though.

src/Operation/FindAndModify.php
```diff
@@ -216,21 +216,14 @@ public function execute(Server $server)
...
 
         $cursor = $server->executeWriteCommand($this->databaseName, new Command($this->createCommandDocument($server)), $this->createOptions());
-        $result = current($cursor->toArray());
-
-        if ( ! isset($result->value)) {
-            return null;
-        }
-
-        if ( ! is_object($result->value)) {
-            throw new UnexpectedValueException('findAndModify command did not return a "value" document');
```
kvwalker: Do we no longer need to throw this exception?

alcaeus: `findAndModify` will either return a document or `null` depending on the operation and matched values. We're guarding against a missing `value` property in the return statement, but from what I can see we don't need to guard against non-object values being returned from `findAndModify`.

PR #659, opened 2019-08-05
PHPLIB-461: Fix functional tests for change streams on sharded clusters

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-461

Due to the way `getMore` works on sharded clusters, we can't rely on the first `getMore` call after an insert to return the data we expect. I'll quote the response from the query team to my questions about why `getMore` behaves this way.

> First, it will be useful to understand the semantics of `getMore` on a tailable `awaitData` cursor such as that created by `$changeStream`:
> 1. If there are documents available when the `getMore` starts running, then `getMore` will read up to `batchSize` OR `EOF`, whichever comes first. It will then return immediately without waiting for any further results.
> 2. If there are no documents available when the `getMore` starts running, it will wait up to the specified `awaitData` timeout to see if any documents become available.
> 3. If documents become available while the `getMore` is waiting, it will adopt the behaviour outlined in (1). In other words: once the `getMore` has obtained at least one result, it cancels its `awaitData` timeout and does not wait again.
> 
> The second thing to understand is the behaviour of change streams in a sharded cluster. For reasons [outlined in this comment](https://jira.mongodb.org/browse/SERVER-36526?focusedCommentId=1972933&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-1972933), change streams must open a cursor on every shard even when monitoring an unsharded collection. Because each change stream is globally-sorted, it cannot return result A from shard X until it has obtained at least one result from every other shard that sorts **later than** result A. In other words, it cannot return A to the client until every other shard has promised that they will never produce a result that sorts before A. To guard against the possibility that one or more shards never return any results, each shard has a no-op writer that updates its oplog every `periodicNoopIntervalSecs` seconds. This allows the shard to return a high-water-mark promising that it will never return a future result which sorts before that point, so we are guaranteed that the change stream will eventually be allowed to return result A to the client.

With an `maxAwaitTimeMS` of 500 ms we stay below the minimum `periodicNoopIntervalSecs` of 1 second, which can cause our `getMore` commands on shards without data to time out before a no-op is read from the oplog. One way to solve this would be to increase the `maxAwaitTimeMS` to a value above 1000 ms to ensure that `getMore` calls don't time out before another no-op is written, but this causes tests to be significantly slower. Furthermore, we should never make any assumptions as to what batch a change stream event is returned in, only that it will eventually be returned in the correct order. To ensure the tests don't unnecessarily fail on sharded clusters, they will now execute multiple `getMore` calls if necessary, only failing if a given number (currently 5) of `getMore` calls fails to return any data. This was also necessary since resume operations would still yield empty batches on the initial call to `aggregate`.

This PR also contains a workaround for [PHPC-1419](https://jira.mongodb.org/browse/PHPC-1419), where a `NonResumableChangeStreamError` error label is not applied to the `ServerException` create for a `ChangeStreamFatalError`. This causes the iterator to repeatedly retry resuming the change stream for an unresumable error. This workaround can be removed once the above issue was fixed.

### DIFF
--- a/tests/FunctionalTestCase.php
+++ b/tests/FunctionalTestCase.php
@@ -336,9 +336,6 @@ abstract class FunctionalTestCase extends TestCase
                 if (! $this->isShardedClusterUsingReplicasets()) {
                     $this->markTestSkipped('$changeStream is only supported with replicasets');
                 }
-
-                // Temporarily skip tests because of an issue with change streams in the driver
-                $this->markTestSkipped('$changeStreams currently don\'t on replica sets');
                 break;
 
             case Server::TYPE_RS_PRIMARY:
--- a/tests/Operation/WatchFunctionalTest.php
+++ b/tests/Operation/WatchFunctionalTest.php
@@ -3,6 +3,7 @@
 namespace MongoDB\Tests\Operation;
 
 use Closure;
+use Iterator;
 use MongoDB\BSON\TimestampInterface;
 use MongoDB\ChangeStream;
 use MongoDB\Driver\Cursor;
@@ -19,6 +20,7 @@ use MongoDB\Operation\DatabaseCommand;
 use MongoDB\Operation\InsertOne;
 use MongoDB\Operation\Watch;
 use MongoDB\Tests\CommandObserver;
+use PHPUnit\Framework\ExpectationFailedException;
 use ReflectionClass;
 use stdClass;
 use Symfony\Bridge\PhpUnit\SetUpTearDownTrait;
@@ -71,8 +73,7 @@ class WatchFunctionalTest extends FunctionalTestCase
         $this->insertDocument(['x' => 1]);
         $this->insertDocument(['x' => 2]);
 
-        $changeStream->next();
-        $this->assertTrue($changeStream->valid());
+        $this->advanceCursorUntilValid($changeStream);
         $this->assertSameDocument($changeStream->current()->_id, $changeStream->getResumeToken());
 
         $changeStream->next();
@@ -81,8 +82,7 @@ class WatchFunctionalTest extends FunctionalTestCase
 
         $this->insertDocument(['x' => 3]);
 
-        $changeStream->next();
-        $this->assertTrue($changeStream->valid());
+        $this->advanceCursorUntilValid($changeStream);
         $this->assertSameDocument($changeStream->current()->_id, $changeStream->getResumeToken());
     }
 
@@ -134,23 +134,21 @@ class WatchFunctionalTest extends FunctionalTestCase
         $this->insertDocument(['x' => 1]);
         $this->insertDocument(['x' => 2]);
 
-        $events = [];
+        $lastEvent = null;
 
         (new CommandObserver())->observe(
             function () use ($changeStream) {
-                $changeStream->next();
+                $this->advanceCursorUntilValid($changeStream);
             },
-            function (array $event) use (&$events) {
-                $events[] = $event;
+            function (array $event) use (&$lastEvent) {
+                $lastEvent = $event;
             }
         );
 
-        $this->assertCount(1, $events);
-        $this->assertSame('getMore', $events[0]['started']->getCommandName());
-        $postBatchResumeToken = $this->getPostBatchResumeTokenFromReply($events[0]['succeeded']->getReply());
+        $this->assertNotNull($lastEvent);
+        $this->assertSame('getMore', $lastEvent['started']->getCommandName());
+        $postBatchResumeToken = $this->getPostBatchResumeTokenFromReply($lastEvent['succeeded']->getReply());
 
-        $changeStream->next();
-        $this->assertTrue($changeStream->valid());
         $this->assertSameDocument($changeStream->current()->_id, $changeStream->getResumeToken());
 
         $changeStream->next();
@@ -171,8 +169,7 @@ class WatchFunctionalTest extends FunctionalTestCase
 
         $this->insertDocument(['_id' => 1, 'x' => 'foo']);
 
-        $changeStream->next();
-        $this->assertTrue($changeStream->valid());
+        $this->advanceCursorUntilValid($changeStream);
 
         $expectedResult = [
             '_id' => $changeStream->current()->_id,
@@ -188,8 +185,7 @@ class WatchFunctionalTest extends FunctionalTestCase
 
         $this->insertDocument(['_id' => 2, 'x' => 'bar']);
 
-        $changeStream->next();
-        $this->assertTrue($changeStream->valid());
+        $this->advanceCursorUntilValid($changeStream);
 
         $expectedResult = [
             '_id' => $changeStream->current()->_id,
@@ -395,6 +391,8 @@ class WatchFunctionalTest extends FunctionalTestCase
 
     public function testRewindMultipleTimesWithResults()
     {
+        $this->skipIfIsShardedCluster('Cursor needs to be advanced multiple times and can\'t be rewound afterwards.');
+
         $operation = new Watch($this->manager, $this->getDatabaseName(), $this->getCollectionName(), [], $this->defaultOptions);
         $changeStream = $operation->execute($this->getPrimaryServer());
 
@@ -487,8 +485,7 @@ class WatchFunctionalTest extends FunctionalTestCase
 
         $this->insertDocument(['_id' => 1, 'x' => 'foo']);
 
-        $changeStream->next();
-        $this->assertTrue($changeStream->valid());
+        $this->advanceCursorUntilValid($changeStream);
 
         $expectedResult = [
             '_id' => $changeStream->current()->_id,
@@ -507,8 +504,7 @@ class WatchFunctionalTest extends FunctionalTestCase
 
         $this->insertDocument(['_id' => 2, 'x' => 'bar']);
 
-        $changeStream->next();
-        $this->assertTrue($changeStream->valid());
+        $this->advanceCursorUntilValid($changeStream);
 
         $expectedResult = [
             '_id' => $changeStream->current()->_id,
@@ -523,6 +519,8 @@ class WatchFunctionalTest extends FunctionalTestCase
 
     public function testResumeMultipleTimesInSuccession()
     {
+        $this->skipIfIsShardedCluster('getMore may return empty response before periodicNoopIntervalSecs on sharded clusters.');
+
         $operation = new Watch($this->manager, $this->getDatabaseName(), $this->getCollectionName(), [], $this->defaultOptions);
         $changeStream = $operation->execute($this->getPrimaryServer());
 
@@ -656,8 +654,7 @@ class WatchFunctionalTest extends FunctionalTestCase
 
         $this->insertDocument(['_id' => 1, 'x' => 'foo']);
 
-        $changeStream->next();
-        $this->assertTrue($changeStream->valid());
+        $this->advanceCursorUntilValid($changeStream);
         $this->assertSame(0, $changeStream->key());
 
         $changeStream->next();
@@ -676,8 +673,7 @@ class WatchFunctionalTest extends FunctionalTestCase
 
         $this->insertDocument(['_id' => 2, 'x' => 'bar']);
 
-        $changeStream->next();
-        $this->assertTrue($changeStream->valid());
+        $this->advanceCursorUntilValid($changeStream);
         $this->assertSame(1, $changeStream->key());
     }
 
@@ -693,8 +689,7 @@ class WatchFunctionalTest extends FunctionalTestCase
         $changeStream->rewind();
         $this->assertFalse($changeStream->valid());
 
-        $changeStream->next();
-        $this->assertTrue($changeStream->valid());
+        $this->advanceCursorUntilValid($changeStream);
 
         $expectedResult = [
             '_id' => $changeStream->current()->_id,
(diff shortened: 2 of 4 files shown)


alcaeus:
Note: build failure is because of a segmentation fault. I highly suspect it's for the same reason that @jmikola is investigating in #638.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -380,6 +377,8 @@ private function assertStartAtOperationTime(TimestampInterface $expectedOperatio
 
     public function testRewindMultipleTimesWithResults()
     {
+        $this->skipIfIsShardedCluster('Cursor needs multiple getMore commands prohibiting rewinding it.');
```
jmikola: Can this be rephrased? It's not clear to me why we're actually skipping this test. In particular, "commands prohibiting rewinding it" is difficult to parse.

I assume the conflict is that the test expects a single call to `next()` to advance to a valid position, and that doesn't fly with sharded clusters. But if `rewind()` is always a NOP, why can't we change the first `next()` call to `advanceCursorUntilValid()`? Wouldn't that allow up to five attempts to call `next()` and return if we reach a valid position (i.e. first result document)? The subsequent `next()` could also use `advanceCursorUntilValid()`, but I think we could also rely on `next()` _not_ to invoke a `getMore` there since both result documents should have been returned in the same batch.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -123,19 +124,17 @@ function (array $event) use (&$events) {
...
             },
             function (array $event) use (&$events) {
                 $events[] = $event;
             }
         );
 
-        $this->assertCount(1, $events);
-        $this->assertSame('getMore', $events[0]['started']->getCommandName());
-        $postBatchResumeToken = $this->getPostBatchResumeTokenFromReply($events[0]['succeeded']->getReply());
+        $lastEvent = array_pop($events);
```
jmikola: Alternatively, we could just swap `$events = [];` for `$lastEvent` and let the second callable overwrite the variable, but I'm fine sticking to the array for a minimal diff and consistency with other methods.

I might suggest a non-empty assertion before we call `array_pop()`, though.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -508,6 +505,8 @@ public function testNoChangeAfterResumeBeforeInsert()
 
     public function testResumeMultipleTimesInSuccession()
     {
+        $this->skipIfIsShardedCluster('Cursor needs multiple getMore commands prohibiting rewinding it.');
```
jmikola: In contrast to `testRewindMultipleTimesWithResults()`, this test calls `next()` and asserts that the position is _not_ valid. That requires us to trust the exact nature of executing a single `getMore` on a replica set. I can see why that wouldn't be reliable with sharded clusters, since `getMore` could come back empty even if there are results.

I'm still a bit confused by this message, though. Would "getMore may return empty response before periodicNoopIntervalSecs on sharded clusters" be more accurate?

tests/Operation/WatchFunctionalTest.php
```diff
@@ -749,9 +745,10 @@ public function testNonResumableErrorCodes($errorCode)
     public function provideNonResumableErrorCodes()
     {
         return [
-            [136], // CappedPositionLost
-            [237], // CursorKilled
-            [11601], // Interrupted
+            'CappedPositionLost' => [136],
+            'CursorKilled' => [237],
+            'Interrupted' => [11601],
+            'ChangeStreamFatalError' => [280],
```
jmikola: Reminder to remove this once https://github.com/mongodb/mongo-php-driver/pull/1018 is merged and [PHPC-1419](https://jira.mongodb.org/browse/PHPC-1419) is released.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -900,12 +897,23 @@ public function testMaxAwaitTimeMS()
...
-        $startTime = microtime(true);
-        $changeStream->next();
-        $duration = microtime(true) - $startTime;
-        $this->assertLessThan($pivot, $duration);
+         * server should not block since a document has been inserted.
+         * For sharded clusters, we have to repeat the getMore iteration until
+         * the cursor is valid since the first getMore commands after an insert
+         * may not return any data. Only the time of the last getMore command is
+         * taken. */
+        for ($i = 0; $i < 10; $i++) {
```
jmikola: Any reason to use 10 attempts here vs. the five that you used for `advanceCursorUntilValid()`?

On a separate note, I think a more conservative change would be something like:

```
$attempts = $this->isShardedCluster() ? 5 : 1;
```

And then use `$attempts` in the `for` loop. That would ensure replica sets only ever make one attempt. Granted, I don't expect that a replica set would ever execute multiple `getMore` commands unless we broke this test in some other way.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -924,17 +932,22 @@ public function testRewindExtractsResumeTokenAndNextResumes()
...
         $operation = new Watch($this->manager, $this->getDatabaseName(), $this->getCollectionName(), [], $options);
         $changeStream = $operation->execute($this->getPrimaryServer());
-        $this->assertSame($resumeToken, $changeStream->getResumeToken());
+        $this->assertEquals($resumeToken, $changeStream->getResumeToken());
 
         $changeStream->rewind();
-        $this->assertTrue($changeStream->valid());
+
+        if ($this->isShardedCluster()) {
+            $this->advanceCursorUntilValid($changeStream);
```
jmikola: Even though `rewind()` should never execute a `getMore` after [PHPLIB-451](https://jira.mongodb.org/browse/PHPLIB-451), the `aggregate` command on the sharded cluster is not guaranteed to return any initial results for the same reason that `getMore` may not (due to waiting for a periodic noop oplog entry). Is that correct? If so, I think a comment might be prudent here (and in the two instances below), since this is a bit more complicated than other places where you were able to simply replace `next()` and a valid assertion with `advanceCursorUntilValid()` directly.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -1010,18 +1026,22 @@ public function testStartAfterOption()
...
-        $this->assertTrue($changeStream->valid());
+        $this->advanceCursorUntilValid($changeStream);
 
         $resumeToken = $changeStream->current()->_id;
 
         $options = $this->defaultOptions + ['startAfter' => $resumeToken];
         $operation = new Watch($this->manager, $this->getDatabaseName(), $this->getCollectionName(), [], $options);
         $changeStream = $operation->execute($this->getPrimaryServer());
-        $this->assertSame($resumeToken, $changeStream->getResumeToken());
+        $this->assertEquals($resumeToken, $changeStream->getResumeToken());
```
jmikola: Was this change actually necessary to resolve a failure or is this just a preemptive change?

I believe `assertSame()` would have worked here, since the spec calls for initializing the resume token from the corresponding option and [`Watch::getInitialResumeToken()`](https://github.com/mongodb/mongo-php-library/blob/e0a2b89974139834e1737f658f58e7f77c6314ae/src/Operation/Watch.php#L339) returns the option directly. The only way I imagine `assertEquals()` to be necessary here is if the `startAfter` option was also returned via the `postBatchResumeToken` of the initial, empty response from `aggregate`. But even if a sharded cluster returned an empty `aggregate` response for some reason (e.g. waiting for periodic noop), I don't see how it'd return our resume token without also include the very same document from which we pulled it.

On a separate note: This test closely mirrors `testResumeAfterOption()`, which doesn't assert that `$changeStream->getResumeToken()` matches the `resumeAfter` option provided to `watch()`. It only asserts that the first document returned is what we expect. I don't object to asserting that the initial resume token is also what we expect, but I'd propose you also make the same change to `testResumeAfterOption()` for consistency. Lastly, if `assertSame()` really doesn't work here I'd suggest using `assertSameDocument()` over `assertEquals()`, since we use that in various other places throughout this test file.

tests/Operation/WatchFunctionalTest.php
alcaeus: I'll reword the skip message. Using `advanceCursorUntilValid` causes the internal cursor to throw an exception once we call `rewind`, as it has advanced already.

We may be able to fix this by changing the `isRewindNop` functionality in `ChangeStreamIterator`, but the underlying cursor definitely counts as advanced, prohibiting a `rewind` after multiple calls to `next` on the change stream. I'll investigate this separately to get to the bottom of it

tests/Operation/WatchFunctionalTest.php
alcaeus: Good point. Not storing all events definitely makes sense in this test.

tests/Operation/WatchFunctionalTest.php
alcaeus: Your analysis is correct, just like your suggestion for a skip message conveys this a lot better than my attempt.

tests/Operation/WatchFunctionalTest.php
alcaeus: This change was necessary, as the initial aggregate command returns a `postBatchResumeToken`, which always takes precedence over `startAfter` or `resumeAfter` options. While this resume token has the same value as the one in the `startAfter` option, it is not the same object as it was returned by the database. Hence the change to `assertEquals` instead of `assertSame`.

alcaeus:
Review feedback addressed in separate commit. I'll squash once the build is stable and after updating to a newer driver release that fixes PHPC-1419.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -924,17 +933,25 @@ public function testRewindExtractsResumeTokenAndNextResumes()
...
-        $changeStream->next();
-        $this->assertTrue($changeStream->valid());
+        $this->advanceCursorUntilValid($changeStream);
 
         $resumeToken = $changeStream->current()->_id;
         $options = ['resumeAfter' => $resumeToken] + $this->defaultOptions;
         $operation = new Watch($this->manager, $this->getDatabaseName(), $this->getCollectionName(), [], $options);
         $changeStream = $operation->execute($this->getPrimaryServer());
-        $this->assertSame($resumeToken, $changeStream->getResumeToken());
+        $this->assertEquals($resumeToken, $changeStream->getResumeToken());
```
jmikola: `assertSameDocument()`?

tests/FunctionalTestCase.php
```diff
@@ -334,9 +334,6 @@ protected function skipIfChangeStreamIsNotSupported()
                 if (! $this->isShardedClusterUsingReplicasets()) {
                     $this->markTestSkipped('$changeStream is only supported with replicasets');
                 }
-
-                // Temporarily skip tests because of an issue with change streams in the driver
-                $this->markTestSkipped('$changeStreams currently don\'t on replica sets');
```
jmikola: The PR itself LGTM, but I realized that removing this skip re-introduced the test failures.

I assume those were not happening in previous Travis builds since this skip was first introduced in https://github.com/mongodb/mongo-php-library/commit/2008bc7733c993d80055accc62cddd6d50b5d0ac as part of https://github.com/mongodb/mongo-php-library/pull/655?

We can discuss offline about whether to merge as-is for now and defer sorting out the double-free for replica set and sharded replica set tests in [PHPLIB-432](https://jira.mongodb.org/browse/PHPLIB-432).

alcaeus: Yes and no. Not sure what made the errors appear consistently, but the skip only affected sharded clusters, not the replicaset build. With this branch, the replicaset build always fails while the build on a sharded cluster at least passes sometimes.

Since this entire PR deals with fixing tests on sharded clusters, I don't see a reason why to merge this without running the tests on a sharded cluster. I'd first like to investigate the cause for the double free some more before just plain skipping those tests that cause the failures.

tests/Operation/WatchFunctionalTest.php
alcaeus: Removed along with the corresponding functionality in `ChangeStream` now that we're using RC-1.

jmikola:
@alcaeus: Regarding the unexpected passing builds despite the lingering double-free issue ([PHPLIB-432](https://jira.mongodb.org/browse/PHPLIB-432)), I noticed that the job for this most recent, passing [build](https://travis-ci.org/mongodb/mongo-php-library/jobs/580855357) ran on PHP 7.3.9 (built August 30). IIRC, previous [failed builds](https://travis-ci.org/mongodb/mongo-php-library/jobs/533836053) were using 7.3.2 (built February 11).

I wasn't able to find the Travis build history for the branch you were using to research this issue, but you may want to check that to see if your recently passing builds also display a new version. If so, it's possible this was fixed upstream; however, I don't see anything in https://github.com/travis-ci/travis-ci/issues or https://github.com/travis-ci/php-src-builder to suggest it was an intentional fix.

alcaeus:
I'm still observing the failure on PHP 7.3.9, see the most recent build log for example: https://travis-ci.com/alcaeus/mongo-php-library/jobs/231388623#L239. This branch was rebased to pull in latest changes from this PR, which interestingly enough no longer shows this behaviour, neither in the PR build on travis-ci.org nor in my branch build on travis-ci.com. I was afraid that the behaviour might only exhibit itself on travis-ci.com (since that uses different infrastructure IIRC), but that doesn't seem to be the case either.

Restarting the build a couple of times did not bring those failures back, so I'm even more at a loss than before.

alcaeus:
@jmikola it looks like #694 got rid of the use-after-free issue we were seeing on previous builds. I re-ran the offending builds multiple times and was no longer able to provoke that behaviour.

PR #662, opened 2019-08-07
PHPLIB-457: Audit and implement any missing change stream spec prose tests

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-457

### DIFF
--- a/src/Operation/Watch.php
+++ b/src/Operation/Watch.php
@@ -250,7 +250,7 @@ class Watch implements Executable, /* @internal */ CommandSubscriber
     {
         return new ChangeStream(
             $this->createChangeStreamIterator($server),
-            function($resumeToken) { return $this->resume($resumeToken); }
+            function($resumeToken, $hasAdvanced) { return $this->resume($resumeToken, $hasAdvanced); }
         );
     }
 
@@ -333,10 +333,11 @@ class Watch implements Executable, /* @internal */ CommandSubscriber
      *
      * @see https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst#resume-process
      * @param array|object|null $resumeToken
+     * @param bool              $hasAdvanced
      * @return ChangeStreamIterator
      * @throws InvalidArgumentException
      */
-    private function resume($resumeToken = null)
+    private function resume($resumeToken = null, $hasAdvanced = false)
     {
         if (isset($resumeToken) && ! is_array($resumeToken) && ! is_object($resumeToken)) {
             throw InvalidArgumentException::invalidType('$resumeToken', $resumeToken, 'array or object');
@@ -347,12 +348,14 @@ class Watch implements Executable, /* @internal */ CommandSubscriber
         // Select a new server using the original read preference
         $server = $this->manager->selectServer($this->aggregateOptions['readPreference']);
 
+        $resumeOption = isset($this->changeStreamOptions['startAfter']) && !$hasAdvanced ? 'startAfter' : 'resumeAfter';
+
         unset($this->changeStreamOptions['resumeAfter']);
         unset($this->changeStreamOptions['startAfter']);
         unset($this->changeStreamOptions['startAtOperationTime']);
 
         if ($resumeToken !== null) {
-            $this->changeStreamOptions['resumeAfter'] = $resumeToken;
+            $this->changeStreamOptions[$resumeOption] = $resumeToken;
         }
 
         if ($resumeToken === null && $this->operationTime !== null) {
--- a/tests/Operation/WatchFunctionalTest.php
+++ b/tests/Operation/WatchFunctionalTest.php
@@ -1382,6 +1382,105 @@ class WatchFunctionalTest extends FunctionalTestCase
         $this->assertSame($resumeToken, $changeStream->getResumeToken());
     }
 
+    /**
+     * Prose test: "$changeStream stage for ChangeStream started with startAfter
+     * against a server >=4.1.1 that has not received any results yet MUST
+     * include a startAfter option and MUST NOT include a resumeAfter option
+     * when resuming a change stream."
+     */
+    public function testResumingChangeStreamWithoutPreviousResultsIncludesStartAfterOption()
+    {
+        if (version_compare($this->getServerVersion(), '4.1.1', '<')) {
+            $this->markTestSkipped('Testing resumeAfter and startAfter can only be tested on servers >= 4.1.1');
+        }
+
+        $operation = new Watch($this->manager, $this->getDatabaseName(), $this->getCollectionName(), [], $this->defaultOptions);
+        $changeStream = $operation->execute($this->getPrimaryServer());
+
+        $this->insertDocument(['x' => 1]);
+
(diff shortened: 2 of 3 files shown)


tests/Operation/WatchFunctionalTest.php
```diff
@@ -1190,6 +1197,266 @@ public function testSessionFreed()
...
+        );
+
+        $this->assertSame(1, $commandCount);
+    }
+
+    /**
+     * Prose test: "ChangeStream will perform server selection before attempting
+     * to resume, using initial readPreference"
+     */
+    public function testOriginalReadPreferenceIsPreservedOnResume()
```
jmikola: Since this test requires a secondary, you may want to lead with:

```php
try {
    $this->manager->selectServer(new ReadPreference('secondary'));
} catch (ConnectionTimeoutException $e) {
    $this->markTestSkipped('Secondary is not available');
}
```

We do this in DocumentationExamplesTest once for something that uses a secondary read preference. I believe that's the only place, so up to you if you'd like to make a common skip method in FunctionalTestCase for this.

Other instances of secondary-related read preferences are in ClientTest.php, CollectionFunctionalTest, and DatabaseFunctionalTest, but they don't actually require a secondary to exist.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -1190,6 +1197,266 @@ public function testSessionFreed()
...
+
+    /**
+     * Prose test
+     * For a ChangeStream under these conditions:
+     * - Running against a server <4.0.7.
+     * - The batch is empty or has been iterated to the last document.
+     * Expected result:
+     * - getResumeToken must return the _id of the last document returned if one exists.
+     * - getResumeToken must return resumeAfter from the initial aggregate if the option was specified.
+     * - If resumeAfter was not specified, the getResumeToken result must be empty.
```
jmikola: I see the first two assertions in the test, but is this third case being tested?

tests/Operation/WatchFunctionalTest.php
```diff
@@ -1190,6 +1197,266 @@ public function testSessionFreed()
...
+
+        $changeStream->next();
+        $this->assertNotSame($previousCursorId, $changeStream->getCursorId());
+
+        $getCursor = Closure::bind(
+            function () {
+                return $this->iterator->getInnerIterator();
+            },
+            $changeStream,
+            ChangeStream::class
```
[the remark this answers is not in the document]
alcaeus: Yes, this allows private access without having to deal with reflection :)

src/Operation/Watch.php
```diff
@@ -347,12 +348,14 @@ private function resume($resumeToken = null)
         // Select a new server using the original read preference
         $server = $this->manager->selectServer($this->aggregateOptions['readPreference']);
 
+        $resumeOption = isset($this->changeStreamOptions['startAfter']) && !$hasAdvanced ? 'startAfter' : 'resumeAfter';
```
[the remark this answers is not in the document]
alcaeus: That is correct. The spec change clarified that drivers have to use `startAfter` if the original `aggregate` command used that option and no documents were returned. I'll reference PHPLIB-411 in the commit message for that second commit.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -1190,6 +1197,266 @@ public function testSessionFreed()
...
+    {
+        if (version_compare($this->getServerVersion(), '4.0.0', '<')) {
+            $this->markTestSkipped('failCommand is not supported');
+        }
+
+        $operation = new Watch($this->manager, $this->getDatabaseName(), $this->getCollectionName(), [], $this->defaultOptions);
+
+        $aggregateCommands = [];
+
+        $this->configureFailPoint([
```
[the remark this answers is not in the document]
alcaeus: With #657, the `\MongoDB\Tests\FunctionalTestCase::getUri` helper was changed to extract a single mongos node from a URL containing multiple nodes to avoid issues with `failCommand` being run on a different server than any subsequent commands. Spec tests that specify `useMultipleMongoses: true` recreate the client with the original connection string to properly test that behaviour. There's also #659 which re-enables change stream tests on sharded clusters which required some more changes to the way we handle change streams in tests. Since that pull request currently triggers a segfault on shutdown in the driver, I'd revisit tests on sharded clusters after merging this PR.

Furthermore, changes necessary to test sharded transactions for [PHPLIB-374](https://jira.mongodb.org/browse/PHPLIB-374) will allow us to create fail points on a specific server, which can also be used to work around the limitation of `failCommand` on sharded clusters.

tests/Operation/WatchFunctionalTest.php
alcaeus: That is checked in the first assertion to `$changeStream->getResumeToken()` directly after executing the `watch` operation. It was easier testing that directly after running the operation since we have a change stream with an empty batch and without a `resumeAfter` option.

tests/Operation/WatchFunctionalTest.php
alcaeus: I added a `catch` block to the `selectServer` statement below where we select the secondary server for the operation. While we could add a `getSecondaryServer()` method similar to `getPrimaryServer`, I'd rather make skipping the test if no secondary was found explicit instead of hiding it in a helper method.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -1190,6 +1197,266 @@ public function testSessionFreed()
...
+
+        $this->assertCount(2, $aggregateCommands);
+
+        $aggregateCommands = array_map(
+            function (array $aggregateCommand) {
+                // Remove resume options from the changestream document
+                if (isset($aggregateCommand['pipeline'][0]->{'$changeStream'})) {
+                    $aggregateCommand['pipeline'][0]->{'$changeStream'} = array_diff_key(
+                        (array) $aggregateCommand['pipeline'][0]->{'$changeStream'},
+                        ['resumeAfter' => false, 'startAfter' => false, 'startAtOperationTime' => false]
```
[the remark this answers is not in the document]
alcaeus: I've added assertions that the first command contains none of the `resumeAfter`, `startAfter` and `startAtOperationTime`, while the second command needs to contain at least one of them. After that, we still compare the rest of the pipeline to make sure they're identical, minus the aforementioned options.

tests/Operation/WatchFunctionalTest.php
jmikola: > I'd rather make skipping the test if no secondary was found explicit instead of hiding it in a helper method.

Ah, I was suggesting a `skipIfNoSecondary()` helper, which would be similar to `skipIfChangeStreamIsNotSupported()` or `skipIfTransactionsAreNotSupported()`. I wasn't suggesting a new `getSecondaryServer()` method that skips the test if it fails to select a secondary.

In any event, this change looks good.

tests/Operation/WatchFunctionalTest.php
jmikola: Ah, I missed that line. LGTM!

PR #665, opened 2019-08-09
PHPLIB-462: ensure compatibility with PHP 7.4

alcaeus:
While we're not going to be running any PHP 7.4 features for a while, we at least want to make sure that we test the library on PHP 7.4. This requires upgrading to newer PHPUnit releases to avoid errors due to the deprecation of the `__toString` methods in reflection classes.

Upgrading to PHPUnit 7 or newer is non-trivial, as support for PHP 5 was dropped and PHPUnit introduces type hints. To cope with that, we use `symfony/phpunit-bridge` to provide compatibility layers for the `TestCase` class. Once https://github.com/symfony/symfony/pull/33078 is merged, we can also drop the constraint traits from this packages. However, since the traits are marked as `@internal` we don't need to wait for a merge of that pull request and can drop them once they're available in `symfony/phpunit-bridge`.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -2,15 +2,10 @@
 
 <phpunit
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/4.3/phpunit.xsd"
-         backupGlobals="false"
-         backupStaticAttributes="false"
+         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/8.3/phpunit.xsd"
+         beStrictAboutOutputDuringTests="true"
+         beStrictAboutChangesToGlobalState="true"
          colors="true"
-         convertErrorsToExceptions="true"
-         convertNoticesToExceptions="true"
-         convertWarningsToExceptions="true"
-         stopOnFailure="false"
-         syntaxCheck="false"
          bootstrap="tests/bootstrap.php"
 >
 
--- a/tests/Collection/CrudSpecFunctionalTest.php
+++ b/tests/Collection/CrudSpecFunctionalTest.php
@@ -300,7 +300,7 @@ class CrudSpecFunctionalTest extends FunctionalTestCase
                 break;
 
             case 'bulkWrite':
-                $this->assertInternalType('array', $expectedResult);
+                $this->assertIsArray($expectedResult);
                 $this->assertInstanceOf(\MongoDB\BulkWriteResult::class, $actualResult);
 
                 if (isset($expectedResult['deletedCount'])) {
@@ -357,7 +357,7 @@ class CrudSpecFunctionalTest extends FunctionalTestCase
 
             case 'deleteMany':
             case 'deleteOne':
-                $this->assertInternalType('array', $expectedResult);
+                $this->assertIsArray($expectedResult);
                 $this->assertInstanceOf(\MongoDB\DeleteResult::class, $actualResult);
 
                 if (isset($expectedResult['deletedCount'])) {
@@ -375,7 +375,7 @@ class CrudSpecFunctionalTest extends FunctionalTestCase
                 break;
 
             case 'insertMany':
-                $this->assertInternalType('array', $expectedResult);
+                $this->assertIsArray($expectedResult);
                 $this->assertInstanceOf(\MongoDB\InsertManyResult::class, $actualResult);
 
                 if (isset($expectedResult['insertedCount'])) {
@@ -391,7 +391,7 @@ class CrudSpecFunctionalTest extends FunctionalTestCase
                 break;
 
             case 'insertOne':
-                $this->assertInternalType('array', $expectedResult);
+                $this->assertIsArray($expectedResult);
                 $this->assertInstanceOf(\MongoDB\InsertOneResult::class, $actualResult);
 
                 if (isset($expectedResult['insertedCount'])) {
@@ -409,7 +409,7 @@ class CrudSpecFunctionalTest extends FunctionalTestCase
             case 'replaceOne':
             case 'updateMany':
             case 'updateOne':
-                $this->assertInternalType('array', $expectedResult);
+                $this->assertIsArray($expectedResult);
                 $this->assertInstanceOf(\MongoDB\UpdateResult::class, $actualResult);
 
                 if (isset($expectedResult['matchedCount'])) {
--- a/tests/Collection/CrudSpecFunctionalTest.php
+++ b/tests/Collection/CrudSpecFunctionalTest.php
@@ -11,6 +11,7 @@ use MongoDB\Operation\FindOneAndReplace;
 use IteratorIterator;
 use LogicException;
 use MultipleIterator;
+use Symfony\Bridge\PhpUnit\SetUpTearDownTrait;
 
 /**
  * CRUD spec functional tests.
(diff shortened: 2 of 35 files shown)


phpunit.xml.dist
```diff
@@ -2,15 +2,10 @@
 
 <phpunit
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/4.3/phpunit.xsd"
-         backupGlobals="false"
-         backupStaticAttributes="false"
+         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/8.3/phpunit.xsd"
+         beStrictAboutOutputDuringTests="true"
+         beStrictAboutChangesToGlobalState="true"
```
jmikola: Am I correct that this option was introduced in 4.6 (see: [blog post](https://thephp.cc/news/2015/04/phpunit-4-6-and-global-state)), even though it was never [documented](https://phpunit.de/manual/4.6/en/appendixes.configuration.html)? I'm not sure about `beStrictAboutOutputDuringTests`, but there are some online references to that dating back to [2015](https://www.drupal.org/project/drupal/issues/2105583#comment-8966483) so I assume that's fine.

Just wanted to confirm that these will apply to all PHPUnit versions we use for testing.

tests/Compat/PolyfillAssertTrait.php
```diff
@@ -0,0 +1,25 @@
...
+use Symfony\Bridge\PhpUnit\Legacy\PolyfillAssertTrait as SymfonyPolyfillAssertTrait;
+
+$r = new ReflectionClass(Assert::class);
+if (! $r->hasMethod('assertEqualsWithDelta')) {
+    /**
+     * @internal
+     */
+    trait PolyfillAssertTrait
+    {
+        use SymfonyPolyfillAssertTrait;
```
jmikola: Is there a particular reason that [`PolyfillAssertTrait`](https://github.com/symfony/symfony/blob/4.4/src/Symfony/Bridge/PhpUnit/Legacy/PolyfillAssertTrait.php) doesn't have a compat shim, like [`SetUpTearDownTrait`](https://github.com/symfony/symfony/blob/4.4/src/Symfony/Bridge/PhpUnit/SetUpTearDownTrait.php)?

Is that because it really is only relevant to older PHPUnit versions, so Symfony is careful not to use it in newer branches?

tests/Collection/CrudSpecFunctionalTest.php
```diff
@@ -19,9 +20,11 @@
  */
 class CrudSpecFunctionalTest extends FunctionalTestCase
 {
+    use SetUpTearDownTrait;
+
     private $expectedCollection;
 
-    public function setUp()
+    private function doSetUp()
```
jmikola: Just to confirm, can we still chain to `parent::setUp()` below despite this method being renamed to `doSetUp()`?

phpunit.xml.dist
alcaeus: The XSD option was added in https://github.com/sebastianbergmann/phpunit/commit/9101f51cad93f97bea7b277007127635653eadbf#diff-365e6053dce410a51198fc699549f0ae, which was first included in 4.0.0. The comments in the code indicate that the functionality itself was added in 3.8.0, so we should be safe here.

The reason this was added is that I regenerated a new configuration file using `phpunit --generate-configuration` to account for changes in the default configuration.

tests/Compat/PolyfillAssertTrait.php
alcaeus: I talked to @nicolas-grekas about this. The assertion polypill is marked as `@internal` because they only use it to patch PHPUnit classes in their simple-phpunit binary: https://github.com/symfony/symfony/blob/a9ace363899373f4fed96f32cd38fc94187666d2/src/Symfony/Bridge/PhpUnit/bin/simple-phpunit.php#L154..L157. With the myriad of PHPUnit versions Symfony needs to support, it's too risky for them to have a single trait and let people use that in classes extending from `PHPUnit\Framework\Assert`, as it would then override the methods already shipped by PHPUnit. Patching the original class works around this issue as the code in `PHPUnit\Framework\Assert` then overrides the trait code.

In our case, I didn't want to patch PHPUnit code and with the PHPUnit versions we currently support, the single trait is good enough for us as all new assertion methods were added in or after PHPUnit 7. That's why we declare our own trait, which is empty if the PHPUnit version is new enough, as it allows us to always include our polyfill trait in the test classes.

There is still a certain risk using internal code from Symfony, but given that they need to keep the trait around as long as they support PHPUnit < 7 I feel rather confident there: dropping older PHPUnit versions for them won't be possible until Symfony 4.4 is EOL, which will be in November 2023. I hope we'll be able to get rid of support for PHP 5 before then. However, if you'd rather not use an internal class we can always duplicate the polyfill to our library, of course keeping the original copyright intact.

tests/Collection/CrudSpecFunctionalTest.php
alcaeus: Yes, that is still possible. The `parent::setUp()` call in `doSetUp` will always go to the parent class. In this case, the parent (`FunctionalTestCase`) also includes the trait, so the call will again land in the shim and call the `doSetUp` method that is written in `FunctionalTestCase`. Chaining the setup/teardown methods is still possible as before, we just have to remember to use the polyfill trait and write `doSetUp` methods until we can get rid of PHPUnit < 8.

alcaeus:
> [symfony/symfony#33078](https://github.com/symfony/symfony/pull/33078) appears to have been merged already.

Yes, I originally included the code I PR'ed to Symfony in this library, but now that it has been merged we no longer need to include it ourselves.
 
> Was that related to using `^4.4@dev` in `composer.json`, or do we still need that until a tagged release is made upstream?
Symfony 4.4 will be released in November, so we have to use an `@dev` release until then. I don't like relying on pre-release versions, but in this case it's the only option we have if we want to test on PHP 7.4 without PHPUnit causing deprecation errors and tripping over them itself. PHPUnit 7.5.15 will fix this (once released), but that version wouldn't be installable without changes either so I decided to update to the newest version where possible. Once Symfony 4.4 is released, I'll update the version constraint to require a stable release.

tests/Collection/CrudSpecFunctionalTest.php
nicolas-grekas: I confirm: the trait has been designed so that only the signature of the method is changed, and not its body.

tests/Compat/PolyfillAssertTrait.php
nicolas-grekas: Or you could use simple-phpunit :)
But I agree, the risk is low.

jmikola: Thanks for the deep explanation! I'm content with this.

@nicolas-grekas: Thanks for chiming in. `simple-phpunit` looks interesting (first I've seen of it) and is probably something (along with the bridge) that I could have benefited from a good while back :smile:. I'll defer to @alcaeus about whether it's worth migrating over to that at some later point.

jmikola:
> Once Symfony 4.4 is released, I'll update the version constraint to require a stable release.

Please make a PHPLIB ticket as a reminder to do so (if you haven't already).

alcaeus:
Created [PHPLIB-465](https://jira.mongodb.org/browse/PHPLIB-465) to track changing the dev requirement after the release of Symfony 4.4.

PR #667, opened 2019-08-13
PHPLIB-467: Connections survive primary step-down

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-467

### DIFF
--- a/mongo-orchestration/replica_sets/replicaset-old.json
+++ b/mongo-orchestration/replica_sets/replicaset-old.json
@@ -15,7 +15,6 @@
                 "smallfiles": true
             },
             "rsParams": {
-                "priority": 99,
                 "tags": {
                     "ordinal": "one",
                     "dc": "pa"
@@ -36,7 +35,6 @@
                 "smallfiles": true
             },
             "rsParams": {
-                "priority": 1.1,
                 "tags": {
                     "ordinal": "two",
                     "dc": "nyc"
--- a/tests/SpecTests/PrimaryStepDownSpecTest.php
+++ b/tests/SpecTests/PrimaryStepDownSpecTest.php
@@ -1,0 +1,281 @@
+<?php
+
+namespace MongoDB\Tests\SpecTests;
+
+use IteratorIterator;
+use MongoDB\Client;
+use MongoDB\Collection;
+use MongoDB\Driver\Command;
+use MongoDB\Driver\Exception\BulkWriteException;
+use MongoDB\Driver\Exception\Exception as DriverException;
+use MongoDB\Driver\ReadPreference;
+use MongoDB\Driver\Server;
+use MongoDB\Driver\WriteConcern;
+use MongoDB\Operation\BulkWrite;
+use MongoDB\Tests\CommandObserver;
+use Symfony\Bridge\PhpUnit\SetUpTearDownTrait;
+use UnexpectedValueException;
+use function current;
+
+/**
+ * @see https://github.com/mongodb/specifications/tree/master/source/connections-survive-step-down/tests
+ */
+class PrimaryStepDownSpecTest extends FunctionalTestCase
+{
+    use SetUpTearDownTrait;
+
+    const INTERRUPTED_AT_SHUTDOWN = 11600;
+    const NOT_MASTER = 10107;
+    const SHUTDOWN_IN_PROGRESS = 91;
+
+    /** @var Client */
+    private $client;
+
+    /** @var Collection */
+    private $collection;
+
+    private function doSetUp()
+    {
+        parent::setUp();
+
+        $this->client = new Client(static::getUri(), ['retryWrites' => false, 'heartbeatFrequencyMS' => 500, 'serverSelectionTimeoutMS' => 20000, 'serverSelectionTryOnce' => false]);
+
+        $this->dropAndRecreateCollection();
+        $this->collection = $this->client->selectCollection($this->getDatabaseName(), $this->getCollectionName());
+    }
+
+    /**
+     * @see https://github.com/mongodb/specifications/tree/master/source/connections-survive-step-down/tests#id10
+     */
+    public function testNotMasterKeepsConnectionPool()
+    {
+        $runOn = [(object) ['minServerVersion' => '4.1.11', 'topology' => [self::TOPOLOGY_REPLICASET]]];
+        $this->checkServerRequirements($runOn);
+
+        // Set a fail point
+        $this->configureFailPoint([
+            'configureFailPoint' => 'failCommand',
+            'mode' => ['times' => 1],
+            'data' => [
+                'failCommands' => ['insert'],
+                'errorCode' => self::NOT_MASTER,
+            ],
+        ]);
+
+        $totalConnectionsCreated = $this->getTotalConnectionsCreated();
+
+        // Execute an insert into the test collection of a {test: 1} document.
+        try {
+            $this->insertDocuments(1);
+        } catch (BulkWriteException $e) {
+            // Verify that the insert failed with an operation failure with 10107 code.
+            $this->assertSame(self::NOT_MASTER, $e->getCode());
+        }
+
+        // Execute an insert into the test collection of a {test: 1} document and verify that it succeeds.
+        $result = $this->insertDocuments(1);
+        $this->assertSame(1, $result->getInsertedCount());
+
+        // Verify that the connection pool has not been cleared
+        $this->assertSame($totalConnectionsCreated, $this->getTotalConnectionsCreated());
+    }
+
+    /**
+     * @see https://github.com/mongodb/specifications/tree/master/source/connections-survive-step-down/tests#id11
+     */
+    public function testNotMasterResetConnectionPool()
+    {
+        $runOn = [(object) ['minServerVersion' => '4.0.0', 'maxServerVersion' => '4.0.999', 'topology' => [self::TOPOLOGY_REPLICASET]]];
+        $this->checkServerRequirements($runOn);
+
+        // Set a fail point
+        $this->configureFailPoint([
+            'configureFailPoint' => 'failCommand',
+            'mode' => ['times' => 1],
+            'data' => [
+                'failCommands' => ['insert'],
+                'errorCode' => self::NOT_MASTER,
+            ],
+        ]);
+
+        $totalConnectionsCreated = $this->getTotalConnectionsCreated();
+
+        // Execute an insert into the test collection of a {test: 1} document.
+        try {
+            $this->insertDocuments(1);
+        } catch (BulkWriteException $e) {
+            // Verify that the insert failed with an operation failure with 10107 code.
+            $this->assertSame(self::NOT_MASTER, $e->getCode());
+        }
+
+        // Verify that the connection pool has been cleared
+        $this->assertSame($totalConnectionsCreated + 1, $this->getTotalConnectionsCreated());
+    }
+
+    /**
+     * @see https://github.com/mongodb/specifications/tree/master/source/connections-survive-step-down/tests#id12
+     */
+    public function testShutdownResetConnectionPool()
+    {
+        $runOn = [(object) ['minServerVersion' => '4.0.0']];
+        $this->checkServerRequirements($runOn);
+
+        // Set a fail point
+        $this->configureFailPoint([
+            'configureFailPoint' => 'failCommand',
+            'mode' => ['times' => 1],
+            'data' => [
+                'failCommands' => ['insert'],
+                'errorCode' => self::SHUTDOWN_IN_PROGRESS,
+            ],
+        ]);
+
+        $totalConnectionsCreated = $this->getTotalConnectionsCreated();
+
+        // Execute an insert into the test collection of a {test: 1} document.
+        try {
+            $this->insertDocuments(1);
+        } catch (BulkWriteException $e) {
+            // Verify that the insert failed with an operation failure with 91 code.
+            $this->assertSame(self::SHUTDOWN_IN_PROGRESS, $e->getCode());
+        }
+
+        // Verify that the connection pool has been cleared
+        $this->assertSame($totalConnectionsCreated + 1, $this->getTotalConnectionsCreated());
+    }
+
+    /**
+     * @see https://github.com/mongodb/specifications/tree/master/source/connections-survive-step-down/tests#id13
+     */
+    public function testInterruptedAtShutdownResetConnectionPool()
+    {
+        $runOn = [(object) ['minServerVersion' => '4.0.0']];
+        $this->checkServerRequirements($runOn);
+
+        // Set a fail point
+        $this->configureFailPoint([
+            'configureFailPoint' => 'failCommand',
+            'mode' => ['times' => 1],
+            'data' => [
+                'failCommands' => ['insert'],
+                'errorCode' => self::INTERRUPTED_AT_SHUTDOWN,
+            ],
+        ]);
+
+        $totalConnectionsCreated = $this->getTotalConnectionsCreated();
+
+        // Execute an insert into the test collection of a {test: 1} document.
+        try {
+            $this->insertDocuments(1);
+        } catch (BulkWriteException $e) {
+            // Verify that the insert failed with an operation failure with 11600 code.
+            $this->assertSame(self::INTERRUPTED_AT_SHUTDOWN, $e->getCode());
+        }
+
+        // Verify that the connection pool has been cleared
+        $this->assertSame($totalConnectionsCreated + 1, $this->getTotalConnectionsCreated());
+    }
+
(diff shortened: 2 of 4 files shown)


tests/SpecTests/PrimaryStepDownSpecTest.php
```diff
@@ -0,0 +1,251 @@
...
+        }
+
+        throw new UnexpectedValueException('Could not determine number of total connections');
+    }
+
+    private function waitForMasterReelection()
+    {
+        for ($i = 0; $i < 10; $i++) {
+            try {
+                $this->insertDocuments(1);
```
jmikola: Alternatively, could we just invoke server selection until a primary is returned?

https://github.com/mongodb/specifications/tree/master/source/connections-survive-step-down/tests#do-we-need-to-wait-for-re-election-after-the-first-test suggests that might be sufficient.

Also, I would suggest using a sleep of 0.5 seconds, as that is the non-configurable [minHeartbeatFrequencyMS](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#minheartbeatfrequencyms) value and theoretically smallest interval that a single-threaded driver would pick up topology changes.

tests/SpecTests/PrimaryStepDownSpecTest.php
```diff
@@ -0,0 +1,251 @@
...
+            $this->assertSame(10107, $e->getCode());
+        }
+
+        // Verify that the connection pool has been cleared
+        $this->assertSame($totalConnectionsCreated + 1, $this->getTotalConnectionsCreated());
+    }
+
+    public function testShutdownResetConnectionPool()
+    {
+        $runOn = [(object) ['minServerVersion' => '4.0.0', 'topology' => [self::TOPOLOGY_REPLICASET]]];
```
jmikola: https://github.com/mongodb/specifications/tree/master/source/connections-survive-step-down/tests#shutdown-in-progress-reset-connection-pool states that this and "Interrupted at shutdown - Reset Connection Pool" should be run on all server versions. Other tests only report a version number, which is a bit vague.

I assume that the getMore test requires a replica set since it invokes a `replSetStepDown` command, but should these other tests run on standalones and sharded clusters? For a sharded cluster, the test will need to use a single mongos in its URI to ensure the fail point is properly applied/used.

tests/SpecTests/PrimaryStepDownSpecTest.php
```diff
@@ -0,0 +1,251 @@
...
+use MongoDB\Driver\Command;
+use MongoDB\Driver\Exception\BulkWriteException;
+use MongoDB\Driver\Exception\Exception as DriverException;
+use MongoDB\Driver\ReadPreference;
+use MongoDB\Driver\WriteConcern;
+use MongoDB\Operation\BulkWrite;
+use MongoDB\Tests\CommandObserver;
+use UnexpectedValueException;
+
+class PrimaryStepDownSpecTest extends FunctionalTestCase
```
jmikola: I might suggest a `@see` link to https://github.com/mongodb/specifications/tree/master/source/connections-survive-step-down/tests in a class-level doc block.

tests/SpecTests/PrimaryStepDownSpecTest.php
alcaeus: I've previously called `Manager::selectServer()` to fetch a server, but this always returns the previous primary even when the replica set does not have a primary yet. I'm not sure why libmongoc isn't picking up on the topology change. Even calling `Manager::selectServer()` directly after the `insert` failed with a `not master` error returns the same server, with it still identifying as a primary. This also happens when sleeping for a while to ensure `minHeartBeatFrequencyMS` passed. We should definitely investigate this to ensure we don't miss any topology changes.

tests/SpecTests/PrimaryStepDownSpecTest.php
alcaeus: > https://github.com/mongodb/specifications/tree/master/source/connections-survive-step-down/tests#shutdown-in-progress-reset-connection-pool states that this and "Interrupted at shutdown - Reset Connection Pool" should be run on all server versions. Other tests only report a version number, which is a bit vague.

I was confused about that as well. The [original spec commit](https://github.com/mongodb/specifications/commit/cb1c5fea75a3d44f1368b073ea91aa00ba3d338c#diff-8caf393de24930695605b96dec90e5baR113) specified "version 4.0 or lower" for `testNotMasterResetConnectionPool` and "all supported server versions" for the two shutdown tests. This was later changed in https://github.com/mongodb/specifications/commit/1e043d971e9f7793e5e5d4e68985a17313df643a to exclude server versions below 4.0 as `failCommand` was only added in 4.0. That's why this there is a single version requirement instead of a range.

> I assume that the getMore test requires a replica set since it invokes a replSetStepDown command, but should these other tests run on standalones and sharded clusters? For a sharded cluster, the test will need to use a single mongos in its URI to ensure the fail point is properly applied/used.

I'd limit the `getMore` and both `not master` tests to replica sets as the `not master` error shouldn't ever appear on non-replicase topologies. The other tests definitely should run on standalones and sharded clusters as well.

tests/SpecTests/PrimaryStepDownSpecTest.php
jmikola: @alcaeus: The behavior you're describing would seem to conflict with ["not master" and "node is recovering"](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#not-master-and-node-is-recovering) in the SDAM spec.

> When the client sees a "not master" or "node is recovering" error it MUST replace the server's description with a default ServerDescription of type Unknown.
> ...
> For single-threaded clients, in the case of a "not master" or "node is shutting down" error, the client MUST mark the topology as "stale" so the next server selection scans all servers.

Feel free to open a CDRIVER ticket so we can look into that further.

tests/SpecTests/PrimaryStepDownSpecTest.php
```diff
@@ -0,0 +1,269 @@
...
+        // Verify that the connection pool has not been cleared
+        $this->assertSame($totalConnectionsCreated, $this->getTotalConnectionsCreated());
+    }
+
+    /**
+     * @see https://github.com/mongodb/specifications/tree/master/source/connections-survive-step-down/tests#id11
+     */
+    public function testNotMasterResetConnectionPool()
+    {
+        $runOn = [(object) ['maxServerVersion' => '4.0.999', 'topology' => [self::TOPOLOGY_REPLICASET]]];
```
jmikola: Since the test states "This test requires a replica set with server version 4.0.", should this also have `minServerVersion: 4.0.0` to ensure we only consider 4.0.x servers?

alcaeus: It is skipped implicitly in configureFailpoint (see https://github.com/mongodb/mongo-php-library/blob/e42136525c78020edde72c494cf6c91a2759ab33/tests/FunctionalTestCase.php#L144), but I’ll make it explicit 👍

jmikola:
FYI: I opened https://github.com/mongodb/specifications/pull/634 to clarify topology requirements in the spec test.

tests/SpecTests/PrimaryStepDownSpecTest.php
alcaeus: Correction: this was a bit of a failure on my part. The initial bulkWrite after a stepdown fails with a `not master` error, which correctly invalidates the topology and causes subsequent calls to `selectServer` to fail with a `ConnectionTimeoutException` until a new master has been elected. This slipped through the cracks since this test actually opens two connections to the server: one by instantiating the client itself, the other by `MongoDB\Tests\FunctionalTestCase` creating a new `Manager` instance in its own `setUp` method. Calling `MongoDB\Tests\FunctionalTestCase::getPrimaryServer` always uses this connection, which is not aware of the topology change since no writes are sent there.

I can change the logic in question, but this still needs an initial write operation on the old primary to get libmongoc to invalidate the topology and check for a new master. After this initial failure we can then perform server selection until a primary is returned.

One more thing to note is that I've observed elections taking quite long - sometimes up to 20 seconds. Is this expected behaviour?

jmikola: I now remember why the write is required. MongoDB 4.2 no longer drops the connection when executing `replSetStepDown` and instead returns a response to the client. This doesn't give SDAM a chance to change the server's state to unknown. Since the connection remains active, you effectively need to execute a write operation to trigger a "not master" error (or otherwise coerce some error with a fail point or network timeout).

I actually wrote about this in [DOCSP-4920](https://jira.mongodb.org/browse/DOCSP-4920), as it came up while we were writing this spec.

If you'd like to make a contribution to the spec, I think we'd do well to mention this in the answer for [Do we need to wait for re-election after the first test?](https://github.com/mongodb/specifications/tree/master/source/connections-survive-step-down/tests#do-we-need-to-wait-for-re-election-after-the-first-test). It currently instructs implementations to invoke server selection until a primary is available; however, that obviously won't work for MongoDB 4.2+.

> One more thing to note is that I've observed elections taking quite long - sometimes up to 20 seconds. Is this expected behaviour?

That certainly seems high. You can try tuning [`settings.electionTimeoutMillis`](https://docs.mongodb.com/manual/reference/replica-configuration/#rsconf.settings.electionTimeoutMillis). [Factors and Conditions that Affect Elections](https://docs.mongodb.com/manual/core/replica-set-elections/#factors-and-conditions-that-affect-elections) may have additional pointers.

You can probably also shave some more time by reducing the current 5-second timeout for the [`replSetStepDown`](https://docs.mongodb.com/manual/reference/command/replSetStepDown/) command.

tests/SpecTests/PrimaryStepDownSpecTest.php
```diff
@@ -0,0 +1,278 @@
...
+                try {
+                    $this->client->getManager()->selectServer(new ReadPreference('primary'));
+                    return;
+                } catch (ConnectionTimeoutException $e) {
+                    usleep(500000);
+                }
+            }
+        }
+
+        $this->fail('Expected primary to be re-elected within 20 seconds.');
```
jmikola: I don't believe the preceding loop is actually time-boxed to 20 seconds. The custom client uses the default URI, which may or may not have a custom `serverSelectionTimeoutMS` value set. I think it may make more sense to specifying the following URI array options:

 * heartbeatFrequencyMS = 500 (minimum allowed value, without patching libmongoc :wink:)
 * serverSelectionTimeoutMS = 20000 (if you really want to keep the 20-second timeout)
 * serverSelectionTryOnce = false (to allow `selectServer()` to loop internally instead of using `for` ourselves)

That should allow us to rewrite this method as:

```php
try {
    $this->insertDocuments(1);
    return;
} catch (DriverException $e) {
    $this->client->getManager()->selectServer(new ReadPreference('primary'));
}
```

tests/SpecTests/PrimaryStepDownSpecTest.php
```diff
@@ -0,0 +1,278 @@
...
+{
+    use SetUpTearDownTrait;
+
+    /** @var Client */
+    private $client;
+
+    /** @var Collection */
+    private $collection;
+
+    private function doSetUp()
```
jmikola: I was going to suggest dropping the collection under test on tearDown, since the tests write to it; however, I realized that we don't consistently do that across the test suite. For example: [`MongoDB\Tests\Collection\FunctionalTestCase::tearDown()`[(https://github.com/mongodb/mongo-php-library/blob/1.4.3/tests/Collection/FunctionalTestCase.php#L25) drops collections after a successful test and leaves behind data after a failure in case we need to inspect it.

I don't think this is worth doing here, but I did want to share the thought. Long term, I'd be more interested in auditing/profiling the test suite to see where we can improve execution time by _reducing_ things like drops. I know database drops can be fairly expensive, so those would be worth looking at to see if they serve a useful purpose.

tests/SpecTests/PrimaryStepDownSpecTest.php
alcaeus: > You can probably also shave some more time by reducing the current 5-second timeout for the [`replSetStepDown`](https://docs.mongodb.com/manual/reference/command/replSetStepDown/) command.

The 5 seconds are required by the spec, see https://github.com/mongodb/specifications/blob/master/source/connections-survive-step-down/tests/README.rst#getmore-iteration.

tests/SpecTests/PrimaryStepDownSpecTest.php
alcaeus: I'll leave it for now. As you mentioned, leaving data behind to check out after test failures can make sense, and the spec tests drop collections during setup anyways. At a previous job I've played around with only dropping collections that were written to during tests (using APM to figure out which collections were written to), but the performance gain from that was negligible IIRC.

alcaeus:
@jmikola it looks like the `getMoreIteration` test causes failures down the line, which unfortunately only appear when running the entire test suite. These errors disappear when skipping that test. Looking at the replica set configuration, it seems likely that there is in fact another step down sometime after running that initial step down: the first replica set member has a significantly higher priority, and in my local tests I was able to confirm that the first server becomes primary a short while after stepping down.

I'm not sure how to best fix this: giving the replica set members the same priority would definitely prevent another step down. The alternative is making the server step down for significantly longer than 5 seconds while hoping that there is no other step down during the tests. What do you think?

jmikola:
> the first replica set member has a significantly higher priority

IIRC, the priorities were configured as such for PHPC since it had some tests that select primary or secondary nodes and assert their tags. Those are inherently fragile tests and I'm dubious of how much value they provide, but beyond that I don't think they have any relevance to PHPLIB. It's only coming up now because we evidently based the PHPLIB cluster configs on what we already had for PHPC.

I'd suggest removing the priority definitions and confirming that it has no adverse affects on the rest of the PHPLIB test suite (AFAIK, it should not). Bumping the `replSetStepDown` value higher may work today, but IMO it would just be kicking the can further down the road and invite a failure if our test suite runs beyond that delay.

jmikola:
I just realized that this was originally a PHPC ticket but you chose to implement in PHPLIB. If so, let's move the JIRA issue over to PHPLIB and revise the commit message.

It bears mentioning that we do have some existing tests in PHPC that start their own servers with MO, but those are mainly creating standalones to test fail points. I think you made the right call to use our existing cluster for these tests. This could have probably been implemented as phpt tests in PHPC, but your decision to use PHPLIB does avoid a possible conflict with fragile replica set tag tests I mentioned above, so all is well there :)

tests/SpecTests/PrimaryStepDownSpecTest.php
```diff
@@ -0,0 +1,274 @@
...
+        $runOn = [(object) ['minServerVersion' => '4.1.11', 'topology' => [self::TOPOLOGY_REPLICASET]]];
+        $this->checkServerRequirements($runOn);
+
+        // Set a fail point
+        $this->configureFailPoint([
+            'configureFailPoint' => 'failCommand',
+            'mode' => ['times' => 1],
+            'data' => [
+                'failCommands' => ['insert'],
+                'errorCode' => 10107,
```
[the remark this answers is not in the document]
alcaeus: FYI, the file seems to have been moved to https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.yml.

mongo-orchestration/replica_sets/replicaset-old.json
```diff
@@ -15,7 +15,6 @@
                 "smallfiles": true
             },
             "rsParams": {
-                "priority": 99,
```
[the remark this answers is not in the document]
alcaeus: They pass on travis-ci, but fail on my machine. I'm currently investigating the cause of these failures.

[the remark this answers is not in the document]
alcaeus: I wasn't able to figure out build failures, but at least they're not confined to my machine anymore. Re-adding the priority fixes the current failures but reintroduces others.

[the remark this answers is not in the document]
alcaeus: I've taken a look at the libmongoc implementation of these tests, and they work exactly the same: grab the connection count from the current primary, perform an operation that fails, check connection count again. This works fine, unless the first test that triggers a stepdown ran. If it ran, connection counts are **sometimes** off, without knowing exactly why. Moving the stepdown test to the end solves this.

I feel bad moving the test, but I'm not sure what could be the cause of this. We don't manage connections at all, and the corresponding tests pass for libmongoc: https://evergreen.mongodb.com/task_log_raw/mongo_c_driver_gcc54_test_latest_replica_set_noauth_sasl_openssl_0acb4a539a62248c2a7f1e0b4a866340c463359c_19_08_27_00_33_53/0?type=T#L3694.

PR #668, opened 2019-08-21
PHPLIB-330: Implement checks for coding standard

alcaeus:
This introduces a coding standard that is automatically checked during the build using phpcs. The coding standard itself is derived from doctrine/coding-standard. I've tried to keep :
* All sniffs that would produce code incompatible with PHP 5.6 are excluded (strict typing, null coalesce operator, class constant visibilities, and most notably type hints)
* All sniffs that cause BC breaks are excluded (again, mostly type hints as well as class naming sniffs)
* Sniffs that would cause functional changes are excluded, amongst other:
  * Using known aliases of common PHP functions (e.g. `sizeof` instead of `count`)
  * Unused private elements: side effects of this have to be carefully reviewed
  * Yoda comparisons
  * Reducing code nesting in favour of early exit
  * Useless if conditions instead of returning directly
  * Static closures
  * Unused variables in closures
  * Using the equal operator
* A follow-up pull request will add type hints to all properties in classes. This was extracted from this pull request to keep the diffs manageable.

This PR was designed to not really have to be reviewed. Most fixes were applied automatically (see last commit), with only some changes done manually.

Note: I've made a point of disabling some sniffs in `DocumentationExamplesTest`. This was done to keep the examples "portable" and allow people to simply copy/paste them to their code, e.g. by preserving fully qualified namespaces.

### DIFF
--- a/.phpcs/autoload.php
+++ b/.phpcs/autoload.php
@@ -1,0 +1,15 @@
+<?php
+
+// Since doctrine/coding-standard requires PHP 7, we can't add it as a dependency
+// yet. This autoload file adds more information to the phpcs error message,
+// telling the user how they can fix the error presented to them by phpcs.
+if (! file_exists(__DIR__ . '/../vendor/doctrine/coding-standard')) {
+    echo <<<ERRORMESSAGE
+==============================================================================
+ERROR: Doctrine coding standard is not installed. To rectify this, please run:
+composer require --dev doctrine/coding-standard=^6.0
+==============================================================================
+
+
+ERRORMESSAGE;
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,20 @@ jobs:
 
     - stage: Smoke Testing
       php: "7.3"
+      env:
+        - CHECKS=phpunit
+    - stage: Smoke Testing
+      php: "7.1"
+      before_install: []
+      before_script:
+        - pecl install -f mongodb-${DRIVER_VERSION}
+        - composer require --no-update doctrine/coding-standard=^6.0
+        - composer install --no-interaction --no-progress --no-suggest ${COMPOSER_OPTIONS}
+      script: vendor/bin/phpcs
+      after_script: []
+      after_failure: []
+      env:
+        - CHECKS=phpcs
 
     # Test remaining supported PHP versions
     - stage: Test
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -1,0 +1,101 @@
+<?xml version="1.0"?>
+<ruleset>
+    <arg name="basepath" value="."/>
+    <arg name="extensions" value="php"/>
+    <arg name="parallel" value="80"/>
+    <arg name="cache" value=".phpcs-cache"/>
+    <arg name="colors" />
+
+    <!-- Ignore warnings, show progress of the run, and show sniff names -->
+    <arg value="nps"/>
+
+    <autoload>.phpcs/autoload.php</autoload>
+
+    <file>.phpcs</file>
+    <file>src</file>
+    <file>tests</file>
+
+    <rule ref="Doctrine">
+        <!-- Exclude sniffs that require newer PHP versions -->
+        <!-- Available with PHP 7.0 -->
+        <exclude name="SlevomatCodingStandard.TypeHints.DeclareStrictTypes" />
+        <!-- In addition to requiring PHP 7.0, this sniff will cause a significant amount of BC breaks. Proceed with caution! -->
+        <exclude name="SlevomatCodingStandard.TypeHints.TypeHintDeclaration" />
+        <exclude name="SlevomatCodingStandard.Exceptions.ReferenceThrowableOnly" />
+        <exclude name="SlevomatCodingStandard.ControlStructures.RequireNullCoalesceOperator" />
+
+        <!-- Available with PHP 7.1 -->
+        <exclude name="SlevomatCodingStandard.Classes.ClassConstantVisibility" />
+        <exclude name="SlevomatCodingStandard.PHP.ShortList.LongListUsed" />
+        <exclude name="SlevomatCodingStandard.TypeHints.NullableTypeForNullDefaultValue" />
+
+        <!-- No statement alignment so far -->
+        <exclude name="Generic.Formatting.MultipleStatementAlignment" />
+
+        <!-- Class naming sniffs are excluded to preserve BC -->
+        <exclude name="SlevomatCodingStandard.Classes.SuperfluousAbstractClassNaming" />
+        <exclude name="SlevomatCodingStandard.Classes.SuperfluousExceptionNaming" />
+        <exclude name="SlevomatCodingStandard.Classes.SuperfluousInterfaceNaming" />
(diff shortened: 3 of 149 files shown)


tests/FunctionalTestCase.php
```diff
@@ -341,13 +349,12 @@ protected function skipIfChangeStreamIsNotSupported()
 
     protected function skipIfCausalConsistencyIsNotSupported()
     {
-        switch ( $this->getPrimaryServer()->getType() )
-        {
+        switch ( $this->getPrimaryServer()->getType()) {
```
jmikola: I think the space between `(` and `$` can be deleted, since this trimmed the trailing space.

src/Operation/MapReduce.php
```diff
@@ -156,7 +156,7 @@ public function __construct($databaseName, $collectionName, JavascriptInterface
         }
 
         if (isset($options['finalize']) && ! $options['finalize'] instanceof JavascriptInterface) {
-            throw InvalidArgumentException::invalidType('"finalize" option', $options['finalize'], Javascript::class);
+            throw InvalidArgumentException::invalidType('"finalize" option', $options['finalize'], JavascriptInterface::class);
```
jmikola: Good catch!

tests/DocumentationExamplesTest.php
```diff
@@ -1241,12 +1266,12 @@ public function testTransactions_intro_example_1()
...
-        $client->hr->dropCollection('employees', ['writeConcern' => new \MongoDB\Driver\WriteConcern('majority')]);
-        $client->reporting->dropCollection('events', ['writeConcern' => new \MongoDB\Driver\WriteConcern('majority')]);
+        $client->hr->dropCollection('employees', ['writeConcern' => new WriteConcern('majority')]);
+        $client->reporting->dropCollection('events', ['writeConcern' => new WriteConcern('majority')]);
 
         /* Collections need to be created before a transaction starts */
-        $client->hr->createCollection('employees', ['writeConcern' => new \MongoDB\Driver\WriteConcern('majority')]);
-        $client->reporting->createCollection('events', ['writeConcern' => new \MongoDB\Driver\WriteConcern('majority')]);
+        $client->hr->createCollection('employees', ['writeConcern' => new WriteConcern('majority')]);
+        $client->reporting->createCollection('events', ['writeConcern' => new WriteConcern('majority')]);
```
jmikola: I think these should be left as-is.

tests/DocumentationExamplesTest.php
```diff
@@ -1404,12 +1441,12 @@ public function testTransactions_retry_example_3()
...
```
jmikola: I think these should be left as-is.

phpcs.xml.dist
```diff
@@ -0,0 +1,104 @@
...
+    <!-- Ignore warnings, show progress of the run, and show sniff names -->
+    <arg value="nps"/>
+
+    <autoload>.phpcs/autoload.php</autoload>
+
+    <file>.phpcs</file>
+    <file>src</file>
+    <file>tests</file>
+
+    <rule ref="Doctrine">
```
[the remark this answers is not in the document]
alcaeus: This pulls in all rules from the Doctrine coding standard and allows us to exclude/modify some of them. Since we just configure phpcs.xml our ruleset doesn't have a name, that would only happen if we create a ruleset.xml in the standardised directory structure, which isn't necessary if we're not going to distribute the coding standard.

.phpcs/autoload.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+// Since doctrine/coding-standard requires PHP 7, we can't add it as a dependency
+// yet. This autoload file adds more information to the phpcs error message,
+// telling the user how they can fix the error presented to them by phpcs.
+if (! file_exists(__DIR__ . '/../vendor/doctrine/coding-standard')) {
```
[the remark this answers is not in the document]
alcaeus: Since doctrine/coding-standard requires PHP 7.2, including it in `require-dev` would implicitly bump our PHP requirement to 7.2 on CI systems, breaking tests for older PHP versions. That's why the check stage requires it manually.

To make the generic error message from phpcs more helpful I decided to use the autoload functionality of phpcs to include this error message if doctrine/coding-standard wasn't installed. Now that you mention it, it's probably best to also mention this in the contribution docs.

.travis.yml
```diff
@@ -25,6 +25,20 @@ jobs:
 
     - stage: Smoke Testing
       php: "7.3"
+      env:
+        - CHECKS=phpunit
+    - stage: Smoke Testing
+      php: "7.1"
```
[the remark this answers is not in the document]
alcaeus: Some of the sniffs we use have feature detection where they disable/enable functionality based on PHP version. For example, the sniff requiring type hints does not enforce an `object` type hint on methods unless running on PHP 7.2 or later. Thus, it's always a good idea to target the lowest supported version for the coding standard in CI.

src/ChangeStream.php
```diff
@@ -202,7 +203,7 @@ private function onIteration($incrementKey)
 
         /* Return early if there is not a current result. Avoid any attempt to
          * increment the iterator's key. */
-        if (!$this->valid()) {
+        if (! $this->valid()) {
```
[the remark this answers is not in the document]
alcaeus: There is only a sniff to enforce a space after the negation operator, but not to enforce it before it, hence the different styles. The space after the opening parenthesis in function calls is forbidden by PSR-2, which serves as the base of doctrine/coding-standard. The space before is left over from me excluding a sniff for now (see comment below about control structure spacing)

src/GridFS/Bucket.php
```diff
@@ -19,17 +19,40 @@
...
+use function MongoDB\apply_type_map_to_document;
+use function MongoDB\BSON\fromPHP;
+use function MongoDB\BSON\toJSON;
+use function property_exists;
+use function sprintf;
+use function stream_context_create;
+use function stream_copy_to_stream;
+use function stream_get_meta_data;
+use function stream_get_wrappers;
+use function urlencode;
```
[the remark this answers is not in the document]
alcaeus: You'll get used to them. It greatly reduces eyestrain from seeing too many `\` all over the place. FWIW, if you give in and start using PhpStorm, it will do this for you automagically.

tests/FunctionalTestCase.php
alcaeus: I excluded that rule for now since I didn't know whether you want to keep the space before the negation operator. I'll drop the exclude and enforce the rule as per PSR-2.

tests/DocumentationExamplesTest.php
alcaeus: I've excluded namespace changes for all code within the doc examples, but changed everything outside of docs examples to match our coding standard. This code is outside of the examples so I've updated it to match the rest of the code base.

jmikola: My mistake. I originally thought this was inside rendered example code.

tests/DocumentationExamplesTest.php
jmikola: Likewise, I also wrongfully thought these were in rendered example code.

PR #669, opened 2019-08-22
Update to latest ext-mongodb and MongoDB server versions

alcaeus:
This pulls in alpha3 of the driver for travis-ci builds and updates the build pipeline to test against MongoDB 4.2.0 by default. MongoDB 4.0.12 is still tested as part of the "older server versions" sections in travis-ci.

This PR also fixes some leftover warnings from the PHPUnit upgrade that was done in #665.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ cache:
 
 env:
   global:
-    - DRIVER_VERSION=1.6.0alpha2
+    - DRIVER_VERSION=1.6.0alpha3
     - SERVER_DISTRO=ubuntu1604
     - SERVER_VERSION=4.0.10
     - DEPLOYMENT=STANDALONE
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ env:
   global:
     - DRIVER_VERSION=1.6.0alpha3
     - SERVER_DISTRO=ubuntu1604
-    - SERVER_VERSION=4.0.10
+    - SERVER_VERSION=4.2.0
     - DEPLOYMENT=STANDALONE
     - COMPOSER_OPTIONS=
 
@@ -44,7 +44,7 @@ jobs:
       env:
         - COMPOSER_OPTIONS=--prefer-lowest
 
-      # Test older standalone server versions (3.0-3.6)
+      # Test older standalone server versions (3.0-4.0)
     - stage: Test
       php: "7.0"
       dist: trusty
@@ -66,6 +66,10 @@ jobs:
       php: "7.0"
       env:
         - SERVER_VERSION=3.6.13
+    - stage: Test
+      php: "7.3"
+      env:
+        - SERVER_VERSION=4.0.12
 
     # Test other server configurations
     - stage: Test
@@ -93,13 +97,6 @@ jobs:
       php: "7.3"
       env:
         - DEPLOYMENT=SHARDED_CLUSTER_RS
-
-    # Test upcoming server versions
-    - stage: Test
-      php: "7.3"
-      env:
-        - SERVER_VERSION=4.2.0-rc1
-        - DEPLOYMENT=REPLICASET
 
 before_install:
   - pip install "mongo-orchestration>=0.6.7,<1.0" --user `whoami`
--- a/tests/Operation/AggregateFunctionalTest.php
+++ b/tests/Operation/AggregateFunctionalTest.php
@@ -217,7 +217,17 @@ class AggregateFunctionalTest extends FunctionalTestCase
                 $results = iterator_to_array($operation->execute($this->getPrimaryServer()));
 
                 $this->assertCount(1, $results);
-                $this->assertObjectHasAttribute('stages', current($results));
+                $result = current($results);
+
+                if ($this->isShardedCluster()) {
+                    $this->assertObjectHasAttribute('shards', $result);
+
+                    foreach ($result->shards as $shard) {
+                        $this->assertObjectHasAttribute('stages', $shard);
+                    }
+                } else {
+                    $this->assertObjectHasAttribute('stages', $result);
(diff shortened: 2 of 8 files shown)


.travis.yml
```diff
@@ -137,7 +135,7 @@ before_script:
   - ulimit -c unlimited -S
 
 script:
-  - export MONGODB_URI=`cat /tmp/uri.txt`
+  - export MONGODB_URI=`cat /tmp/uri.txt`$APPEND_URI
```
jmikola: PHPC does this in [`setup_mo.sh`](https://github.com/mongodb/mongo-php-driver/blob/1.6.0alpha3/.travis.scripts/setup_mo.sh#L12) as it writes to `/tmp/uri.txt`. That was based on prior art from some URI modifications it had to do for the `STANDALONE_SSL` environment.

I think one benefit of keeping this in `setup_mo.sh` is that we can ensure the options are appended if `setup_mo.sh` is executed from something other than `.travis.yml`. This may be relevant if we end up standardizing the MO scripts for Evergreen and Travis down the line. I've also run `setup_mo.sh` myself to launch the sever from time to time.

Beyond that, I'd actually love to make a change to MO itself so we can specify extra URI options in the cluster's JSON config file, but that's something further down the road.

tests/Operation/AggregateFunctionalTest.php
```diff
@@ -217,7 +217,17 @@ function() use ($pipeline, $options) {
                 $results = iterator_to_array($operation->execute($this->getPrimaryServer()));
 
                 $this->assertCount(1, $results);
-                $this->assertObjectHasAttribute('stages', current($results));
+                $result = current($results);
+
+                if ($this->isShardedCluster()) {
```
jmikola: We had sharded clusters in Travis before this PR, so was this test previously failing?

tests/Operation/WatchFunctionalTest.php
```diff
@@ -1295,7 +1296,7 @@ public function testErrorDuringAggregateCommandDoesNotCauseResume()
         $this->configureFailPoint([
             'configureFailPoint' => 'failCommand',
             'mode' => ['times' => 1],
-            'data' => ['failCommands' => ['aggregate'], 'errorCode' => self::NOT_MASTER],
+            'data' => ['failCommands' => ['aggregate'], 'errorCode' => self::INTERRUPTED],
```
jmikola: See: https://github.com/mongodb/specifications/pull/636

tests/SpecTests/ChangeStreamsSpecTest.php
```diff
@@ -18,6 +18,10 @@
  */
 class ChangeStreamsSpecTest extends FunctionalTestCase
 {
+    private static $incompleteTests = [
+        'change-streams-errors: Change Stream should error when _id is projected out' => 'PHPC-1419',
+    ];
```
jmikola: Was this test previously passing?

tests/Operation/AggregateFunctionalTest.php
alcaeus: I'm not sure why it started failing, I'll try to narrow it down to a certain commit. The difference made sense so I thought I'd update the test, but it did seem strange. I'll see what I can find out.

tests/SpecTests/ChangeStreamsSpecTest.php
alcaeus: It was passing, then it was failing after the update. Now after trying to reproduce it, it's no longer failing, so go figure. I'll unskip it and let travis decide what's going on.

FWIW, until [PHPC-1419](https://jira.mongodb.org/browse/PHPC-1419) is fixed we'll try to resume the change stream after `getMore` failed with the expected error, with the subsequent `aggregate` command failing as expected. This is fixed when we finish up sharing tests in #659.

tests/Operation/AggregateFunctionalTest.php
alcaeus: That's why I originally wanted to leave a comment on this yesterday, because I knew that I'd forget overnight. The test change is necessary as the output of `explain` is different in MongoDB 4.2. We might want to further adapt the test and check the server version as well, but for now this will do.

tests/SpecTests/ChangeStreamsSpecTest.php
alcaeus: The test fails on sharded clusters, but for different reasons.

On a sharded cluster backed by standalone servers, `$changeStream` obviously isn't supported. This causes an exception when trying to run the watch operation from the spec test, but that exception does not contain the `NonResumableChangeStreamError` (because it's not a change stream error) label, causing the test to fail.
This is caused because the spec requires this test to run on both replicaset and sharded topologies. Unfortunately, the spec does not differentiate between the two types of sharded clusters (backed entirely by replicasets or not), so this is something we may want to address in the specification in the future. However, since the spec test also require testing change streams on a standalone server to ensure an exception is thrown, we can't just skip these tests on unsupported platforms. I may have to hardcode an exception for this one particular test.

On a sharded cluster backed by replica sets, we're seeing the behaviour described in https://github.com/mongodb/mongo-php-library/pull/659#issue-304180039. In a nutshell, the first `getMore` call does not return any results as mongos hasn't received returnable results from all shards yet. A subsequent `getMore` would return a document, causing the `getMore` call to fail with the expected error. Even then, the test still fails because of [PHPC-1419](https://jira.mongodb.org/browse/PHPC-1419): the `NonResumableChangeStreamError` is not present in the exception.

With that in mind, I'd skip the test for now and properly fix it in #659 where we take care of other change stream tests as well.

FWIW, debugging this issue made me realise that PHPC-1419 only appears on sharded clusters. I'll investigate that further separately to figure out the cause of the bug.

PR #670, opened 2019-08-22
PHPLIB-418: Add the ability to specify a pipeline to an update command

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-418

### DIFF
--- a/src/Operation/ReplaceOne.php
+++ b/src/Operation/ReplaceOne.php
@@ -25,6 +25,7 @@ use MongoDB\UpdateResult;
 use function is_array;
 use function is_object;
 use function MongoDB\is_first_key_operator;
+use function MongoDB\is_pipeline;
 
 /**
  * Operation for replacing a single document with the update command.
@@ -77,6 +78,10 @@ class ReplaceOne implements Executable
 
         if (is_first_key_operator($replacement)) {
             throw new InvalidArgumentException('First key in $replacement argument is an update operator');
+        }
+
+        if (is_pipeline($replacement)) {
+            throw new InvalidArgumentException('$replacement argument is a pipeline');
         }
 
         $this->update = new Update(
--- a/src/functions.php
+++ b/src/functions.php
@@ -120,6 +120,47 @@ function is_first_key_operator($document)
 }
 
 /**
+ * Returns whether an update specification is a valid aggregation pipeline.
+ *
+ * @internal
+ * @param mixed $pipeline
+ * @return boolean
+ */
+function is_pipeline($pipeline)
+{
+    if (! is_array($pipeline)) {
+        return false;
+    }
+
+    if ($pipeline === []) {
+        return false;
+    }
+
+    $expectedKey = 0;
+
+    foreach ($pipeline as $key => $stage) {
+        if (! is_array($stage) && ! is_object($stage)) {
+            return false;
+        }
+
+        if ($expectedKey !== $key) {
+            return false;
+        }
+
+        $expectedKey++;
+        $stage = (array) $stage;
+        reset($stage);
+        $key = key($stage);
+
+        if (! isset($key[0]) || $key[0] !== '$') {
+            return false;
+        }
+    }
+
+    return true;
+}
+
+/**
  * Returns whether we are currently in a transaction.
  *
  * @internal
--- a/tests/FunctionsTest.php
+++ b/tests/FunctionsTest.php
@@ -10,6 +10,7 @@ use function MongoDB\create_field_path_type_map;
 use function MongoDB\generate_index_name;
 use function MongoDB\is_first_key_operator;
 use function MongoDB\is_mapreduce_output_inline;
+use function MongoDB\is_pipeline;
 
 /**
  * Unit tests for utility functions.
@@ -222,6 +223,35 @@ class FunctionsTest extends TestCase
                 ],
                 'field.$',
             ],
+        ];
+    }
+
+    /**
+     * @dataProvider providePipelines
+     */
+    public function testIsPipeline($expected, $pipeline)
+    {
+        $this->assertSame($expected, is_pipeline($pipeline));
+    }
+
+    public function providePipelines()
+    {
+        return [
+            'Not an array' => [false, (object) []],
+            'Empty array' => [false, []],
+            'Non-sequential indexes in array' => [false, [1 => ['$group' => []]]],
+            'Update document instead of pipeline' => [false, ['$set' => ['foo' => 'bar']]],
+            'Invalid pipeline stage' => [false, [['group' => []]]],
+            'Update with DbRef' => [false, ['x' => ['$ref' => 'foo', '$id' => 'bar']]],
+            'Valid pipeline' => [
+                true,
+                [
+                    ['$match' => ['foo' => 'bar']],
+                    ['$group' => ['_id' => 1]],
+                ],
+            ],
+            'False positive with DbRef in numeric field' => [true, ['0' => ['$ref' => 'foo', '$id' => 'bar']]],
+            'DbRef in numeric field as object' => [false, (object) ['0' => ['$ref' => 'foo', '$id' => 'bar']]],
         ];
     }
 }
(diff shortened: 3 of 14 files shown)


tests/SpecTests/crud/updateWithPipelines.json
```diff
@@ -1,4 +1,9 @@
 {
+  "runOn": [
```
alcaeus: See https://github.com/mongodb/specifications/pull/635.

src/Operation/ReplaceOne.php
```diff
@@ -76,6 +76,10 @@ public function __construct($databaseName, $collectionName, $filter, $replacemen
             throw new InvalidArgumentException('First key in $replacement argument is an update operator');
         }
 
+        if (\MongoDB\is_pipeline($replacement)) {
```
jmikola: I _think_ this makes sense, although I didn't find this discussed in the CRUD spec. In hindsight, I think https://github.com/mongodb/specifications/commit/7246066275c34a3609defd27527121a350cce3da#diff-5efd2e4ce61f4ccda0ce031099303996 for [SPEC-1267](https://jira.mongodb.org/browse/SPEC-1267) should have addressed this in the CRUD spec's [Update vs. Replace Validation](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#update-vs-replace-validation) section. It could also have added a failing test for `replaceOne` and replace via `BulkWrite`. Would you mind opening a spec ticket? Hopefully we can get to it during SPEC triage Monday afternoon (NYC time).

Although documents cannot contain keys prefixed with `$`, I was wondering if the following replacement document might be ambiguous (at least for our `is_pipeline()` function):

```php
[
    [ '$ref' => '<any string>' ],
    [ '$id' => '<any non-array value>' ],
]
```

This could be interpreted as a replacement document assigning a [DBRef](https://docs.mongodb.com/manual/reference/database-references/#dbrefs) to the document's top-level "0" field, or a pipeline with `$ref` and `$id` stages. I decided to test this in the `mongo` shell:

```
rs0:PRIMARY> db.foo.drop()
false

rs0:PRIMARY> db.foo.insert({"0":1})
WriteResult({ "nInserted" : 1 })

rs0:PRIMARY> db.foo.findOne()
{ "_id" : ObjectId("5d61d496637adc23cd5c4215"), "0" : 1 }

rs0:PRIMARY> db.foo.replaceOne({},[{"$ref":"foo","$id":"bar"}])
2019-08-24T20:26:51.491-0400 E  QUERY    [js] uncaught exception: Error: Cannot use pipeline-style updates in a replacement operation :
DBCollection.prototype.replaceOne@src/mongo/shell/crud_api.js:461:15
@(shell):1:1
```

That supports the validation you have here. Still, it's a client-side error and I was wondering what the server-side error might be if we actually passed this along to the `update` command. Does MongoDB 4.2 accept a BSON array as a valid replacement document in the [`update` command's `updates[].u` option](https://docs.mongodb.com/manual/reference/command/update/#update-command-u)? Recent server versions added stricter validation, but I believe in some older servers (perhaps in the 3.x range) a BSON array/document were accepted interchangeably. Let's find out:

```
rs0:PRIMARY> db.runCommand({update:"foo",updates:[{q:{},u:[{"$ref":"foo","$id":"bar"}]}]});
{
	"n" : 0,
	"nModified" : 0,
	"writeErrors" : [
		{
			"index" : 0,
			"code" : 40323,
			"errmsg" : "A pipeline stage specification object must contain exactly one field."
		}
	],
        ...
}
```

My DBRef example is a total edge case, but let's suppose an application actually did want to perform that replacement for some reason. Passing an actual BSON document (i.e. object cast on the PHP array) does the trick:

```
rs0:PRIMARY> db.runCommand({update:"foo",updates:[{q:{},u:{"0":{"$ref":"foo","$id":"bar"}}}]}); 
{
	"n" : 1,
	"nModified" : 1,
        ...
}
```

I also traced the code path for constructing the `update` command and we can see in [`_mongoc_write_command_update_append()`](https://github.com/mongodb/mongo-c-driver/blob/1.15.0/src/libmongoc/src/mongoc/mongoc-write-command.c#L121) that libmongoc always appends the `u` option as a document unless it's a pipeline. So even if a PHP array was used as a replacement document and managed to sneak past our update/replace validation, we could trust libmongoc to actually send a BSON document to the server.

src/functions.php
```diff
@@ -110,6 +110,40 @@ function is_first_key_operator($document)
...
+ * @param mixed $pipeline
+ * @return boolean
+ */
+function is_pipeline($pipeline)
+{
+    if (! is_array($pipeline)) {
+        return false;
+    }
+
+    if ($pipeline === []) {
```
jmikola: Continuing from my previous shell testing, it looks like MongoDB 4.2 accepts an empty pipeline and simply NOPs:

```
rs0:PRIMARY> db.runCommand({update:"foo",updates:[{q:{},u:[]}]});
{
	"n" : 1,
	"nModified" : 0,
        ...
}
```

Contrast this with an empty replacement document:

```
rs0:PRIMARY> db.runCommand({update:"foo",updates:[{q:{},u:{}}]});
{
	"n" : 1,
	"nModified" : 1,
	...
}
```

Is this function based on libmongoc's [`_mongoc_document_is_pipeline()`](https://github.com/mongodb/mongo-c-driver/blob/1.15.0/src/libmongoc/src/mongoc/mongoc-util.c#L530)? I see that also returns `false` if the argument has no elements.

I'm not arguing that this is incorrect. I actually think it's the most intuitive behavior, especially for PHP where users would most commonly pass `[]` as an empty replacement document; however, I'm curious if this was dictated by the spec. If not, perhaps we should mention it in the SPEC ticket I proposed opening in my earlier comment.

src/functions.php
```diff
@@ -110,6 +110,40 @@ function is_first_key_operator($document)
...
+        return false;
+    }
+
+    foreach ($pipeline as $stage) {
```
jmikola: Should we also check that the array's keys follow a continuous sequence from zero like `_mongoc_document_is_pipeline()`? Otherwise, I think this function might return `true` for an associative array (nevermind numeric) consisting of a single DBRef (e.g. `[ 'x' => [ '$ref' => 'foo', '$id' => 'bar' ]]`). That seems like a bug.

I may be wrong, but either way I think it'd be worth adding some unit tests in FunctionsTest.php.

src/Operation/ReplaceOne.php
alcaeus: Thank you for that detailed analysis. I've created a spec ticket to ask for clarification in the spec. From the result of your tests, I deduce that our handling is correct in this case with users having to explicitly pass an object for a replacement that you suggested above.

I agree that it's an edge case, especially considering that DBRef objects normally contain both `$ref` and `$id` keys. With that in mind, we could add logic to `MongoDB\is_pipeline` to assert that each stage in the potential pipeline only contains a single element; this would at least remove the client-side assertion when replacing a document with a single DBRef object, as we would no longer accept that as a valid aggregation pipeline. What do you think?

src/functions.php
alcaeus: > Is this function based on libmongoc's [`_mongoc_document_is_pipeline()`](https://github.com/mongodb/mongo-c-driver/blob/1.15.0/src/libmongoc/src/mongoc/mongoc-util.c#L530)? I see that also returns `false` if the argument has no elements.

That is correct. I assumed that it'd be wrong to treat an empty array as an aggregation pipeline, but I also specifically remember one test failing because of that, which is why I added the check after all.

src/functions.php
alcaeus: Good point. I've added checks for the expected key and also added this specific example as a test (among others).

tests/FunctionsTest.php
```diff
@@ -224,4 +225,31 @@ public function provideTypeMapValues()
...
+
+    public function providePipelines()
+    {
+        return [
+            'noArray' => [false, (object) []],
+            'emptyArray' => [false, []],
+            'missingIndexes' => [false, [1 => ['$group' => []]]],
+            'updateDocument' => [false, ['$set' => ['foo' => 'bar']]],
+            'invalidStage' => [false, [['group' => []]]],
+            'updateWithDbRef' => [false, ['x' => ['$ref' => 'foo', '$id' => 'bar']]],
```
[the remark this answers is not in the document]
alcaeus: Added. I also made the test names in the data provider more readable.

PR #675, opened 2019-09-02
PHPLIB-457: Add missing change stream prose test 14

alcaeus:
The logic of this is the same as in the test for the C driver ([see link](https://github.com/mongodb/mongo-c-driver/blob/9994d4568176ec1503c50c92e68a6b0f2ee2bf2a/src/libmongoc/tests/test-mongoc-change-stream.c#L2291..L2295)), which also uses `startAtOperationTime` to ensure a non-empty batch is returned when running without either `startAfter` and `resumeAfter` option.

### DIFF
--- a/tests/Operation/WatchFunctionalTest.php
+++ b/tests/Operation/WatchFunctionalTest.php
@@ -11,6 +11,7 @@ use MongoDB\Driver\Exception\ConnectionTimeoutException;
 use MongoDB\Driver\Exception\LogicException;
 use MongoDB\Driver\Exception\ServerException;
 use MongoDB\Driver\Manager;
+use MongoDB\Driver\Monitoring\CommandSucceededEvent;
 use MongoDB\Driver\ReadPreference;
 use MongoDB\Driver\WriteConcern;
 use MongoDB\Exception\ResumeTokenException;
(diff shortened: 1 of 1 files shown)


tests/Operation/WatchFunctionalTest.php
```diff
@@ -1404,6 +1405,67 @@ public function testGetResumeTokenReturnsOriginalResumeTokenOnEmptyBatch()
...
+            $this->assertInstanceOf(CommandSucceededEvent::class, $event['succeeded']);
+            $reply = $event['succeeded']->getReply();
+
+            $this->assertObjectHasAttribute('operationTime', $reply);
+            $lastOpTime = $reply->operationTime;
+        });
+
+        $this->insertDocument(['x' => 1]);
+
+        $changeStream->next();
```
jmikola: I think this should use `advanceCursorUntilValid()` from #659, since this test will also run on sharded clusters.

alcaeus: I was going to merge this before #659 since I'm still working on figuring out the cause for the double free. I'd like to go ahead with the merge then change this while rebasing #659.

jmikola: SGTM.

PR #677, opened 2019-09-02
PHPLIB-466: Support mongos pinning for sharded transactions

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-466

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -170,7 +170,7 @@ class Client
             $options['typeMap'] = $this->typeMap;
         }
 
-        $server = $this->manager->selectServer(new ReadPreference(ReadPreference::RP_PRIMARY));
+        $server = select_server($this->manager, new ReadPreference(ReadPreference::RP_PRIMARY), extract_session_from_options($options));
 
         if (! isset($options['writeConcern']) && server_supports_feature($server, self::$wireVersionForWritableCommandWriteConcern)) {
             $options['writeConcern'] = $this->writeConcern;
@@ -246,7 +246,7 @@ class Client
     public function listDatabases(array $options = [])
     {
         $operation = new ListDatabases($options);
-        $server = $this->manager->selectServer(new ReadPreference(ReadPreference::RP_PRIMARY));
+        $server = select_server($this->manager, new ReadPreference(ReadPreference::RP_PRIMARY), extract_session_from_options($options));
 
         return $operation->execute($server);
     }
@@ -311,7 +311,7 @@ class Client
             $options['readPreference'] = $this->readPreference;
         }
 
-        $server = $this->manager->selectServer($options['readPreference']);
+        $server = select_server($this->manager, $options['readPreference'], extract_session_from_options($options));
 
         if (! isset($options['readConcern']) && server_supports_feature($server, self::$wireVersionForReadConcern)) {
             $options['readConcern'] = $this->readConcern;
--- a/src/Operation/Watch.php
+++ b/src/Operation/Watch.php
@@ -40,6 +40,8 @@ use function is_object;
 use function is_string;
 use function MongoDB\Driver\Monitoring\addSubscriber;
 use function MongoDB\Driver\Monitoring\removeSubscriber;
+use function MongoDB\extract_session_from_options;
+use function MongoDB\select_server;
 use function MongoDB\server_supports_feature;
 
 /**
@@ -375,8 +377,11 @@ class Watch implements Executable, /* @internal */ CommandSubscriber
 
         $this->hasResumed = true;
 
-        // Select a new server using the original read preference
-        $server = $this->manager->selectServer($this->aggregateOptions['readPreference']);
+        /* Select a new server using the original read preference. While watch
+         * is not usable within transactions, we still check if there is a
+         * pinned session. This is to avoid an ambiguous error message about
+         * running a command on the wrong server. */
+        $server = select_server($this->manager, $this->aggregateOptions['readPreference'], extract_session_from_options($this->aggregateOptions));
 
         $resumeOption = isset($this->changeStreamOptions['startAfter']) && ! $hasAdvanced ? 'startAfter' : 'resumeAfter';
 
--- a/src/functions.php
+++ b/src/functions.php
@@ -19,6 +19,8 @@ namespace MongoDB;
 
 use Exception;
 use MongoDB\BSON\Serializable;
+use MongoDB\Driver\Manager;
+use MongoDB\Driver\ReadPreference;
 use MongoDB\Driver\Server;
 use MongoDB\Driver\Session;
 use MongoDB\Exception\InvalidArgumentException;
@@ -372,4 +374,33 @@ function with_transaction(Session $session, callable $callback, array $transacti
 {
     $operation = new WithTransaction($callback, $transactionOptions);
     $operation->execute($session);
+}
+
+/**
+ * Returns the session option if it is set and valid.
+ *
+ * @internal
+ * @param array $options
+ * @return Session|null
+ */
+function extract_session_from_options(array $options)
+{
+    if (! isset($options['session']) || ! $options['session'] instanceof Session) {
+        return null;
+    }
+
+    return $options['session'];
+}
+
+/**
+ * Performs server selection, respecting the server a session may be pinned to
+ *
+ * @internal
+ * @return Server
+ */
+function select_server(Manager $manager, ReadPreference $readPreference = null, Session $session = null)
+{
+    $server = $session !== null ? $session->getServer() : null;
+
+    return $server ?: $manager->selectServer($readPreference);
 }
--- a/tests/FunctionalTestCase.php
+++ b/tests/FunctionalTestCase.php
@@ -384,9 +384,16 @@ abstract class FunctionalTestCase extends TestCase
             $this->markTestSkipped('Transactions are not supported on standalone servers');
         }
 
-        // TODO: MongoDB 4.2 should support sharded clusters (see: PHPLIB-374)
         if ($this->isShardedCluster()) {
-            $this->markTestSkipped('Transactions are not supported on sharded clusters');
+            if (! $this->isShardedClusterUsingReplicasets()) {
+                $this->markTestSkipped('Transactions are not supported on sharded clusters without replica sets');
+            }
+
+            if (version_compare($this->getFeatureCompatibilityVersion(), '4.2', '<')) {
+                $this->markTestSkipped('Transactions are only supported on FCV 4.2 or higher');
+            }
+
+            return;
         }
 
         if (version_compare($this->getFeatureCompatibilityVersion(), '4.0', '<')) {
--- a/tests/FunctionalTestCase.php
+++ b/tests/FunctionalTestCase.php
@@ -151,7 +151,7 @@ abstract class FunctionalTestCase extends TestCase
      * @param array|stdClass $command configureFailPoint command document
      * @throws InvalidArgumentException if $command is not a configureFailPoint command
      */
-    protected function configureFailPoint($command)
+    public function configureFailPoint($command, Server $server = null)
     {
         if (! $this->isFailCommandSupported()) {
             $this->markTestSkipped('failCommand is only supported on mongod >= 4.0.0 and mongos >= 4.1.5.');
@@ -173,14 +173,16 @@ abstract class FunctionalTestCase extends TestCase
             throw new InvalidArgumentException('$command is not a configureFailPoint command');
         }
 
+        $failPointServer = $server ?: $this->getPrimaryServer();
+
         $operation = new DatabaseCommand('admin', $command);
-        $cursor = $operation->execute($this->getPrimaryServer());
+        $cursor = $operation->execute($failPointServer);
         $result = $cursor->toArray()[0];
 
         $this->assertCommandSucceeded($result);
 
         // Record the fail point so it can be disabled during tearDown()
-        $this->configuredFailPoints[] = $command->configureFailPoint;
+        $this->configuredFailPoints[] = [$command->configureFailPoint, $failPointServer];
     }
 
     /**
@@ -408,9 +410,7 @@ abstract class FunctionalTestCase extends TestCase
             return;
         }
 
-        $server = $this->getPrimaryServer();
-
-        foreach ($this->configuredFailPoints as $failPoint) {
+        foreach ($this->configuredFailPoints as list($failPoint, $server)) {
             $operation = new DatabaseCommand('admin', ['configureFailPoint' => $failPoint, 'mode' => 'off']);
             $operation->execute($server);
         }
--- a/tests/SpecTests/CommandExpectations.php
+++ b/tests/SpecTests/CommandExpectations.php
@@ -10,6 +10,7 @@ use MongoDB\Driver\Monitoring\CommandSubscriber;
 use MongoDB\Driver\Monitoring\CommandSucceededEvent;
 use MultipleIterator;
 use function count;
+use function in_array;
 use function key;
 use function MongoDB\Driver\Monitoring\addSubscriber;
 use function MongoDB\Driver\Monitoring\removeSubscriber;
@@ -36,6 +37,9 @@ class CommandExpectations implements CommandSubscriber
 
     /** @var boolean */
     private $ignoreExtraEvents = false;
+
+    /** @var string[] */
+    private $ignoredCommandNames = [];
 
     private function __construct(array $events)
     {
@@ -110,6 +114,14 @@ class CommandExpectations implements CommandSubscriber
         $o->ignoreCommandFailed = true;
         $o->ignoreCommandSucceeded = true;
 
+        /* Ignore the buildInfo and getParameter commands as they are used to
+         * check for the availability of configureFailPoint and are not expected
+         * to be called by any spec tests.
+         * configureFailPoint needs to be ignored as the targetedFailPoint
+         * operation will be caught by command monitoring and is also not
+         * present in the expected commands in spec tests. */
+        $o->ignoredCommandNames = ['buildInfo', 'getParameter', 'configureFailPoint'];
+
         return $o;
     }
 
@@ -120,7 +132,7 @@ class CommandExpectations implements CommandSubscriber
      */
     public function commandFailed(CommandFailedEvent $event)
     {
-        if ($this->ignoreCommandFailed || ($this->ignoreExtraEvents && count($this->actualEvents) === count($this->expectedEvents))) {
+        if ($this->ignoreCommandFailed || $this->isEventIgnored($event)) {
             return;
         }
 
@@ -134,7 +146,7 @@ class CommandExpectations implements CommandSubscriber
      */
     public function commandStarted(CommandStartedEvent $event)
     {
-        if ($this->ignoreCommandStarted || ($this->ignoreExtraEvents && count($this->actualEvents) === count($this->expectedEvents))) {
+        if ($this->ignoreCommandStarted || $this->isEventIgnored($event)) {
             return;
         }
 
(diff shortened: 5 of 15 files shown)


tests/SpecTests/RetryableWritesSpecTest.php
```diff
@@ -29,15 +28,13 @@ public function testRetryableWrites(stdClass $test, array $runOn = null, array $
             $this->markTestSkipped('Transaction numbers are only allowed on a replica set member or mongos (PHPC-1415)');
         }
 
-        if (isset($test->useMultipleMongoses) && $test->useMultipleMongoses && $this->isShardedCluster()) {
-            $this->manager = new Manager(static::getUri(true));
```
alcaeus: I realised way too late that we don't use `$this->manager` in these tests. This was exposed when testing session pinning, as running with a single mongos won't trigger pinning logic in libmongoc.

tests/SpecTests/TransactionsSpecTest.php
```diff
@@ -36,6 +37,9 @@ class TransactionsSpecTest extends FunctionalTestCase
     private static $incompleteTests = [
         'error-labels: add unknown commit label to MaxTimeMSExpired' => 'PHPC-1382',
         'error-labels: add unknown commit label to writeConcernError MaxTimeMSExpired' => 'PHPC-1382',
+        'mongos-recovery-token: commitTransaction retry fails on new mongos' => 'isMaster failpoints cannot be disabled',
```
alcaeus: I was not able to reliably clear the `failCommand` fail point if it's configured to fail a certain amounts of times. It seems that we don't trigger the fail point the expected number of times, which causes issues when removing the fail point (as `isMaster` will cause an error). However, even trying to remove the fail point in a loop did not always clear the fail point, which causes failures in following tests.

tests/SpecTests/TransactionsSpecTest.php
```diff
@@ -47,9 +51,9 @@ class TransactionsSpecTest extends FunctionalTestCase
         'transaction-options: readConcern snapshot in startTransaction options' => 'PHPLIB does not properly inherit readConcern for transactions',
     ];
 
-    private static function doSetUpBeforeClass()
+    private function doSetUp()
     {
-        parent::setUpBeforeClass();
+        parent::setUp();
```
alcaeus: This change makes the test more compliant with the spec, which requires that all sessions be killed before running a test. We still kill all sessions after a failing test.

tests/SpecTests/TransactionsSpecTest.php
jmikola: Reading the [script for this test](https://github.com/mongodb/specifications/blob/c08e1be322477aaefb8d42f8d50e30fbc15cc1d0/source/transactions/tests/mongos-recovery-token.yml#L177) and the comments therein, I wonder if it was written with multi-threaded SDAM in mind. I wasn't aware we had any tests failing `isMaster` commands before reading this.

It looks like this test dates back to the original spec change for recovery tokens in [SPEC-1168](https://jira.mongodb.org/browse/SPEC-1168).

Perhaps you can reach out to @bazile-clyde to see whether the C driver is able to test this with a multi- or single-threaded client. Multi-threaded clients would offload all `isMaster` activity to a separate background thread and socket, which means these `isMaster` failures might not affect the client under test at all.

tests/SpecTests/CommandExpectations.php
```diff
@@ -109,6 +113,7 @@ public static function fromTransactions(array $expectedEvents)
 
         $o->ignoreCommandFailed = true;
         $o->ignoreCommandSucceeded = true;
+        $o->ignoredCommandNames = ['buildInfo', 'getParameter', 'configureFailPoint'];
```
jmikola: The [transaction spec test README](https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst) doesn't talk about needing to filter out APM events for the `targetedFailPoint` operation. 

I wonder if that should be clarified in either [Command Started Events](https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst#command-started-events) or the [`targetedFailPoint`](https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst#id7) section. If the latter, the operation could specify that the fail point should be set via a different client to avoid polluting the list of observed APM events for the client under test.

Granted, PHP would still need to work around this on its own because we can only listen to events from a global context, but that may still be a worthwhile spec improvement.

As for the other two command here, I assume they're needed because `configureFailPoint()` will not execute a version check each time. I'd suggest adding a comment block above this line to explain why all three of these events are ignored.

tests/SpecTests/Operation.php
```diff
@@ -500,6 +504,35 @@ private function executeForSession(Session $session, Context $context)
...
+        switch ($this->name) {
+            case 'assertSessionPinned':
+                $session = $context->{$this->arguments['session']};
+                $test->assertNotNull($session->getServer());
+
+                return null;
+            case 'assertSessionTransactionState':
+                $session = $context->{$this->arguments['session']};
+                $expected = in_array($this->arguments['state'], ['in_progress', 'starting']);
+                $test->assertSame($expected, $session->isInTransaction());
```
jmikola: I previously brought up this issue for PHP in https://github.com/mongodb/specifications/pull/548#discussion_r285253695 for [SPEC-1195](https://jira.mongodb.org/browse/SPEC-1195). I think this is a good compromise, but we'd probably do well to leave a TODO and create a new PHPC ticket to suggest implementing some API to expose the session state. Whether we ultimately decide to do that, and lobby for the necessary libmongoc change, can be decided later but it'd be good to record the proposal and any decision in a formal record.

tests/SpecTests/Operation.php
```diff
@@ -500,6 +504,35 @@ private function executeForSession(Session $session, Context $context)
         }
     }
 
+    private function executeForTestRunner(FunctionalTestCase $test, Context $context)
+    {
+        switch ($this->name) {
+            case 'assertSessionPinned':
+                $session = $context->{$this->arguments['session']};
+                $test->assertNotNull($session->getServer());
```
jmikola: I realize it's just as effective in practice, but would an `assertInstanceOf` check be more appropriate here?

tests/SpecTests/Operation.php
```diff
@@ -500,6 +504,35 @@ private function executeForSession(Session $session, Context $context)
         }
     }
 
+    private function executeForTestRunner(FunctionalTestCase $test, Context $context)
+    {
+        switch ($this->name) {
+            case 'assertSessionPinned':
+                $session = $context->{$this->arguments['session']};
```
jmikola: Rather than dynamically access the Context property, I think this method should start by calling `replaceArgumentSessionPlaceholder()` as is done in other `execute*()` methods within this class.

That will replace the placeholder value (e.g. "session0") with the appropriate session object and also raise a meaningful exception if the placeholder is invalid. After calling that method, we can trust that the argument is a Session instance.

Should this have been picked up by CS, or do the sniffs ignore dynamic attribute access?

tests/SpecTests/TransactionsSpecTest.php
```diff
@@ -128,17 +132,17 @@ public static function assertCommandMatches(stdClass $expected, stdClass $actual
      */
     public function testTransactions(stdClass $test, array $runOn = null, array $data, $databaseName = null, $collectionName = null)
     {
+        $this->skipIfTransactionsAreNotSupported();
```
jmikola: Shouldn't we be able to rely on the `runOn` requirements to enforce this? I don't believe we used the common skip method in the change stream spec test runner, which is otherwise used by `WatchFunctionalTest`.

tests/SpecTests/TransactionsSpecTest.php
```diff
@@ -214,7 +218,7 @@ protected function createTestCollection()
         $context = $this->getContext();
 
         $database = $context->getDatabase();
-        $database->createCollection($context->collectionName, $context->defaultWriteOptions);
+        $database->createCollection($context->collectionName, ['writeConcern' => new WriteConcern('majority')] + $context->defaultWriteOptions);
```
jmikola: Why was this necessary? Doesn't `Context::fromTransactions()` already apply `w:majority` for the default WC options?

tests/SpecTests/TransactionsSpecTest.php
```diff
@@ -151,7 +155,7 @@ public function testTransactions(stdClass $test, array $runOn = null, array $dat
         $databaseName = isset($databaseName) ? $databaseName : $this->getDatabaseName();
         $collectionName = isset($collectionName) ? $collectionName : $this->getCollectionName();
 
-        $context = Context::fromTransactions($test, $databaseName, $collectionName);
+        $context = Context::fromTransactions($test, $databaseName, $collectionName, $useMultipleMongoses);
```
jmikola: This looks fine. Alternatively, we could inject the `$uri` instead and insulate `Context::fromTransactions()` (and `fromRetryableWrites()`) from having to think about `$useMultipleMongoses`. Up to you.

tests/SpecTests/RetryableWritesSpecTest.php
jmikola: Had we used `$this->manager`, which is set from `MongoDB\Tests\FunctionalTestCase::setUp()`, I don't think that would have worked either. That manager is constructed purely from the URI environment variable and wouldn't have considered the spec tests' `useMultipleMongoses` option.

alcaeus: That's why I originally added this conditional to recreate the manager with multiple mongooses if specified in the spec test, not realising that it still didn't have any effect.

src/Client.php
```diff
@@ -170,7 +170,7 @@ public function dropDatabase($databaseName, array $options = [])
             $options['typeMap'] = $this->typeMap;
         }
 
-        $server = $this->manager->selectServer(new ReadPreference(ReadPreference::RP_PRIMARY));
+        $server = select_server($this->manager, new ReadPreference(ReadPreference::RP_PRIMARY), extract_session_from_options($options));
```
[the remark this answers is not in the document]
alcaeus: While I'm all for reducing line width, I'm not sure if we want to switch the constants for strings. Using the constants allows additional features in IDEs, such as "Find usages" to highlight usages of that specific string without encountering false positives in comments or other strings (e.g. exception messages).

Either way, I'd like to do this in a separate pull request, so I've created [PHPLIB-475](https://jira.mongodb.org/browse/PHPLIB-475) to track this and consolidate discussion on whether we want to use strings or constants consistently throughout the code.

src/Operation/Watch.php
```diff
@@ -376,7 +378,7 @@ private function resume($resumeToken = null, $hasAdvanced = false)
         $this->hasResumed = true;
 
         // Select a new server using the original read preference
-        $server = $this->manager->selectServer($this->aggregateOptions['readPreference']);
+        $server = select_server($this->manager, $this->aggregateOptions['readPreference'], extract_session_from_options($this->aggregateOptions));
```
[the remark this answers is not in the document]
alcaeus: A watch operation is currently not allowed in a transaction:
> MongoDB\Driver\Exception\CommandException : Operation not permitted in transaction :: caused by :: Operation not permitted in transaction :: caused by :: $changeStream cannot run with a readConcern other than 'majority', or in a multi-document transaction. Current readConcern: { readConcern: { level: "snapshot" } }

I originally changed all instances of server selection within the library with the new logic so that it's consistent. In this case, changing server selection back would mean that running the watch operation on a pinned session will error because we're running the command on the wrong server, instead of informing the user that the operation can't be done in a transaction. Even in this technically impossible case, I'd still pick the pinned server to have the correct exception shown to the user.

This will also avoid problems down the line if watch ever becomes available in a transaction.

src/functions.php
```diff
@@ -338,3 +340,32 @@ function create_field_path_type_map(array $typeMap, $fieldPath)
...
+ * Performs server selection, respecting the server a session may be pinned to
+ *
+ * @internal
+ * @return Server
+ */
+function select_server(Manager $manager, ReadPreference $readPreference = null, Session $session = null)
+{
+    $server = $session !== null ? $session->getServer() : null;
+
+    return $server ?: $manager->selectServer($readPreference);
```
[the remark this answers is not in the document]
alcaeus: This is not checked by phpcs, but might eventually be checked by phpstan or psalm once we introduce static analysis (for phpstan, this also means requiring a special phpstan-strict package).

I normally prefer ternary operators for such instances, but I'll defer to your preference here.

tests/FunctionalTestCase.php
```diff
@@ -408,9 +410,7 @@ private function disableFailPoints()
             return;
         }
 
-        $server = $this->getPrimaryServer();
-
-        foreach ($this->configuredFailPoints as $failPoint) {
+        foreach ($this->configuredFailPoints as list($failPoint, $server)) {
```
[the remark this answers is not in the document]
alcaeus: Took me a while to figure out that we don't have short list syntax yet - I liked the following even better:
```php
foreach ($this->configuredFailPoints as [$failPoint, $server])
```

tests/SpecTests/CommandExpectations.php
alcaeus: The `buildInfo` and `getParameter` commands are because of us checking for the availability of the `configureFailPoint` command. With the introduction of the `targetedFailPoint` operation in spec tests, it may become necessary to filter these commands as they will inevitably be caught up in the command loggers enabled for spec tests. I'll create a SPEC ticket to add this to the "Command Started Events" section using similar language.

I'll also add a comment here as to why we're ignoring `buildInfo` and `getParameter` commands 👍

tests/SpecTests/Operation.php
alcaeus: I've added a comment and created [PHPC-1438](https://jira.mongodb.org/browse/PHPC-1438) to track exposing the internal transaction state.

tests/SpecTests/Operation.php
alcaeus: It's just as effective if we rely on the return type of `getServer`. Since this isn't strict at all, it's better to be defensive here and use `assertInstanceOf`. Good catch! 👍

tests/SpecTests/Operation.php
alcaeus: phpcs can't inspect this as it has no knowledge of types. This would be caught by using static analysis with phpstan or psalm at a stricter level.

tests/SpecTests/TransactionsSpecTest.php
alcaeus: The change stream spec tests only run on the `replicaset` topology, while `WatchFunctionalTest` skips tests if change streams are not supported (which checks whether a sharded cluster uses replica sets in the background).

The spec test can't specify to only run tests on a sharded cluster backed by replica set, as the topology will be `sharded` in both cases. The `topology` key of the `runOn` object is an array with the requirement being that the topology in use is listed in the topology array. IMO, this is a shortcoming of the current `runOn` syntax, but then from a topology perspective there is no difference between a sharded cluster using standalone servers and a cluster using replicasets.

FWIW, I added this skip after encountering test errors on a sharded cluster with standalone servers as that does not support transactions. This is due to the `sharded` topology requirement also matching for that cluster type. It could be that other drivers are not testing that particular use-case, which is why this went uncaught before.

tests/SpecTests/TransactionsSpecTest.php
alcaeus: I first ran into test failures and added this to avoid issues but forgot to remove all instances of that. Thanks for pointing it out 👍

tests/SpecTests/TransactionsSpecTest.php
alcaeus: I thought about this as well, but ultimately decided on this solution. Since `useMultipleMongoses` is a context option for the tests, it makes most sense to handle this internally and not leak it into the tests themselves. Since the context creates its own client and currently knows where to get the URL from, I thought it was best to leave that logic encapsulated and simply expose the `useMultipleMongoses` argument in the factories.

tests/SpecTests/TransactionsSpecTest.php
alcaeus: Ah, the comments aren't present in the JSON files, which is why I missed the detailed explanation. The test is implemented in the C driver, but I couldn't find out whether it uses a single- or multi-threaded client. I'll check with Clyde on this.

tests/FunctionalTestCase.php
```diff
@@ -382,9 +384,16 @@ protected function skipIfTransactionsAreNotSupported()
             $this->markTestSkipped('Transactions are not supported on standalone servers');
         }
 
-        // TODO: MongoDB 4.2 should support sharded clusters (see: PHPLIB-374)
         if ($this->isShardedCluster()) {
-            $this->markTestSkipped('Transactions are not supported on sharded clusters');
+            if (!$this->isShardedClusterUsingReplicasets()) {
```
alcaeus: We can't easily figure out what storage engine is in use on sharded clusters, so we're skipping that check when running a sharded setup. I don't think we even run any tests on the MMAPv1 storage engine at all in PHPLIB. This may change once we move library tests to Evergreen, but I'd revisit this then and figure out a more robust solution.

alcaeus:
@jmikola I've added the prose tests for mongos pinning. I'm not sure if they provide additional value to the spec tests (especially since they cover a range of pin/unpin scenarios), but I've added them nonetheless 👍

src/Operation/Watch.php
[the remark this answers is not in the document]
jmikola: I concur that this would ensure the user gets more accurate error message. Would you mind adding a comment to document this reasoning?

src/functions.php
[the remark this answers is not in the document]
jmikola: I'm ambivalent. Fine to leave as-is.

tests/SpecTests/TransactionsSpecTest.php
jmikola: There is one change stream spec test that runs on a sharded cluster ([here](https://github.com/mongodb/specifications/blob/4d5c5c78f68eb1c3c1a4462e197fdc5fd2312bf0/source/change-streams/tests/change-streams-errors.yml#L54)), but I'm not sure if that error test even gets as far as caring that the sharded cluster may not be backed by a replica set.

tests/FunctionalTestCase.php
jmikola: MMAPv1 was completely [removed in 4.2](https://docs.mongodb.com/master/release-notes/4.2/#removed-mmapv1-storage-engine), so this shouldn't be a concern.

tests/SpecTests/TransactionsSpecTest.php
```diff
@@ -202,6 +208,84 @@ public function provideTests()
...
     }
 
+    /**
+     * Prose test 1: Test that starting a new transaction on a pinned
+     * ClientSession unpins the session and normal server selection is performed
+     * for the next operation.
+     */
+    public function testStartingNewTransactionOnPinnedSessionUnpinsSession()
+    {
+        $this->skipIfTransactionsAreNotSupported();
```
jmikola: Perhaps we should just move this to `setUp()`.

tests/SpecTests/TransactionsSpecTest.php
```diff
@@ -202,6 +208,84 @@ public function provideTests()
         return $testArgs;
     }
 
+    /**
+     * Prose test 1: Test that starting a new transaction on a pinned
+     * ClientSession unpins the session and normal server selection is performed
+     * for the next operation.
+     */
+    public function testStartingNewTransactionOnPinnedSessionUnpinsSession()
```
jmikola: FWIW, I opened [SPEC-1433](https://jira.mongodb.org/browse/SPEC-1433) to determine if these prose tests are still relevant. I see no harm in adding them, though.

tests/SpecTests/TransactionsSpecTest.php
bazile-clyde: All of our fail points are activated using single-threaded clients. Also, to disable the fail point we didn't use any loops but instead just sent the command `{'configureFailPoint': 'failCommand' 'mode': 'off'}`. Both activating and deactivating fail points were done via a new client, pinned to the same server ID as the one we used for testing.

tests/SpecTests/CommandExpectations.php
alcaeus: Created [SPEC-1434](https://jira.mongodb.org/browse/SPEC-1434) to track this.

PR #678, opened 2019-09-04
PHPLIB-473: Ensure read concern is inherited for transactions

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-473

### DIFF
--- a/src/Database.php
+++ b/src/Database.php
@@ -511,7 +511,7 @@ class Database
 
         $server = select_server($this->manager, $options);
 
-        if (! isset($options['readConcern']) && server_supports_feature($server, self::$wireVersionForReadConcern)) {
+        if (! isset($options['readConcern']) && server_supports_feature($server, self::$wireVersionForReadConcern) && ! is_in_transaction($options)) {
             $options['readConcern'] = $this->readConcern;
         }
 
--- a/src/Database.php
+++ b/src/Database.php
@@ -198,7 +198,7 @@ class Database
     {
         $hasWriteStage = is_last_pipeline_operator_write($pipeline);
 
-        if (! isset($options['readPreference'])) {
+        if (! isset($options['readPreference']) && ! is_in_transaction($options)) {
             $options['readPreference'] = $this->readPreference;
         }
 
@@ -206,7 +206,7 @@ class Database
             $options['readPreference'] = new ReadPreference(ReadPreference::RP_PRIMARY);
         }
 
-        $server = select_server($this->manager, $options['readPreference'], extract_session_from_options($options));
+        $server = select_server($this->manager, $options);
 
         /* MongoDB 4.2 and later supports a read concern when an $out stage is
          * being used, but earlier versions do not.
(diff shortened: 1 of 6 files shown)


alcaeus:
PR was reworked to fix all failing tests. I found a couple of occurences where we still applied `readConcern` and `writeConcern` options even when in a transactions, those where fixed.

`readPreference` is now also no longer inherited when in a transaction. Since using the transaction's read preference has upstream dependencies on [libmongoc](https://jira.mongodb.org/browse/CDRIVER-3358) and the [PHP extension](https://jira.mongodb.org/browse/PHPC-1439), this was extracted to [PHPLIB-476](https://jira.mongodb.org/browse/PHPLIB-476). For now we fall back to a primary read preference, as that is the only read preference currently allowed in a transaction.

src/Database.php
```diff
@@ -280,9 +280,9 @@ public function createCollection($collectionName, array $options = [])
             $options['typeMap'] = $this->typeMap;
         }
 
-        $server = select_server($this->manager, new ReadPreference(ReadPreference::RP_PRIMARY), extract_session_from_options($options));
+        $server = select_server($this->manager, $options);
 
-        if (! isset($options['writeConcern']) && server_supports_feature($server, self::$wireVersionForWritableCommandWriteConcern)) {
+        if (! isset($options['writeConcern']) && server_supports_feature($server, self::$wireVersionForWritableCommandWriteConcern) && ! is_in_transaction($options)) {
```
jmikola: IIRC, these checks were not originally added in the PR for [PHPLIB-368](https://jira.mongodb.org/browse/PHPLIB-368) because the operations themselves weren't supported within a transaction. I think this makes sense, though, as it makes it less likely we'll get a client-side error for providing an explicit RC/WC (be it from our driver or libmongoc) and more likely the user will get a server-side error for running an unsupported command within a transaction.

I suppose this is similar in some respects to your earlier PR to consider the pinned session when resuming a change stream.

src/functions.php
```diff
@@ -393,14 +393,40 @@ function extract_session_from_options(array $options)
...
+function select_server(Manager $manager, array $options)
 {
-    $server = $session !== null ? $session->getServer() : null;
+    $session = extract_session_from_options($options);
+    if ($session !== null && $session->getServer() !== null) {
+        return $session->getServer();
+    }
+
+    $readPreference = extract_read_preference_from_options($options);
+    if (! $readPreference instanceof ReadPreference) {
```
jmikola: It makes no technical difference, but this seems inconsistent with the previous `if` condition. Would it make sense to use `$readPreference === null` here?

src/Database.php
alcaeus: Yep, I thought it best to add it for consistency, and to avoid failures down the line when we realise that transactions allow more operations than I envisioned. I'm not sure if 4.4 will support DDL commands in transactions, but this will have us prepared either way.

src/functions.php
alcaeus: Ha...I thought I'd be smart and be type safe here since we don't have return types yet. If you don't mind I'd rather change the call above to also check for `$session` being an `MongoDB\Driver\Session` instance instead of just checking for it not being `null`.

jmikola: I'm totally on board with `instanceof` over `!== null`. Go for it.

PR #679, opened 2019-09-04
PHPLIB-474: Provide transaction example using new withTransaction API

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-474

### DIFF
--- a/tests/DocumentationExamplesTest.php
+++ b/tests/DocumentationExamplesTest.php
@@ -1541,6 +1541,73 @@ class DocumentationExamplesTest extends FunctionalTestCase
     }
 
     /**
+     * @doesNotPerformAssertions
+     */
+    public function testWithTransactionExample()
+    {
+        $this->skipIfTransactionsAreNotSupported();
+
+        $uriString = static::getUri(true);
+
+        // phpcs:disable SlevomatCodingStandard.Namespaces.ReferenceUsedNamesOnly
+        // Start Transactions withTxn API Example 1
+        /*
+         * For a replica set, include the replica set name and a seedlist of the members in the URI string; e.g.
+         * uriString = 'mongodb://mongodb0.example.com:27017,mongodb1.example.com:27017/?replicaSet=myRepl'
+         * For a sharded cluster, connect to the mongos instances; e.g.
+         * uriString = 'mongodb://mongos0.example.com:27017,mongos1.example.com:27017/'
+         */
+
+        $client = new \MongoDB\Client($uriString);
+
+        // Prerequisite: Create collections. CRUD operations in transactions must be on existing collections.
+        $client->selectCollection(
+            'mydb1',
+            'foo',
+            [
+                'writeConcern' => new \MongoDB\Driver\WriteConcern(\MongoDB\Driver\WriteConcern::MAJORITY, 1000),
+            ]
+        )->insertOne(['abc' => 0]);
+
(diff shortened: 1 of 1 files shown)


tests/DocumentationExamplesTest.php
```diff
@@ -1540,6 +1540,85 @@ public function testCausalConsistency()
...
+
+        $uriString = static::getUri(true);
+
+        // phpcs:disable SlevomatCodingStandard.Namespaces.ReferenceUsedNamesOnly
+        // Start Transactions withTxn API Example 1
+        /*
+         * For a replica set, include the replica set name and a seedlist of the members in the URI string; e.g.
+         * uriString = 'mongodb://mongodb0.example.com:27017,mongodb1.example.com:27017/admin?replicaSet=myRepl'
+         * For a sharded cluster, connect to the mongos instances; e.g.
+         * uriString = 'mongodb://mongos0.example.com:27017,mongos1.example.com:27017/admin'
```
jmikola: `/admin` looks redundant in both examples. I left a note about this in [DRIVERS-655](https://jira.mongodb.org/browse/DRIVERS-655) for @kay-kim, so maybe we can get that removed. If not, I suppose this will need to remain as-is for consistency.

tests/DocumentationExamplesTest.php
```diff
@@ -1540,6 +1540,85 @@ public function testCausalConsistency()
...
+         */
+
+        $client = new \MongoDB\Client($uriString);
+
+        // Prerequisite: Create collections. CRUD operations in transactions must be on existing collections.
+        $client->selectCollection(
+            'test',
+            'foo',
```
jmikola: The examples in the DRIVERS ticket use "mydb1" and "mydb2" for each collection, rather than "test" for both. I'm not sure if that's pertinent to demonstrate a transaction across two databases, but I think we should what's specified.

tests/DocumentationExamplesTest.php
```diff
@@ -1540,6 +1540,85 @@ public function testCausalConsistency()
...
+            [
+                'readConcern' => new \MongoDB\Driver\ReadConcern(\MongoDB\Driver\ReadConcern::MAJORITY),
```
jmikola: The example also doesn't set a read concern for the collection used to insert both example documents. I don't think we need this.

tests/DocumentationExamplesTest.php
```diff
@@ -1540,6 +1540,85 @@ public function testCausalConsistency()
...
+                'writeConcern' => new \MongoDB\Driver\WriteConcern(\MongoDB\Driver\WriteConcern::MAJORITY, 1000),
+            ]
+        )->insertOne(['xyz' => 0]);
+
+        // Step 1: Define the callback that specifies the sequence of operations to perform inside the transactions.
+
+        $callback = function (\MongoDB\Driver\Session $session) use ($client) {
+            $client->selectCollection(
```
jmikola: The clients created in the callback also use `mydb1.foo` and `mydb2.bar` as their namespaces, respectively. They also don't take any read or write concern options.

tests/DocumentationExamplesTest.php
jmikola: [DRIVERS-655](https://jira.mongodb.org/browse/DRIVERS-655) was updated and no longer includes "/admin".

PR #683, opened 2019-09-24
PHPLIB-482: Add missing error expectation to crud tests

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-482

While looking at [PHPLIB-481](https://jira.mongodb.org/browse/PHPLIB-481), I realised that the crud spec tests don't check error expectations. While we're unlikely to miss any errors because of this (since the return value of the operation call is most likely not going to match the expectation), I still think it's sensible to add these to the 1.5 branch to ensure we're correctly running the spec tests.

### DIFF
--- a/tests/SpecTests/Operation.php
+++ b/tests/SpecTests/Operation.php
@@ -158,6 +158,7 @@ final class Operation
     {
         $o = new self($operation);
 
+        $o->errorExpectation = ErrorExpectation::fromCrud($operation);
         $o->resultExpectation = ResultExpectation::fromCrud($operation, $o->getResultAssertionType());
 
         if (isset($operation->collectionOptions)) {
--- a/tests/SpecTests/ErrorExpectation.php
+++ b/tests/SpecTests/ErrorExpectation.php
@@ -75,6 +75,17 @@ final class ErrorExpectation
         return $o;
     }
 
+    public static function fromCrud(stdClass $result)
+    {
+        $o = new self();
+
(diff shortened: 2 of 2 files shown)


tests/SpecTests/Operation.php
```diff
@@ -158,6 +158,7 @@ public static function fromCrud(stdClass $operation)
     {
         $o = new self($operation);
 
+        $o->errorExpectation = ErrorExpectation::fromCrud($operation);
```
jmikola: I see that `ResultExpectation::fromCrud()` did check for an error result, but that only changed the assertion type to `ASSERT_NOTHING`. Is this fixing an outstanding bug where the CRUD spec tests were never asserting error exceptions (since being ported over to the `MongoDB\Tests\SpecTests\FunctionalTestCase` runner?

If so: (a) oops, and (b) can we track this with a JIRA ticket?

alcaeus: Yes and no. For one, the error message is more sensible when having `ErrorExpectation` in place, as it fails if an error was recorded even though none was expected. At the same time, we didn't check for the presence of an error for the two tests that throw one, so those tests were definitely incomplete. I've created [PHPLIB-482](https://jira.mongodb.org/browse/PHPLIB-482) to track this and I've also amended the commit to reference the ticket.

jmikola: That all makes sense. Thanks for catching this!

PR #688, opened 2019-10-08
PHPLIB-485: Inherit collection's typeMap in distinct

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-485

### DIFF
--- a/tests/Collection/CollectionFunctionalTest.php
+++ b/tests/Collection/CollectionFunctionalTest.php
@@ -16,7 +16,10 @@ use MongoDB\Operation\Count;
 use MongoDB\Tests\CommandObserver;
 use function array_filter;
 use function call_user_func;
+use function is_scalar;
+use function json_encode;
 use function strchr;
+use function usort;
 use function version_compare;
 
 /**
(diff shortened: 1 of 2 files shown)


tests/Collection/CollectionFunctionalTest.php
```diff
@@ -170,6 +173,88 @@ function (array $event) {
...
+        ]);
+        $this->manager->executeBulkWrite($this->getNamespace(), $bulkWrite);
+
+        $values = $this->collection->withOptions(['typeMap' => $typeMap])->distinct('x');
+
+        /* This sort callable sorts all scalars to the front of the list. All
+         * non-scalar values are sorted by running json_encode on them and
+         * comparing their string representations.
+         */
+        $sort = function ($a, $b) {
```
jmikola: Would it be easier to just assert that the count of `$expectedDocuments` and `$values` is equal and then loop over `$expectedDocuments` and assert that each of its elements exists in `$values`?

I realize the only scalar is `4` at the moment, but this seems like it could be fragile if additional scalars were added to the test data.

alcaeus: I shamelessly copied this from DistinctFunctionalTest, so I had to double-check what it does. This method generally sorts scalars to the top and non-scalars to the bottom, but each group is sorted within itself, so adding multiple scalars shouldn't pose an issue. Non-scalars are encoded to JSON before sorting, so they are sorted by their JSON representation.

PR #690, opened 2019-10-29
PHPLIB-477: Deprecate CodeWScope for use within the mapReduce command

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-477

Contrary to the original wording suggestion, I replaced the usage of "CodeWScope" for "Javascript instance with a scope". I don't think users will be familiar with the internal nomenclature of BSON types, so this may be more understandable to the average user.

Apart from deprecating these in the documentation, we also emit an `E_USER_DEPRECATED` PHP error that can be logged with libraries like [symfony/phpunit-bridge](https://github.com/symfony/phpunit-bridge).

### DIFF
--- a/docs/includes/apiargs-MongoDBCollection-method-mapReduce-option.yaml
+++ b/docs/includes/apiargs-MongoDBCollection-method-mapReduce-option.yaml
@@ -13,6 +13,11 @@ name: finalize
 type: :php:`MongoDB\\BSON\\Javascript <class.mongodb-bson-javascript>`
 description: |
   Follows the reduce method and modifies the output.
+
+  .. note::
+
+     Passing a Javascript instance with a scope is deprecated. Put all scope
+     variables in the ``scope`` option of the MapReduce operation.
 interface: phpmethod
 operation: ~
 optional: true
--- a/src/Operation/MapReduce.php
+++ b/src/Operation/MapReduce.php
@@ -40,6 +40,8 @@ use function is_string;
 use function MongoDB\create_field_path_type_map;
 use function MongoDB\is_mapreduce_output_inline;
 use function MongoDB\server_supports_feature;
+use function trigger_error;
+use const E_USER_DEPRECATED;
 
 /**
  * Operation for the mapReduce command.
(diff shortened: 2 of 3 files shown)


docs/includes/apiargs-MongoDBCollection-method-mapReduce-option.yaml
```diff
@@ -13,6 +13,11 @@ name: finalize
 type: :php:`MongoDB\\BSON\\Javascript <class.mongodb-bson-javascript>`
 description: |
   Follows the reduce method and modifies the output.
+
+  .. note::
+
+     Passing a Javascript instance with a scope is deprecated. Put all scope
+     variables in the "scope" option of the MapReduce operation.
```
jmikola: I think our convention for referring to options uses monospace formatting instead of quotes, so we should write `scope` here. Feel free to double-check, though.

This made we wonder if the `scope` option for `mapReduce` was being preserved with the aggregation changes. According to the design document linked from [PM-766](https://jira.mongodb.org/browse/PM-766), it will remain available.

src/Operation/MapReduce.php
```diff
@@ -242,6 +253,19 @@ public function __construct($databaseName, $collectionName, JavascriptInterface
             unset($options['writeConcern']);
         }
 
+        // Handle deprecation of CodeWScope
+        if ($map->getScope() !== null) {
+            @trigger_error('Use of Javascript with scope in "$map" argument for MapReduce is deprecated. Put all scope variables in the "scope" option of the MapReduce operation.', E_USER_DEPRECATED);
+        }
+
+        if ($reduce->getScope() !== null) {
+            @trigger_error('Use of Javascript with scope in "$map" argument for MapReduce is deprecated. Put all scope variables in the "scope" option of the MapReduce operation.', E_USER_DEPRECATED);
```
jmikola: "$reduce" argument

docs/includes/apiargs-MongoDBCollection-method-mapReduce-option.yaml
alcaeus: Monospace formatting is the way to go in the rest of the docs, so I've adjusted it.

The deprecation message was suggested in the DRIVERS ticket, along with the "CodeWScope" wording which I decided to drop. This is why I assumed that the `scope` option would remain available. Thanks for confirming this.

src/Operation/MapReduce.php
alcaeus: 🤦‍♂️

PR #694, opened 2019-11-08
PHPLIB-501: Prevent memory leaks from generators

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-501

This was discovered while investigating https://github.com/doctrine/mongodb-odm/issues/2092. The different behaviour can be seen in action here: https://3v4l.org/UIfIb. Note that when we don't specifically unset the generator, it stays around much longer even though the wrapping iterator has already been garbage collected.

### DIFF
--- a/src/Model/CachingIterator.php
+++ b/src/Model/CachingIterator.php
@@ -18,8 +18,8 @@
 namespace MongoDB\Model;
 
 use Countable;
-use Generator;
 use Iterator;
+use IteratorIterator;
 use Traversable;
 use function count;
 use function current;
@@ -41,7 +41,7 @@ class CachingIterator implements Countable, Iterator
     /** @var array */
     private $items = [];
 
-    /** @var Generator */
+    /** @var IteratorIterator */
     private $iterator;
 
     /** @var boolean */
@@ -61,7 +61,9 @@ class CachingIterator implements Countable, Iterator
      */
     public function __construct(Traversable $traversable)
     {
-        $this->iterator = $this->wrapTraversable($traversable);
+        $this->iterator = new IteratorIterator($traversable);
+
+        $this->iterator->rewind();
         $this->storeCurrentItem();
     }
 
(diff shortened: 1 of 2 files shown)


tests/Model/CachingIteratorTest.php
```diff
@@ -53,7 +53,7 @@ public function testIterationWithEmptySet()
 
     public function testPartialIterationDoesNotExhaust()
     {
-        $traversable = $this->getTraversableThatThrows([1, 2, new Exception()]);
+        $traversable = $this->getTraversable([1, 2, new Exception()]);
```
jmikola: I assume this is an easy change because we never iterate to the third element?

src/Model/CachingIterator.php
```diff
@@ -61,8 +64,7 @@ class CachingIterator implements Countable, Iterator
      */
     public function __construct(Traversable $traversable)
     {
-        $this->iterator = $this->wrapTraversable($traversable);
-        $this->storeCurrentItem();
+        $this->iterator = new IteratorIterator($traversable);
```
jmikola: Would it make sense to just prepare the iterator here and store the current item? That might allow you to remove the `prepareIterator()` calls from other methods.

As-is, there might be a subtle behavioral change, since the original implementation did rewind the traversable, create the generator, and iterate once to store the current item?

jmikola:
Renamed the issue to reflect the actual change and title in JIRA.

Also, re: your comment in https://github.com/doctrine/mongodb-odm/pull/2100#issuecomment-552385561, can you elaborate on the subtle behavioral change by switching to IteratorIterator? Does it relate to my earlier comment about removing the `prepareIterator()` call?

alcaeus:
That is correct. When using the generator, we immediately start iterating the wrapped traversable. When switching to IteratorIterator, we can still do that (by calling `rewind` in the constructor), but we may want to defer starting iteration of the wrapped traversable until the wrapping iterator is starting its iteration.

tests/Model/CachingIteratorTest.php
alcaeus: Almost. The difference between the methods is that `getTraversableThatThrows` checks an element and throws it if it's an Exception. Since we don't need to iterate over exceptions anywhere, I decided to simplify this.

alcaeus:
PR adapted to replicate previous behaviour: iteration of the traversable starts when `CachingIterator` is constructed, not when iteration of `CachingIterator` starts.

PR #696, opened 2019-11-11
PHPLIB-494: Fix validation of options with default values

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-494

This fixes a few validation errors where it is possible to use `null` as a valid value for an option, resulting in undefined behaviour. All other options that have a default value are directly checked for their type, without an `isset` check to guard against a missing option since it should always be present due to the array union.

### DIFF
--- a/src/Operation/Watch.php
+++ b/src/Operation/Watch.php
@@ -178,8 +178,12 @@ class Watch implements Executable, /* @internal */ CommandSubscriber
             'readPreference' => new ReadPreference(ReadPreference::RP_PRIMARY),
         ];
 
-        if (isset($options['fullDocument']) && ! is_string($options['fullDocument'])) {
+        if (! is_string($options['fullDocument'])) {
             throw InvalidArgumentException::invalidType('"fullDocument" option', $options['fullDocument'], 'string');
+        }
+
+        if (! $options['readPreference'] instanceof ReadPreference) {
+            throw InvalidArgumentException::invalidType('"readPreference" option', $options['readPreference'], ReadPreference::class);
         }
 
         if (isset($options['resumeAfter']) && ! is_array($options['resumeAfter']) && ! is_object($options['resumeAfter'])) {
--- a/src/Client.php
+++ b/src/Client.php
@@ -95,7 +95,7 @@ class Client
     {
         $driverOptions += ['typeMap' => self::$defaultTypeMap];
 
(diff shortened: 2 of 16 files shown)


src/Operation/Watch.php
```diff
@@ -178,7 +178,7 @@ public function __construct(Manager $manager, $databaseName, $collectionName, ar
             'readPreference' => new ReadPreference(ReadPreference::RP_PRIMARY),
         ];
 
-        if (isset($options['fullDocument']) && ! is_string($options['fullDocument'])) {
+        if (! is_string($options['fullDocument'])) {
```
jmikola: Do you think we should also check `readPreference` here? Otherwise, a user could pass `null` for ReadPreference and it will [be ignored in Aggregate](https://github.com/mongodb/mongo-php-library/blob/1.5.1/src/Operation/Aggregate.php#L388). In that case, it's possible the client-level read preference (from a connection string) may be applied by libmongoc.

I think my original approach to this issue was going to create a utility function to replace the `$options +=` operation and ensure that `null` values from the user are entirely ignored by unsetting them before we `+=`. That would also circumvent this issue. It would also allow the `isset()` checks to be left as-is; however, I agree that they would serve no purpose since the default options would always be non-null.

alcaeus: If a user calls `Collection::watch` or `Database::watch` with a `null` `readPreference` in the options array, the helper method will inject its own read preference into the options array, overriding the default `primary` read preference from the watch operation class.

If a user instantiates the `Watch` operation class directly, they would be able to pass a `null` `readPreference` directly and cause a `null` `readPreference` being passed to the `Aggregate` command. This read preference is only used when resuming the change stream, at which point the `select_server` function will use a primary read preference since it isn't set.

Regardless, I can add a check for completeness to ensure that users don't pass in a `null` `readPreference`.

---

I was thinking about changing all `+=` calls to a method but saw that in those cases, the checks weren't conditional except in the instances fixed here. That's why I decided to update the few occurrences where there was an `isset` instead of changing all array unions.

jmikola: Points noted. I agree with adding the check to not allow a `null` `readPreference` option for Watch, though.

alcaeus: I've added the check and also added a test to submit a `null` read preference to the Operation

PR #706, opened 2020-01-10
PHPLIB-519: Allow testing against development driver versions

alcaeus:
This allows us to later switch to always test against dev driver versions until the RC phase of the extension, saving us the trouble of releasing alpha builds of the driver for internal testing purposes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -112,6 +112,18 @@ jobs:
       env:
         - DEPLOYMENT=SHARDED_CLUSTER_RS
 
+    # Test next patch release for driver
+    - stage: Test
+      php: "7.3"
+      env:
+        - DRIVER_BRANCH="v1.6"
+
+    # Test next minor release for driver
+    - stage: Test
+      php: "7.3"
+      env:
+        - DRIVER_BRANCH="master"
+
 before_install:
   - pip install "mongo-orchestration>=0.6.7,<1.0" --user `whoami`
   - export SERVER_FILENAME=mongodb-linux-x86_64-${SERVER_DISTRO}-${SERVER_VERSION}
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       php: "7.1"
       before_install: []
       before_script:
-        - pecl install -f mongodb-${DRIVER_VERSION}
```
jmikola: I actually don't recall why we stopped using the `tpecl` command, which caches the install. It looks like it was removed in https://github.com/mongodb/mongo-php-library/commit/8774590ef047b9997cd245ece7dfcd83edc63e16.

`.travis.yml` still has the `tpecl` definition, but it looks like it's no longer being used since that commit. Do you want to try relocating it to `install-extension.sh` and use it when installing based on `$DRIVER_VERSION`? Otherwise, I suppose we should just delete it entirely (can do so in a separate issue/PR).

.travis/install-extension.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+
+if [ "x${DRIVER_VERSION}" != "x" ]; then
```
jmikola: As a sanity check, I think it would make sense to raise an error at the top of this script if both `$DRIVER_VERSION` and `$DRIVER_BRANCH` are non-empty. As-is, we're likely to compile the driver twice and end up with the branch version (last one wins).

.travis.yml
alcaeus: I'll change it back to using `tpecl`, makes sense to cache it when possible.

.travis/install-extension.sh
alcaeus: I'll change this around to use an `else if` and give precedence to a branch build. This allows us to omit setting `DRIVER_VERSION=""` in the env for branch builds. If none of the drivers can be installed, the following `php --ri` call will fail, but I can also exit with a non-zero code in this script to break the build earlier.

PR #708, opened 2020-01-20
PHPLIB-511: Handle autoEncryptionOpts in MongoDB\Client

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-511

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ phpunit.xml
 # phpcs
 .phpcs-cache
 phpcs.xml
+
+mongocryptd.pid
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,9 @@ cache:
 
 env:
   global:
-    - DRIVER_VERSION=1.6.0
+    - DRIVER_VERSION=1.7.0
+    # TODO: remove once a 1.7 driver release has been tagged
+    - DRIVER_BRANCH="master"
     - SERVER_DISTRO=ubuntu1604
     - SERVER_VERSION=4.2.0
     - DEPLOYMENT=STANDALONE
@@ -112,17 +114,19 @@ jobs:
       env:
         - DEPLOYMENT=SHARDED_CLUSTER_RS
 
+    # TODO: re-enable once v1.7 has been branched in PHPC
     # Test next patch release for driver
-    - stage: Test
-      php: "7.3"
-      env:
-        - DRIVER_BRANCH="v1.6"
+#    - stage: Test
+#      php: "7.3"
+#      env:
+#        - DRIVER_BRANCH="v1.7"
 
+    # TODO: re-enable once v1.7 has been branched in PHPC
     # Test next minor release for driver
-    - stage: Test
-      php: "7.3"
-      env:
-        - DRIVER_BRANCH="master"
+#    - stage: Test
+#      php: "7.3"
+#      env:
+#        - DRIVER_BRANCH="master"
 
 before_install:
   - pip install "mongo-orchestration>=0.6.7,<1.0" --user `whoami`
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
         "php": "^5.6 || ^7.0",
         "ext-hash": "*",
         "ext-json": "*",
-        "ext-mongodb": "^1.6"
+        "ext-mongodb": "^1.7"
     },
     "require-dev": {
         "phpunit/phpunit": "^5.7.27 || ^6.4 || ^8.3",
--- a/docs/includes/apiargs-MongoDBClient-method-construct-driverOptions.yaml
+++ b/docs/includes/apiargs-MongoDBClient-method-construct-driverOptions.yaml
@@ -112,4 +112,18 @@ description: |
 interface: phpmethod
 operation: ~
 optional: true
+---
+arg_name: option
+name: autoEncryption
+type: array
+description: |
+  Options to configure client-side field-level encryption in the driver. The
+  encryption options are documented in the :php:`extension documentation
+  <manual/en/mongodb-driver-manager.construct.php#mongodb-driver-manager.construct-driveroptions>`.
+  For the ``keyVaultClient`` option, you may pass a :phpclass:`MongoDB\\Client`
+  instance, which will be unwrapped to provide a :php:`MongoDB\\Driver\\Manager <class.mongodb-driver-manager>`
+  to the extension.
+interface: phpmethod
+operation: ~
+optional: true
 ...
--- a/src/Client.php
+++ b/src/Client.php
@@ -99,6 +99,14 @@ class Client
             throw InvalidArgumentException::invalidType('"typeMap" driver option', $driverOptions['typeMap'], 'array');
         }
 
+        if (isset($driverOptions['autoEncryption']['keyVaultClient'])) {
+            if ($driverOptions['autoEncryption']['keyVaultClient'] instanceof self) {
+                $driverOptions['autoEncryption']['keyVaultClient'] = $driverOptions['autoEncryption']['keyVaultClient']->manager;
+            } elseif (! $driverOptions['autoEncryption']['keyVaultClient'] instanceof Manager) {
+                throw InvalidArgumentException::invalidType('"keyVaultClient" autoEncryption option', $driverOptions['autoEncryption']['keyVaultClient'], [self::class, Manager::class]);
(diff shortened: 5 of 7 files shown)


.travis.yml
```diff
@@ -14,7 +14,8 @@ cache:
 
 env:
   global:
-    - DRIVER_VERSION=1.6.0
+    - DRIVER_VERSION=1.7.0
+    - DRIVER_BRANCH="master"
```
jmikola: Given that [`install-extension.sh`](https://github.com/mongodb/mongo-php-library/blob/master/.travis/install-extension.sh) checks for `DRIVER_BRANCH` with `if` and then `DRIVER_VERSION` via an `elif`, wouldn't defining both result in `DRIVER_VERSION` never getting used?

composer.json
```diff
@@ -13,7 +13,7 @@
         "php": "^5.6 || ^7.0",
         "ext-hash": "*",
         "ext-json": "*",
-        "ext-mongodb": "^1.6"
+        "ext-mongodb": "^1.7"
```
jmikola: Just confirming, but this works on the `master` branch of the driver because `phongo_version.h` already reports itself as `1.7.0-dev`, correct?

src/Client.php
```diff
@@ -99,6 +99,20 @@ public function __construct($uri = 'mongodb://127.0.0.1/', array $uriOptions = [
             throw InvalidArgumentException::invalidType('"typeMap" driver option', $driverOptions['typeMap'], 'array');
         }
 
+        if (isset($driverOptions['autoEncryption'])) {
+            if (! is_array($driverOptions['autoEncryption'])) {
+                throw InvalidArgumentException::invalidType('"autoEncryption" driver option', $driverOptions['autoEncryption'], 'array');
+            }
+
+            if (isset($driverOptions['autoEncryption']['keyVaultClient'])) {
+                if (! $driverOptions['autoEncryption']['keyVaultClient'] instanceof self) {
```
jmikola: I wonder if we can just reduce the entire block of code added to this constructor to:

```php
if (isset($driverOptions['autoEncryption']['keyVaultClient']) &&
    $driverOptions['autoEncryption']['keyVaultClient'] instanceof self) {
    $driverOptions['autoEncryption']['keyVaultClient'] = $driverOptions['autoEncryption']['keyVaultClient']->manager;
}
```

IIRC, PHPC will already check that `autoEncryption` is an array and go on to check that `keyVaultClient` is a Manager instance. The only responsibility of PHPLIB would be to "unwrap" a Client instance into a Manager. I also don't think we need to prohibit folks from passing a Manager here.

I'll note that the [`Client::__construct()`](https://docs.mongodb.com/php-library/current/reference/method/MongoDBClient__construct/#definition) docs originally didn't go into detail about driver-specific options (i.e. everything other than `typeMap`), although we do appear to duplicate those docs in the current text. If users were to go right to the PHPC docs for information on driver options, they might actually think they need to pass a Manager instance. I suppose this PR should also revise the documentation to add an entry for `autoEncryption` with a brief description that (a) links to the PHPC docs and (b) notes that users can pass a `MongoDB\Client` in lieu of a `MongoDB\Driver\Manager` if they wish, which the library will unwrap for them.

Side note: I just realized we missed deprecating SSL driver options -- I've opened [PHPLIB-521](https://jira.mongodb.org/browse/PHPLIB-521) to track that.

.travis.yml
alcaeus: Yes, which is intended until we've got a stable driver release, since PECL won't be able to install the 1.7 driver.

jmikola: A "todo" comment similar to what you added below for the new matrix configs would work well here.

src/Client.php
alcaeus: > I wonder if we can just reduce the entire block of code added to this constructor to:

Technically, yes. However, that changes the exception message to be "Expected "keyVaultClient" encryption option to be MongoDB\Driver\Manager, string given", which can be misleading as we allow passing a `MongoDB\Client` as well.

I can change the code to check if `$driverOptions['autoEncryption']['keyVaultClient']` is set, handle unwrapping if necessary and throw only if the `keyVaultClient` is not an instance of `MongoDB\Client` or `MongoDB\Driver\Manager`. As is, we disallow passing a manager directly, which may still be a valid use-case for some people.

src/Client.php
```diff
@@ -99,6 +99,14 @@ public function __construct($uri = 'mongodb://127.0.0.1/', array $uriOptions = [
             throw InvalidArgumentException::invalidType('"typeMap" driver option', $driverOptions['typeMap'], 'array');
         }
 
+        if (isset($driverOptions['autoEncryption']['keyVaultClient'])) {
+            if ($driverOptions['autoEncryption']['keyVaultClient'] instanceof self) {
+                $driverOptions['autoEncryption']['keyVaultClient'] = $driverOptions['autoEncryption']['keyVaultClient']->manager;
+            } elseif (! $driverOptions['autoEncryption']['keyVaultClient'] instanceof Manager) {
+                throw InvalidArgumentException::invalidType('"keyVaultClient" autoEncryption option', $driverOptions['autoEncryption']['keyVaultClient'], [self::class, Manager::class]);
```
jmikola: Note that `functions.php` already happened to pass "array or object" for this sort of error. I'll defer to you if you'd like to keep the new code in `invalidType()` (kudos for handling the case for 3+ types :smile:) or replace this with a string.

tests/ClientTest.php
```diff
@@ -20,6 +20,20 @@ public function testConstructorDefaultUri()
...
 
+    /**
+     * @doesNotPerformAssertions
+     */
+    public function testConstructorAutoEncryptionOpts()
+    {
+        $autoEncryptionOpts = [
+            'keyVaultClient' => new Client(static::getUri()),
+            'keyVaultNamespace' => 'default.keys',
+            'kmsProviders' => ['aws' => (object) ['accessKeyId' => 'abc', 'secretAccessKey' => 'def']],
```
jmikola: Does the `kmsProviders.aws` option actually need to be cast to an object?

docs/includes/apiargs-MongoDBClient-method-construct-driverOptions.yaml
```diff
@@ -112,4 +112,18 @@ description: |
...
+---
+arg_name: option
+name: autoEncryption
+type: array
+description: |
+  Options to configure client-side field-level encryption in the driver. The
+  encryption options are documented in the :php:`extension documentation
+  <manual/en/mongodb-driver-manager.construct.php#mongodb-driver-manager.construct-driveroptions>`.
+  For the ``keyVaultClient`` option, you may pass a ``MongoDB\Client`` instance,
+  which will be unwrapped to provide a ``MongoDB\Driver\Manager`` to the
```
jmikola:
```
:phpclass:`MongoDB\\Driver\\Manager`
```

docs/includes/apiargs-MongoDBClient-method-construct-driverOptions.yaml
```diff
@@ -112,4 +112,18 @@ description: |
...
 optional: true
+---
+arg_name: option
+name: autoEncryption
+type: array
+description: |
+  Options to configure client-side field-level encryption in the driver. The
+  encryption options are documented in the :php:`extension documentation
+  <manual/en/mongodb-driver-manager.construct.php#mongodb-driver-manager.construct-driveroptions>`.
+  For the ``keyVaultClient`` option, you may pass a ``MongoDB\Client`` instance,
```
jmikola:
```
:phpclass:`MongoDB\\Client`
```

.gitignore
```diff
@@ -11,3 +11,5 @@ phpunit.xml
 # phpcs
 .phpcs-cache
 phpcs.xml
+
+mongocryptd.pid
```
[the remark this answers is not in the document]
alcaeus: Yes, this shows up wherever the current working directory is when invoking phpunit.

src/Client.php
alcaeus: Since we're dealing with class names, the string concatenation when calling `invalidType` wouldn't have looked nice.
(I've been reminded of proper Oxford comma usage often enough to remember it now, hence the special case for 3+ types)

tests/ClientTest.php
alcaeus: No, that must have been left over from some testing. I've removed the cast.

PR #709, opened 2020-01-22
PHPLIB-512: Expose ClientEncryption API in MongoDB\Client

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-512

Note: contains commits from #708, only the last two commits are for the ticket in question.

### DIFF
--- a/docs/reference/method/MongoDBClient-createClientEncryption.txt
+++ b/docs/reference/method/MongoDBClient-createClientEncryption.txt
@@ -1,0 +1,51 @@
+=========================================
+MongoDB\\Client::createClientEncryption()
+=========================================
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Definition
+----------
+
+.. phpmethod:: MongoDB\\Client::createClientEncryption()
+
+   Returns a :php:`MongoDB\\Driver\\ClientEncryption <class.mongodb-driver-clientencryption>`
+   object for manual encryption and decryption of values.
+
+   .. code-block:: php
+
+      function createClientEncryption(array $options): MongoDB\Driver\ClientEncryption
+
+   This method has the following parameters:
+
+   .. include:: /includes/apiargs/MongoDBClient-method-createClientEncryption-param.rst
+
+   The ``$options`` parameter supports all options documented in the
+   :php:`extension manual <manual/en/mongodb-driver-manager.createclientencryption.php>`.
+   For the ``keyVaultClient`` option, an instance of :phpclass:`MongoDB\\Client`
+   is automatically unwrapped and the :php:`MongoDB\\Driver\\Manager <class.mongodb-driver-manager>`
+   instance is passed to the extension.
+
+Return Values
+-------------
+
+A :php:`MongoDB\\Driver\\ClientEncryption <class.mongodb-driver-clientencryption>`
+instance which can be used to encrypt and decrypt values.
+
+Errors/Exceptions
+-----------------
+
+.. include:: /includes/extracts/error-invalidargumentexception.rst
+.. include:: /includes/extracts/error-driver-invalidargumentexception.rst
+
+See Also
+--------
+
+- :php:`MongoDB\\Driver\\Manager::createClientEncryption()
+  <manual/en/mongodb-driver-manager.createclientencryption.php>`
--- a/src/Client.php
+++ b/src/Client.php
@@ -17,6 +17,7 @@
 
 namespace MongoDB;
 
+use MongoDB\Driver\ClientEncryption;
 use MongoDB\Driver\Exception\InvalidArgumentException as DriverInvalidArgumentException;
 use MongoDB\Driver\Exception\RuntimeException as DriverRuntimeException;
 use MongoDB\Driver\Manager;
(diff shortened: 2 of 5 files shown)


src/Client.php
```diff
@@ -161,6 +162,24 @@ public function __toString()
         return $this->uri;
     }
 
+    /**
+     * Returns a ClientEncryption instance for explicit encryption and decryption
+     *
+     * @return ClientEncryption
```
jmikola: Missing `@param array $options`.

tests/ClientTest.php
```diff
@@ -173,4 +174,55 @@ public function testSelectDatabasePassesOptions()
...
         $this->assertSame(WriteConcern::MAJORITY, $debug['writeConcern']->getW());
     }
+
+    public function testCreateClientEncryption()
+    {
+        $client = new Client(static::getUri());
+
+        $options = [
+            'keyVaultNamespace' => 'default.keys',
+            'kmsProviders' => ['aws' => (object) ['accessKeyId' => 'abc', 'secretAccessKey' => 'def']],
```
jmikola: Same question as https://github.com/mongodb/mongo-php-library/pull/708#discussion_r369545255. Is the `(object)` cast needed?

docs/reference/method/MongoDBClient-createClientEncryption.txt
```diff
@@ -0,0 +1,50 @@
...
+   .. code-block:: php
+
+      function createClientEncryption(array $options): MongoDB\Driver\ClientEncryption
+
+   This method has the following parameters:
+
+   .. include:: /includes/apiargs/MongoDBClient-method-createClientEncryption-param.rst
+
+   The ``$options`` parameter supports all options documented in the extension
+   manual. For the ``keyVaultClient`` option, an instance of ``MongoDB\Client``
```
jmikola:
```
:phpclass:`MongoDB\\Client`
```

Should "extension manual" link to the PHP.net docs, as you did in #708 for the `autoEncryption` driver option? I realize https://www.php.net/manual/en/mongodb-driver-manager.createclientencryption.php does not yet exist, but I suppose you can still add this now since you're linking to the same URL in the "See Also" below.

docs/reference/method/MongoDBClient-createClientEncryption.txt
```diff
@@ -0,0 +1,50 @@
...
+   is automatically unwrapped and the ``MongoDB\Driver\Manager`` instance is
+   passed to the extension.
+
+Return Values
+-------------
+
+A MongoDB\Driver\ClientEncryption instance which can be used to encrypt and
```
jmikola:
```
:phpclass:`MongoDB\\Driver\\ClientEncryption`
```

docs/reference/method/MongoDBClient-createClientEncryption.txt
```diff
@@ -0,0 +1,50 @@
...
+
+      function createClientEncryption(array $options): MongoDB\Driver\ClientEncryption
+
+   This method has the following parameters:
+
+   .. include:: /includes/apiargs/MongoDBClient-method-createClientEncryption-param.rst
+
+   The ``$options`` parameter supports all options documented in the extension
+   manual. For the ``keyVaultClient`` option, an instance of ``MongoDB\Client``
+   is automatically unwrapped and the ``MongoDB\Driver\Manager`` instance is
```
jmikola:
```
:php:`MongoDB\\Driver\\Manager <class.mongodb-driver-manager>`
```

alcaeus: The suggestion above was to use `:phpclass:` instead of `php:` - is there a significant difference between these two styles?

jmikola: `:php:` is used for links to PHP.net, while `:phpclass:` is used for internal links within the PHPLIB docs. I made a typo above, so I'll fix.

docs/reference/method/MongoDBClient-createClientEncryption.txt
jmikola: Sorry, this should be `:php:` and you'll need to add `<class.mongodb-driver-clientencryption>` (preemptively).

docs/reference/method/MongoDBClient-createClientEncryption.txt
```diff
@@ -0,0 +1,51 @@
...
+   .. code-block:: php
+
+      function createClientEncryption(array $options): MongoDB\Driver\ClientEncryption
+
+   This method has the following parameters:
+
+   .. include:: /includes/apiargs/MongoDBClient-method-createClientEncryption-param.rst
+
+   The ``$options`` parameter supports all options documented in the
+   :php:`extension manual<<manual/en/mongodb-driver-manager.createclientencryption.php>`.
```
jmikola: Double `<<` looks like a typo. I think the first `<` should be a space.

docs/reference/method/MongoDBClient-createClientEncryption.txt
```diff
@@ -0,0 +1,51 @@
...
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Definition
+----------
+
+.. phpmethod:: MongoDB\\Client::createClientEncryption()
+
+   Returns a ClientEncryption object for manual encryption and decryption of
```
jmikola: I think it'd be reasonable to use

```
:php:`MongoDB\\Driver\\ClientEncryption <class.mongodb-driver-clientencryption>`
```

here just to provide users another direct link to the actual API they'll end up using. I've noted that it's also in the return section below.

PR #716, opened 2020-01-29
PHPLIB-492: Create tutorial for client side encryption

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-492
https://jira.mongodb.org/browse/PHPLIB-517

### DIFF
--- a/docs/tutorial/client-side-encryption.txt
+++ b/docs/tutorial/client-side-encryption.txt
@@ -1,0 +1,241 @@
+======================
+Client-Side Encryption
+======================
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Client-Side Field Level Encryption allows administrators and developers to
+encrypt specific data fields in addition to other MongoDB encryption features.
+
+
+Automatic Encryption and Decryption
+-----------------------------------
+
+.. note::
+
+   Auto encryption is an enterprise only feature.
+
+The following example uses a local key, however using AWS Key Management Service
+is also an option. The data in the ``encryptedField`` field is automatically
+encrypted on insertion and decrypted when querying on the client side.
+
+.. code-block:: php
+
+   <?php
+
+   use MongoDB\BSON\Binary;
+   use MongoDB\Client;
+
+   $localKey = new Binary('<binary key data (96 bytes)>', Binary::TYPE_GENERIC);
+
+   $encryptionOpts = [
+       'keyVaultNamespace' => 'admin.datakeys',
+       'kmsProviders' => [
+           'local' => ['key' => $localKey],
+       ],
+   ];
+
+   $client = new Client('mongodb://127.0.0.1');
+   $clientEncryption = $client->createClientEncryption($encryptionOpts);
+
+   $database = $client->selectDatabase('test');
+   $database->dropCollection('coll'); // remove old data
+
+   // Create new key in the key vault and store its ID for later use
+   $keyId = $clientEncryption->createDataKey('local');
+
+   $database->createCollection('coll', [
+       'validator' => [
+           '$jsonSchema' => [
+               'bsonType' => 'object',
+               'properties' => [
+                   'encryptedField' => [
+                       'encrypt' => [
+                           'keyId' => [$keyId],
+                           'bsonType' => 'string',
+                           'algorithm' => ClientEncryption::AEAD_AES_256_CBC_HMAC_SHA_512_DETERMINISTIC,
+                       ],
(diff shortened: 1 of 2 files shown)


docs/tutorial/fle.txt
```diff
@@ -0,0 +1,164 @@
...
+
+   $collection->insertOne(['encryptedField' => '123456789']);
+
+   var_dump($collection->findOne([]));
+
+
+Manually encrypting and decrypting values
+-----------------------------------------
+
+In the Community edition of MongoDB, you will have to manually encrypt and
```
jmikola: Perhaps we should refer to this as "MongoDB Community Edition", per https://docs.mongodb.com/manual/administration/install-community/.

docs/tutorial/fle.txt
```diff
@@ -0,0 +1,164 @@
...
+
+
+Automatic encryption and decryption
+-----------------------------------
+
+.. note::
+
+   Auto encryption is an enterprise only feature.
+
+The following example assumes the key and schema have already been created in
```
jmikola: Should there be an example of creating the key and schema? I assume this differs from the second example where we use a `schemaMap`.

docs/tutorial/fle.txt
```diff
@@ -0,0 +1,164 @@
...
+.. code-block:: php
+
+   <?php
+
+   use MongoDB\BSON\Binary;
+   use MongoDB\Client;
+   use MongoDB\Driver\ClientEncryption;
+
+   $localKey = new Binary('<binary key data>', Binary::TYPE_GENERIC);
+   $keyId    = new Binary('<UUID of encryption key', Binary::TYPE_UUID);
```
jmikola: Should we add a comment above this to make the connection to `ClientEncryption::createDataKey()`? I believe we only show that being called in the previous `schemaMap` example, which is also for the enterprise, automatic encryption. That all seems a bit odd, as ClientEncryption seems like it's only useful for the explicit encryption but we end up using `createDataKey` alone for automatic encryption as well.

docs/tutorial/fle.txt
alcaeus: Added key and schema creation using `$jsonValidator` to the example. I would also suggest that the example in the Java documentation is improved the same way, as that's where this example was copied from.

docs/tutorial/fle.txt
alcaeus: I've changed the example and added creating an encryption key using `createDataKey`, so there is no confusion as to where this is coming from. However, this might lead users to believe that they should create a new encryption key anytime they want to encrypt data, which might not be desirable. We may want to address this in the tutorial if we indeed don't recommend doing this.

jmikola: @kevinAlbs I assume we don't expect users to create new data keys each time they need to invoke encryption. Do we expect them to use this method once and then either (a) directly query the key vault collection for future use or (b) save the ID and then hard-code it into their application/environment? I read through most of [ClientEncryption](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#clientencryption) and related references to `keyId` but I don't see much elaboration on this.

kevinAlbs: > I assume we don't expect users to create new data keys each time they need to invoke encryption.

Yes, but they may be creating keys on the fly as part of their application. E.g. if they're creating a per-user data key.

> Do we expect them to use this method once and then either (a) directly query the key vault collection for future use or (b) save the ID and then hard-code it into their application/environment?

That is one way. But they may also associate one or more keyAltNames based on other data (e.g. a user identifier) so would not need to query or hard code the ID.

alcaeus: alternative name is a good point. I've added an example to show how this is used and also explained that encryption keys shouldn't be created every time a value is encrypted. I hope this clarifies the solution.

docs/tutorial/client-side-encryption.txt
```diff
@@ -0,0 +1,241 @@
...
+Referencing Encryption Keys by an Alternative Name
+--------------------------------------------------
+
+While it is possible to create an encryption key every time data is encrypted,
+this is not the recommended approach. Instead, you should create your encryption
+keys depending on your use-case, e.g. by creating a user-specific encryption
+key. To reference keys in your software, you can use the keyAltName attribute
+specified when creating the key. The following example creates an encryption key
+with an alternative name, which could be done when deploying the application.
+The software then emcrypts data by referencing the key by its alternative name.
```
jmikola: Ah, missed this earlier but "emcrypts" is a typo. May want to run spell-check on this file if you've not already.

alcaeus:
Merged manually in 02bd585.

PR #718, opened 2020-02-04
PHPLIB-499: Handle absence of 'ns' field in index specifications returned from listIndexes

alcaeus:
This serves as POC for the change. When creating an `IndexInfoIterator`, we provide it with an optional namespace, which will be used for the `ns` field in the index information if it wasn't provided by the server (which will be the case for 4.4).

This PR also deprecates the `getNamespace` method for consistency with the server. We may want to keep the method, but considering that the namespace is no longer part of the index information returned by the server, we may opt to drop the method as well.

### DIFF
--- a/src/Model/IndexInfoIteratorIterator.php
+++ b/src/Model/IndexInfoIteratorIterator.php
@@ -18,6 +18,8 @@
 namespace MongoDB\Model;
 
 use IteratorIterator;
+use Traversable;
+use function array_key_exists;
 
 /**
  * IndexInfoIterator for both listIndexes command and legacy query results.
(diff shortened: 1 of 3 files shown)


src/Model/IndexInfo.php
```diff
@@ -95,9 +97,12 @@ public function getName()
      * Return the index namespace (e.g. "db.collection").
      *
      * @return string
+     * @deprecated Deprecated in 1.7.0
      */
     public function getNamespace()
     {
+        @trigger_error(sprintf('The "%s" method is deprecated since mongodb/mongodb 1.7 and will be dropped in 2.0.', __METHOD__));
```
jmikola: IIRC, we used this approach in other cases where the server stopped supporting an option (e.g. `mapReduce`). In this case, we know the server is no longer supporting this but there's no danger of using this method since we'll still populate the value.

Would it make sense to defer this until the final 1.x release before 2.0, similar to what is done in Doctrine?

alcaeus: I've removed the deprecation for now, we can always deprecate in a later release.

jmikola: SGTM. I also just realized that https://github.com/mongodb/specifications/commit/8ed5d562f073f2e43e24c092492fa5ffe1a6b115 doesn't actually require deprecation.

jmikola:
FYI, this originally failed with the intermittent double-free error on shutdown in the sharded replica set cluster ([PHPC-1541](https://jira.mongodb.org/browse/PHPC-1541)). Re-kicking the build passed on the second attempt.

alcaeus:
Merged manually in 23e32dd.

PR #720, opened 2020-02-11
PHPLIB-529: Drop support for PHP 5.6

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-529

Follows https://github.com/mongodb/mongo-php-driver/pull/1096.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,14 +10,14 @@
         { "name": "Katherine Walker", "email": "katherine.walker@mongodb.com" }
     ],
     "require": {
-        "php": "^5.6 || ^7.0",
+        "php": "^7.0",
         "ext-hash": "*",
         "ext-json": "*",
         "ext-mongodb": "^1.7"
     },
     "require-dev": {
-        "phpunit/phpunit": "^5.7.27 || ^6.4 || ^8.3",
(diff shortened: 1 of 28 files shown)


composer.json
```diff
@@ -10,15 +10,15 @@
...
+        "php": "^7.0",
         "ext-hash": "*",
         "ext-json": "*",
         "ext-mongodb": "^1.7"
     },
     "require-dev": {
-        "phpunit/phpunit": "^5.7.27 || ^6.4 || ^8.3",
-        "sebastian/comparator": "^1.0 || ^2.0 || ^3.0",
-        "squizlabs/php_codesniffer": "^3.4",
+        "phpunit/phpunit": "^6.4 || ^8.3",
```
jmikola: Are there any compatibility shims in place for PHPUnit 5.x that can now be removed? If so, that doesn't need to happen in this PR if you'd rather defer that to another issue and backlog the task.

alcaeus: The ConstraintTrait we use is needed to differentiate between PHPUnit 6 and 7+, while the SetupTearDownTrait is necessary until we go to PHP 7.1 as minimum version, since 7.0 doesn't support the `void` return type hint.

As for PolyfillAssertTrait, some of those methods were only introduced in PHPUnit 8, so it's also still necessary.

PR #721, opened 2020-02-11
 PHPLIB-500: Support for allowDiskUse on find operations

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-500

PHP 5 build failures will be fixed by #720.

### DIFF
--- a/docs/includes/apiargs-MongoDBCollection-method-findOne-option.yaml
+++ b/docs/includes/apiargs-MongoDBCollection-method-findOne-option.yaml
@@ -11,6 +11,10 @@ source:
   ref: skip
 ---
 source:
+  file: apiargs-MongoDBCollection-method-find-option.yaml
+  ref: allowDiskUse
+---
+source:
   file: apiargs-MongoDBCollection-common-option.yaml
   ref: collation
 ---
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,9 @@ cache:
 
 env:
   global:
-    - DRIVER_VERSION=1.7.0
+    - DRIVER_VERSION=1.8.0
+# Todo: Remove when v1.8 has been branched
+    - DRIVER_BRANCH="master"
     - SERVER_DISTRO=enterprise-ubuntu1604
     - SERVER_VERSION=4.2.0
     - DEPLOYMENT=STANDALONE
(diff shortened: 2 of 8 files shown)


tests/SpecTests/DocumentsMatchConstraint.php
```diff
@@ -276,6 +276,10 @@ private function assertEquals(ArrayObject $expected, ArrayObject $actual, $ignor
         }
 
         foreach ($expected as $key => $expectedValue) {
+            if ($expectedValue === null) {
+                continue;
```
jmikola: I assume this was done to satisfy the [Fields that must NOT be present in Actual Documents](https://github.com/mongodb/specifications/commit/92255e0e57027646e8ef23399d3f33fdc114221f#diff-c4a2b694daca296938e31db2cd9b5206R289) spec test change introduced by mongodb/specifications#692.

I think this is technically incorrect, as it's skipping the assertion entirely if the expected value is `null`. At the very least, I think you'd want to assert that `$actualHasKey` (set below) is `false`.\\But rather than modify DocumentsMatchConstraint at all (which inadvertently affect other assertions, such as change stream results), I'd propose handling this in `CrudSpecTest::assertCommandMatches()`, similar to as is done in [TransactionSpecTest here](https://github.com/mongodb/mongo-php-library/blob/1.6.0/tests/SpecTests/TransactionsSpecTest.php#L107). I believe the transaction spec tests are the only other tests that use the same "null means unset" concept for command assertions.

Lastly, I believe this PR is also missing `find-allowdiskuse.json` from https://github.com/mongodb/specifications/commit/92255e0e57027646e8ef23399d3f33fdc114221f.

docs/includes/apiargs-MongoDBCollection-method-findOne-option.yaml
```diff
@@ -10,6 +10,10 @@ source:
   file: apiargs-MongoDBCollection-method-find-option.yaml
   ref: skip
 ---
+source:
+  file: apiargs-MongoDBCollection-method-find-option.yaml
+  ref: allowDiskUse
```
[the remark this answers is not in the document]
alcaeus: Added it there and to `apiargs-MongoDBGridFSBucket-method-findOne-option.yaml` as well.

tests/SpecTests/DocumentsMatchConstraint.php
alcaeus: Good point. I've moved the check to not affect other tests. The `find-allowdiskuse.json` test was added in #717, but didn't get executed because we're not testing against a 4.3 server. Since some tests currently fail on 4.3, I'll add this in a separate PR and hold off on merging this until we're properly testing this functionality.

jmikola: Ah, I totally forgot that was included in my PR.

alcaeus: Yep, I forgot to push this change 😅

alcaeus:
@jmikola holding off on merging this until [PHPLIB-530](https://jira.mongodb.org/browse/PHPLIB-530) is done as the functionality introduced here currently isn't tested for lack of a 4.3 build stage.

PR #726, opened 2020-03-10
PHPLIB-536: Add spec tests for default write concern

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-536

### DIFF
--- a/tests/SpecTests/CommandExpectations.php
+++ b/tests/SpecTests/CommandExpectations.php
@@ -102,6 +102,16 @@ class CommandExpectations implements CommandSubscriber
         return $o;
     }
 
+    public static function fromReadWriteConcern(array $expectedEvents)
+    {
+        $o = new self($expectedEvents);
+
+        $o->ignoreCommandFailed = true;
+        $o->ignoreCommandSucceeded = true;
+
+        return $o;
+    }
+
     public static function fromRetryableReads(array $expectedEvents)
     {
         $o = new self($expectedEvents);
--- a/tests/SpecTests/Context.php
+++ b/tests/SpecTests/Context.php
@@ -166,6 +166,21 @@ final class Context
         return $o;
     }
 
+    public static function fromReadWriteConcern(stdClass $test, $databaseName, $collectionName)
+    {
+        $o = new self($databaseName, $collectionName);
+
+        if (isset($test->outcome->collection->name)) {
+            $o->outcomeCollectionName = $test->outcome->collection->name;
+        }
+
+        $clientOptions = isset($test->clientOptions) ? (array) $test->clientOptions : [];
+
+        $o->client = new Client(FunctionalTestCase::getUri(), $clientOptions);
+
+        return $o;
+    }
+
     public static function fromRetryableReads(stdClass $test, $databaseName, $collectionName, $bucketName)
     {
         $o = new self($databaseName, $collectionName);
@@ -253,12 +268,13 @@ final class Context
         return $this->useEncryptedClient && $this->encryptedClient ? $this->encryptedClient : $this->client;
     }
 
-    public function getCollection(array $collectionOptions = [])
+    public function getCollection(array $collectionOptions = [], array $databaseOptions = [])
     {
         return $this->selectCollection(
             $this->databaseName,
             $this->collectionName,
-            $this->prepareOptions($collectionOptions)
+            $collectionOptions,
+            $databaseOptions
         );
     }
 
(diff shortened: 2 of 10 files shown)


tests/SpecTests/ReadWriteConcernSpecTest.php
```diff
@@ -0,0 +1,121 @@
...
+namespace MongoDB\Tests\SpecTests;
+
+use stdClass;
+use function basename;
+use function dirname;
+use function file_get_contents;
+use function glob;
+
+/**
+ * Transactions spec tests.
```
jmikola: Copypasta

tests/SpecTests/CommandExpectations.php
```diff
@@ -102,6 +102,21 @@ public static function fromCrud(array $expectedEvents)
...
     }
 
+    public static function fromReadWriteConcern(array $expectedEvents)
+    {
+        $o = new self($expectedEvents);
+
+        $o->ignoreCommandFailed = true;
+        $o->ignoreCommandSucceeded = true;
+
+        /* Retryable read spec tests don't include extra commands, e.g. the
```
jmikola: Is this copypasta?

alcaeus: Yep. I really need to refactor these tests to avoid copy pasting all this code every time we add a new spec test runner.

tests/SpecTests/ReadWriteConcernSpecTest.php
```diff
@@ -0,0 +1,119 @@
...
+
+        if (isset($test->skipReason)) {
+            $this->markTestSkipped($test->skipReason);
+        }
+
+        if (isset($runOn)) {
+            $this->checkServerRequirements($runOn);
+        }
+
+        if (isset($test->skipReason)) {
```
p-mongo: This seems to have already been checked earlier on line 55.

tests/SpecTests/ReadWriteConcernSpecTest.php
```diff
@@ -0,0 +1,119 @@
...
+use stdClass;
+use function basename;
+use function dirname;
+use function file_get_contents;
+use function glob;
+
+/**
+ * @see https://github.com/mongodb/specifications/tree/master/source/read-write-concern
+ */
+class ReadWriteConcernSpecTest extends FunctionalTestCase
```
p-mongo: This entire class/file appears to be identical to CrudSpecTest. The only difference I could see was the argument to glob() call, giving a different path. Would it not be better to reuse this code somehow and only specify the different path to where the spec test files are stored?

tests/SpecTests/Operation.php
```diff
@@ -184,6 +184,16 @@ public static function fromCrud(stdClass $operation)
         return $o;
     }
 
+    public static function fromReadWriteConcern(stdClass $operation)
```
p-mongo: Unlike `fromCrud`, this method has no mention of `collectionOptions`. Does it actually pass collection options to the collection? Looking at the test json files, the write concern always seems to be the empty document, and I am wondering whether the tests pass because they do not attempt to set the default write concern when performing operations rather than because the driver does not send the empty write concern document to the server.

If the `collectionOptions` do need to be set here, it seems sensible to simply call `fromCrud`.

tests/SpecTests/Context.php
```diff
@@ -312,13 +327,17 @@ public function prepareOptions(array $options)
...
 
-            $options['writeConcern'] = isset($j)
-                ? new WriteConcern($w, $wtimeout, $j)
-                : new WriteConcern($w, $wtimeout);
+                $options['writeConcern'] = isset($j)
+                    ? new WriteConcern($w, $wtimeout, $j)
+                    : new WriteConcern($w, $wtimeout);
+            } else {
+                unset($options['writeConcern']);
+            }
```
p-mongo: This block does not seem to handle additional fields existing in `$writeConcern`. If a spec test has any such fields, they would be silently ignored instead of producing an error because the spec runner does not handle them.

tests/SpecTests/ReadWriteConcernSpecTest.php
alcaeus: Yes, there is a lot of duplication in these tests. I plan to refactor them to extract common functionality and build extension points, since some of the spec tests use a slightly different format.

tests/SpecTests/Context.php
alcaeus: The code block a couple of lines above handles this:
```
$writeConcern = (array) $options['writeConcern'];
$diff = array_diff_key($writeConcern, ['w' => 1, 'wtimeout' => 1, 'j' => 1]);

if (! empty($diff)) {
    throw new LogicException('Unsupported writeConcern args: ' . implode(',', array_keys($diff)));
}
```

tests/SpecTests/Operation.php
alcaeus: Great catch. I've fixed this and added handling for `databaseOptions` as well, which was previously ignored.

tests/SpecTests/ReadWriteConcernSpecTest.php
jmikola: The unified test runner project might also allow us to consolidate everything into a common `UnifiedSpecTest.php` file, and then have extending classes only implement a data provider to pull in relevant JSON files.

p-mongo: :+1:

PR #729, opened 2020-03-24
PHPLIB-537: Use whitelist to check if a change stream is resumable

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-537


src/ChangeStream.php
```diff
@@ -180,15 +198,15 @@ private function isResumableError(RuntimeException $exception)
...
         }
 
-        if ($exception->hasErrorLabel('NonResumableChangeStreamError')) {
-            return false;
+        if (server_supports_feature($this->iterator->getServer(), self::$wireVersionForResumableChangeStreamError)) {
+            return $exception->hasErrorLabel('ResumableChangeStreamError');
         }
 
-        if (in_array($exception->getCode(), self::$nonResumableErrorCodes)) {
+        if ($exception->hasErrorLabel('NonResumableChangeStreamError')) {
```
p-mongo: I do not see anything in the [spec](https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst) about "NonResumableChangeStreamError" label, is this something PHP POCed?

p-mongo: Based on https://github.com/mongodb/specifications/commit/f93d78191f3db2898a59013a7ed5650352ef6da8#diff-875f237661962b4ca9b2e58d97449831L49 and the conversation in driver-devs I am under the impression that this error label should no longer be consulted, what do you think @alcaeus ?

alcaeus: Yes, that is correct. I've removed the label.

PR #731, opened 2020-03-24
PHPLIB-516: Append PHPLIB version information to handshake data

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-516

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,8 @@
         "php": "^7.0",
         "ext-hash": "*",
         "ext-json": "*",
-        "ext-mongodb": "^1.8"
+        "ext-mongodb": "^1.8",
+        "jean85/pretty-package-versions": "^1.2"
     },
     "require-dev": {
         "phpunit/phpunit": "^6.4 || ^8.3",
--- a/docs/includes/apiargs-MongoDBClient-method-construct-driverOptions.yaml
+++ b/docs/includes/apiargs-MongoDBClient-method-construct-driverOptions.yaml
@@ -142,6 +142,34 @@ description: |
   For the ``keyVaultClient`` option, you may pass a :phpclass:`MongoDB\\Client`
   instance, which will be unwrapped to provide a :php:`MongoDB\\Driver\\Manager <class.mongodb-driver-manager>`
   to the extension.
+  .. versionadded:: 1.6
+interface: phpmethod
+operation: ~
+optional: true
+---
+arg_name: option
+name: driver
+type: array
+description: |
+  Additional driver metadata to be passed on to the server handshake. This is an
+  array containing ``name``, ``version``, and ``platform`` fields:
+
+  .. code-block:: php
+
+     [
+         'name' => 'my-driver',
+         'version' => '1.2.3-dev',
+         'platform' => 'some-platform',
+     ]
+
+  .. note::
+
+     This feature is primarily designed for custom drivers and ODMs, which may
+     want to identify themselves to the server for diagnostic purposes.
+     Applications should use the ``appName`` URI option instead of driver
+     metadata.
+
+  .. versionadded:: 1.7
 interface: phpmethod
 operation: ~
 optional: true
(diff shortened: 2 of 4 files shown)


composer.json
```diff
@@ -13,7 +13,8 @@
         "php": "^7.0",
         "ext-hash": "*",
         "ext-json": "*",
-        "ext-mongodb": "^1.8"
+        "ext-mongodb": "^1.8",
+        "jean85/pretty-package-versions": "^1.2"
```
alcaeus: This provides a better interface around version strings returned by [ocramius/package-versions](https://packagist.org/packages/ocramius/package-versions). It is used to provide the current version of this library without having to maintain it in a version constant.

src/Client.php
```diff
@@ -354,4 +365,47 @@ public function watch(array $pipeline = [], array $options = [])
 
         return $operation->execute($server);
     }
+
+    private static function getVersion() : string
+    {
+        if (self::$version === null) {
+            try {
+                self::$version = PrettyVersions::getVersion('mongodb/mongodb')->getPrettyVersion();
```
alcaeus: While the package-versions package caches package information when they are installed, I thought it would be best to still cache this data to ensure we're not unnecessarily causing delays on connecting.

tests/ClientTest.php
```diff
@@ -54,6 +54,14 @@ public function provideInvalidConstructorDriverOptions()
 
         $options[][] = ['autoEncryption' => ['keyVaultClient' => 'foo']];
 
+        foreach ($this->getInvalidStringValues() as $value) {
+            $options[][] = ['driver' => ['name' => $value]];
+        }
+
+        foreach ($this->getInvalidStringValues() as $value) {
+            $options[][] = ['driver' => ['version' => $value]];
```
jmikola: Should we also test `platform` here?

src/Client.php
```diff
@@ -354,4 +365,47 @@ public function watch(array $pipeline = [], array $options = [])
...
+            } catch (OutOfBoundsException $e) {
+                return 'unknown';
```
jmikola: I took a look through PrettyVersions and it seems that it doesn't throw anything on its own, so I assume this comes from Ocramius' library. If so, I think there's also a case where [UnexpectedValueException could be thrown](https://github.com/Ocramius/PackageVersions/blob/651c372efc914aea8223e049f85afaf65e09ba23/src/PackageVersions/Versions.php#L27). Would it be safer to just catch any RuntimeException?

Also, I'm totally on board with _not_ exposing the [original exception](https://github.com/Ocramius/PackageVersions/blob/651c372efc914aea8223e049f85afaf65e09ba23/src/PackageVersions/FallbackVersions.php#L95). I don't think "fix your shit" would go over well with users, especially when it might not be clear which library is generating that message :laughing:

docs/includes/apiargs-MongoDBClient-method-construct-driverOptions.yaml
```diff
@@ -142,6 +142,26 @@ description: |
...
+  array containing ``name``, ``version``, and ``platform`` fields:
+
+  .. code-block:: php
+
+     [
+         'name' => 'my-driver',
+         'version' => '1.2.3-dev',
+         'platform' => 'some-platform',
+     ]
+  .. versionadded:: 1.7
```
jmikola: Should we add a note that this is best used for things like wrapping libraries and ODMs rather than applications?

tests/ClientTest.php
alcaeus: Added, but requires a little extra work since we only pass platform information through to the driver.

src/Client.php
alcaeus: Changed to catch any throwable. That exception message (along with not wanting to break a connection to MongoDB because we couldn't determine the version) was the main reason for adding a try-catch block.

tests/ClientTest.php
jmikola: Oh, I didn't realize that when I left my first comment. I see now that PHPLIB is only throwing its own `InvalidArgumentException` for the two options _it_ merges. Thanks for adding this just the same.

PR #735, opened 2020-04-06
PHPLIB-434: Add note about performance of countDocuments

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-434

### DIFF
--- a/docs/reference/method/MongoDBCollection-countDocuments.txt
+++ b/docs/reference/method/MongoDBCollection-countDocuments.txt
@@ -52,6 +52,14 @@ obtain the result. If a ``filter`` parameter is given, this is converted into
 a ``$match`` pipeline operator. Optional ``$skip`` and ``$limit`` stages are
 added between ``$match`` and ``group`` if present in the options.
 
+.. note::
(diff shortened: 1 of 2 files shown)


docs/reference/method/MongoDBCollection-countDocuments.txt
```diff
@@ -52,6 +52,14 @@ obtain the result. If a ``filter`` parameter is given, this is converted into
 a ``$match`` pipeline operator. Optional ``$skip`` and ``$limit`` stages are
 added between ``$match`` and ``group`` if present in the options.
 
+.. note::
+
+   This method counts documents on the server side. To obtain the total number
```
p-mongo, line 57: I think it would be  clearer to say "To obtain an approximate total number of documents ...". When I read the current language, I first see  "the total number" which to me is the exact number, and when I read the rest of the sentence I mentally go back and revise "the total number" to become "an estimated number".

alcaeus: Good point, updated!

PR #736, opened 2020-04-06
PHPLIB-522: Loosen index equality checks for GridFS

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-522

Instead of using a strict array comparison (which checks order of elements as well as type-safe value checks), we are now using a `MultipleIterator` to iterate over the expected and actual keys in order. Since we assume numeric values for index sort only, we check if the actual value is numeric, then cast to int to cover for some drivers (e.g. the shell) creating indexes with float values for sort.

For the chunks index, we attempt to create the index even if an index with the same keys but non-unique exists. This will cause an exception, but there is no other way to work around this issue. Other index options are not considered for either index.

### DIFF
--- a/src/GridFS/CollectionWrapper.php
+++ b/src/GridFS/CollectionWrapper.php
@@ -17,14 +17,18 @@
 
 namespace MongoDB\GridFS;
 
+use ArrayIterator;
 use MongoDB\Collection;
 use MongoDB\Driver\Cursor;
 use MongoDB\Driver\Manager;
 use MongoDB\Driver\ReadPreference;
 use MongoDB\Exception\InvalidArgumentException;
 use MongoDB\UpdateResult;
+use MultipleIterator;
 use stdClass;
 use function abs;
+use function count;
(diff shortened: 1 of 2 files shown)


src/GridFS/CollectionWrapper.php
```diff
@@ -334,6 +342,32 @@ private function ensureIndexes()
...
+
+        foreach ($iterator as $key => $value) {
+            list($expectedKey, $actualKey)     = $key;
+            list($expectedValue, $actualValue) = $value;
+
+            if ($expectedKey !== $actualKey) {
+                return false;
+            }
+
+            if (! is_numeric($actualValue) || (int) $expectedValue !== (int) $actualValue) {
```
p-mongo, line 363: If the actual value is not numeric, this function will return false even if values are identical, right? Maybe worth noting that somewhere.

alcaeus: Added a clarifying comment. Thanks!

PR #741, opened 2020-04-28
PHPLIB-544: Add commitQuorum option to createIndexes

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-544

### DIFF
--- a/docs/includes/apiargs-MongoDBCollection-method-createIndex-option.yaml
+++ b/docs/includes/apiargs-MongoDBCollection-method-createIndex-option.yaml
@@ -1,4 +1,23 @@
 arg_name: option
+name: commitQuorum
+type: string|integer
+description: |
+  Specifies how many data-bearing members of a replica set, including the
+  primary, must complete the index builds successfully before the primary marks
+  the indexes as ready.
+
+  This option accepts the same values for the "w" field in a write concern plus
+  "votingMembers", which indicates all voting data-bearing nodes.
+
+  This is not supported for server versions prior to 4.4 and will result in an
+  exception at execution time if used.
+
+  .. versionadded:: 1.7
+interface: phpmethod
+operation: ~
+optional: true
+---
+arg_name: option
 name: unique
 type: boolean
 description: |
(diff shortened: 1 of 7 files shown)


docs/includes/apiargs-MongoDBCollection-method-createIndex-option.yaml
```diff
@@ -1,4 +1,23 @@
 arg_name: option
+name: commitQuorum
+type: string|integer
+description: |
+  Specifies how many data-bearing members of a replica set, including the
+  primary, must complete the index builds successfully before the primary marks
+  the indexes as ready.
+
+  This option accepts the same values for the "w" field in a write concern plus
+  "votingMembers", which indicates all voting data-bearing nodes.
```
jmikola, line 10: I'd revise as:

> ... in a write concern in addition to the string `"votingMembers"`, which...

Monospace formatting helps indicate that we're talking about a string value here. I think we do that elsewhere in our docs.

The "w" field could probably just be reformatted as `w` (monospace, but no quotes), since we're only referring to a key.

docs/includes/apiargs-MongoDBCollection-method-createIndexes-option.yaml
```diff
@@ -1,3 +1,22 @@
+arg_name: option
+name: commitQuorum
+type: string|integer
+description: |
```
jmikola, line 4: Is there any way these can share the same include? Rather than add this to `apiargs-MongoDBCollection-common-option.yaml` perhaps we could just include the content from `apiargs-MongoDBCollection-method-createIndex-option.yaml`.

src/Operation/CreateIndexes.php
```diff
@@ -67,6 +72,8 @@ class CreateIndexes implements Executable
      *
      * Supported options:
      *
+     *  * commitQuorum (integer|string):
```
jmikola, line 75: Missing a short description here.

src/Operation/CreateIndexes.php
```diff
@@ -202,6 +213,16 @@ private function executeCommand(Server $server)
             'indexes' => $this->indexes,
         ];
 
+        if (isset($this->options['commitQuorum'])) {
+            /* Drivers MUST manually raise an error if this option is specified
+             * when creating an index on a pre 4.4 server. */
+            if (! server_supports_feature($server, self::$wireVersionForCommitQuorum)) {
+                throw new CommandException('Invalid field specified for createIndexes command: commitQuorum', 2);
```
jmikola, line 220: Where is code `2` come from?

I think this should throw an UnsupportedException via a new factory method for consistency.

PR #749, opened 2020-05-26
PHPLIB-556: Throw client-side error when allowDiskUse is not supported by a server

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-556

### DIFF
--- a/src/Operation/Find.php
+++ b/src/Operation/Find.php
@@ -55,6 +55,9 @@ class Find implements Executable, Explainable
     /** @var integer */
     private static $wireVersionForReadConcern = 4;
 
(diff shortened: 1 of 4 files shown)


src/Operation/Find.php
```diff
@@ -55,6 +55,9 @@ class Find implements Executable, Explainable
     /** @var integer */
     private static $wireVersionForReadConcern = 4;
 
+    /** @var integer */
+    private static $wireVersionForAllowDiskUseServerSideError = 4;
```
jmikola, line 59: Thanks for clarifying "server-side error" here.

PR #758, opened 2020-06-29
PHPLIB-547: Add helpers to list database and collection names

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-547

### DIFF
--- a/docs/reference/method/MongoDBClient-listDatabaseNames.txt
+++ b/docs/reference/method/MongoDBClient-listDatabaseNames.txt
@@ -1,6 +1,8 @@
-================================
-MongoDB\\Client::listDatabases()
-================================
+====================================
+MongoDB\\Client::listDatabaseNames()
+====================================
+
+.. versionadded:: 1.7
 
 .. default-domain:: mongodb
 
@@ -13,13 +15,13 @@ MongoDB\\Client::listDatabases()
 Definition
 ----------
 
-.. phpmethod:: MongoDB\\Client::listDatabases()
+.. phpmethod:: MongoDB\\Client::listDatabaseNames()
 
-   Returns information for all databases on the server.
+   Returns names for all databases on the server.
 
    .. code-block:: php
 
-      function listDatabases(array $options = []): MongoDB\Model\DatabaseInfoIterator
+      function listDatabaseNames(array $options = []): Iterator
 
    This method has the following parameters:
 
@@ -32,9 +34,8 @@ Definition
 Return Values
 -------------
 
-A traversable :phpclass:`MongoDB\\Model\\DatabaseInfoIterator`, which contains
-a :phpclass:`MongoDB\\Model\\DatabaseInfo` object for each database on the
-server.
+An :php:`Iterator <class.iterator.php>`, which provides the name of each
+database on the server.
 
 Errors/Exceptions
 -----------------
@@ -54,33 +55,19 @@ The following example lists all databases on the server:
 
    $client = new MongoDB\Client;
 
-   foreach ($client->listDatabases() as $databaseInfo) {
-       var_dump($databaseInfo);
+   foreach ($client->listDatabaseNames() as $databaseName) {
+       var_dump($databaseName);
    }
 
 The output would then resemble::
 
-   object(MongoDB\Model\DatabaseInfo)#4 (3) {
-     ["name"]=>
-     string(5) "local"
-     ["sizeOnDisk"]=>
-     float(65536)
-     ["empty"]=>
-     bool(false)
-   }
-   object(MongoDB\Model\DatabaseInfo)#7 (3) {
-     ["name"]=>
-     string(4) "test"
-     ["sizeOnDisk"]=>
-     float(32768)
-     ["empty"]=>
-     bool(false)
-   }
+   string(5) "local"
+   string(4) "test"
 
 See Also
 --------
 
-- :phpmethod:`MongoDB\\Database::listDatabaseNames()`
+- :phpmethod:`MongoDB\\Database::listDatabases()`
 - :manual:`listDatabases </reference/command/listDatabases>` command reference
   in the MongoDB manual
 - `Enumerating Databases
--- a/docs/reference/method/MongoDBDatabase-listCollectionNames.txt
+++ b/docs/reference/method/MongoDBDatabase-listCollectionNames.txt
@@ -1,0 +1,98 @@
+========================================
+MongoDB\\Database::listCollectionNames()
+========================================
+
+.. versionadded:: 1.7
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Definition
+----------
+
+.. phpmethod:: MongoDB\\Database::listCollectionNames()
+
+   Returns names for all collections in this database.
+
+   .. code-block:: php
+
+      function listCollectionNames(array $options = []): Iterator
+
+   This method has the following parameters:
+
+   .. include:: /includes/apiargs/MongoDBDatabase-method-listCollections-param.rst
+
+   The ``$options`` parameter supports the following options:
+
+   .. include:: /includes/apiargs/MongoDBDatabase-method-listCollections-option.rst
+
+Return Values
+-------------
+
+An :php:`Iterator <class.iterator.php>`, which provides the name of each
+collection in the database.
+
+Example
+-------
+
+The following example lists all of the collections in the ``test`` database:
+
+.. code-block:: php
+
+   <?php
+
+   $database = (new MongoDB\Client)->test;
+
+   foreach ($database->listCollectionNames() as $collectionName) {
+       var_dump($collectionName);
+   }
+
+The output would then resemble::
+
+   string(11) "restaurants"
+   string(5) "users"
+   string(6) "restos"
+
+The following example lists all collections whose name starts with ``"rest"``
+in the ``test`` database:
+
+.. code-block:: php
+
+   <?php
+
+   $database = (new MongoDB\Client)->test;
+
+   $collections = $database->listCollectionNames([
+       'filter' => [
+           'name' => new MongoDB\BSON\Regex('^rest.*'),
+       ],
+   ]);
+
+   foreach ($collections as $collectionName) {
+       var_dump($collectionName);
+   }
+
+The output would then resemble::
+
+   string(11) "restaurants"
+   string(6) "restos"
+
+.. note::
+
+   When enumerating collection names, a filter expression can only filter based
+   on a collection's name and type. No other fields are available.
+
+See Also
+--------
+
+- :phpmethod:`MongoDB\\Database::listCollections()`
+- :manual:`listCollections </reference/command/listCollections` command
+  reference in the MongoDB manual
+- `Enumerating Collections
+  <https://github.com/mongodb/specifications/blob/master/source/enumerate-collections.rst>`_
+  specification
--- a/src/Client.php
+++ b/src/Client.php
@@ -17,6 +17,7 @@
 
 namespace MongoDB;
 
+use Iterator;
 use Jean85\PrettyVersions;
 use MongoDB\Driver\ClientEncryption;
 use MongoDB\Driver\Exception\InvalidArgumentException as DriverInvalidArgumentException;
@@ -33,6 +34,7 @@ use MongoDB\Model\BSONArray;
 use MongoDB\Model\BSONDocument;
 use MongoDB\Model\DatabaseInfoIterator;
 use MongoDB\Operation\DropDatabase;
+use MongoDB\Operation\ListDatabaseNames;
 use MongoDB\Operation\ListDatabases;
 use MongoDB\Operation\Watch;
 use Throwable;
@@ -271,6 +273,22 @@ class Client
     public function getWriteConcern()
     {
         return $this->writeConcern;
+    }
+
+    /**
+     * List database names.
+     *
+     * @see ListDatabaseNames::__construct() for supported options
+     * @throws UnexpectedValueException if the command response was malformed
+     * @throws InvalidArgumentException for parameter/option parsing errors
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function listDatabaseNames(array $options = []) : Iterator
+    {
+        $operation = new ListDatabaseNames($options);
+        $server = select_server($this->manager, $options);
+
+        return $operation->execute($server);
     }
 
     /**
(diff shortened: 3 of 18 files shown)


tests/SpecTests/RetryableReadsSpecTest.php
```diff
@@ -18,9 +18,7 @@ class RetryableReadsSpecTest extends FunctionalTestCase
 {
     /** @var array */
     private static $skippedOperations = [
-        'listCollectionNames' => 'Not implemented',
         'listCollectionObjects' => 'Not implemented',
-        'listDatabaseNames' => 'Not implemented',
```
jmikola, line 23: Yay for free test coverage.

src/Command/ListCollections.php
```diff
@@ -0,0 +1,135 @@
...
+     * Constructs a listCollections command.
+     *
+     * Supported options:
+     *
+     *  * filter (document): Query by which to filter collections.
+     *
+     *  * maxTimeMS (integer): The maximum amount of time to allow the query to
+     *    run.
+     *
+     *  * session (MongoDB\Driver\Session): Client session.
```
jmikola, line 56: Should `nameOnly` be documented here? I realize it's not in the public-facing docs for the operations themselves, but this is the raw command.

docs/reference/method/MongoDBDatabase-listCollectionNames.txt
```diff
@@ -0,0 +1,97 @@
...
+
+The output would then resemble::
+
+   string(11) "restaurants"
+   string(6) "restos"
+
+.. note::
+
+   When enumerating collection names, a filter expression can only filter based
+   on a collection's name and type. No other fields are available.
```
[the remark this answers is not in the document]
alcaeus: This restriction only applies when `nameOnly` is `true`. From the [documentation](https://docs.mongodb.com/manual/reference/command/listCollections/):
> When nameOnly is true, your filter expression can only filter based on a collection’s name and type. No other fields are available.

This restriction does not apply when listing full collection information, but applies because `nameOnly` is applied before any filters.

src/Client.php
```diff
@@ -273,6 +275,24 @@ public function getWriteConcern()
...
+     * List database names.
+     *
+     * @see ListDatabaseNames::__construct() for supported options
+     * @param array $options
+     * @return Generator
+     * @throws UnexpectedValueException if the command response was malformed
+     * @throws InvalidArgumentException for parameter/option parsing errors
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function listDatabaseNames(array $options = [])
```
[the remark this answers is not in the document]
alcaeus: I've started adding them to new methods, but forgot them in this case. I added type hints now.

src/Command/ListCollections.php
```diff
@@ -0,0 +1,135 @@
...
+use function is_integer;
+use function is_object;
+
+/**
+ * Wrapper for the listCollections command.
+ *
+ * @internal
+ * @see http://docs.mongodb.org/manual/reference/command/listCollections/
+ */
+class ListCollections implements Executable
```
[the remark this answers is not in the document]
alcaeus: The main advantage of using `nameOnly` is that the server does not need to lock each collection in the database, which it needs to do when grabbing full statistics. In the interest of not creating `CollectionInfo` documents with partial information, I opted against re-using the `ListCollection` operation even when we don't expose the custom iterator.

Another main reason for extracting to its own namespace was to avoid conflicts between the command and the `ListCollections` operation, which does not accept the `nameOnly` option due to the reason above.

src/Operation/ListCollectionNames.php
```diff
@@ -0,0 +1,79 @@
+<?php
+/*
+ * Copyright 2015-2017 MongoDB, Inc.
```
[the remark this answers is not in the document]
alcaeus: I've updated new files with "2020-present". We may want to look into either changing the year for all files or omitting the second portion, as the copyright period extends from when the code was created to a certain time after the author's death (which may be more complicated since the Copyright is owned by MongoDB).

src/Operation/ListCollectionNames.php
```diff
@@ -0,0 +1,79 @@
...
+     * Execute the operation.
+     *
+     * @see Executable::execute()
+     * @param Server $server
+     * @return Generator
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function execute(Server $server)
+    {
+        yield from new CallbackIterator(
```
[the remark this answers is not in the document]
alcaeus: No functional difference, just a different style. I find this style easier to read, but can change to using a loop if you prefer to yield from a loop.

src/Command/ListCollections.php
alcaeus: Yes, that was an oversight. Added.

src/Command/ListCollections.php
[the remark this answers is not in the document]
jmikola: Works for me.

src/Operation/ListCollectionNames.php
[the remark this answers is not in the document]
jmikola: See https://jira.mongodb.org/browse/DRIVERS-409 for context. I think that instructs using "-present" so we never have to update these again, and it also permits using the project inception date as the original date, although we can certainly use the current year as files are introduced.

src/Operation/ListCollectionNames.php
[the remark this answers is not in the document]
jmikola: Good to know they're equivalent, as I wasn't sure if I was missing some Generator subtly here. I just figured this would remove the need for that class, but no objection since CallbackIterator is internal anyway.

src/Command/ListDatabases.php
```diff
@@ -0,0 +1,154 @@
...
+
+        $this->options = $options;
+    }
+
+    /**
+     * Execute the operation.
+     *
+     * @see Executable::execute()
+     * @param Server $server
+     * @return array An array of database names or an array of database info structures, depending on the nameOnly option
```
jmikola, line 99: If I understand correctly, a server supporting `nameOnly` has the effect of only limiting fields in the returned objects to the `name` field (omitting `sizeOnDisk`, etc.). The server response structure is otherwise the name.

I missed this in the first review, but I think it'd be preferable to do the `name` extraction in the ListDatabaseNames method. That way, this `execute` method is always returning the array of `databases` in the command result. Having conditional return types (even though they're both technically arrays) feels like code smell.

src/Command/ListDatabases.php
```diff
@@ -0,0 +1,154 @@
...
+     *  * authorizedDatabases (boolean): Determines which databases are returned
+     *    based on the user privileges.
+     *
+     *    For servers < 4.0.5, this option is ignored.
+     *
+     *  * filter (document): Query by which to filter databases.
+     *
+     *    For servers < 3.6, this option is ignored.
+     *
+     *  * maxTimeMS (integer): The maximum amount of time to allow the query to
```
jmikola, line 59: I think `nameOnly` should be documented here as well.

src/Command/ListDatabases.php
```diff
@@ -0,0 +1,154 @@
...
+        if (! empty($this->options['filter'])) {
+            $cmd['filter'] = (object) $this->options['filter'];
+        }
+
+        if (isset($this->options['maxTimeMS'])) {
+            $cmd['maxTimeMS'] = $this->options['maxTimeMS'];
+        }
+
+        if (isset($this->options['nameOnly'])) {
+            $cmd['nameOnly'] = $this->options['nameOnly'];
```
jmikola, line 120: I was going to ask whether we should check wire versions, but I realized that `filter` and `nameOnly` appear to both be ignored if unsupported. That's likely because pre-3.6 servers weren't strict about raising errors for unsupported options.

Just thinking aloud here. No change needed.

src/Command/ListDatabases.php
```diff
@@ -0,0 +1,154 @@
...
+
+        $cursor = $server->executeReadCommand('admin', new Command($cmd), $this->createOptions());
+        $cursor->setTypeMap(['root' => 'array', 'document' => 'array']);
+        $result = current($cursor->toArray());
+
+        if (! isset($result['databases']) || ! is_array($result['databases'])) {
+            throw new UnexpectedValueException('listDatabases command did not return a "databases" array');
+        }
+
+        // Using array_column here emulates nameOnly behavior for servers that don't support it and ensures a consistent return type
```
jmikola, line 131: This is what I'd suggest moving to ListDatabaseNames, which currently just yields from this result directly. I suppose doing `yield from array_column($result['databases'], 'name')` would be cheaper than using the CallbackIterator, especially since the result is already an array (unlike a potentially large listCollections response, where iterators may conserve memory).

src/Database.php
```diff
@@ -408,6 +410,23 @@ public function getWriteConcern()
         return $this->writeConcern;
     }
 
+    /**
+     * Returns the names of all collections in this database
+     *
+     * @see ListCollectionNames::__construct() for supported options
+     * @param array $options
+     * @return Generator
```
jmikola, line 418: In the interest of a conservative public API, I wonder if we should just document that we return a Traversable or Iterator. This would be similar to how I only documented a generic CollectionInfoIterator return type for the `listCollections` method. Granted, that was the abstract the user from thinking about legacy and cursor-based responses, but it was also suggesting that the only thing they need care about was that the current element is a CollectionInfo object.

Generator has additional methods like `send()` that I expect would have no use, so the fact that we happen to use a generator here seems like an implementation detail best not revealed.

src/Operation/ListDatabaseNames.php
```diff
@@ -0,0 +1,81 @@
...
+     *
+     * @see Executable::execute()
+     * @param Server $server
+     * @return Generator
+     * @throws UnexpectedValueException if the command response was malformed
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function execute(Server $server)
+    {
+        yield from $this->listDatabases->execute($server);
```
jmikola, line 79: This is where I'd suggest:

```php
$result = $this->listDatabases->execute($server);

yield from array_column($result['databases'], 'name');
```

tests/Command/ListCollectionsTest.php
```diff
@@ -1,9 +1,10 @@
 <?php
 
-namespace MongoDB\Tests\Operation;
+namespace MongoDB\Tests\Command;
```
jmikola, line 3: Since these were renamed, does that mean there is no longer a test for `MongoDB\Operation\ListCollections` and `ListDatabases`? I suppose that's not a problem since they really don't do much at all, and the operation functional tests remained in place.

I'll also note that in the case of FindAndModify and its derivative operations, the operation tests only deal with special options where validation logic is in the derivative class. There's nothing left in the ListCollections operation to actually test.

LGTM, just wanted to point this out.

src/Client.php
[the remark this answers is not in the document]
jmikola: Just left a related comment about this on the `Database::listCollectionNames` method. Perhaps it makes sense to only document that we return a Traversable or Iterator.

docs/reference/method/MongoDBClient-listDatabaseNames.txt
```diff
@@ -0,0 +1,74 @@
...
+   .. include:: /includes/apiargs/MongoDBClient-method-listDatabases-param.rst
+
+   The ``$options`` parameter supports the following options:
+
+   .. include:: /includes/apiargs/MongoDBClient-method-listDatabases-option.rst
+
+Return Values
+-------------
+
+A generator containing the name of each database on the server.
```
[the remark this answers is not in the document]
jmikola: Since this hasn't been changed yet, please see my other comment about only documenting a Traversable or Iterator. There may not be any reason to inform users that we're using Generators internally.

src/Command/ListDatabases.php
alcaeus: Absolutely agree, interestingly enough I "forgot" that idea by the time I built `ListCollectionNames`. I refactored it.

src/Command/ListDatabases.php
alcaeus: `array_column` won't work here since the `ListCollections` command returns a CachingIterator over the result cursor, which is why I added the iterator on top of that.

src/Database.php
alcaeus: Generators can only be iterated once and can't be rewound, so I think the distinction is important to make sure people don't expect to repeatedly iterate over the same list of names.

tests/Command/ListCollectionsTest.php
alcaeus: `ListCollectionsTest` only tests parameter parsing, which is why I moved the tests to the command namespace. Functional testing of the operations is done in `ListCollectionsFunctionalTest` in the `Operation` test namespace.

src/Command/ListDatabases.php
jmikola: We're talking about ListDatabases here, though, which always returns an array. Am I misunderstanding?

src/Database.php
jmikola: Ah, good point. That sounds like a possible deviation from the rewindable iterators returned by `listDatabases` and `listCollections`, though. Is it possible to wrap the generator with our CachingIterator. That would save users from having to ever care about this.

Two side notes:

 * https://github.com/mongodb/mongo-php-library/commit/52df7a112e760b595a3485f50c23adf0591b95e8 refactored our CachingIterator to use IteratorIterator instead of Generators, but I believe it forgot to update comments in the file that still refer to `yield`
 * I just found https://github.com/bpolaszek/rewindable-generator, which is a project that addresses this issue with Generators, but it unfortunately uses SPL's CachingIterator, which has its own problems (see [this blog comment](https://blog.calevans.com/2013/12/19/the-php-cachingiterator/#comment-110609), which I found a while ago in [this tweet](https://twitter.com/jmikola/status/923656201610592257)). As such, it actually breaks if you don't fully iterator the Generator before rewinding, which is something our own CachingIterator addresses (an early `rewind()` exhausts the inner iterator).

src/Command/ListDatabases.php
alcaeus: Apologies, I got confused. `ListDatabaseNames` uses `array_column` since we're already dealing with an array.

src/Database.php
alcaeus: Good point. `ListCollectionNames` now returns the `CallbackIterator` directly since we already have a `CachingIterator` below. `ListDatabaseNames` returns an `ArrayIterator` to iterate over the array of names. This can then always be changed to return a `CachingIterator` or other iterator should the `listDatabases` command ever be changed to returning a cursor.

PR #759, opened 2020-06-30
PHPLIB-561: Retry replSetStepDown in connection failover test

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-561

While the test is currently skipped due to issues with subsequent tests (see [PHPLIB-471](https://jira.mongodb.org/browse/PHPLIB-471)), we can still make the change to retry `replSetStepDown` up to 10 times.

### DIFF
--- a/tests/SpecTests/PrimaryStepDownSpecTest.php
+++ b/tests/SpecTests/PrimaryStepDownSpecTest.php
@@ -16,6 +16,7 @@ use MongoDB\Tests\CommandObserver;
 use Symfony\Bridge\PhpUnit\SetUpTearDownTrait;
 use UnexpectedValueException;
 use function current;
+use function sprintf;
 
 /**
  * @see https://github.com/mongodb/specifications/tree/master/source/connections-survive-step-down/tests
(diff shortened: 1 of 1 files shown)


tests/SpecTests/PrimaryStepDownSpecTest.php
```diff
@@ -215,7 +217,20 @@ public function testGetMoreIteration()
...
-        $primary->executeCommand('admin', new Command(['replSetStepDown' => 5, 'force' => true]));
+
+        $success = false;
+        $attempts = 0;
+        do {
+            try {
+                $attempts++;
+                $primary->executeCommand('admin', new Command(['replSetStepDown' => 5, 'force' => true]));
+                $success = true;
+            } catch (Exception $e) {
```
jmikola, line 228: Would it make sense to just use `DriverException` here? That's what's used in `waitForMasterReelection`, and I think this similar because we wouldn't expect a library exception from any of this logic.

alcaeus: Absolutely correct. This is a driver exception, but PhpStorm decided to import the class even though `DriverException` was already imported. I've rectified this.

PR #761, opened 2020-06-30
PHPLIB-558: Don't inherit read preference in Database::command

alcaeus:
https://jira.mongodb.org/browse/PHPLIB-558

~Same as with https://github.com/mongodb/mongo-php-driver/pull/1144, I've targeted the upcoming minor release for this fix, but can back port to the current stable release for a July patch release.~

### DIFF
--- a/tests/Database/DatabaseFunctionalTest.php
+++ b/tests/Database/DatabaseFunctionalTest.php
@@ -7,6 +7,7 @@ use MongoDB\Driver\BulkWrite;
 use MongoDB\Driver\Cursor;
 use MongoDB\Driver\ReadConcern;
 use MongoDB\Driver\ReadPreference;
+use MongoDB\Driver\Server;
 use MongoDB\Driver\WriteConcern;
 use MongoDB\Exception\InvalidArgumentException;
 use MongoDB\Operation\CreateIndexes;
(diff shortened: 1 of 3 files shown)


tests/Database/DatabaseFunctionalTest.php
```diff
@@ -100,6 +101,22 @@ public function testCommand()
         $this->assertTrue($commandResult->ismaster);
     }
 
+    public function testCommandDoesNotInheritReadPreference()
+    {
+        if ($this->getPrimaryServer()->getType() !== Server::TYPE_RS_PRIMARY) {
```
jmikola, line 106: What do you think about creating an `isReplicaSet` method in the base `FunctionalTestCase` class (alongside `isShardedCluster`). Then you can also refactor `CreateIndexesFunctionalTest::testCommitQuorumOption` to use the same method.

tests/Database/DatabaseFunctionalTest.php
```diff
@@ -100,6 +101,22 @@ public function testCommand()
...
+            $this->markTestSkipped('Test only applies to replica sets');
+        }
+
+        $this->database = new Database($this->manager, $this->getDatabaseName(), ['readPreference' => new ReadPreference(ReadPreference::RP_SECONDARY)]);
```
jmikola, line 110: I like that this test should also pass even if the replica set has a single primary member :+1:

tests/Database/DatabaseFunctionalTest.php
alcaeus: Done. The second test is only in 1.7, so I'll replace that occurrence when merging up.

alcaeus:
@jmikola I'll ignore the phpcs warning in this PR as tests are stable. I'll fix coding style in a separate pull request.

PR #765, opened 2020-07-09
PHPLIB-566, PHPLIB-567: Treat CursorNotFound as a resumable change stream error and sync tests

jmikola:
https://jira.mongodb.org/browse/PHPLIB-566
https://jira.mongodb.org/browse/PHPLIB-567

POC of https://github.com/mongodb/specifications/pull/834 for [DRIVERS-1308](https://jira.mongodb.org/browse/DRIVERS-1308). Language ticket is pending.

### DIFF
--- a/src/ChangeStream.php
+++ b/src/ChangeStream.php
@@ -42,6 +42,9 @@ class ChangeStream implements Iterator
      */
     const CURSOR_NOT_FOUND = 43;
 
+    /** @var int */
+    private static $cursorNotFound = 43;
+
     /** @var int[] */
     private static $resumableErrorCodes = [
         6, // HostUnreachable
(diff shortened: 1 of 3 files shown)


src/ChangeStream.php
```diff
@@ -42,6 +42,9 @@ class ChangeStream implements Iterator
      */
     const CURSOR_NOT_FOUND = 43;
 
+    /** @var int */
+    private static $cursorNotFound = 43;
```
alcaeus, line 46: There's `CURSOR_NOT_FOUND` a couple of lines above that could be used. Alternatively, I'm fine deprecating the public constant in favour of a private one (class constant visibility will be available to us once we drop PHP 7.0, so we'd continue using `private static` until then).

jmikola: Indeed. I didn't want to use the `CURSOR_NOT_FOUND` because it's already deprecated (at least in documentation). Even if we dropped PHP 7.0 before PHPLIB 2.0, we couldn't change `CURSOR_NOT_FOUND` to `private` without it being a BC break.

PR #771, opened 2020-07-20
PHPLIB-569: Atlas Data Lake spec tests

jmikola:
https://jira.mongodb.org/browse/PHPLIB-569

Tests synced with mongodb/specifications@bdc283c5279490389ef1b2c36fc5058cae95f755

This is waiting on Evergreen migration ([PHPLIB-469](https://jira.mongodb.org/browse/PHPLIB-469)). In the meantime, I've been testing locally with the following `phpunit.xml` modification:

```
<env name="MONGODB_URI" value="mongodb://mhuser:pencil@127.0.0.1:27017"/>
```

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -7,6 +7,7 @@
          beStrictAboutChangesToGlobalState="true"
          colors="true"
          bootstrap="tests/bootstrap.php"
+         defaultTestSuite="MongoDB CRUD Test Suite"
 >
 
     <php>
@@ -18,6 +19,11 @@
     <testsuites>
         <testsuite name="MongoDB CRUD Test Suite">
             <directory>./tests/</directory>
+            <exclude>tests/SpecTests/AtlasDataLakeSpecTest.php</exclude>
+        </testsuite>
+
+        <testsuite name="Atlas Data Lake Test Suite">
+            <file>tests/SpecTests/AtlasDataLakeSpecTest.php</file>
         </testsuite>
     </testsuites>
 </phpunit>
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -7,7 +7,7 @@
          beStrictAboutChangesToGlobalState="true"
          colors="true"
          bootstrap="tests/bootstrap.php"
-         defaultTestSuite="MongoDB CRUD Test Suite"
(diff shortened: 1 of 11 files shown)


alcaeus:
@jmikola I know you can't review your own PR, but if you could give the evergreen changes a quick look I'd appreciate it.

Evergreen patch: https://spruce.mongodb.com/version/5f55e63a57e85a7177ac7fd7/tasks

phpunit.xml.dist
```diff
@@ -7,6 +7,7 @@
          beStrictAboutChangesToGlobalState="true"
          colors="true"
          bootstrap="tests/bootstrap.php"
+         defaultTestSuite="MongoDB CRUD Test Suite"
```
jmikola, line 10: I don't believe the test suite was used previously. Perhaps it makes sense to rename this, since it's very much beyond CRUD. "Default Test Suite" would be fine.

phpunit.xml.dist
```diff
@@ -18,6 +19,11 @@
     <testsuites>
         <testsuite name="MongoDB CRUD Test Suite">
             <directory>./tests/</directory>
+            <exclude>tests/SpecTests/AtlasDataLakeSpecTest.php</exclude>
```
jmikola, line 22: I suppose this is optional because `AtlasDataLakeSpecTest.php` should already skip if the server doesn't identify as ADL. Up to you if you'd prefer to keep it, but I think removing this would ensure that the skip logic actually gets tested.

jmikola (commented):
Two comments/suggestions but LGTM.

phpunit.xml.dist
alcaeus: That is correct, we could remove this and rely on the skip logic. This has also become unnecessary with the introduction of the `defaultTestSuite` option. Before that, invoking PHPUnit without a `--testsuite` argument would run all test suites, so the ADL tests would get run twice if it wasn't excluded here.

I'm removing the exclude since it's not necessary anymore.

phpunit.xml.dist
alcaeus: Done - I believe this was the original name, but it makes sense to finally change it.

jmikola:
LGTM

alcaeus (approved):

PR #773, opened 2020-07-30
PHPLIB-574: Fix failing SSL test on travis-ci

alcaeus:
PHPLIB-574

This PR updates the expired client certificate which was the reason why the standalone_ssl cluster could not be started. In addition to that, this changes the logic in setup_mo.sh to abort the build if the server could not be started. It will also print the log files in question to aid with debugging.

Note that due to us using an old stable version of mongo-orchestration (which has since not been replaced), we need to overwrite the built-in client certificate with the new certificate to ensure that mo can connect to the server to verify that it has started correctly. This will not be necessary on evergreen, as we will use a different set of cluster configurations there.


.travis/setup_mo.sh
```diff
@@ -6,45 +6,55 @@ if [[ -z $TRAVIS_BUILD_DIR ]]; then
     export TRAVIS_BUILD_DIR=`pwd`;
 fi
 
+# Replace the default client certificate with the new one to make sure mo keeps working
+cp ${TRAVIS_BUILD_DIR}/mongo-orchestration/ssl/client.pem ${MO_PATH}/lib/client.pem
```
jmikola, line 10: > Note that due to us using an old stable version of mongo-orchestration (which has since not been replaced), we need to overwrite the built-in client certificate with the new certificate to ensure that mo can connect to the server to verify that it has started correctly. This will not be necessary on evergreen, as we will use a different set of cluster configurations there.

Do I assume correctly that after Evergreen is added, we'll share the same configs for both Evergreen and Travis? IIRC, that's the case for PHPC.

Separate question: should there be a PHPLIB ticket to bump the mongo-orchestration version we use, or is that also happening in the Evergreen migration?

alcaeus: > Do I assume correctly that after Evergreen is added, we'll share the same configs for both Evergreen and Travis? IIRC, that's the case for PHPC.

Yes. I plan on consolidating these scripts and fetching them from the repo directly when building on travis-ci in the future.

> Separate question: should there be a PHPLIB ticket to bump the mongo-orchestration version we use, or is that also happening in the Evergreen migration?

The last tagged release is 0.6.12, which was tagged in December 2018. I've created PHPLIB-576 to track switching to a non-tagged version that may be more up-to-date.

mongo-orchestration/ssl/ca.pem
```diff
@@ -1,48 +1,21 @@
...
-NLmvqWbGlnqWywNcMihutJKbDCdp/Km5olUPkiNbB3sWsOkViXoiU/V0pK6BZvir
-d67EZpGwydpogyH9kVVCEQKBgGHXc3Q7SmCBRbOyQrQQk0m6i+V8328W1S5m2bPg
-M62aWXMOMn976ZRT1pBDSwz1Y5yJ3NDf7gTZLjEwpgCNrFCJRcc4HLL0NDL8V5js
-VjvpUU5GyYdsJdb+M4ZUPHi/QEaqzqPQumwJSLlJEdfWirZWVj9dDA8XcpGwQjjy
-psHRAoGBAJUTgeJYhjK7k5sgfh+PRqiRJP0msIH8FK7SenBGRUkelWrW6td2Riey
-EcOCMFkRWBeDgnZN5xDyWLBgrzpw9iHQQIUyyBaFknQcRUYKHkCx+k+fr0KHHCUb
-X2Kvf0rbeMucb4y/h7950HkBBq83AYKMAoI8Ql3cx7pKmyOLXRov
------END RSA PRIVATE KEY-----
\ No newline at end of file
+-----BEGIN CERTIFICATE-----
```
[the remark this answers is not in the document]
alcaeus: Updated. I'll make sure to include this in the squashed message.

PR #776, opened 2020-08-05
PHPLIB-425: GridFS methods should throw if stream copying fails

alcaeus:
PHPLIB-425

I'm not too happy with throwing a generic `StreamException`, but there really isn't more information we can reliably add to the error messages (except for mentioning whether it happened on up- or download). Targeting 1.8 for this change, since we're throwing new exceptions in three methods, which I don't want to add in a patch release. Happy to retarget the PR to 1.7 if we're not worried about the new exceptions.

### DIFF
--- a/src/GridFS/Bucket.php
+++ b/src/GridFS/Bucket.php
@@ -28,6 +28,7 @@ use MongoDB\Exception\InvalidArgumentException;
 use MongoDB\Exception\UnsupportedException;
 use MongoDB\GridFS\Exception\CorruptFileException;
 use MongoDB\GridFS\Exception\FileNotFoundException;
+use MongoDB\GridFS\Exception\StreamException;
 use MongoDB\Model\BSONArray;
 use MongoDB\Model\BSONDocument;
 use MongoDB\Operation\Find;
@@ -238,6 +239,7 @@ class Bucket
      * @param resource $destination Writable Stream
      * @throws FileNotFoundException if no file could be selected
      * @throws InvalidArgumentException if $destination is not a stream
+     * @throws StreamException if the file could not be uploaded
      * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
      */
     public function downloadToStream($id, $destination)
@@ -246,7 +248,9 @@ class Bucket
             throw InvalidArgumentException::invalidType('$destination', $destination, 'resource');
         }
 
-        stream_copy_to_stream($this->openDownloadStream($id), $destination);
+        if (stream_copy_to_stream($this->openDownloadStream($id), $destination) === false) {
+            throw StreamException::downloadFailed();
+        }
     }
 
     /**
@@ -273,6 +277,7 @@ class Bucket
      * @param array    $options     Download options
      * @throws FileNotFoundException if no file could be selected
      * @throws InvalidArgumentException if $destination is not a stream
+     * @throws StreamException if the file could not be uploaded
      * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
      */
     public function downloadToStreamByName($filename, $destination, array $options = [])
@@ -281,7 +286,9 @@ class Bucket
             throw InvalidArgumentException::invalidType('$destination', $destination, 'resource');
         }
 
-        stream_copy_to_stream($this->openDownloadStreamByName($filename, $options), $destination);
+        if (stream_copy_to_stream($this->openDownloadStreamByName($filename, $options), $destination) === false) {
+            throw StreamException::downloadFailed();
+        }
     }
 
     /**
@@ -607,6 +614,7 @@ class Bucket
      * @param array    $options  Stream options
      * @return mixed ID of the newly created GridFS file
      * @throws InvalidArgumentException if $source is not a GridFS stream
+     * @throws StreamException if the file could not be uploaded
      * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
      */
     public function uploadFromStream($filename, $source, array $options = [])
@@ -616,7 +624,9 @@ class Bucket
         }
 
         $destination = $this->openUploadStream($filename, $options);
-        stream_copy_to_stream($source, $destination);
+        if (stream_copy_to_stream($source, $destination) === false) {
+            throw StreamException::uploadFailed();
+        }
 
         return $this->getFileIdForStream($destination);
     }
--- a/src/GridFS/Bucket.php
+++ b/src/GridFS/Bucket.php
@@ -249,7 +249,7 @@ class Bucket
         }
 
         if (@stream_copy_to_stream($this->openDownloadStream($id), $destination) === false) {
-            throw StreamException::downloadFailed();
+            throw StreamException::downloadFromIdFailed($id, $destination);
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/GridFS/Bucket.php
```diff
@@ -246,7 +248,9 @@ public function downloadToStream($id, $destination)
             throw InvalidArgumentException::invalidType('$destination', $destination, 'resource');
         }
 
-        stream_copy_to_stream($this->openDownloadStream($id), $destination);
+        if (stream_copy_to_stream($this->openDownloadStream($id), $destination) === false) {
+            throw StreamException::downloadFailed();
```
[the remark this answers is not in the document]
alcaeus: In this case (and the one in `downloadToStreamByName` below, we could print the `uri` associated with the stream retrieved using `stream_get_meta_data()`. Printing the ID can be troublesome, as we don't know the identifier type. While using `(string) $id` would work for scalars and any string able object, this would fail if someone used an array for an ID, as the string cast would then only return "Array" without any further information. This is why I originally skipped adding this information.

However, we can definitely print a filename in `downloadToStreamByName`. I'll make that change and see if I can come up with reliably printing the identifier.

src/GridFS/Bucket.php
```diff
@@ -246,7 +248,9 @@ public function downloadToStream($id, $destination)
             throw InvalidArgumentException::invalidType('$destination', $destination, 'resource');
         }
 
-        stream_copy_to_stream($this->openDownloadStream($id), $destination);
+        if (stream_copy_to_stream($this->openDownloadStream($id), $destination) === false) {
```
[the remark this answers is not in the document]
alcaeus: Using error suppression here is not controversial, I just didn't think of using it in the first place. I'll add it.

src/GridFS/Bucket.php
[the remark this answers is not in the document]
alcaeus: I noted that we use `toJson(fromPHP(...))` in `FileNotFoundException`, so I've duplicated that logic 👍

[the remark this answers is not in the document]
jmikola: Ah, the `uri` is a good idea.

> Printing the ID can be troublesome, as we don't know the identifier type.

IIRC, this came up somewhere else in the driver/library and I've just relied on JSON encoding.

> I noted that we use `toJson(fromPHP(...))` in `FileNotFoundException`

That might have been what I was thinking of.

The URI would also be a good idea, but I'm fine with either and it seems like the ID is more consistent.

src/GridFS/Exception/StreamException.php
```diff
@@ -0,0 +1,40 @@
...
+use function stream_get_meta_data;
+
+class StreamException extends RuntimeException
+{
+    /** @param resource $destination */
+    public static function downloadFromFilenameFailed(string $filename, $destination) : self
+    {
+        $metadata = stream_get_meta_data($destination);
+
+        return new static(sprintf('Downloading GridFS file "%s" to resource "%s" failed: stream error.', $filename, $metadata['uri']));
```
jmikola, line 18: Apologies for digging into this further, but some new thoughts on this.

Does "stream error" add anything useful here, given that all StreamException variants use the same suffix? Perhaps remove it to keep the message shorter. Since we're suppressing the PHP error, it occurs to me that we really don't have any contextual message to report other than "something went wrong" -- and even if we didn't suppress the error we aren't in a position to capture it for inclusion here.

Neither the filename nor ID really uniquely identify the file since we don't know the bucket namespace -- and even a filename isn't unique within a bucket namespace. Would it make sense to also use the `$metadata['uri']` of the GridFS download stream here? By this point, our Bucket method (i.e. `openDownloadStream`, `openDownloadStreamByName`) will already have created a file path (i.e. `createPathForFile`) and that should uniquely identify the GridFS file being downloaded (bucket namespace appended with ID). It conveniently converts ObjectId identifiers to a string (only deferring to extended JSON if necessary because a `__toString()` doesn't exist).

We can still include the user-specified ID or filename as well, as in the following examples:

    Downloading file from "<gridfs-uri>" to "<destination-uri>" failed. GridFS identifier: <id>
    Downloading file from "<gridfs-uri>" to "<destination-uri>" failed. GridFS filename: <filename>

More on `openUploadStream` below.

src/GridFS/Exception/StreamException.php
```diff
@@ -0,0 +1,40 @@
...
+    public static function downloadFromIdFailed($id, $destination) : self
+    {
+        $stringId = toJSON(fromPHP(['_id' => $id]));
+        $metadata = stream_get_meta_data($destination);
+
+        return new static(sprintf('Downloading GridFS file with identifier "%s" to resource "%s" failed: stream error.', $stringId, $metadata['uri']));
+    }
+
+    /** @param resource $source */
+    public static function uploadFailed(string $filename, $source) : self
```
jmikola, line 34: In the case of `openUploadStream`, the path created by `createPathForUpload` only points to the bucket namespace since the filename and ID aren't assembled into a file document until later in the WritableStream constructor; however, by the point you'd throw this exception, that underlying stream would already have been created. If you wanted to provide a path in the same form as a downloaded file exception, we _could_ access the raw file document for the resource with `Bucket::getRawFileDocumentForStream()` and use `createPathForFile` to assemble a URI.

That said, the user knowingly provided a filename string so it's probably best to still include that in the error message, too. Perhaps:

    Uploading file from "<source-uri>" to "<gridfs-uri>" failed. GridFS filename: <filename>

The GridFS URI will always have a "gridfs://" prefix, so it's clear that the destination is a GridFS file (although they technically could both be).

src/GridFS/Exception/StreamException.php
```diff
@@ -0,0 +1,40 @@
...
+        return new static(sprintf('Downloading GridFS file "%s" to resource "%s" failed: stream error.', $filename, $metadata['uri']));
+    }
+
+    /**
+     * @param mixed    $id
+     * @param resource $destination
+     */
+    public static function downloadFromIdFailed($id, $destination) : self
+    {
+        $stringId = toJSON(fromPHP(['_id' => $id]));
```
jmikola, line 27: This logic may lead to more concise (and consistent) output: https://github.com/mongodb/mongo-php-library/blob/b3a4c2412d38f709babdf7d54f51914062eb49bc/src/GridFS/Bucket.php#L632-L636

Although, if you end up using the error messages I suggested above, this might not even be necessary if you decide "GridFS identifier" is redundant since it'd already be reported in the `<gridfs-uri>`.

src/GridFS/Exception/StreamException.php
alcaeus: Good point. Updated and added tests

src/GridFS/Exception/StreamException.php
alcaeus: Updated to use the error message above 👍

src/GridFS/Exception/StreamException.php
alcaeus: Now that you mention it, that logic contains a flaw: passing an array for `_id` will result in the string "Array" being used, since it triggers the first check (it's not an object). I'll investigate the effect of using this logic in `createPathForFile` separately, as using an identifier like `['foo' => 'bar']` is technically fine.

tests/GridFS/UnusableStream.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+namespace MongoDB\Tests\GridFS;
+
+final class UnusableStream
```
jmikola, line 12: I don't recall if this was in the original PR (missed it if so), but great idea!

PR #778, opened 2020-08-28
PHPLIB-562: Add basic matrix to test PHPLIB on evergreen

alcaeus:
PHPLIB-562

This adds a baseline of builds to evergreen. Note that PHP 7.0 is not available on all environments (which may be due to us using e.g. "rhel71-power8" over "rhel71-ppc", but I'll investigate as part of subsequent tickets.

@jmikola while reviewing, keep in mind that most of the new files in .evergreen are from the drivers tools repository. You'll want to focus on config.yml and install-dependencies.sh.


.evergreen/config.yml
```diff
@@ -0,0 +1,655 @@
...
+        remote_file: ${UPLOAD_BUCKET}/${build_variant}/${revision}/${BUILD_NAME}/${build_id}.tar.gz
+        bucket: mciuploads
+        local_file: build.tar.gz
+    - command: shell.exec
+      params:
+        continue_on_err: true
+        # EVG-1105: Use s3.get extract_to: ./
+        script: "set -o xtrace && cd .. && rm -rf ${PROJECT_DIRECTORY} && mkdir ${PROJECT_DIRECTORY}/ && tar xf build.tar.gz -C ${PROJECT_DIRECTORY}/"
+
+  "exec compile script" :
```
jmikola, line 152: Should we move the space between the key and colon? I realize this is probably copied over, but the YAML syntax highlighting revealed this for a few functions.

.evergreen/config.yml
```diff
@@ -0,0 +1,655 @@
...
+        aws_key: ${aws_key}
+        aws_secret: ${aws_secret}
+        local_file:  ${PROJECT_DIRECTORY}/doc/html/index.html
+        remote_file: ${UPLOAD_BUCKET}/docs/${CURRENT_VERSION}/index.html
+        bucket: mciuploads
+        permissions: public-read
+        content_type: text/html
+        display_name: "Rendered docs"
+
+  "upload coverage" :
```
jmikola, line 189: If any functions aren't being used, would it make more sense to remove them? I don't see this referenced anywhere in the file. Likewise for "upload docs" and perhaps others. If you don't feel strongly, I don't object to keeping the dead code around as I expect we'll probably use these functions at some point.

.evergreen/config.yml
```diff
@@ -0,0 +1,655 @@
...
+          vars:
+            TOPOLOGY: "replica_set"
+        - func: "run tests"
+
+    - name: "test-sharded_cluster"
+      tags: ["sharded_cluster"]
+      commands:
+        - func: "bootstrap mongo-orchestration"
+          vars:
+            VERSION: "2.4"
```
jmikola, line 463: Is this server version 2.4?

Looking below, I think only the "test-standalone" task is used at the moment. Either way, I think we should probably get rid of this line because we surely won't want to use server version 2.4 when we add sharded cluster tests.

.evergreen/config.yml
```diff
@@ -0,0 +1,655 @@
...
+            TOPOLOGY: "sharded_cluster"
```
jmikola, line 464: Looking at `sharded_clusters/basic.json` and other MO configs in this PR, none appear to use replica sets. This means that PHPLIB won't actually be able to test change streams and transactions on sharded clusters.

Is there a separate PHPLIB ticket to add configs like we have in PHPC's [scripts/presets/travis/sharded_clusters](https://github.com/mongodb/mongo-php-driver/tree/master/scripts/presets/travis/sharded_clusters) directory?

.evergreen/config.yml
```diff
@@ -0,0 +1,655 @@
...
+        variables:
+           STORAGE_ENGINE: "inmemory"
+
+
+buildvariants:
+
+- matrix_name: "test-baseline"
+  matrix_spec: {"os-php7": ["debian92-test", "ubuntu1804-arm64-test", "rhel70-test"], "php-versions": "*", "versions": "4.4", "driver-versions": "latest-stable" }
+  exclude_spec:
+    - { "os-php7": ["debian92-test", "rhel70-test"], "php-versions": "7.0", versions: "*", "driver-versions": "*" }
```
jmikola, line 652: Is `versions: "*"` overkill since the matrix is only testing 4.4? Or is this just to be defensive in case `matrix_spec` is changed? I assume it may be the latter as you did the same for `driver-version` here.

.evergreen/install-dependencies.sh
```diff
@@ -0,0 +1,85 @@
...
+      echo "Install Windows dependencies"
+      ;;
+
+   darwin*)
+      echo "Install macOS dependencies"
+      ;;
+
+   linux-rhel*)
+      echo "Install RHEL dependencies"
+      ;;
```
jmikola, line 55: Is this only empty because you're ignoring RHEL in the test matrix, or are no additional dependencies needed? I assume this might have been copied from PHPC?

.evergreen/install-dependencies.sh
```diff
@@ -0,0 +1,85 @@
...
+install_extension
+
+php --ri mongodb
+
+# Install composer
+php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
+php composer-setup.php
+php -r "unlink('composer-setup.php');"
+
+php composer.phar update $COMPOSER_FLAGS
```
jmikola, line 85: I don't see `$COMPOSER_FLAGS` defined anywhere. Should the "install dependencies" function provide a default or is that unnecessary?

Where would you expect to define `$COMPOSER_FLAGS` if at all? Perhaps an axis for testing the lowest and highest compatible deps?

.evergreen/config.yml
alcaeus: I cleaned out unused functions. We may need to bring them back in the future, but for the time being there's no point keeping them around. 👍

.evergreen/config.yml
alcaeus: This must have been left over from previous iterations. Removed for the time being.

.evergreen/config.yml
alcaeus: We don't have a dedicated tracking ticket for topologies, but I'll include them in PHPLIB-565 when expanding the version axis. I'll see if we need to add separate configs (similar to PHPC) and whether these configs should be contributed upstream to ensure all drivers can test sharded clusters backed by replica sets.

.evergreen/config.yml
alcaeus: Purely being defensive here as I noticed 7.0 is not available on these platforms, regardless of server version.

.evergreen/install-dependencies.sh
alcaeus: This is the default file as copied over from the tools repository. We can remove this if we don't want dead code from the basic file here.

.evergreen/install-dependencies.sh
alcaeus: COMPOSER_FLAGS will be used once we expand to testing against lowest dependencies as well as dev dependencies.

.evergreen/install-dependencies.sh
jmikola: No objections.

alcaeus:
For the sake of completeness, one more patch build: https://spruce.mongodb.com/version/5f4de9b52fbabe0c4a292eb6/tasks

PR #779, opened 2020-08-31
PHPLIB-345: Fix PHP shutdown error with dangling writable streams

alcaeus:
PHPLIB-345

This applies a suggestion by @sgolemon to avoid PHP trying to use previously destroyed stream objects. The behaviour can be shown by leaving a dangling writable stream, which causes errors on shutdown. The workaround ensures that the stream wrapper instance "forgets" that it owned a stream wrapper, as `StreamWrapper::stream_close` is called after `StreamWrapper::__destruct`.

### DIFF
--- a/src/GridFS/StreamWrapper.php
+++ b/src/GridFS/StreamWrapper.php
@@ -57,6 +57,14 @@ class StreamWrapper
     /** @var ReadableStream|WritableStream|null */
     private $stream;
 
+    public function __destruct()
+    {
+        /* This destructor is a workaround for PHP trying to use the stream well
+         * after all objects have been destructed. This causes autoloading
+         * issues and possibly segmentation faults during PHP shutdown. */
+        $this->stream = null;
+    }
+
     /**
      * Return the stream's file document.
      *
@@ -88,6 +96,10 @@ class StreamWrapper
      */
     public function stream_close()
     {
+        if (! $this->stream) {
+            return;
+        }
+
         $this->stream->close();
     }
 
--- a/src/GridFS/StreamWrapper.php
+++ b/src/GridFS/StreamWrapper.php
@@ -60,7 +60,7 @@ class StreamWrapper
     public function __destruct()
     {
         /* This destructor is a workaround for PHP trying to use the stream well
-         * after all objects have been destructed. This causes autoloading
+         * after all objects have been destructed. This can cause autoloading
          * issues and possibly segmentation faults during PHP shutdown. */
         $this->stream = null;
     }
--- a/tests/GridFS/BucketFunctionalTest.php
+++ b/tests/GridFS/BucketFunctionalTest.php
@@ -720,10 +720,12 @@ class BucketFunctionalTest extends FunctionalTestCase
         }
 
         $path = __DIR__ . '/../../vendor/autoload.php';
-        @exec(
-            <<<CMD
+        $command = <<<CMD
 php -r "require '$path'; \\\$stream = (new MongoDB\Client)->test->selectGridFSBucket()->openUploadStream('filename', ['disableMD5' => true]);" 2>&1
-CMD,
+CMD;
+
+        @exec(
+            $command,
             $output,
             $return
         );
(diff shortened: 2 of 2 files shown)


tests/GridFS/BucketFunctionalTest.php
```diff
@@ -708,6 +711,21 @@ public function testExistingIndexIsReused()
         $this->assertIndexNotExists($this->chunksCollection->getCollectionName(), 'files_id_1_n_1');
     }
 
+    public function testDanglingOpenWritableStream()
+    {
+        $path = __DIR__ . '/../../vendor/autoload.php';
+        @exec(
+            <<<CMD
+php -r "require '$path'; class_exists(MongoDB\Operation\FindOne::class); \\\$stream = (new MongoDB\Client)->test->selectGridFSBucket()->openUploadStream('filename', ['disableMD5' => true]);" 2>&1
+CMD,
```
jmikola, line 720: Why is `class_exists(MongoDB\Operation\FindOne::class);` necessary?

IIRC, one of the relevant bits of [PHPLIB-345](https://jira.mongodb.org/browse/PHPLIB-345) was that if the class wasn't _already_ loaded, then we'd see an error when our GridFS code attempted to load it during shutdown. I would think that triggering autoloading ahead of time might defeat the point of a regression test.

tests/GridFS/BucketFunctionalTest.php
```diff
@@ -708,6 +711,21 @@ public function testExistingIndexIsReused()
...
+            $output
+        );
+
+        $output = implode(PHP_EOL, $output);
+
+        $this->assertSame('', $output);
```
jmikola, line 726: I assume we don't care about verifying the exit code (not that `exec` can do that anyway), as a segfault would _always_ product output?

Separate question: would we need to skip this test on Windows platforms?

src/GridFS/StreamWrapper.php
```diff
@@ -88,6 +96,10 @@ public static function register($protocol = 'gridfs')
      */
     public function stream_close()
     {
+        if (! $this->stream) {
```
jmikola, line 99: Is `stream_close` the only handler that might be called after StreamWrapper destructs? Just want to make sure we don't need `if (! $this->stream)` checks elsewhere.

src/GridFS/StreamWrapper.php
```diff
@@ -57,6 +57,14 @@ class StreamWrapper
     /** @var ReadableStream|WritableStream|null */
     private $stream;
 
+    public function __destruct()
+    {
+        /* This destructor is a workaround for PHP trying to use the stream well
+         * after all objects have been destructed. This causes autoloading
+         * issues and possibly segmentation faults during PHP shutdown. */
```
jmikola, line 64: Nit: "This can causes", since apparently this isn't an issue for PHP 8 (if I understand Sara's response in the JIRA issue correctly).

tests/GridFS/BucketFunctionalTest.php
alcaeus: This is left over from a previous attempt at fixing this. I thought I had removed it, but apparently not.

tests/GridFS/BucketFunctionalTest.php
alcaeus: Yes, we always assume there will be some output, but I'll check if we can also assert against the exit code to be on the safe side.

Regarding Windows, skipping that would make sense.

src/GridFS/StreamWrapper.php
alcaeus: PHP only calls `stream_close` after calling `__destruct`. Userland code shouldn't manage to make such a call in the first place.

tests/GridFS/BucketFunctionalTest.php
alcaeus: Return code is now checked as well, and we skip the test on windows.

tests/GridFS/BucketFunctionalTest.php
alcaeus: Removed.

tests/GridFS/BucketFunctionalTest.php
jmikola: Oh, I missed that `exec()` _does_ capture the exit code as its third argument. Nice.

PR #780, opened 2020-09-02
PHPLIB-565: Add MongoDB version axis to build matrix

alcaeus:
PHPLIB-565

This PR expands evergreen test coverage by testing against different driver versions, MongoDB server versions, and topologies. Summary:
* The first matrix tests all PHP versions on all operating systems (dropping a few where we don't have binaries)
* The second matrix tests all driver versions on all PHP versions. Uses Ubuntu 18.04 (since it offers all PHP versions) and only latest stable MongoDB server version
* The third matrix tests all MongoDB server versions (3.0-latest dev) on PHP 7.4 and RHEL 7.0 (since that's the only distro that offers all server versions)

In all cases, we test against standalone, replica set, and sharded cluster topologies. The latter uses a replica set by default, but we override the provided file to make sure the replica set has secondaries, which is required for some tests. All other files are removed as they are not used.

Patch build: https://spruce.mongodb.com/version/5f4fa2e73e8e863a703f0b4e/tasks. Note that replica set builds fail intermittently due to segmentation faults (double free when closing cursors).


.evergreen/config.yml
```diff
@@ -174,7 +174,7 @@ functions:
       params:
         script: |
           ${PREPARE_SHELL}
-          MONGODB_VERSION=${VERSION} TOPOLOGY=${TOPOLOGY} AUTH=${AUTH} SSL=${SSL} STORAGE_ENGINE=${STORAGE_ENGINE} sh ${DRIVERS_TOOLS}/.evergreen/run-orchestration.sh
+          MONGODB_VERSION=${VERSION} ORCHESTRATION_FILE=${ORCHESTRATION_FILE} TOPOLOGY=${TOPOLOGY} AUTH=${AUTH} SSL=${SSL} STORAGE_ENGINE=${STORAGE_ENGINE} sh ${PROJECT_DIRECTORY}/.evergreen/run-orchestration.sh
```
alcaeus, line 177: Using our own file allows us to add logic to override the used file. Passing `ORCHESTRATION_FILE` as well allows configurations to select a specific file instead of relying on the filename being assembled by the logic included.

.evergreen/config.yml
```diff
@@ -417,9 +417,7 @@ axes:
     display_name: Driver Version
     values:
       - id: "latest-stable"
-        display_name: "1.8.0beta2"
-        variables:
-          DRIVER_VERSION: "1.8.0beta2"
+        display_name: "1.8-stable"
```
alcaeus, line 420: This was left over from earlier tests where we didn't have 1.8 stable yet. We can now omit the `DRIVER_VERSION` constant to always install the current stable release.

tests/Operation/AggregateFunctionalTest.php
```diff
@@ -221,9 +221,7 @@ function () use ($pipeline, $options) {
                 $this->assertCount(1, $results);
                 $result = current($results);
 
-                if ($this->isShardedCluster()) {
-                    $this->assertObjectHasAttribute('shards', $result);
-
+                if (isset($result->shards)) {
```
alcaeus, line 224: This was necessary as the output of `explain` on a cluster with a single shard does not contain the `shards` key but rather looks like the result on a standalone.

tests/SpecTests/FunctionalTestCase.php
```diff
@@ -198,6 +198,10 @@ protected function dropTestAndOutcomeCollections()
     {
         $context = $this->getContext();
 
+        if ($context->databaseName === 'admin') {
```
alcaeus, line 201: This fixes failures when running the crud/db-aggregate tests on the latest MongoDB version.

.evergreen/config.yml
jmikola: Was this borrowed from PHPC or is it new for PHPLIB (in which case it may be something you'd like to add back to PHPC)?

.evergreen/run-tests.sh
```diff
@@ -22,9 +22,4 @@ OLD_PATH=$PATH
 PATH=/opt/php/${PHP_VERSION}-64bit/bin:$OLD_PATH
 
 # Run the tests, and store the results in a Evergreen compatible JSON results file
```
jmikola, line 24: I assume "store the results in a Evergreen compatible JSON results file" doesn't happen here, nor in PHPC, which uses `run-tests.php`. Should we revise this comment?

.evergreen/run-tests.sh
```diff
@@ -22,9 +22,4 @@ OLD_PATH=$PATH
...
-case "$OS" in
-   *)
-      php vendor/bin/phpunit -v
-      ;;
-esac
-
+php vendor/bin/phpunit
```
jmikola, line 25: I assume removing verbose output was intentional. So long as we still get detail for skipped tests and failures, that works for me.

tests/SpecTests/FunctionalTestCase.php
jmikola: Is latest 4.4.x, or 4.5-dev?

.evergreen/orchestration/configs/sharded_clusters/basic.json
```diff
@@ -1,30 +1,48 @@
...
-      "id": "sh01",
-      "shardParams": {
-        "members": [{
-          "procParams": {
+    "id": "shard_cluster_1",
+    "shards": [
+        {
+            "id": "sh01",
+            "shardParams": {
+                "members": [
```
jmikola, line 7: Your OP says:

> In all cases, we test against standalone, replica set, and sharded cluster topologies. The latter uses a replica set by default, but we override the provided file to make sure the replica set has secondaries, which is required for some tests.

It's been a while since I looked an MO config for sharding, but I assume config servers aren't specified at all? In that case, I assume `members` here comprises the replica set for the first shard, and two of these three nodes will end up as secondaries. Correct?

Would it make sense to rename this file to `sharded-rs.json` instead of `basic.json`?

.evergreen/run-tests.sh
alcaeus: We get details for failures, but not for skipped tests. I can add this back if you'd like.

.evergreen/run-tests.sh
alcaeus: I was going to look at the result file format and see if this makes test results easier to analyse. Going through the build log to see what's going on isn't really sustainable.

.evergreen/orchestration/configs/sharded_clusters/basic.json
alcaeus: I can rename it to `sharded-rs.json` and use a different `ORCHESTRATION_FILE` config in evergreen. The default `basic.yml` already specifies a replica set as shard, albeit with a single server which causes one test to fail as it attempts to select a secondary.

tests/SpecTests/FunctionalTestCase.php
alcaeus: Latest is 4.5-dev

.evergreen/run-tests.sh
jmikola: Realistically, I don't think we're ever going to look at build logs and investigate whether tests are skipped when they shouldn't be, so I'm fine leaving this as-is.

AFAIK PHPLIB does not suffer from the same issue that PHPC did to necessitate [PHPC-1343](https://jira.mongodb.org/browse/PHPC-1343) (a canary test that fails when no servers are available).

.evergreen/orchestration/configs/sharded_clusters/basic.json
jmikola: Oh, I missed that the original was a single-node replica set. I'm fine leaving this as-is, then.

I don't imagine we'll actually need a test environment for a shard with a single-node RS or standalone.

.evergreen/config.yml
jmikola: I didn't find `ORCHESTRATION_FILE` in https://github.com/mongodb/mongo-php-driver/blob/master/.evergreen/config.yml, so I assume this is new. Feel free to create a ticket if you think it's worth porting this over to PHPC down the line.

jmikola:
Looking at the five errors in the patch build, four are related to:

```
1) MongoDB\Tests\SpecTests\ClientSideEncryptionSpecTest::testExternalKeyVault with data set #0 (false)
MongoDB\Driver\Exception\BulkWriteException: Bulk write failed due to previous MongoDB\Driver\Exception\EncryptionException: not all keys requested were satisfied
/data/mci/be31210cdc31fd73c4e4fedf4e369a8f/src/src/Operation/InsertOne.php:134
/data/mci/be31210cdc31fd73c4e4fedf4e369a8f/src/src/Collection.php:931
/data/mci/be31210cdc31fd73c4e4fedf4e369a8f/src/tests/SpecTests/ClientSideEncryptionSpecTest.php:346
Caused by
MongoDB\Driver\Exception\EncryptionException: not all keys requested were satisfied
/data/mci/be31210cdc31fd73c4e4fedf4e369a8f/src/src/Operation/InsertOne.php:134
/data/mci/be31210cdc31fd73c4e4fedf4e369a8f/src/src/Collection.php:931
/data/mci/be31210cdc31fd73c4e4fedf4e369a8f/src/tests/SpecTests/ClientSideEncryptionSpecTest.php:346
```

And one is the "double free or corruption" error (oddly no FLE error for that environment).

alcaeus:
Yep, I'm not sure what the cause for the encryption errors is. @kevinAlbs are you aware of anything that could cause these "not all keys requested were satisfied" errors on replica sets? Note that these don't always show up.

.evergreen/config.yml
alcaeus: Created PHPC-1676 to track this.

alcaeus:
Thanks @kevinAlbs, that was the reason for the failures. Builds now pass, save for one double-free: https://spruce.mongodb.com/version/5f50e8d432f41749d8020f87/tasks 🎉

jmikola:
@alcaeus did you re-kick the build such that the intermittent double-free error disappeared? If so, I'm OK to merge this as-is and revisit the double-free error in [PHPC-1541](https://jira.mongodb.org/browse/PHPC-1541).

alcaeus:
@jmikola I restarted the build this morning to see if the failure is intermittent or more permanent. Merging as-is sounds good as we still haven't been able to reproduce this in real-life conditions.

jmikola (approved):

PR #782, opened 2020-09-08
Remove duplicate skipReason check

jmikola:
This was introduced in 1245eb94e97d4958a6ed124b4a8b5453f2d60be3

### DIFF
--- a/tests/SpecTests/TransactionsSpecTest.php
+++ b/tests/SpecTests/TransactionsSpecTest.php
@@ -130,10 +130,6 @@ class TransactionsSpecTest extends FunctionalTestCase
             $this->markTestIncomplete(self::$incompleteTests[$this->dataDescription()]);
         }
 
-        if (isset($test->skipReason)) {
-            $this->markTestSkipped($test->skipReason);
-        }
-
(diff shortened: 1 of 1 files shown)


jmikola:
@alcaeus: It looks like `phpcs` was enabled on v1.5 in https://github.com/mongodb/mongo-php-library/commit/339492ff76b7f3c60d78624264211ae1c50793dc. Was the branch not updated at the time to satisfy the code checks? I wouldn't have expected this to fail as it did.

alcaeus:
> @alcaeus: It looks like `phpcs` was enabled on v1.5 in [339492f](https://github.com/mongodb/mongo-php-library/commit/339492ff76b7f3c60d78624264211ae1c50793dc). Was the branch not updated at the time to satisfy the code checks? I wouldn't have expected this to fail as it did.

This could be due to recent changes in phpcs or sniffs: the tool has a history of fixing bugs or introducing sniffs that create new errors on a run, even when upgrading patch or minor releases. In this case, this is a regression that we've ignored in a later branch, so feel free to ignore (and maybe run phpcs after merging up to the current stable branch).

alcaeus (approved):

PR #787, opened 2020-09-18
PHPLIB-582: Create JSON test result file for evergreen

alcaeus:
PHPLIB-582

This ticket unfortunately is incomplete due to a few issues with the build system.

Reporting test results using `attach.results` does not allow us to easily add logs (e.g. information about assertion failures, traces, test output, etc.) to the test results. This is possible using `attach.xunit_results`, but I could not get that command to read the results files written by PHPUnit. For one, PHPUnit reports nested test suites (a feature of JUnit), which the parser in evergreen does not understand. This should be fixed by [EVG-12951](https://jira.mongodb.org/browse/EVG-12951).

While I had a temporary listener that worked around this by flattening the file (and to report a single test suite containing all tests), this still didn't yield any results. This went as far as Evergreen happily accepting a non-existent or invalid xUnit result file. So, for the time being, we report test results using the JSON format. This will at least cause them to show up in a "Tests" tab under individual tasks, allowing us to see which test failed without consulting the log. Once the problems mentioned above have been solved, we can revisit this and add more detail to the test reporting.

Note that due to how PHPUnit approaches its PHP version support, we cannot support both PHPUnit 6 and PHPUnit 8 anymore. While we could add return type hints in any class implementing the listener interface, we can't do so until we drop PHP 7.0 as the listener interface requires a `void` return type which was only introduced in 7.1. The easiest way to fix this is to drop PHPUnit 8 and rely on 6 only, which will work up to PHP 7.4. We can then revisit this once we start removing our custom listeners, drop support for PHP 7.0, or add support for PHP 8.

### DIFF
--- a/tests/EvergreenLogListener.php
+++ b/tests/EvergreenLogListener.php
@@ -1,0 +1,205 @@
+<?php
+
+namespace MongoDB\Tests;
+
+use Exception;
+use PHPUnit\Framework\AssertionFailedError;
+use PHPUnit\Framework\SelfDescribing;
+use PHPUnit\Framework\Test;
+use PHPUnit\Framework\TestFailure;
+use PHPUnit\Framework\TestListener;
+use PHPUnit\Framework\TestSuite;
+use PHPUnit\Framework\Warning;
+use PHPUnit\Util\Filter;
+use PHPUnit\Util\Printer;
+use function get_class;
+use function json_encode;
+use function round;
+
+// phpcs:disable SlevomatCodingStandard.Exceptions.ReferenceThrowableOnly.ReferencedGeneralException
+class EvergreenLogListener extends Printer implements TestListener
+{
+    /** @var array[] */
+    private $tests = [];
+
+    /** @var array|null */
+    private $currentTestCase = null;
+
+    /**
+     * Flush buffer and close output.
+     */
+    public function flush()
+    {
+        $this->write($this->getJson());
+
+        parent::flush();
+    }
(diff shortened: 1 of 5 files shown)


tests/EvergreenLogListener.php
```diff
@@ -0,0 +1,205 @@
...
+use PHPUnit\Framework\TestSuite;
+use PHPUnit\Framework\Warning;
+use PHPUnit\Util\Filter;
+use PHPUnit\Util\Printer;
+use function get_class;
+use function json_encode;
+use function round;
+
+// phpcs:disable SlevomatCodingStandard.Exceptions.ReferenceThrowableOnly.ReferencedGeneralException
+class EvergreenLogListener extends Printer implements TestListener
```
jmikola, line 20: Since we already use [symfony/phpunit-bridge](https://github.com/symfony/phpunit-bridge), I noticed it has an alias for a SymfonyTestsListener interface. I think you could leverage that instead of implementing PHPUnit's interface directly and preserve our ability to test with PHPUnit 8.x.

alcaeus: Unfortunately, that class is not a generic implementation for test listeners, but rather a specific one to support the use-case of the PHPUnit bridge.

jmikola: Ah, I just poked my head into SymfonyTestsListenerTrait.

jmikola (approved):

PR #788, opened 2020-09-18
PHPLIB-563: Finalise PHP runtime axis

alcaeus:
PHPLIB-563

This PR makes a few last changes to the pipeline:
- Exclude rhel74-zseries and rhel71-power8 because they don't pass or don't even start. Since this also affects PHPC, I'd take care of this in the PHPC follow-up epic.
- Add tests against lowest supported versions. We test the base driver release and oldest composer dependencies against all server versions.
- PHP 7.0 now also works on Debian 9.2 an RHEL 7.0. I was able to work around the pecl errors that prevented running on those platforms

Pull requests currently test the "test-php-versions" pipeline, which tests the driver on all PHP versions and several operating systems against the latest server. I'm open to adding additional variants/tasks to the mix, but thought that those were a good starting point.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -538,6 +538,11 @@ buildvariants:
 # Only tests against latest MongoDB and ext-mongodb versions
 - matrix_name: "test-php-versions"
   matrix_spec: {"os-php7": "*", "php-versions": "*", "edge-versions": "latest-stable", "driver-versions": "latest-stable" }
+  exclude_spec:
+    # rhel71-power8 fails due to not reaching pecl
+    - { "os-php7": "rhel71-power8", "php-versions": "*", edge-versions: "*", "driver-versions": "*" }
+    # rhel74-zseries doesn't start in a timely fashion - most likely missing executors
+    - { "os-php7": "rhel74-zseries", "php-versions": "*", edge-versions: "*", "driver-versions": "*" }
   display_name: "${os-php7}, PHP ${php-versions}, MongoDB ${edge-versions}, ext-mongodb ${driver-versions}"
   tasks:
     - name: "test-standalone"
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -261,7 +261,7 @@ functions:
           ${PREPARE_SHELL}
           file="${PROJECT_DIRECTORY}/.evergreen/install-dependencies.sh"
           # Don't use ${file} syntax here because evergreen treats it as an empty expansion.
-          [ -f "$file" ] && PHP_VERSION=${PHP_VERSION} DRIVER_VERSION=${DRIVER_VERSION} DRIVER_REPO=${DRIVER_REPO} DRIVER_BRANCH=${DRIVER_BRANCH} sh $file || echo "$file not available, skipping"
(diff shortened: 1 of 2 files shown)


.evergreen/config.yml
```diff
@@ -528,16 +532,26 @@ axes:
...
-  matrix_spec: {"os-php7": ["debian92-test", "ubuntu1804-arm64-test", "rhel70-test"], "php-versions": "*", "edge-versions": "latest-stable", "driver-versions": "latest-stable" }
+  matrix_spec: {"os-php7": "*", "php-versions": "*", "edge-versions": "latest-stable", "driver-versions": "latest-stable" }
   exclude_spec:
-    - { "os-php7": ["debian92-test", "rhel70-test"], "php-versions": "7.0", edge-versions: "*", "driver-versions": "*" }
-  display_name: "PHP ${php-versions}, ${os-php7}, MongoDB ${edge-versions}, ext-mongodb ${driver-versions}"
+    # rhel71-power8 fails due to not reaching pecl
+    - { "os-php7": "rhel71-power8", "php-versions": "*", edge-versions: "*", "driver-versions": "*" }
+    # rhel74-zseries doesn't start in a timely fashion - most likely missing executors
+    - { "os-php7": "rhel74-zseries", "php-versions": "*", edge-versions: "*", "driver-versions": "*" }
+  display_name: "* ${os-php7}, PHP ${php-versions}, MongoDB ${edge-versions}, ext-mongodb ${driver-versions}"
```
jmikola, line 554: Why does the display name start with an asterisk?

Should we use a consistent order with other display names? I noticed the "test-dependencies" matrix below has some of the components here in a different order.

.evergreen/install-dependencies.sh
```diff
@@ -4,6 +4,10 @@ set -o errexit  # Exit the script with error if any of the commands fail
 
 install_extension ()
 {
+   # Workaround to get PECL running on PHP 7.0
```
jmikola, line 7: Would it make sense to guard this in a conditional for `PHP_VERSION` starting with "7.0."?

https://stackoverflow.com/a/2172367/162228 suggests you could use `if [[ $PHP_VERSION == 7.0.* ]]`

.evergreen/config.yml
alcaeus: The asterisk is there to make sure these variants are sorted to the top of the list, as they are the ones we require for a PR to pass.

As for ordering in display names, I moved the main variant (e.g. PHP version in this case, MongoDB server version in other cases) to the front of the list. I have no objections against changing this if you prefer to have them alphabetically sorted in general.

jmikola: Oh, I see now that you're prioritizing the variable axes over the fixed ones, and in the "test-dependencies" matrix only the MongoDB version varies. That all makes sense to me.

jmikola (approved):

.evergreen/install-dependencies.sh
alcaeus: Despite me trying to use `bash` as a shell, I couldn't get evergreen to play nice:
```
[2020/09/22 08:21:34.150] + [[ 7.0.32 == 7.0.* ]]
 [2020/09/22 08:21:34.150] /data/mci/b4ada7bfe7e6167ba91c9754dd42a065/src/.evergreen/install-dependencies.sh: 8: /data/mci/b4ada7bfe7e6167ba91c9754dd42a065/src/.evergreen/install-dependencies.sh: [[: not found
```

Since this has no negative impact on other PHP versions, I'd use this unconditionally, especially since 7.0 is the next PHP version we'll drop support for (although no timeline for this exists).

PR #791, opened 2020-10-21
PHPLIB-597: Replace direct PHPUnit dependency with symfony/phpunit-bridge

alcaeus:
PHPLIB-597

This is a prerequisite to get tests going on PHP 8. PHPUnit makes it impossible to run tests on PHP 7 - PHP 8, and the PHPUnit bridge fixes that problem for us.

This PR introduces a new trait for the constraint polyfill, as PHPUnit 9 changes the method signature for `Constraint::evaluate` which we override. These changes will be contributed back to Symfony (hopefully for inclusion in 5.2).


composer.json
```diff
@@ -29,11 +27,17 @@
...
+            "tests/PHPUnit/FunctionsPolyfill.php"
+        ]
     },
     "extra": {
         "branch-alias": {
             "dev-master": "1.8.x-dev"
         }
+    },
+    "scripts": {
+        "pre-autoload-dump": "if test -f vendor/bin/simple-phpunit; then vendor/bin/simple-phpunit --version > /dev/null; fi"
```
jmikola, line 41: What's the purpose of dumping the version of `simple-phpunit` to `/dev/null`? Is this something you're instructed to do as a consumer of `simple-phpunit`? I'd also be curious if the embedded script poses a problem for Windows.

tests/SpecTests/ClientSideEncryptionSpecTest.php
```diff
@@ -395,65 +482,7 @@ public function testBSONSizeLimitsAndBatchSplitting()
...
-        }
-
-        // Test 5
-        $collection->insertOne(['_id' => 'under_16mib', 'unencrypted' => str_repeat('a', 16777216 - 2000)]);
-
-        // Test 6
-        $this->expectException(BulkWriteException::class);
-        $this->expectExceptionMessageRegExp('#object to insert too large#');
-        $collection->insertOne(['_id' => 'encryption_exceeds_16mib', 'unencrypted' => str_repeat('a', 16777216 - 2000)] + $document);
+        $test($this, $collection, $document);
```
jmikola, line 485: I assume there's no problem for the closures that omit a third argument?

jmikola (approved):
LGTM. Glad we're able to replace `assertInternalType` with the newer methods.

composer.json
alcaeus: I'll add a comment, but in order to get our autoload-dev directive to work, we need to ensure phpunit is installed, which can be accomplished by dumping the version.

This will most likely be changed once again since PHPUnit 9 moves the `assert*` functions to a different namespace and autoloads them automatically, so I'll have to create a separate polyfill for that. For now I'm just happy to finally get this working properly.

tests/SpecTests/ClientSideEncryptionSpecTest.php
alcaeus: This might become a problem in PHP 8, in which case my next PR will fix that part. PHP 8 will require some more work, but I wanted to extract this to keep the diffs concise.

composer.json
jmikola: Noted. If this is temporary, perhaps a JSON comment would be prudent (like I did for `autoload-dev`). Up to you, though.

tests/SpecTests/ClientSideEncryptionSpecTest.php
jmikola: ArgumentCountError? Yes, I am familiar with that :rofl:

composer.json
alcaeus: Murphy's law suggests that if I don't add a comment, the code would stay in for more than a few days. I'll add one ;)

stof: The simple-phpunit script supports `vendor/bin/simple-phpunit install` to install the corresponding PHPUnit version

PR #786, opened 2020-09-15
Allow PHP 8 to be installed

vasilvestre:
Allowing PHP 8 on master branch could allow people on dev channel to report issues on it.

What do you think about this ?

Actual problem accoding to alcaeus is that mongo-php-driver isn't ready yet. https://github.com/mongodb/mongo-php-driver/issues/1161

PR #793, opened 2020-11-04
PHP 8 support

alcaeus:
This covers PHPLIB-595 and PHPLIB-588.

Due to PHP 8 not being available on Evergreen yet, we test standalone, replica set, and sharded topologies on PHP 8 as well, in addition to testing them on PHP 7.4. While a new version of the extension is required on PHP 8, we will not bump the minimum ext-mongodb version as existing versions should continue to work as expected.

Notable changes:
* PHPUnit 9 moves the global assertions functions to a PHPUnit namespace. To provide compatibility with earlier PHPUnit versions (required to continue supporting PHP 7.0 and newer) we include a polyfill here which falls back to the PHPUnit functions if they exist.
* travis-ci tested on PHP 7.3 by default. This was changed to PHP 7.4, including simplifying the configuration to make future changes easier (we expect to always test on the newest stable version)
* The BC layer for constraints has been removed, as the changes were incorporated into the PHPUnit Bridge provided by Symfony. We'll update the dependency to a stable version once Symfony 5.2 has been released. Since it's a development dependency, using a branch version for it has no stability impact on users.


vasilvestre:
Fix https://github.com/mongodb/mongo-php-library/pull/786

PR #796, opened 2020-11-10
Update build pipeline

alcaeus:
With travis-ci experiencing long queues, it makes sense to reduce the size of the build pipeline and exclude all variants that are being tested on evergreen. We now run a set of smokescreen tests on PHP 7.4, `phpcs` checks on PHP 7.1 (lowest version supported by doctrine/coding-standard) and PHP 8 builds for each topologies (as PHP 8 is not supported on Evergreen yet).

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -440,12 +440,18 @@ axes:
           DRIVER_VERSION: "1.8.1"
       - id: "latest-stable"
         display_name: "1.8-stable"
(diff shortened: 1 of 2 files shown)


alcaeus:
While travis-ci.org is still waiting for build executors, I was able to confirm that the commit passes on travis-ci.com: https://travis-ci.com/github/alcaeus/mongo-php-library/builds/199822180.

PR #801, opened 2021-01-15
PHPLIB-598: Clean up cursor usages to account for iterator changes

alcaeus:
PHPLIB-598

@jmikola I decided to clean up the tailable cursor docs as well and add a note about needing to wrap cursors in an `IteratorIterator` before manually iterating. Let me know if you'd rather keep the original documentation intact and add a note that wrapping is no longer necessary with 1.9.0+.

### DIFF
--- a/docs/tutorial/tailable-cursor.txt
+++ b/docs/tutorial/tailable-cursor.txt
@@ -35,25 +35,21 @@ While normal cursors can be iterated once with ``foreach``, that approach will
 not work with tailable cursors. When ``foreach`` is used with a tailable cursor,
 the loop will stop upon reaching the end of the initial result set. Attempting
 to continue iteration on the cursor with a second ``foreach`` would throw an
-exception, since PHP attempts to rewind the cursor.
-
-In order to continuously read from a tailable cursor, we will need to directly
-control the cursor's iteration (e.g. call ``next()``), avoid inadvertently
-rewinding the cursor, and decide when to wait for new results or stop iteration
-entirely.
+exception, since PHP attempts to rewind the cursor. Therefore, reading from a
+tailable cursor will require direct usage of the :php:`Iterator <iterator>` API.
 
 .. note::
 
     Before version 1.9.0 of the ``ext-mongodb`` extension, the cursor class does
     not implement the :php:`Iterator <iterator>` interface. To manually iterate
-    cursors using the method described below, cursors must be wrapped in a
+    a cursor using the method below, it must first be wrapped with an
     :php:`IteratorIterator <iteratoriterator>`.
 
 Manually Iterating a Normal Cursor
 ----------------------------------
 
 Before looking at how a tailable cursor can be iterated, we'll start by
-examining how the class interacts with a normal cursor.
+examining how the ``Iterator`` methods interact with a normal cursor.
 
 The following example finds five restaurants and uses ``foreach`` to view the
 results:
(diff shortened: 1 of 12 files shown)


docs/tutorial/tailable-cursor.txt
```diff
@@ -37,18 +37,23 @@ the loop will stop upon reaching the end of the initial result set. Attempting
...
+entirely.
 
-Wrapping a Normal Cursor
-------------------------
+.. note::
+
+    Before version 1.9.0 of the ``ext-mongodb`` extension, the cursor class does
+    not implement the :php:`Iterator <iterator>` interface. To manually iterate
+    cursors using the method described below, cursors must be wrapped in a
+    :php:`IteratorIterator <iteratoriterator>`.
```
jmikola, line 50: Suggest rephrasing second sentence:

> To manually iterate a cursor using the method below, it must first be wrapped with an `IteratorIterator`.

docs/tutorial/tailable-cursor.txt
```diff
@@ -37,18 +37,23 @@ the loop will stop upon reaching the end of the initial result set. Attempting
...
+
+Manually Iterating a Normal Cursor
+----------------------------------
 
-Before looking at how a tailable cursor can be wrapped with
-:php:`IteratorIterator <iteratoriterator>`, we'll start by examining how the
-class interacts with a normal cursor.
+Before looking at how a tailable cursor can be iterated, we'll start by
+examining how the class interacts with a normal cursor.
```
jmikola, line 56: > we'll start by examining how the `Iterator` methods interact with a normal cursor.

src/Model/CachingIterator.php
```diff
@@ -53,14 +53,14 @@ class CachingIterator implements Countable, Iterator
...
-     * to omit an explicit call * to rewind() before using the other methods.
+     * Additionally, this mimics behavior of the SPL iterators and allows users
+     * to omit an explicit call to rewind() before using the other methods.
      *
      * @param Traversable $traversable
      */
     public function __construct(Traversable $traversable)
     {
-        $this->iterator = new IteratorIterator($traversable);
+        $this->iterator = $traversable instanceof Iterator ? $traversable : new IteratorIterator($traversable);
```
jmikola, line 63: Correct me if I'm wrong, but this could always have been done independent of this ticket (or PHPC 1.9.0), correct?

Seems like just a smarter change to avoid an extra IteratorIterator composition when we don't need it.

tests/UnifiedSpecTests/CollectionData.php
```diff
@@ -3,7 +3,6 @@
 namespace MongoDB\Tests\UnifiedSpecTests;
 
 use ArrayIterator;
-use IteratorIterator;
```
jmikola, line 6: Does this PR basically remove all references to IteratorIterator in the library's test suite -- or at least any place we used it to wrap a Cursor?

src/Model/CachingIterator.php
alcaeus: That's correct, we could've always done this. I noticed this and made the change as most iterators passed now won't need to be wrapped as they already implement the `Iterator` interface.

tests/UnifiedSpecTests/CollectionData.php
alcaeus: Yes, we did this in a number of tests to be able to work with the cursor directly (e.g. by using a `MultipleIterator`, which requires an `Iterator`).

Other places where we wrap a cursor are the various `*InfoIterators` which remain as they add functionality, and `ChangeStreamIterator`, which also attaches special handling and thus can't not wrap the original cursor.

alcaeus:
@jmikola I made the changes you suggested to the docs. Note that the three failing builds are due to a recent change to the Symfony PHPUnit Bridge which prevents it from working on PHP 7.0. I'll take care of this in a separate PR.

jmikola (approved):

PR #802, opened 2021-01-18
PHPLIB-606: Move CI checks from travis-ci to GitHub Actions

alcaeus:
PHPLIB-606

Creating this PR against the current stable branch so we can take advantage of these checks already (evergreen currently only tests the master branch).

The change in DocumentationExamplesTest.php were necessary as drivers-evergreen-tools starts a sharded cluster backed by a single-member replica set. I'll take a look at either changing the orchestration file to start a three-member replica set or alternatively add support for defining a custom orchestration file.


alcaeus:
Note: GitHub does not seem to pick up on the files until they've been merged. Build results are available by checking the last commit in my repository: https://github.com/alcaeus/mongo-php-library/commit/3f9c717540bcd14d47e9b227a7bc13aeb380357f.

.github/workflows/coding-standards.yml
```diff
@@ -0,0 +1,68 @@
...
+        php-version:
+          - "7.4"
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v2"
+
+      - name: Setup cache environment
+        id: extcache
+        uses: shivammathur/cache-extensions@v1
```
jmikola, line 29: Are there any concerns with depending on third-party repositories for action recipes? I imagine if this was taken down our CI builds would break (among many others). Might there be security concerns we were using encrypted environment variables (e.g. Atlas testing) and an action dependency was compromised?

AFAIK, secrets are only tested with Evergreen so this may be a hypothetical question.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,97 @@
...
+            topology: "sharded_cluster"
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v2"
+        with:
+          fetch-depth: 2
+
+      - id: setup-mongodb
+        uses: mongodb-labs/drivers-evergreen-tools@master
```
jmikola, line 51: I assume versioning is a separate can of worms, since those tags would also apply to the entire repository and not just actions? OK with using `master` here, but just curious.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,97 @@
...
+          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
+
+      - name: "Install dependencies with composer"
+        run: "composer update --no-interaction --no-progress"
+
+      - name: "Run PHPUnit"
+        run: "vendor/bin/simple-phpunit -v"
+        env:
+          SYMFONY_DEPRECATIONS_HELPER: 999999
```
jmikola, line 96: Would it make more sense to just use "disabled" here? See: https://symfony.com/blog/new-in-symfony-3-1-deprecation-helper-improvements#ignoring-triggered-deprecations

tests/DocumentationExamplesTest.php
```diff
@@ -1486,6 +1480,12 @@ public function testCausalConsistency()
             [ 'sku' => '111', 'name' => 'Peanuts', 'start' => new UTCDateTime() ]
         );
 
+        try {
```
jmikola, line 1483: Can you add a comment explaining why this was moved down from the top of the method (like we do for other skip conditions)? I think the long-term solution would be to create a new utility method that ensures our environment is a sharded cluster backed by multi-member replica sets, but we can punt on that if this is the only use case.

.github/workflows/coding-standards.yml
alcaeus: > Are there any concerns with depending on third-party repositories for action recipes? I imagine if this was taken down our CI builds would break (among many others).

Yes, our CI builds would break in that case, but I'd expect a fork to spring up relatively soon considering how widely used these actions are.

> Might there be security concerns we were using encrypted environment variables (e.g. Atlas testing) and an action dependency was compromised?

Good question, I haven't considered yet. Quoting from the [secrets documentation](https://docs.github.com/en/actions/reference/encrypted-secrets):
> To make a secret available to an action, you must set the secret as an input or environment variable in the workflow file. [...] You can use and read encrypted secrets in a workflow file if you have access to edit the file.

Reading this, any secrets defined in our org or repo would have to be passed on to third-party actions as input or environment variables (either using `with` or `env` keys). I'd assume that a compromised third-party action file would not gain access to secrets defined in our repository, but I concede I haven't tested this (yet).

That said, I haven't configured any secrets yet, so we're not running any AWS tests or anything. We can decide to do this at a later date, or we can leave testing with secrets to our evergreen pipeline.

.github/workflows/tests.yml
alcaeus: Yes, this uses git tags, which apply to the whole repository. Since we don't version the repo at all I've decided to use master until we decide that we want to start tagging releases in that repository.

.github/workflows/tests.yml
alcaeus: `disabled` disables logging of deprecations, while this setting logs deprecations but doesn't set a non-zero exit code for `phpunit` until 1000000 deprecations are reached in a run.

tests/DocumentationExamplesTest.php
alcaeus: Added:
> In sharded clusters, server selection ignores the read preference mode, so using $manager->selectServer does not work here. To work around this, we run a query on a secondary and rely on an exception to let us know that no secondary is available.

alcaeus:
One more note: I haven't explored how we can test against upcoming driver versions (which requires compiling the extension ourselves rather than installing it via the setup-php action). This will most likely become relevant at some point in the future when we start working on features that require bumping to a development driver version.

.github/workflows/coding-standards.yml
jmikola: > That said, I haven't configured any secrets yet, so we're not running any AWS tests or anything. We can decide to do this at a later date, or we can leave testing with secrets to our evergreen pipeline.

I'd say nothing to do here, but something to keep in mind if you have grander plans for moving drivers onto GitHub Actions down the line. Maybe you want to add a line about this in your scope doc, but I reckon it's potentially future work since you were only focused on a subset of tests for quick PR feedback.

.github/workflows/tests.yml
jmikola: Got it. So this gives you a chance to view the deprecations in build logs without actually failing the build. Works for me if you'll find that useful.

jmikola:
> One more note: I haven't explored how we can test against upcoming driver versions (which requires compiling the extension ourselves rather than installing it via the setup-php action).

Would something like `.travis/install-extension.sh` not port over nicely to GitHub Actions? I think it's reasonable to make that follow-up work, although it shouldn't be urgent since such upcoming features would likely be developed internally and our PRs would be able to target Evergreen. Perhaps it'd be problematic if that meant GH Action builds would perpetually fail, though?

alcaeus:
> Would something like `.travis/install-extension.sh` not port over nicely to GitHub Actions?

Potentially - although I don't want to use this all the time as the cache layer speeds up builds when the selected PHP version has already been installed.

> I think it's reasonable to make that follow-up work [...] Perhaps it'd be problematic if that meant GH Action builds would perpetually fail, though?

That's correct: it's follow-up work, but once we get PHPC-1684 done this will be required to keep the build stable. Either way, nothing to worry about for the time being.

.github/workflows/coding-standards.yml
alcaeus: Will do to make sure other engineers are aware, although I want this to be a subset for quick PR feedback in all drivers for the time being.

jmikola (approved):

PR #803, opened 2021-01-21
PHPLIB-608 Fix wrong validity check in CachingIterator

alcaeus:
PHPLIB-608

Note that this does not affect 1.8, as `CachingIterator` always wraps the given traversable in an `IteratorIterator`, which corrects this wrong behaviour in underlying iterators. In the upcoming 1.9 release, this is no longer the case for any traversable that implements the `Iterator` interface, so this wrong behaviour can lead to extra elements given (hence the test failures).

### DIFF
--- a/src/Model/CachingIterator.php
+++ b/src/Model/CachingIterator.php
@@ -157,8 +157,6 @@ class CachingIterator implements Countable, Iterator
             return;
         }
 
-        $key = $this->iterator->key();
-
-        $this->items[$key] = $this->iterator->current();
+        $this->items[$this->iterator->key()] = $this->iterator->current();
     }
 }
--- a/src/Model/CachingIterator.php
+++ b/src/Model/CachingIterator.php
@@ -153,11 +153,11 @@ class CachingIterator implements Countable, Iterator
      */
     private function storeCurrentItem()
     {
-        $key = $this->iterator->key();
-
-        if ($key === null) {
+        if (! $this->iterator->valid()) {
             return;
         }
+
+        $key = $this->iterator->key();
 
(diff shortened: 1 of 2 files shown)


tests/Model/CachingIteratorTest.php
```diff
@@ -110,6 +111,46 @@ public function testCountWithEmptySet()
         $this->assertCount(0, $iterator);
     }
 
+    /**
+     * This protects against iterators that return valid keys on invalid
+     * positions, which was the case in ext-mongodb until PHPC-1748 was fixed.
+     */
+    public function testWithWrongIterator()
+    {
+        $nestedIterator = new class implements Iterator {
```
jmikola, line 120: What a lovely use of anonymous classes!

src/Model/CachingIterator.php
```diff
@@ -153,12 +153,12 @@ private function exhaustIterator()
...
     private function storeCurrentItem()
     {
-        $key = $this->iterator->key();
-
-        if ($key === null) {
+        if (! $this->iterator->valid()) {
             return;
         }
 
+        $key = $this->iterator->key();
```
jmikola, line 160: With this change `$key` is only used once, so perhaps we can just do:

```php
$this->items[$this->iterator->key()] = $this->iterator->current();
```

jmikola (approved):
Suggested refactoring but LGTM.

Side note: reviewing this made me realize that CachingIterator would break if the wrapped Traversable had non-unique keys. I opened [PHPLIB-609](https://jira.mongodb.org/browse/PHPLIB-609) to track that.

PR #810, opened 2021-02-10
PHPLIB-617 Continuous Matrix Testing for 4.0 era drivers

alcaeus:
PHPLIB-617

Since this version is not tested on evergreen yet, this includes the minimal scripts required to test on evergreen without adding an evergreen configuration.

### DIFF
--- a/.evergreen/install-dependencies.sh
+++ b/.evergreen/install-dependencies.sh
@@ -1,11 +1,36 @@
 #!/bin/sh
 set -o errexit  # Exit the script with error if any of the commands fail
 
+set_php_version ()
+{
+   PHP_VERSION=$1
+
+   if [ ! -d "/opt/php" ]; then
+      echo "PHP is not available"
+      exit 1
+   fi
+
+   if [ -d "/opt/php/${PHP_VERSION}-64bit/bin" ]; then
+      export PHP_PATH="/opt/php/${PHP_VERSION}-64bit/bin"
+   else
+      # Try to find the newest version matching our constant
+      export PHP_PATH=`find /opt/php/ -maxdepth 1 -type d -name "${PHP_VERSION}.*-64bit" -print | sort -V -r | head -1`
+   fi
+
+   if [ ! -d "$PHP_PATH" ]; then
+      echo "Could not find PHP binaries for version ${PHP_VERSION}. Listing available versions..."
+      ls -1 /opt/php
+      exit 1
+   fi
+
+   export PATH=$PHP_PATH/bin:$PATH
+}
+
 install_extension ()
 {
    # Workaround to get PECL running on PHP 7.0
-   export PHP_PEAR_PHP_BIN=${PHP_PATH}/bin/php
-   export PHP_PEAR_INSTALL_DIR=${PHP_PATH}/lib/php
+#   export PHP_PEAR_PHP_BIN=${PHP_PATH}/bin/php
+#   export PHP_PEAR_INSTALL_DIR=${PHP_PATH}/bin/php
 
    rm -f ${PHP_PATH}/lib/php.ini
 
@@ -100,10 +125,7 @@ case "$DEPENDENCIES" in
       ;;
 esac
 
-PHP_PATH=/opt/php/${PHP_VERSION}-64bit
-OLD_PATH=$PATH
-PATH=$PHP_PATH/bin:$OLD_PATH
-
+set_php_version $PHP_VERSION
 install_extension
 install_composer
 
--- a/.evergreen/install-dependencies.sh
+++ b/.evergreen/install-dependencies.sh
@@ -1,5 +1,4 @@
 #!/bin/sh
-set -o xtrace   # Write all commands first to stderr
 set -o errexit  # Exit the script with error if any of the commands fail
 
 install_extension ()
--- a/.evergreen/install-dependencies.sh
+++ b/.evergreen/install-dependencies.sh
@@ -1,0 +1,99 @@
+#!/bin/sh
+set -o xtrace   # Write all commands first to stderr
+set -o errexit  # Exit the script with error if any of the commands fail
+
+install_extension ()
+{
+   # Workaround to get PECL running on PHP 7.0
+   export PHP_PEAR_PHP_BIN=${PHP_PATH}/bin/php
+   export PHP_PEAR_INSTALL_DIR=${PHP_PATH}/lib/php
+
+   rm -f ${PHP_PATH}/lib/php.ini
+
+   if [ "x${DRIVER_BRANCH}" != "x" ] || [ "x${DRIVER_REPO}" != "x" ]; then
+      CLONE_REPO=${DRIVER_REPO:-https://github.com/mongodb/mongo-php-driver}
+      CHECKOUT_BRANCH=${DRIVER_BRANCH:-master}
+
+      echo "Compiling driver branch ${CHECKOUT_BRANCH} from repository ${CLONE_REPO}"
+
+      mkdir -p /tmp/compile
+      rm -rf /tmp/compile/mongo-php-driver
+      git clone ${CLONE_REPO} /tmp/compile/mongo-php-driver
+      cd /tmp/compile/mongo-php-driver
+
+      git checkout ${CHECKOUT_BRANCH}
+      git submodule update --init
+      phpize
+      ./configure --enable-mongodb-developer-flags
+      make all -j20 > /dev/null
+      make install
+
+      cd ${PROJECT_DIRECTORY}
+   elif [ "x${DRIVER_VERSION}" != "x" ]; then
+      echo "Installing driver version ${DRIVER_VERSION} from PECL"
+      pecl install -f mongodb-${DRIVER_VERSION}
+   else
+      echo "Installing latest driver version from PECL"
+      pecl install -f mongodb
+   fi
+
+   sudo cp ${PROJECT_DIRECTORY}/.evergreen/config/php.ini ${PHP_PATH}/lib/php.ini
+}
+
+# Functions to fetch MongoDB binaries
+. ${DRIVERS_TOOLS}/.evergreen/download-mongodb.sh
+OS=$(uname -s | tr '[:upper:]' '[:lower:]')
+
+get_distro
+
+case "$DISTRO" in
+   cygwin*)
+      echo "Install Windows dependencies"
+      ;;
+
+   darwin*)
+      echo "Install macOS dependencies"
+      ;;
+
+   linux-rhel*)
+      echo "Install RHEL dependencies"
+      ;;
+
+   linux-ubuntu*)
+      echo "Install Ubuntu dependencies"
+      sudo apt-get install -y awscli || true
+      ;;
+
+   sunos*)
+      echo "Install Solaris dependencies"
+      sudo /opt/csw/bin/pkgutil -y -i sasl_dev || true
+      ;;
+
+   *)
+      echo "All other platforms..."
+      ;;
+esac
+
+case "$DEPENDENCIES" in
+   lowest*)
+      COMPOSER_FLAGS="${COMPOSER_FLAGS} --prefer-lowest"
+      ;;
+
+   *)
+      ;;
+esac
+
+PHP_PATH=/opt/php/${PHP_VERSION}-64bit
+OLD_PATH=$PATH
+PATH=$PHP_PATH/bin:$OLD_PATH
+
+install_extension
+
+php --ri mongodb
+
+# Install composer
+php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
+php composer-setup.php
+php -r "unlink('composer-setup.php');"
+
+php composer.phar update $COMPOSER_FLAGS
--- a/.evergreen/install-dependencies.sh
+++ b/.evergreen/install-dependencies.sh
@@ -10,9 +10,9 @@ install_extension ()
 
    rm -f ${PHP_PATH}/lib/php.ini
 
-   if [ "x${DRIVER_BRANCH}" != "x" ] || [ "x${DRIVER_REPO}" != "x" ]; then
-      CLONE_REPO=${DRIVER_REPO:-https://github.com/mongodb/mongo-php-driver}
-      CHECKOUT_BRANCH=${DRIVER_BRANCH:-master}
+   if [ "x${EXTENSION_BRANCH}" != "x" ] || [ "x${EXTENSION_REPO}" != "x" ]; then
+      CLONE_REPO=${EXTENSION_REPO:-https://github.com/mongodb/mongo-php-driver}
+      CHECKOUT_BRANCH=${EXTENSION_BRANCH:-master}
 
       echo "Compiling driver branch ${CHECKOUT_BRANCH} from repository ${CLONE_REPO}"
 
@@ -29,9 +29,9 @@ install_extension ()
       make install
 
       cd ${PROJECT_DIRECTORY}
-   elif [ "x${DRIVER_VERSION}" != "x" ]; then
-      echo "Installing driver version ${DRIVER_VERSION} from PECL"
-      pecl install -f mongodb-${DRIVER_VERSION}
+   elif [ "x${EXTENSION_VERSION}" != "x" ]; then
+      echo "Installing driver version ${EXTENSION_VERSION} from PECL"
+      pecl install -f mongodb-${EXTENSION_VERSION}
    else
       echo "Installing latest driver version from PECL"
       pecl install -f mongodb
--- a/.evergreen/install-dependencies.sh
+++ b/.evergreen/install-dependencies.sh
@@ -38,6 +38,26 @@ install_extension ()
    fi
 
    sudo cp ${PROJECT_DIRECTORY}/.evergreen/config/php.ini ${PHP_PATH}/lib/php.ini
+
+   php --ri mongodb
+}
+
+install_composer ()
+{
+  EXPECTED_CHECKSUM="$(php -r 'copy("https://composer.github.io/installer.sig", "php://stdout");')"
+  php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
+  ACTUAL_CHECKSUM="$(php -r "echo hash_file('sha384', 'composer-setup.php');")"
+
+  if [ "$EXPECTED_CHECKSUM" != "$ACTUAL_CHECKSUM" ]; then
+    >&2 echo 'ERROR: Invalid installer checksum'
+    rm composer-setup.php
+    exit 1
+  fi
+
+  php composer-setup.php --quiet
+  RESULT=$?
+  rm composer-setup.php
+  exit $RESULT
 }
 
 # Functions to fetch MongoDB binaries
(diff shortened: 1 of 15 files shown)


.evergreen/run-tests.sh
```diff
@@ -0,0 +1,41 @@
...
+MONGODB_URI=${MONGODB_URI:-}
+TESTS=${TESTS:-}
+PHP_VERSION=${PHP_VERSION:-7.2.21}
+
+# The PLATFORM environment variable is used by the matrix testing project.
+# For the time being, we can use that to install the correct extension version
+if [ "x${PLATFORM}" != "x" ]; then
+  DIR=$(dirname $0)
+  DRIVER_VERSION=1.5.5 . $DIR/install-dependencies.sh
+fi
```
alcaeus, line 23: This is necessary since the drivers-matrix-testing project does not allow to call additional scripts. I'll propose changes to call install-dependencies as that's a standard script from drivers-evergreen-tools that all drivers should be using.

Allowing projects to define their own environment variables also removes the need for hardcoding an extension version here.

.evergreen/run-tests.sh
```diff
@@ -0,0 +1,41 @@
...
+#       SSL                     Set to enable SSL. Defaults to "nossl"
+#       MONGODB_URI             Set the suggested connection MONGODB_URI (including credentials and topology info)
+#       MARCH                   Machine Architecture. Defaults to lowercase uname -m
+
+
+AUTH=${AUTH:-noauth}
+SSL=${SSL:-nossl}
+MONGODB_URI=${MONGODB_URI:-}
+TESTS=${TESTS:-}
+PHP_VERSION=${PHP_VERSION:-7.2.21}
```
alcaeus, line 16: The drivers-matrix-testing project does not allow to set additional environment variables. I'll work on adding support for that next, until then we need this fallback.

alcaeus (commented):
@jmikola a couple of notes on the scripts since they introduce changes from the ones we use in master.

.evergreen/install-dependencies.sh
```diff
@@ -0,0 +1,99 @@
...
+      cd ${PROJECT_DIRECTORY}
+   elif [ "x${DRIVER_VERSION}" != "x" ]; then
+      echo "Installing driver version ${DRIVER_VERSION} from PECL"
+      pecl install -f mongodb-${DRIVER_VERSION}
+   else
+      echo "Installing latest driver version from PECL"
+      pecl install -f mongodb
+   fi
+
+   sudo cp ${PROJECT_DIRECTORY}/.evergreen/config/php.ini ${PHP_PATH}/lib/php.ini
```
jmikola, line 40: Would it make more sense to _append_ our `php.ini` to the existing one, or is there nothing of value in the original `php.ini`?

I looked at https://github.com/10gen/mongo-php-toolchain but I don't see any reference to `php.ini` at all. I dunno if that's something we might want to change down the line, as I happen to use the following for my local PHP builds:

```
date.timezone = "America/New_York"
display_errors = On
display_startup_errors = On
error_reporting = -1
log_errors = On
memory_limit = -1
report_memleaks = On
report_zend_debug = 0
short_open_tag = Off
```

Obviously `date.timezone` isn't relevant, but I think the other settings could be helpful to ensure PHP doesn't suppress memory leaks and such by default.

.evergreen/run-tests.sh
```diff
@@ -0,0 +1,41 @@
...
+
+# The PLATFORM environment variable is used by the matrix testing project.
+# For the time being, we can use that to install the correct extension version
+if [ "x${PLATFORM}" != "x" ]; then
+  DIR=$(dirname $0)
+  DRIVER_VERSION=1.5.5 . $DIR/install-dependencies.sh
```
jmikola, line 22: Ah, we lucked out with this being 1.5+, as we only standardized on the `configure` options in 1.5.0: https://github.com/mongodb/mongo-php-driver/commit/da72946e654f7a5afda189795a05a03726c1877f

.evergreen/install-dependencies.sh
```diff
@@ -0,0 +1,99 @@
...
+PHP_PATH=/opt/php/${PHP_VERSION}-64bit
+OLD_PATH=$PATH
+PATH=$PHP_PATH/bin:$OLD_PATH
+
+install_extension
+
+php --ri mongodb
+
+# Install composer
+php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
```
jmikola, line 95: I assume htis was lifted from https://getcomposer.org/download/. Perhaps it makes more sense to follow the instructions in https://getcomposer.org/doc/faqs/how-to-install-composer-programmatically.md so we can check the signature? Looks like it'd be easy to package into another shell script.

.evergreen/run-tests.sh
```diff
@@ -0,0 +1,41 @@
...
+
+
+AUTH=${AUTH:-noauth}
+SSL=${SSL:-nossl}
+MONGODB_URI=${MONGODB_URI:-}
+TESTS=${TESTS:-}
+PHP_VERSION=${PHP_VERSION:-7.2.21}
+
+# The PLATFORM environment variable is used by the matrix testing project.
+# For the time being, we can use that to install the correct extension version
```
jmikola, line 19: I assume `PLATFORM` is actually used for some other purpose. Is there a ticket to introduce another environment variable (e.g. `MATRIX_TESTING`) specifically so scripts can determine if they're running for matrix testing?

.evergreen/run-tests.sh
```diff
@@ -0,0 +1,41 @@
...
+if [ "x${PLATFORM}" != "x" ]; then
+  DIR=$(dirname $0)
+  DRIVER_VERSION=1.5.5 . $DIR/install-dependencies.sh
+fi
+
+OS=$(uname -s | tr '[:upper:]' '[:lower:]')
+[ -z "$MARCH" ] && MARCH=$(uname -m | tr '[:upper:]' '[:lower:]')
+
+echo "Running $AUTH tests over $SSL, connecting to $MONGODB_URI"
```
jmikola, line 28: Should we rephrase this? I imagine this might print: "Running noauth tests over nossl", which seems odd.

What are `AUTH` and `SSL` when they are set (and not defaulting to `noauth` and `nossl`, respectively)?

Perhaps something like:

* `Running tests with AUTH=$AUTH, SSL=$SSL, connecting to $MONGO_URI`

Or if `$AUTH` (and `$SSL` would only ever be) "auth" or "noauth", then perhaps:

* `Running tests with $AUTH and $SSL, connecting to: $MONGODB_URI`

.evergreen/install-dependencies.sh
```diff
@@ -0,0 +1,99 @@
...
+      make all -j20 > /dev/null
+      make install
+
+      cd ${PROJECT_DIRECTORY}
+   elif [ "x${DRIVER_VERSION}" != "x" ]; then
+      echo "Installing driver version ${DRIVER_VERSION} from PECL"
+      pecl install -f mongodb-${DRIVER_VERSION}
+   else
+      echo "Installing latest driver version from PECL"
+      pecl install -f mongodb
```
jmikola, line 37: I know PECL tries to append to `php.ini` after it runs. It looks like that only if `php_ini` is specified in the PEAR/PECL configuration.

If you want to add `pecl config-set php_ini ${PHP_PATH}/lib/php.ini` before invoking `pecl install`, perhaps that would let you do without the custom `php.ini` file in this PR. Although if `${PHP_PATH}/lib/php.ini` doesn't already exist (per my next comment), you'd need to `touch` first.

Maybe that's not worth the trouble since you'd also have to deal with file permissions and/or run `pecl` with `sudo`. Feel free to ignore.

.evergreen/run-tests.sh
```diff
@@ -0,0 +1,41 @@
...
+OLD_PATH=$PATH
+PATH=/opt/php/${PHP_VERSION}-64bit/bin:$OLD_PATH
+
+# Disable failing PHPUnit due to deprecations
+export SYMFONY_DEPRECATIONS_HELPER=999999
+
+# Run the tests, and store the results in a Evergreen compatible JSON results file
+case "$TESTS" in
+   *)
+      php vendor/bin/phpunit --configuration phpunit.evergreen.xml
```
jmikola, line 39: I forget when you introduced `simple-phpunit`, but if any of these branches uses that would we want to add an invocation of `vendor/bin/simple-phpunit --version` (nothing that actually runs tests) to `install_dependencies.sh`? IIRC, it installs PHPUnit lazily on its first run. I don't think we'd want that to happen during `run_tests.sh`, since it technically is a dependency.

phpunit.evergreen.xml
```diff
@@ -0,0 +1,32 @@
...
+        convertErrorsToExceptions="true"
+        convertNoticesToExceptions="true"
+        convertWarningsToExceptions="true"
+        stopOnFailure="false"
+        syntaxCheck="false"
+        bootstrap="tests/bootstrap.php"
+>
+
+    <php>
+        <ini name="error_reporting" value="-1"/>
```
jmikola, line 18: Oh, this may relate to my earlier comment about customizing `php.ini`. Are there other options we might want to include here to ensure things like memory leaks get reported? Alternatively, would it make more sense to just move this into the PR's `php.ini` file that you apply to the entire PHP runtime?

phpunit.evergreen.xml
```diff
@@ -0,0 +1,32 @@
...
+        <env name="MONGODB_URI" value="mongodb://127.0.0.1:27017/?serverSelectionTimeoutMS=100"/>
```
jmikola, line 19: Should this be omitted so we pick up the environment variable from `run_tests.sh`, or is this only a fallback?

.evergreen/install-dependencies.sh
alcaeus: I haven't focused on improving the process just yet, especially since testing in these branches is difficult as they are not tested on evergreen anyways. We could do the following:
* Move a sensible default configuration to mongo-php-toolchain to have it in place by default, and
* set `php_ini` in the pecl configuration so that `pecl install` enables the extension.

With this, we can remove this workaround, first in master to confirm that everything is working correctly, followed by a backport to these branches. WDYT?

.evergreen/run-tests.sh
alcaeus: There's a fine line between careful planning and pure dumb luck 😅

.evergreen/install-dependencies.sh
alcaeus: Ah, very good point. I'll add the script as suggested to the docs here, but this is something we can consider moving to the toolchain (only the script, not the actual composer install)

.evergreen/run-tests.sh
alcaeus: Yes, this is an ugly monkey patch. `PLATFORM` contains the platform being tested on and is only set by the drivers-matrix-testing pipeline. A more elaborate solution removes the need to call this by being able to define environment variables for driver entries and invoking install-dependencies.sh automatically.

.evergreen/run-tests.sh
alcaeus: I've noticed the wording before but it has never bugged me to the point where I wanted to change it. Due to evergreen logging everything, this log line is preceded by the actual env variables a few lines above: https://evergreen.mongodb.com/task_log_raw/mongo_php_library_test_php_versions__php_versions~7.0_edge_versions~latest_stable_driver_versions~latest_stable_os_php7~debian92_test_test_replica_set_2d6fce1d04aac91fdc9d5245d5be738555921842_21_02_09_09_52_54/0?type=T#L2266, so we might as well remove this. I'll defer to your preference on this, I typically skip over this line anyways.

.evergreen/install-dependencies.sh
alcaeus: Also covered by PHPC-1772, and I'd test this in master and then back port.

.evergreen/run-tests.sh
alcaeus: The v1.7-matrix-testing branch switches to simple-phpunit, the previous branches still use phpunit directly. I'll move installing simple-phpunit to the install-dependencies.sh script so this is not part of running tests 👍

phpunit.evergreen.xml
alcaeus: This was copied over from phpunit.xml.dist, but I believe we can remove this once we've dealt with php.ini in our toolchain.

phpunit.evergreen.xml
alcaeus: This is a default so that we don't pass an empty uri if not specified externally. From the [PHPUnit manual on `php.env`](https://phpunit.readthedocs.io/en/9.5/configuration.html#the-env-element):
> By default, environment variables are not overwritten if they exist already. To force overwriting existing variables, use the `force` attribute.

.evergreen/run-tests.sh
alcaeus: After some more playing around, I was able to remove the check on PLATFORM and the hardcoded PHP_VERSION. With recent changes in drivers-matrix-testing, custom variables are passed through, so I can leverage the new `PHP_VERSION` and `PHP_IS_MATRIX_TESTING` variables. Since they can only be set per driver (instead of per driverEra, e.g. "4.4"), I can't use the same method for defining the extension version to install, so that's still hardcoded in run-tests.sh.

.evergreen/run-tests.sh
```diff
@@ -0,0 +1,52 @@
...
+#       SSL                     Set to enable SSL. Defaults to "nossl"
+#       MONGODB_URI             Set the suggested connection MONGODB_URI (including credentials and topology info)
+#       MARCH                   Machine Architecture. Defaults to lowercase uname -m
+
+
+AUTH=${AUTH:-noauth}
+SSL=${SSL:-nossl}
+MONGODB_URI=${MONGODB_URI:-}
+TESTS=${TESTS:-}
+IS_MATRIX_TESTING=${IS_MATRIX_TESTING:-false}
```
alcaeus, line 16: This variable will be introduced with https://github.com/mongodb/drivers-matrix-testing/pull/7

.evergreen/run-tests.sh
```diff
@@ -13,27 +13,21 @@ AUTH=${AUTH:-noauth}
...
-      esac
-   }
-
-   set_extension_version "${DRIVER_MONGODB_VERSION}"
-
-   echo $DRIVER_VERSION
+if [ "$IS_MATRIX_TESTING" == "true" ]; then
+   case "${DRIVER_MONGODB_VERSION}" in
+      '4.4')
+          export DRIVER_VERSION='1.8.2'
```
jmikola, line 22: Won't this clobber the definitions in https://github.com/mongodb/drivers-matrix-testing/pull/6?

alcaeus: Ah, good point. We probably shouldn't use `export`, but this should be fine without. Alternatively, I can change this from `DRIVER_VERSION` to `EXTENSION_VERSION` (which would be more appropriate anyways). I'll just have to remember to update the config file as well once this gets merged up into a branch with evergreen config present.

.evergreen/install-dependencies.sh
jmikola: Fine to punt on this. I left a comment on the issue referring back to this thread for context.

.evergreen/run-tests.sh
jmikola: Should this PR wait on https://github.com/mongodb/drivers-matrix-testing/pull/7 to rename the environment variable to `IS_MATRIX_TESTING`?

.evergreen/run-tests.sh
jmikola: If the actual environment variables are dumped above, I think we can definitely remove this.

.evergreen/run-tests.sh
jmikola: I assume this is just about relaying information to `install-dependencies.sh`, which is called a few lines later. In that case, `EXTENSION_VERSION` seems fine. Using distinct environment variables feels cleaner and could avoid a WTF down the line.

As for merging up to another branch, perhaps you want to lump this todo item in with the simple-phpunit item in https://github.com/mongodb/mongo-php-library/pull/810/files#r574301076?

.evergreen/run-tests.sh
jmikola: I'll leave this comment thread open in case you want to create a separate tracking ticket for this when merging up to the v1.7 branch. Feel free to resolve at your leisure.

.evergreen/run-tests.sh
alcaeus: This has been merged and renamed.

.evergreen/run-tests.sh
alcaeus: Yep, this will produce conflicts anyways.

.evergreen/run-tests.sh
alcaeus: Since the output was due to the `xtrace` option being enabled (which I want to remove across the board for cleaner logs), I've updated the wording to your last suggestion.

alcaeus:
New patch build: https://spruce.mongodb.com/version/602e562f5623434b3f2cc74c/tasks

Testing against 4.2 and 4.4 passes thanks to some well-placed skips, but I have to investigate the failures on latest as they occur in spec tests. I'd call this done for the time being and revisit the failures once I can see how the 4.2 and 4.4 drivers fare.

alcaeus:
@jmikola Tests now all pass, except for the occasional failure on replica sets due to a double free (our old friend has returned): https://spruce.mongodb.com/version/602f72fc9ccd4e172a60e6e8/tasks

tests/FunctionalTestCase.php
```diff
@@ -48,6 +48,10 @@ protected function assertSameObjectId($expectedObjectId, $actualObjectId)
 
     protected function getFeatureCompatibilityVersion(ReadPreference $readPreference = null)
     {
+        if ($this->isShardedCluster()) {
+            return $this->getServerVersion($readPreference);
+        }
+
```
jmikola, line 54: Backport? :)

.evergreen/run-tests.sh
```diff
@@ -0,0 +1,59 @@
...
+   case "${MONGODB_VERSION}" in
+      latest)
+         MONGODB_VERSION_NUMBER='5.0'
+         ;;
+      *)
+         MONGODB_VERSION_NUMBER=$MONGODB_VERSION
+         ;;
+   esac
+
+   PHPUNIT_OPTS="--dont-report-useless-tests --exclude-group matrix-testing-server-${MONGODB_VERSION_NUMBER}-driver-${DRIVER_MONGODB_VERSION},matrix-testing-server-${MONGODB_VERSION_NUMBER}-driver-${DRIVER_MONGODB_VERSION}-topology-${TOPOLOGY}"
```
jmikola, line 40: Very nice.

jmikola (approved):

tests/FunctionalTestCase.php
alcaeus: Yes. Fixed a number of tests 😉

.evergreen/run-tests.sh
alcaeus: Yeah, not happy about the formatting, but your suggestion to use groups to skip tests was great.

PR #806, opened 2021-01-29
PHPLIB-627: Comprehensive Atlas Tests

jmikola:
https://jira.mongodb.org/browse/PHPLIB-613
https://jira.mongodb.org/browse/PHPLIB-627
https://jira.mongodb.org/browse/PHPLIB-640

This includes a commit syncing spec tests with mongodb/specifications#921, which should be revised after that PR is merged so that it instead refers to the merged commit in that repository.

PR #811, opened 2021-02-12
PHPLIB-618: Change estimatedDocumentCount to use $collStats

alcaeus:
PHPLIB-618

### DIFF
--- a/src/Operation/EstimatedDocumentCount.php
+++ b/src/Operation/EstimatedDocumentCount.php
@@ -116,6 +116,7 @@ class EstimatedDocumentCount implements Executable, Explainable
     public function execute(Server $server)
     {
         $command = $this->createCommand($server);
+
         if ($command instanceof Aggregate) {
             try {
                 $cursor = $command->execute($server);
--- a/src/Operation/EstimatedDocumentCount.php
+++ b/src/Operation/EstimatedDocumentCount.php
@@ -17,12 +17,18 @@
 
 namespace MongoDB\Operation;
 
+use MongoDB\Driver\Exception\CommandException;
 use MongoDB\Driver\Exception\RuntimeException as DriverRuntimeException;
+use MongoDB\Driver\ReadConcern;
+use MongoDB\Driver\ReadPreference;
 use MongoDB\Driver\Server;
+use MongoDB\Driver\Session;
 use MongoDB\Exception\InvalidArgumentException;
 use MongoDB\Exception\UnexpectedValueException;
 use MongoDB\Exception\UnsupportedException;
 use function array_intersect_key;
+use function is_integer;
+use function MongoDB\server_supports_feature;
 
 /**
  * Operation for obtaining an estimated count of documents in a collection
@@ -42,11 +48,15 @@ class EstimatedDocumentCount implements Executable, Explainable
     /** @var array */
     private $options;
 
-    /** @var Count */
-    private $count;
+    /** @var int */
+    private static $errorCodeCollectionNotFound = 26;
+
+    /** @var int */
+    private static $wireVersionForCollStats = 12;
 
     /**
-     * Constructs a count command.
+     * Constructs a command to get the estimated number of documents in a
+     * collection.
      *
      * Supported options:
      *
@@ -73,9 +83,24 @@ class EstimatedDocumentCount implements Executable, Explainable
     {
         $this->databaseName = (string) $databaseName;
         $this->collectionName = (string) $collectionName;
-        $this->options = array_intersect_key($options, ['maxTimeMS' => 1, 'readConcern' => 1, 'readPreference' => 1, 'session' => 1]);
 
-        $this->count = $this->createCount();
+        if (isset($options['maxTimeMS']) && ! is_integer($options['maxTimeMS'])) {
+            throw InvalidArgumentException::invalidType('"maxTimeMS" option', $options['maxTimeMS'], 'integer');
+        }
+
+        if (isset($options['readConcern']) && ! $options['readConcern'] instanceof ReadConcern) {
+            throw InvalidArgumentException::invalidType('"readConcern" option', $options['readConcern'], ReadConcern::class);
+        }
+
+        if (isset($options['readPreference']) && ! $options['readPreference'] instanceof ReadPreference) {
+            throw InvalidArgumentException::invalidType('"readPreference" option', $options['readPreference'], ReadPreference::class);
+        }
+
+        if (isset($options['session']) && ! $options['session'] instanceof Session) {
+            throw InvalidArgumentException::invalidType('"session" option', $options['session'], Session::class);
+        }
+
+        $this->options = array_intersect_key($options, ['maxTimeMS' => 1, 'readConcern' => 1, 'readPreference' => 1, 'session' => 1]);
     }
 
     /**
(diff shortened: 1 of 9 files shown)


src/Operation/EstimatedDocumentCount.php
```diff
@@ -73,9 +83,24 @@ public function __construct($databaseName, $collectionName, array $options = [])
     {
         $this->databaseName = (string) $databaseName;
         $this->collectionName = (string) $collectionName;
-        $this->options = array_intersect_key($options, ['maxTimeMS' => 1, 'readConcern' => 1, 'readPreference' => 1, 'session' => 1]);
 
-        $this->count = $this->createCount();
+        if (isset($options['maxTimeMS']) && ! is_integer($options['maxTimeMS'])) {
```
alcaeus, line 87: I decided that checking options for `estimatedDocumentCount` here specifically makes sense now that we no longer know which command (count vs. aggregate) is being executed.

src/Operation/EstimatedDocumentCount.php
```diff
@@ -90,18 +115,53 @@ public function __construct($databaseName, $collectionName, array $options = [])
...
+        if ($command instanceof Aggregate) {
+            try {
+                $cursor = $command->execute($server);
+            } catch (CommandException $e) {
+                if ($e->getCode() == self::$errorCodeCollectionNotFound) {
+                    return 0;
+                }
+
+                throw $e;
+            }
```
alcaeus, line 128: I'm not entirely happy with this part of the logic, but I couldn't come up with a nicer way to ignore nsNotFound errors.

src/Operation/EstimatedDocumentCount.php
```diff
@@ -90,18 +115,53 @@ public function __construct($databaseName, $collectionName, array $options = [])
      */
     public function execute(Server $server)
     {
-        return $this->count->execute($server);
+        $command = $this->createCommand($server);
+        if ($command instanceof Aggregate) {
```
alcaeus, line 119: Given that we expose the command document using `getCommandDocument`, I thought it was best to create the command and attach special handling to the `aggregate` command, as opposed to checking for the correct wire version here again. Let me know if you'd prefer an explicit wire version check here instead of the hidden check in `createCommand`.

tests/FunctionalTestCase.php
```diff
@@ -278,7 +279,7 @@ protected function getServerVersion(ReadPreference $readPreference = null)
         $document = current($cursor->toArray());
 
         if (isset($document['version']) && is_string($document['version'])) {
-            return $document['version'];
+            return preg_replace('#^(\d+\.\d+\.\d+).*$#', '\1', $document['version']);
```
alcaeus, line 282: This change was necessary since my tests were being skipped. My local MongoDB server reported `4.9.0-alpha4-...` as version number, which `version_compare` interprets as being lower than `4.9.0` (due to the alpha flag). The [Version String in the unified test format](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#version-string) makes no mention of stabilities, but instead requires the version to conform to the `<major>.<minor>.<patch>` schema.

tests/UnifiedSpecTests/UnifiedSpecTest.php
```diff
@@ -141,6 +141,19 @@ private function doTestCase(stdClass $test, string $schemaVersion, array $runOnR
         }
     }
 
+    /**
+     * @dataProvider provideCrudTests
+     */
+    public function testCrud(...$args)
```
alcaeus, line 147: Having separate test methods for individual specs allows for easier filtering of what tests to execute (without having to mess with `--filter` and dataset names.

src/Operation/EstimatedDocumentCount.php
jmikola: Both Count and Aggregate do the same checks for these options, so does this change anything?

Perhaps this is needed to ensure we raise these errors in the constructor, since construction of the other operation is now deferred to `execute()`? I suppose the only alternative would be to create both a Count and Aggregate operation up front and decide which is used later.

src/Operation/EstimatedDocumentCount.php
```diff
@@ -90,18 +115,53 @@ public function __construct($databaseName, $collectionName, array $options = [])
...
+            try {
+                $cursor = $command->execute($server);
+            } catch (CommandException $e) {
```
jmikola, line 122: I just noticed that ListIndexes uses `MongoDB\Driver\Exception\RuntimeException` for this logic. Reported that in [PHPLIB-636](https://jira.mongodb.org/browse/PHPLIB-636).

src/Operation/EstimatedDocumentCount.php
jmikola: This is on par with what ListIndexes and DropCollection both do.

src/Operation/EstimatedDocumentCount.php
```diff
@@ -90,18 +115,53 @@ public function __construct($databaseName, $collectionName, array $options = [])
...
+                if ($e->getCode() == self::$errorCodeCollectionNotFound) {
+                    return 0;
+                }
+
+                throw $e;
+            }
+
+            $cursor->rewind();
+
+            return $cursor->current()->n;
```
jmikola, line 132: Is this worth an `(integer)` cast just to be sure? I wouldn't expect anything else to be returned, but it is technically a wild value from the server.

src/Operation/EstimatedDocumentCount.php
jmikola: I realize `getCommandDocument` complicates things a bit. I think this is fine.

I'd probably just suggest a blank line between assigning `$command` and the `if` for readability.

tests/FunctionalTestCase.php
jmikola: This seems related to https://github.com/mongodb/specifications/pull/931.

Works for me. The other alternative would be to use `versionArray` and build the string ourselves. Presumably, we'll need to do this for the test runners as well?

tests/UnifiedSpecTests/UnifiedSpecTest.php
jmikola: Is this change related to anything else in the PR? I don't see any tests added to the `crud/` directory.

Given that UnifiedSpecTest is primarily testing its own tests (e.g. passing/failing), I wonder if other specs using the runner should just have their own test classes.

This would probably be best deferred until after I merge #806, as that PR extracts the runner itself from the test class. If this isn't urgent, I'd suggest pulling this out to avoid merge conflicts and just proposing this in a PHPLIB ticket dependent on PHPLIB-613.

src/Operation/EstimatedDocumentCount.php
alcaeus: Yes, that is the main difference. I considered instantiating both operation objects up front, but avoided this as one object may throw an exception during initialisation even though it wouldn't be used later on.

tests/FunctionalTestCase.php
alcaeus: Yes, that issue was raised afterwards, and we have PHPLIB-634 as a downstream ticket to take care of this. If you don't mind I'd leave this in for now (as it ensures that the tests run as expected) and then the logic for all test runners as part of PHPLIB-634.

tests/UnifiedSpecTests/UnifiedSpecTest.php
alcaeus: Given that the tests for each spec would only contain a single `test<Spec>(...$args)` method, I'm not sure what there is to be gained by having each in a separate class. If we want to separate them somewhat, I'd suggest the following:
* Refactor `UnifiedSpecTest` to an abstract `UnifiedSpecTestCase` class
* Add a `UnifiedRunnerTest` class that only tests the runner itself (e.g. passing/failing)
* From there, add either a common `SpecTest` class or a single `<Spec>Test` class for each spec that has tests in the unified format.

In the meantime, I can extract this change to a separate file until we're done with #806, especially considering that the PR for PHPLIB-622 will add versioned API tests where we'll see the problem. That should avoid merge conflicts in the refactoring PR.

alcaeus: To follow up on the tests themselves, there's a single new test file added to `tests/UnifiedSpecTests/crud/`, which is what is tested here. The other tests have not been ported to the unified format yet.

Thinking of it, I remember discussing where these tests would live, but don't remember coming to a conclusion. I've put them in the `UnifiedSpecTests` folder for the time being, but we may want to not use that directory as a catch-all for unified spec tests and rather use different directories for it.

jmikola: Oh, I must have missed `tests/UnifiedSpecTests/crud/estimatedDocumentCount.json` in the previous review. I see it now.

jmikola (approved):
The coding standards check failed. Once that is addressed, LGTM.

PR #814, opened 2021-03-05
PHPLIB-624 Sync change stream spec tests for delta oplog entries

alcaeus:
PHPLIB-624

### DIFF
--- a/tests/UnifiedSpecTests/UnifiedSpecTest.php
+++ b/tests/UnifiedSpecTests/UnifiedSpecTest.php
@@ -178,6 +178,19 @@ class UnifiedSpecTest extends FunctionalTestCase
         return $this->provideTests(__DIR__ . '/valid-fail');
     }
 
+    /**
+     * @dataProvider provideChangeStreamsTests
+     */
+    public function testChangeStreams(...$args)
+    {
+        $this->doTestCase(...$args);
+    }
+
+    public function provideChangeStreamsTests()
+    {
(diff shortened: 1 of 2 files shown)


tests/UnifiedSpecTests/UnifiedSpecTest.php
```diff
@@ -178,6 +178,19 @@ public function provideFailingTests()
...
+     */
+    public function testChangeStreams(...$args)
+    {
+        $this->doTestCase(...$args);
+    }
+
+    public function provideChangeStreamsTests()
+    {
+        return $this->provideTests(__DIR__ . '/change-streams');
+    }
```
jmikola, line 192: Same comment as https://github.com/mongodb/mongo-php-library/pull/811#discussion_r598996504. I'll note that this PR is actually adding a file to `change-stream/`, so perhaps this is needed to actually run the test. If so, this can stay.

jmikola (approved):

tests/UnifiedSpecTests/UnifiedSpecTest.php
alcaeus: Yes, same story here. Happy to hold off on this until we've come to a resolution or until the changes introduced by #806 are merged.

jmikola: Oh, they can both get merged so these tests can be executed. In #811 I missed that you were adding a unified CRUD test.

jmikola (approved):

PR #815, opened 2021-03-25
PHPLIB-621 Document Versioned API usage

alcaeus:
PHPLIB-621

This is the PHPLIB counterpart to the API documentation drafted in https://github.com/alcaeus/doc-en/pull/1. I wanted to get a head start on these changes, but wait with wrapping up documentation until the features are merged. Nonetheless, the tutorial and API documentation is ready to be reviewed.

Note: The first commit sorts non-deprecated driver options to the top in alphabetical order, with deprecated options documented last. The other two commits document the usage of the `serverApi` options.

### DIFF
--- a/docs/includes/apiargs-MongoDBClient-method-construct-driverOptions.yaml
+++ b/docs/includes/apiargs-MongoDBClient-method-construct-driverOptions.yaml
@@ -41,6 +41,18 @@ operation: ~
 optional: true
 ---
 arg_name: option
+name: serverApi
+type: :php:`MongoDB\\Driver\\ServerApi <class.mongodb-driver-serverapi>`
+description: |
+  Used to declare an API version on the client. See the
+  :manual:`Versioned API tutorial </tutorial/versioned-api>` for usage.
+
+  .. versionadded:: 1.9
+interface: phpmethod
+operation: ~
+optional: true
+---
+arg_name: option
 name: typeMap
 type: array
 description: |
--- a/docs/includes/apiargs-MongoDBClient-method-construct-driverOptions.yaml
+++ b/docs/includes/apiargs-MongoDBClient-method-construct-driverOptions.yaml
@@ -1,4 +1,46 @@
 arg_name: option
+name: autoEncryption
+type: array
+description: |
+  Options to configure client-side field-level encryption in the driver. The
+  encryption options are documented in the :php:`extension documentation
+  <manual/en/mongodb-driver-manager.construct.php#mongodb-driver-manager.construct-driveroptions>`.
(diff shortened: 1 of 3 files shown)


docs/includes/apiargs-MongoDBClient-method-construct-driverOptions.yaml
```diff
@@ -1,4 +1,58 @@
 arg_name: option
+name: autoEncryption
```
jmikola, line 2: Was this moved up to alphabetize all of the options? I noticed `allow_invalid_hostname` is still down below.

docs/tutorial/versioned-api.txt
```diff
@@ -0,0 +1,91 @@
...
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Declaring an API version
+------------------------
+
+To declare an API version, pass a ``serverApi`` driver option when creating your
+manager. The value is a
```
jmikola, line 17: Since we're in the PHPLIB docs, should this be "client"?

docs/tutorial/versioned-api.txt
```diff
@@ -0,0 +1,91 @@
...
+:php:`MongoDB\\Driver\\ServerApi <class.mongodb-driver-serverapi>` instance that
+contains API version information. The only available API version is ``"1"``.
```
jmikola, line 19: I would suggest removing "The only available API version is 1" or replacing it with something more future-proof:

> This feature is introduced in MongoDB 5.0, which will initially support only API version "1". Additional versions may be introduced in future versions of the server.

docs/tutorial/versioned-api.txt
```diff
@@ -0,0 +1,91 @@
...
+   Only declare an API version when connecting to a deployment that has no
+   pre-5.0 members. Older servers will error when encountering commands with a
+   declared API version.
+
+Strict API
+----------
+
+By default, declaring an API version guarantees behavior for commands that are
+part of the versioned API, but does not forbid using commands that are not part
+of the API version. To only allow commands that are part of an API version,
```
jmikola, line 45: Should this say "To only allow commands and options"? In this example, you are again using the `find` command, which we know is in version API 1. The difference is that the `tailable` option is not supported.

docs/includes/apiargs-MongoDBClient-method-construct-driverOptions.yaml
alcaeus: Yes. The first commit moves the deprecated options to the bottom and sorts non-deprecated once alphabetically.

PR #824, opened 2021-04-29
PHPLIB-650 Run versioned API tests with acceptApiVersion2

alcaeus:
PHPLIB-650

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -356,7 +356,7 @@ tasks:
         - func: "run atlas data lake test"
 
     - name: "test-requireApiVersion"
-      tags: ["versioned_api"]
+      tags: ["versioned-api"]
       commands:
         - func: "bootstrap mongo-orchestration"
           vars:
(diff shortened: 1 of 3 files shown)


.evergreen/config.yml
```diff
@@ -356,7 +356,7 @@ tasks:
         - func: "run atlas data lake test"
 
     - name: "test-requireApiVersion"
-      tags: ["versioned_api"]
```
jmikola, line 359: Ah, I didn't realize Evergreen supported tags. Was this just never used?

.evergreen/config.yml
```diff
@@ -356,7 +356,7 @@ tasks:
...
+      tags: ["versioned-api"]
```
jmikola, line 359: Should this test also use `TESTS: "versioned-api"` to select a subset of the PHPUnit test suite, or does it intentionally need to target the entire suite?

.evergreen/config.yml
alcaeus: Yep, we never used this. libmongoc makes heavy use of tags in their config.

.evergreen/config.yml
alcaeus: The Versioned API spec calls for running the entire test suite against a server with `requireApiVersion` and authentication enabled to ensure that all commands declare an API version, as they can't be checked using command monitoring.

PR #829, opened 2021-05-31
PHPLIB-659 Support options for time-series collections

alcaeus:
PHPLIB-659

### DIFF
--- a/docs/includes/apiargs-MongoDBDatabase-method-createCollection-option.yaml
+++ b/docs/includes/apiargs-MongoDBDatabase-method-createCollection-option.yaml
@@ -38,15 +38,16 @@ arg_name: option
 name: expireAfterSeconds
 type: integer
 description: |
-  Available on 5.0+ servers only
-
   Allows specifying a TTL after which documents will be removed from a
   time-series collection.
+
+  This option is available in MongoDB 5.0+ and will result in an exception at
+  execution time if specified for an older server version.
+
+  .. versionadded:: 1.9
 interface: phpmethod
 operation: ~
 optional: true
-post: |
-  .. versionadded:: 1.9
 ---
 arg_name: option
 name: flags
@@ -151,14 +152,15 @@ arg_name: option
 name: timeseries
 type: array|object
 description: |
-  Available on 5.0+ servers only
-
   Allows users to specify options for time-series collections.
+
+  This option is available in MongoDB 5.0+ and will result in an exception at
+  execution time if specified for an older server version.
+
+  .. versionadded:: 1.9
 interface: phpmethod
 operation: ~
 optional: true
-post: |
-  .. versionadded:: 1.9
 ---
 source:
   file: apiargs-MongoDBDatabase-common-option.yaml
--- a/docs/includes/apiargs-MongoDBDatabase-method-createCollection-option.yaml
+++ b/docs/includes/apiargs-MongoDBDatabase-method-createCollection-option.yaml
@@ -35,6 +35,20 @@ pre: |
   </reference/bson-type-comparison-order/#collation>` for the collection.
 ---
 arg_name: option
+name: expireAfterSeconds
+type: integer
+description: |
+  Available on 5.0+ servers only
+
+  Allows specifying a TTL after which documents will be removed from a
+  time-series collection.
+interface: phpmethod
+operation: ~
+optional: true
+post: |
+  .. versionadded:: 1.9
+---
+arg_name: option
 name: flags
 type: integer
 description: |
@@ -132,6 +146,19 @@ description: |
 interface: phpmethod
 operation: ~
 optional: true
+---
+arg_name: option
+name: timeseries
+type: array|object
+description: |
+  Available on 5.0+ servers only
+
+  Allows users to specify options for time-series collections.
+interface: phpmethod
+operation: ~
+optional: true
+post: |
+  .. versionadded:: 1.9
 ---
 source:
   file: apiargs-MongoDBDatabase-common-option.yaml
--- a/src/Operation/CreateCollection.php
+++ b/src/Operation/CreateCollection.php
@@ -82,6 +82,10 @@ class CreateCollection implements Executable
      *    This is not supported for server versions < 3.4 and will result in an
      *    exception at execution time if used.
      *
+     *  * expireAfterSeconds: The TTL for documents in time series collections.
+     *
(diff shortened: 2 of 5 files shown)


docs/includes/apiargs-MongoDBDatabase-method-createCollection-option.yaml
```diff
@@ -133,6 +133,19 @@ interface: phpmethod
 operation: ~
 optional: true
 ---
+arg_name: option
+name: timeseries
+type: array|object
+description: |
+  Available on 5.0+ servers only
+
+  Allows users to specify options for time-series collections.
```
alcaeus, line 142: Note: I haven't found additional documentation for this param publicly yet, so I've only stubbed out documentation. I suggest we revisit this closer to the 5.0 release cycle if at all. In that case, I'd create a follow-up ticket, but I'm also fine linking to the command documentation for more information.

docs/includes/apiargs-MongoDBDatabase-method-createCollection-option.yaml
```diff
@@ -133,6 +133,19 @@ interface: phpmethod
 operation: ~
 optional: true
 ---
+arg_name: option
+name: timeseries
```
jmikola, line 137: I realize this PR pertains to [DRIVERS-1438](https://jira.mongodb.org/browse/DRIVERS-1438), which only added a `timeseries` option, but there is a follow-up ticket ([DRIVERS-1774](https://jira.mongodb.org/browse/DRIVERS-1774)) that moves `expireAfterSeconds` to a top-level option. That spec PR is still being reviewed, so there are not yet any language tickets.

I'd suggest holding off on this PR for the time being, since DRIVERS-1774 will require changing this documentation, `CreateCollection.php`, and resyncing spec tests.

alcaeus: Ah, thanks for the heads up. I'll give the tests from that PR a quick check to provide feedback and will wait until DRIVERS-1774 is ready before merging.

alcaeus:
Updated to incorporate changes from DRIVERS-1774 and DRIVERS-1779. Note that the docs are very vague at this point, but I was not able to find any public documentation for these fields yet. Should we create a ticket to follow up on documentation before the release of 1.9?

jmikola:
> Should we create a ticket to follow up on documentation before the release of 1.9?

Yes. I'd suggest talking to @agolin95 about that in case a separate DRIVERS ticket would be useful. We can then have that block on a relevant ticket for the server manual itself. In this case, I don't think we really need to document the internal structure of `timeseries` in our documentation, but it'd be nice to have an actual URL in the server manual to link to for reference.

alcaeus:
@jmikola rebased on top of master and ensured that the spec test is up to date. Ready for a final review.

docs/includes/apiargs-MongoDBDatabase-method-createCollection-option.yaml
```diff
@@ -35,6 +35,20 @@ pre: |
   </reference/bson-type-comparison-order/#collation>` for the collection.
 ---
 arg_name: option
+name: expireAfterSeconds
+type: integer
+description: |
+  Available on 5.0+ servers only
```
jmikola, line 41: Based on existing docs, this should be phrased as:

> This option is available in MongoDB 5.0+ and will result in an exception at execution time if specified for an older server version.

Alternatively (as is done later in this file for `writeConcern`):

> This is not supported for server versions prior to 5.0 and will result in an exception at execution time if used.

In both cases, these lines are typically added _after_ any documentation describing the option's functionality. For instance, the `writeConcern` docs later in this file add that using `post:`.

docs/includes/apiargs-MongoDBDatabase-method-createCollection-option.yaml
```diff
@@ -133,6 +147,19 @@ interface: phpmethod
 operation: ~
 optional: true
 ---
+arg_name: option
+name: timeseries
+type: array|object
+description: |
+  Available on 5.0+ servers only
```
jmikola, line 154: Same comment as `expireAfterSeconds`.

docs/includes/apiargs-MongoDBDatabase-method-createCollection-option.yaml
```diff
@@ -133,6 +147,19 @@ interface: phpmethod
...
+
+  Allows users to specify options for time-series collections.
+interface: phpmethod
+operation: ~
+optional: true
+post: |
+  .. versionadded:: 1.9
```
jmikola, line 161: Same comment as `expireAfterSeconds`.

docs/includes/apiargs-MongoDBDatabase-method-createCollection-option.yaml
```diff
@@ -35,6 +35,20 @@ pre: |
...
+description: |
+  Available on 5.0+ servers only
+
+  Allows specifying a TTL after which documents will be removed from a
+  time-series collection.
+interface: phpmethod
+operation: ~
+optional: true
+post: |
+  .. versionadded:: 1.9
```
jmikola, line 49: `post:` is generally only helpful if we're inheriting common docs. In this case, it'd be more consistent with existing docs to just append this after two line breaks at the end of `description`.

src/Operation/CreateCollection.php
```diff
@@ -167,6 +179,10 @@ public function __construct($databaseName, $collectionName, array $options = [])
             throw InvalidArgumentException::invalidType('"storageEngine" option', $options['storageEngine'], 'array or object');
         }
 
+        if (isset($options['timeseries']) && ! is_array($options['timeseries']) && ! is_object($options['timeseries'])) {
```
jmikola, line 182: Reading this again with fresh eyes, I find it very odd that the server team didn't use camelCase for this option name (vs. `expireAfterSeconds`).

jmikola (approved):
LGTM with my suggestions for doc changes.

docs/includes/apiargs-MongoDBDatabase-method-createCollection-option.yaml
alcaeus: I've updated the documentation as suggested. Thanks!

alcaeus:
Build failures are addressed in #835, merging despite failing CI.

PR #830, opened 2021-06-01
PHPLIB-663: Support let option for aggregate command

jmikola:
https://jira.mongodb.org/browse/PHPLIB-663


src/Operation/Aggregate.php
```diff
@@ -196,6 +204,10 @@ public function __construct($databaseName, $collectionName, array $pipeline, arr
             throw InvalidArgumentException::invalidType('"hint" option', $options['hint'], 'string or array or object');
         }
 
+        if (isset($options['let']) && ! is_array($options['let']) && ! is_object($options['let'])) {
+            throw InvalidArgumentException::invalidType('"let" option', $options['let'], 'array or object');
```
alcaeus, line 208: `invalidType` supports an array of accepted types but I haven't managed to change all of these.

```suggestion
            throw InvalidArgumentException::invalidType('"let" option', $options['let'], ['array', 'object']);
```

alcaeus (approved):

src/Operation/Aggregate.php
jmikola: Ah, you added this back in https://github.com/mongodb/mongo-php-library/commit/b2cd2866c64690603896b788cd56e45ff4cc69a4. Totally forgot about that.

PR #834, opened 2021-06-28
Support 1.4 schema in unified test runner and sync tests for various 5.0-compat features

jmikola:
* https://jira.mongodb.org/browse/PHPLIB-655
 * https://jira.mongodb.org/browse/PHPLIB-660
 * https://jira.mongodb.org/browse/PHPLIB-665
 * https://jira.mongodb.org/browse/PHPLIB-666
 * https://jira.mongodb.org/browse/PHPLIB-669
 * https://jira.mongodb.org/browse/PHPLIB-670
 * https://jira.mongodb.org/browse/PHPLIB-679
 * https://jira.mongodb.org/browse/PHPLIB-680

This PR updates all unified spec tests to their latest version (across individual commits). This includes more recent changes to the unified transaction tests for [DRIVERS-1539](https://jira.mongodb.org/browse/DRIVERS-1539).

Detection of load balancer and serverless is intentional omitted (pending the respective PHP tickets to implement support for those features). The test runner currently assumes that neither is being used when resolving RunOnRequirements.


tests/UnifiedSpecTests/UnifiedSpecTest.php
```diff
@@ -45,66 +45,6 @@ class UnifiedSpecTest extends FunctionalTestCase
...
-        'crud/updateMany-dots_and_dollars: Updating document to set top-level dollar-prefixed key on 5.0+ server' => 'CDRIVER-3895 and PHPC-1765',
-        'crud/updateMany-dots_and_dollars: Updating document to set top-level dotted key on 5.0+ server' => 'CDRIVER-3895 and PHPC-1765',
-        'crud/updateMany-dots_and_dollars: Updating document to set dollar-prefixed key in embedded doc on 5.0+ server' => 'CDRIVER-3895 and PHPC-1765',
-        'crud/updateMany-dots_and_dollars: Updating document to set dotted key in embedded doc on 5.0+ server' => 'CDRIVER-3895 and PHPC-1765',
-        'crud/updateOne-dots_and_dollars: Updating document to set top-level dollar-prefixed key on 5.0+ server' => 'CDRIVER-3895 and PHPC-1765',
-        'crud/updateOne-dots_and_dollars: Updating document to set top-level dotted key on 5.0+ server' => 'CDRIVER-3895 and PHPC-1765',
-        'crud/updateOne-dots_and_dollars: Updating document to set dollar-prefixed key in embedded doc on 5.0+ server' => 'CDRIVER-3895 and PHPC-1765',
-        'crud/updateOne-dots_and_dollars: Updating document to set dotted key in embedded doc on 5.0+ server' => 'CDRIVER-3895 and PHPC-1765',
```
alcaeus, line 107: Heads up: I added some transactions tests that were failing to this list in #835

jmikola: Thanks. I'll be sure to discard those changes when I rebase this.

tests/UnifiedSpecTests/Operation.php
```diff
@@ -248,17 +261,17 @@ private function executeForClient(Client $client)
...
-                $changeStream = $client->watch(
-                    $args['pipeline'] ?? [],
+                assertArrayHasKey('pipeline', $args);
+                assertIsArray($args['pipeline']);
+
+                return $client->watch(
+                    $args['pipeline'],
                     array_diff_key($args, ['pipeline' => 1])
                 );
-                $changeStream->rewind();
```
jmikola, line 255: It's not clear to me why we were ever rewinding here. When a ChangeStream is first instantiated, its key is `null`. This means that `iteratorUntilDocumentOrError` is going to make an initial `rewind()` call on its own before attempting to check for `valid()` and either return the first result or loop with `next()` until we hit a `valid()` element.

Removing this didn't break any of the existing change stream tests, so I think it's safe to conclude that this was redundant.

tests/UnifiedSpecTests/Operation.php
```diff
@@ -358,6 +420,11 @@ private function executeForCollection(Collection $collection)
 
             case 'updateMany':
             case 'updateOne':
+                assertArrayHasKey('filter', $args);
+                assertArrayHasKey('update', $args);
+                assertInstanceOf(stdClass::class, $args['filter']);
+                assertThat($args['update'], logicalOr(new IsType('array'), new IsType('object')));
```
jmikola, line 426: This is necessary for update pipelines. We do the same when parsing args for bulkWrite requests.

tests/UnifiedSpecTests/Operation.php
```diff
@@ -368,18 +435,31 @@ private function executeForCollection(Collection $collection)
...
                 );
             case 'listIndexes':
-                return $collection->listIndexes($args);
+                return iterator_to_array($collection->listIndexes($args));
             case 'mapReduce':
+                assertArrayHasKey('map', $args);
+                assertArrayHasKey('reduce', $args);
+                assertArrayHasKey('out', $args);
+                assertInstanceOf(Javascript::class, $args['map']);
+                assertInstanceOf(Javascript::class, $args['reduce']);
```
jmikola, line 460: We could technically use JavascriptInterface here (per MapReduce's own signature), but decoding extended JSON should always yield Javascript objects. I don't believe any existing unified tests cover mapReduce, but the tests in retryable reads do use `$code` for this type -- so I'd expect these to still be Javascript whenever those are ported to the unified format.

tests/UnifiedSpecTests/Operation.php
```diff
@@ -473,15 +612,23 @@ private function executeForBucket(Bucket $bucket)
 
         switch ($this->name) {
             case 'delete':
+                assertArrayHasKey('id', $args);
```
jmikola, line 615: No type assertions here since `id` can be anything.

tests/UnifiedSpecTests/Operation.php
```diff
@@ -504,74 +651,108 @@ private function executeForTestRunner()
...
                 $eventObserver = $this->context->getEventObserverForClient($this->arguments['client']);
                 assertNotEquals(...$eventObserver->getLsidsOnLastTwoCommands());
                 break;
+            case 'assertNumberConnectionsCheckedOut':
+                assertArrayHasKey('connections', $args);
+                assertIsInt($args['connections']);
+                /* PHP does not implement connection pooling. Check parameters
+                 * for the sake of valid-fail tests, but otherwise raise an
+                 * error. */
+                Assert::fail('Tests using assertNumberConnectionsCheckedOut should be skipped');
```
jmikola, line 709: FWIW, this failure is no different than an earlier `connections` assertion failure. Either would satisfy the catch in `UnifiedSpecTest::testFailingTests()` (only PHPUnit's SkippedTest and Warning exceptions are disqualifying).

Still, I thought it was worth asserting as much as possible before we bomb out (on the off chance the test file itself was incorrect).

tests/UnifiedSpecTests/Operation.php
```diff
@@ -504,74 +651,108 @@ private function executeForTestRunner()
...
+                assertArrayHasKey('indexName', $args);
                 assertIsString($args['databaseName']);
                 assertIsString($args['collectionName']);
                 assertIsString($args['indexName']);
                 assertNotContains($args['indexName'], $this->getIndexNames($args['databaseName'], $args['collectionName']));
                 break;
             case 'assertSameLsidOnLastTwoCommands':
+                /* Context::getEventObserverForClient() requires the client ID.
+                 * Avoid checking $args['client'], which is already resolved. */
+                assertArrayHasKey('client', $this->arguments);
```
jmikola, line 692: This is one of the few places we assert on `$this->arguments` instead of `$args`, because we need the unresolved entity ID for the Context method. We also do this below for `assertSessionDirty`.

tests/UnifiedSpecTests/RunOnRequirement.php
```diff
@@ -50,16 +82,28 @@ public function __construct(stdClass $o)
         if (isset($o->topologies)) {
             assertIsArray($o->topologies);
             assertContainsOnly('string', $o->topologies);
+            assertEmpty(array_diff($o->topologies, self::$supportedTopologies));
```
jmikola, line 85: Previously, unrecognized topology values would be ignored.

tests/UnifiedSpecTests/RunOnRequirement.php
```diff
@@ -69,17 +113,7 @@ public function isSatisfied(string $serverVersion, string $topology, stdClass $s
...
-                return true;
-            }
-
-            /* Ensure "sharded-replicaset" is also accepted for topologies that
-             * only include "sharded" (agnostic about the shard topology) */
-            if ($topology === self::TOPOLOGY_SHARDED_REPLICASET && in_array(self::TOPOLOGY_SHARDED, $this->topologies)) {
-                return true;
-            }
-
+        if (isset($this->topologies) && ! $this->isTopologySatisfied($topology)) {
```
jmikola, line 116: This fixes a subtle bug introduced by the versioned API work. When topology was the last thing checked, we could `return true`; however, that doesn't work if we need to do other assertions later on in this function.

tests/UnifiedSpecTests/UnifiedTestRunner.php
```diff
@@ -213,93 +215,41 @@ private function doTestCase(stdClass $test, string $schemaVersion, array $runOnR
...
-        $serverParameters = $this->getCachedServerParameters();
+        if (! isset($cachedIsSatisfiedArgs)) {
+            $cachedIsSatisfiedArgs = [
+                $this->getServerVersion(),
+                $this->getTopology(),
+                $this->getServerParameters(),
+                $this->isAuthenticated(),
+                $this->isServerless(),
+            ];
+        }
```
jmikola, line 237: Once I started adding new fields to the requirements, it made a lot more sense to cache them in one place (vs. the two methods we previously had).

tests/UnifiedSpecTests/UnifiedTestRunner.php
```diff
@@ -213,93 +215,41 @@ private function doTestCase(stdClass $test, string $schemaVersion, array $runOnR
...
-                throw new UnexpectedValueException('Toplogy is neither single nor RS nor sharded');
-        }
-
-        return $cachedTopology;
+        Assert::markTestSkipped(sprintf(
+            'Server (version=%s, toplogy=%s, auth=%s) does not meet test requirements',
+            $cachedIsSatisfiedArgs[0],
+            $cachedIsSatisfiedArgs[1],
+            $cachedIsSatisfiedArgs[3] ? 'yes' : 'no'
+        ));
```
jmikola, line 252: I didn't bother to add serverless here, which I think is fine because it's not really supported yet. We can consider doing so when the functionality is actually implemented.

That said, I don't see a nice way to print the server parameters in this message -- since there could be tons of them. One approach might be trying to log the environment details when we first compute/cache them, and then keeping the skip message short and avoiding redundant output. I'm not familiar with what, if any, logging mechanisms PHPUnit makes available.

tests/UnifiedSpecTests/UnifiedTestRunner.php
```diff
@@ -339,6 +289,52 @@ private function getServerVersion() : string
...
     }
 
+    /**
+     * Return the topology type.
+     *
+     * @throws UnexpectedValueException if topology is neither single nor RS nor sharded
+     */
+    private function getTopology() : string
+    {
+        // TODO: detect load-balanced topologies once PHPLIB-671 is implemented
```
jmikola, line 299: IIRC, this will entail checking an environment variable and likely preempt this `switch`.

tests/UnifiedSpecTests/UnifiedTestRunner.php
```diff
@@ -339,6 +289,52 @@ private function getServerVersion() : string
...
+        }
+    }
+
+    /**
+     * Return whether the connection is authenticated.
+     */
+    private function isAuthenticated() : bool
+    {
+        $database = $this->internalClient->selectDatabase('admin');
+        $connectionStatus = $database->command(['connectionStatus' => 1])->toArray()[0];
```
jmikola, line 320: Supported since MongoDB 2.4: https://docs.mongodb.com/v3.0/reference/command/connectionStatus/

I look forward to finding out this isn't implemented properly by the Atlas proxies :smile:

alcaeus: FWIW, libmongoc checks whether the connection string contains authentication info, or if the `MONGOC_TEST_USER` env variable is set. If the proxy gives us trouble, I suggest we default to something like this as a fallback.

alcaeus (approved):
LGTM, thanks for taking care of the new test runner features 🎉

tests/UnifiedSpecTests/UnifiedTestRunner.php
jmikola: I'll add a comment to this method to note that we might want to consider that.

PR #837, opened 2021-07-01
PHPLIB-636: ListIndexes should use CommandException for catching missing namespace and database errors

tanlisu:
https://jira.mongodb.org/browse/PHPLIB-636

### DIFF
--- a/src/Operation/ListIndexes.php
+++ b/src/Operation/ListIndexes.php
@@ -19,7 +19,7 @@ namespace MongoDB\Operation;
 
 use EmptyIterator;
 use MongoDB\Driver\Command;
-use MongoDB\Driver\Exception\RuntimeException as DriverRuntimeException;
+use MongoDB\Driver\Exception\CommandException;
 use MongoDB\Driver\Server;
 use MongoDB\Driver\Session;
 use MongoDB\Exception\InvalidArgumentException;
@@ -90,7 +90,7 @@ class ListIndexes implements Executable
      * @see Executable::execute()
      * @param Server $server
      * @return IndexInfoIterator
-     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     * @throws CommandException for other driver errors (e.g. connection errors)
      */
     public function execute(Server $server)
     {
@@ -123,7 +123,7 @@ class ListIndexes implements Executable
      *
      * @param Server $server
      * @return IndexInfoIteratorIterator
-     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     * @throws CommandException for other driver errors (e.g. connection errors)
      */
     private function executeCommand(Server $server)
     {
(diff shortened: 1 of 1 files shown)


src/Operation/ListIndexes.php
```diff
@@ -123,7 +123,7 @@ private function createOptions()
      *
      * @param Server $server
      * @return IndexInfoIteratorIterator
-     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     * @throws CommandException for other driver errors (e.g. connection errors)
```
jmikola, line 126: These two instances of `@throws` should not have been modified. Did you think to do this because of a phpcs warning that DriverRuntimeException was undefined? IIRC, it does attempt to check class references in doc block comments.

Remember that [ServerException](https://www.php.net/manual/en/class.mongodb-driver-exception-serverexception.php) and its child classes are defined as originating from a server response. Something like a connection error (mentioned on this very line) can still be thrown by [`Server::executeReadCommand()`](https://www.php.net/manual/en/mongodb-driver-server.executereadcommand.php) and `ListIndexes::executeCommand()` will not catch it at all and allow it to propagate up the chain.

See [Errors/Exceptions in `ServerException::executeReadCommand()`](https://www.php.net/manual/en/mongodb-driver-server.executereadcommand.php#refsect1-mongodb-driver-server.executereadcommand-errors) for a list of possible exceptions that the method might throw. Note that it doesn't mention ServerException directly, but that's covered by RuntimeException.

----

On a separate note, if these lines _were_ actually supposed to be modified, we would also want to update the documentation for [Errors/Exceptions in `Collection::listIndexes()`](https://docs.mongodb.com/php-library/master/reference/method/MongoDBCollection-listIndexes/#errors-exceptions), the source of which is in [docs/reference/method/MongoDBCollection-listIndexes.txt](https://github.com/mongodb/mongo-php-library/blob/master/docs/reference/method/MongoDBCollection-listIndexes.txt).

src/Operation/ListIndexes.php
```diff
@@ -19,7 +19,7 @@
 
 use EmptyIterator;
 use MongoDB\Driver\Command;
-use MongoDB\Driver\Exception\RuntimeException as DriverRuntimeException;
```
jmikola, line 22: Related to my comment below, I think we may actually need to leave this line in place to satisfy the reference in `@throws` (phpcs might complain otherwise, even though it's just in a comment). So CommandException should probably be added as a new line.

src/Operation/ListIndexes.php
tanlisu: Ah I got confused and changed it while I was changing the DriverRuntimeException to CommandException, but now I see I shouldn't have modified it in the @throws. Sorry, it makes sense to me now!

jmikola (approved):
Test failures on MongoDB latest will be addressed by https://github.com/mongodb/mongo-php-library/pull/834.

LGTM.

PR #838, opened 2021-07-02
PHPLIB-674 Revise documentation for installing coding standard

alcaeus:
PHPLIB-674

The version of doctrine/coding-standard and phpcs itself were not compatible with PHP 8, which is why @tanlisu wasn't able to run it locally.

Having dropped support for PHP 7.0, we can also remove the manual work required and add a dev requirement for doctrine/coding-standard. This was previously not possible as there were no versions of the coding standard compatible with PHP 7.1.

As a side effect of the upgrade, a few new sniffs have been introduced. Changes were made in separate commits, which makes it easy to drop certain changes if we don't want to have them. A particularly noisy one is return type spacing, which changed with PSR-12. If there are no objections, I'd like to be compatible with PSR-12, hence the changes.

Commits are either `phpcbf:` if all fixes were done by our tooling, and `manual:` for changes that were made manually.

Changes made as a result of dropping PHP 7.0:
* Typed parameters with a `null` default value now use nullable type hints. This is not a BC break, even on PHP 7.1 since the resulting signature is essentially the same.
* Short list syntax: from `list($foo, $bar) = $someArray` to `[$foo, $bar] = $someArray`
* Class constant visibility: all class constants are now explicitly marked as `public` instead of implicitly being public

Other highlights:
* Tests now require native type hints. Since tests are not part of the API, we don't need to worry about BC breaks there
* All methods in the `src` directory must have some documentation for their type (either native or via phpdoc). I've added missing docblocks, which mainly concerned `Explainable::getCommandDocument`

@jmikola please feel free to point out any changes that you're not comfortable with. I'm crossing my fingers for the build to be stable. I suggest "reviewing" the commits separately for convenience.

@tanlisu could you please check out this branch and see if the tooling works for you? I've tested it in PHP 8 and was able to run phpcs successfully, so I hope the problems are solved on your machine as well.


tanlisu:
I was able to run phpcs so it worked for me!

src/Operation/Find.php
```diff
@@ -347,13 +347,22 @@ public function execute(Server $server)
...
+     */
     public function getCommandDocument(Server $server)
     {
         return $this->createCommandDocument();
     }
 
     /**
      * Construct a command document for Find
+     *
+     * @return array
```
jmikola, line 365: Was this preferable to adding an `array` return type hint?

I realize we probably can't do that for `Explainable::getCommandDocument` since it's an interface method, but this is private.

src/GridFS/WritableStream.php
```diff
@@ -262,6 +262,9 @@ private function abort()
         $this->isClosed = true;
     }
 
+    /**
+     * @return mixed
+     */
```
jmikola, line 267: In this case, I expect a doc block is required since there is no return type hint equivalent until [PHP 8.0](https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.mixed), which is also where union types are introduced.

src/functions.php
```diff
@@ -256,6 +256,13 @@ function server_supports_feature(Server $server, $feature)
     return $minWireVersion <= $feature && $maxWireVersion >= $feature;
 }
 
+/**
+ * Return whether the input is an array of strings.
+ *
+ * @internal
+ * @param array $input
+ * @return boolean
```
jmikola, line 264: Technically, `$input` is mixed although we expect it to be an array if this function is to return `true`.

If all of these functions are internal, is there any reason not to use return type hints?

jmikola (commented):
Reviewed all commits except for those prefixed with "phpcbf"

src/Operation/Find.php
alcaeus: There might be more instances like this, but I've fixed this one already. Usually, the type sniffs would suggest using a native return type over `@return` in this case, and even fix it automatically. However, I have disabled this sniff in the `src` folder because of the BC breaks it would cause in public methods. If I were able to only target private members, we could fix this automatically.

Note that with PHP 8.1, the new `#[ReturnTypeWillChange]` attribute allows developers to mark methods that will have a return type in a future major release (to indicate that downstream classes should add the type before then). As part of our 8.1 compatibility effort, we will most likely be annotating those methods with the attribute above, at which point we can also add a suppression for that coding standards violation and turn on the automatic sniff, which will result in these being fixed automatically.

src/GridFS/WritableStream.php
alcaeus: That is correct 👍

src/functions.php
alcaeus: Good point. I've added native types for all internal functions (`with_transaction`) was excluded.

src/functions.php
```diff
@@ -72,7 +72,7 @@ function apply_type_map_to_document($document, array $typeMap)
  * @return string
  * @throws InvalidArgumentException
  */
-function generate_index_name($document)
+function generate_index_name($document): string
```
jmikola, line 75: I think you'll need one more pass with `phpcs`. I recall return type hints requiring a space on either side of the colon.

jmikola (approved):
LGTM after a final pass with `phpcs`.

src/functions.php
alcaeus: That was changed to be in line with PSR-12, which prohibits this. (I've also had too many discussions about that rule back in the day, which is why I took the easy way of using PSR-12 instead)

PR #840, opened 2021-07-07
PHPLIB-518: Provide MongoDB\Collection::rename() method to rename a collection

tanlisu:
https://jira.mongodb.org/browse/PHPLIB-518


docs/reference/method/MongoDBCollection-renameCollection.txt
```diff
@@ -0,0 +1,81 @@
...
+     ["storage":"ArrayObject":private]=>
+     array(3) {
+       ["ns"]=>
+       string(16) "test.restaurants"
+       ["nIndexesWas"]=>
+       int(3)
+       ["ok"]=>
+       float(1)
+     }
+   }
```
tanlisu, line 74: copied from `MongoDBCollection-dropCollection.txt`, I'm not sure what the correct output should be yet

tests/Operation/RenameCollectionFunctionalTest.php
```diff
@@ -0,0 +1,151 @@
...
+use function version_compare;
+
+class RenameCollectionFunctionalTest extends FunctionalTestCase
+{
+    public function testDefaultWriteConcernIsOmitted()
+    {
+        (new CommandObserver())->observe(
+            function () {
+                $operation = new RenameCollection(
+                    'admin',
```
tanlisu, line 22: I was originally using `$this->getDatabaseName()` for the `$databaseName` parameter, but it gave the error `MongoDB\Driver\Exception\CommandException: renameCollection may only be run against the admin database` so I changed all the tests to use the `'admin'` database - not sure if that was the right thing to do.

tests/Operation/RenameCollectionFunctionalTest.php
```diff
@@ -0,0 +1,151 @@
...
+        $server = $this->getPrimaryServer();
+
+        $insertOne = new InsertOne('admin', $this->getCollectionName(), ['x' => 1]);
+        $writeResult = $insertOne->execute($server);
+        $this->assertEquals(1, $writeResult->getInsertedCount());
+
+        $operation = new RenameCollection('admin', $this->getCollectionName(), $renamedCollectionName);
+        $commandResult = $operation->execute($server);
+
+        $this->assertCommandSucceeded($commandResult);
```
tanlisu, line 49: The tests aren’t passing now because the assertion in this line fails; I printed the `$commandResult` object for the error message but am not sure what’s wrong still.

Screen Shot 2021-07-07 at 4 46 25 PM

tests/Operation/RenameCollectionFunctionalTest.php
jmikola: The [server manual]((https://docs.mongodb.com/manual/reference/command/renameCollection/) does state that the command itself must be run against the `admin` database:

> Issue the renameCollection command against the admin database.

However, that shouldn't limit the source and target namespaces involved in the renaming.

src/Operation/RenameCollection.php
```diff
@@ -0,0 +1,196 @@
...
+            throw UnsupportedException::writeConcernNotSupportedInTransaction();
+        }
+
+        $command = new Command([
+            'renameCollection' => $this->collectionName,
+            'to' => $this->newCollectionName,
+        ]);
+
+        try {
+            $cursor = $server->executeWriteCommand($this->databaseName, $command, $this->createOptions());
```
jmikola, line 150: This where you should use `admin`.

docs/reference/method/MongoDBCollection-rename.txt
```diff
@@ -0,0 +1,77 @@
...
+
+The output would then resemble::
+
+   object(MongoDB\Model\BSONDocument)#9 (1) {
+     ["storage":"ArrayObject":private]=>
+     array(1) {
+       ["ok"]=>
+       float(1)
+     }
+   }
```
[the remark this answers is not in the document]
tanlisu: Yup, the only fields in the `rename` response besides "ok" were "$clusterTime" and "operationTime", but I saw that those were omitted in `MongoDBCollection-drop.txt` so I left them out here as well.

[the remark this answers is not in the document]
jmikola: Indeed, those are generic fields that the server adds to everything (re: [causal consistency](https://github.com/mongodb/specifications/blob/master/source/causal-consistency/causal-consistency.rst)), so it's reasonable to omit them from our examples.

tests/Operation/RenameCollectionFunctionalTest.php
```diff
@@ -0,0 +1,131 @@
...
+                $this->assertObjectHasAttribute('lsid', $event['started']->getCommand());
+            }
+        );
+    }
+
+    public function tearDown(): void
+    {
+        if ($this->hasFailed()) {
+            return;
+        }
```
jmikola, line 124: This was presumably copied from `MongoDB\Tests\Operation\FunctionalTestCase::tearDown()`. I think this behavior is actually incorrect as it means we'll never call the `tearDown()` method of the top-level MongoDB\Tests\FunctionalTestCase` class, which disables fail points.

I opened [PHPLIB-693](https://jira.mongodb.org/browse/PHPLIB-693) to track that, so this can remain as-is for now.

tests/Operation/RenameCollectionFunctionalTest.php
```diff
@@ -0,0 +1,131 @@
...
+use MongoDB\Driver\Exception\CommandException;
+use MongoDB\Operation\DropCollection;
+use MongoDB\Operation\FindOne;
+use MongoDB\Operation\InsertOne;
+use MongoDB\Operation\RenameCollection;
+use MongoDB\Tests\CommandObserver;
+
+use function version_compare;
+
+class RenameCollectionFunctionalTest extends FunctionalTestCase
```
jmikola, line 14: I think this file would do well to have a test for the `dropTarget` option. Perhaps two cases:

* Rename an existing collection when the target already exists and expect a server error of some kind.
* Repeat the first test but specify `dropTarget` as true and expect success.

docs/reference/method/MongoDBDatabase-renameCollection.txt
```diff
@@ -0,0 +1,76 @@
...
+Definition
+----------
+
+.. phpmethod:: MongoDB\\Database::renameCollection()
+
+   Changes the name of an existing collection.
+
+   .. code-block:: php
+
+      function renameCollection($fromNamespace, $toNamespace, array $options = []): array|object
```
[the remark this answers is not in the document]
tanlisu: I don't think it works to determine whether a collection name or namespace is being provided by checking whether the argument contains a dot, since the collection names used in the tests contain a dot as well. So I'm not sure how it would be possible to do this actually - any ideas?

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -13,6 +13,8 @@
 use MongoDB\Exception\UnsupportedException;
 use MongoDB\MapReduceResult;
 use MongoDB\Operation\Count;
+use MongoDB\Operation\DropCollection;
+use MongoDB\Operation\FindOne;
```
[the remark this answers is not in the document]
tanlisu: I was using `DropCollection` and `FindOne` for the renamed collection in `testRename()`, so I don't think I'm able to use the Collection class methods in this case right?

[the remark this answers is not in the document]
jmikola: In a later comment, I suggest creating a Collection instance for the renamed class. That would be taboo in the Operation test classes since those shouldn't concern themselves with the Collection class, but it shouldn't be a concern in CollectionFunctionalTest since we're already calling Collection methods elsewhere in this class.

docs/reference/method/MongoDBDatabase-renameCollection.txt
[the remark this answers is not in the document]
jmikola: Very good point. This would work the other way around (e.g. `$databaseNameOrNamespace`) because database names cannot contain dots; however, it's ambiguous if we're trying to differentiate a namespace from a collection name.

In that case, I think it makes sense to leave `$toNamespace` as-is.

`$fromNamespace` can still be changed to `$collectionName` for consistency with `Database::dropCollection()`, and the helper method can build the namespace to send to the Operation internally.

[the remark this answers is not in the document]
jmikola: To add on to this, let's change the API of both the helpers _and_ Operation class to no longer take namespace strings as paramaters:

```
$database->renameCollection($collectionName, $toDatabaseName, $toCollectionName);
$collection->rename($toDatabaseName, $toCollectionName);
```

The Operation class will then have four required parameters:

 * fromDatabaseName
 * fromCollectionName
 * toDatabaseName
 * toCollectionName

It can then build the namespace internally with string concatenation in the constructor (allowing you to keep the internal private properties that store namespaces).

For the Database helper, fromDatabaseName would be pulled from the database instance's own name. And for the Collection helper, both "from" parameters would be pulled from the collection instance.

tests/Database/DatabaseFunctionalTest.php
```diff
@@ -211,6 +213,29 @@ public function testModifyCollection(): void
...
+        $bulkWrite = new BulkWrite();
+        $bulkWrite->insert(['_id' => 1, 'x' => 'foo']);
+
+        $writeResult = $this->manager->executeBulkWrite($this->getNamespace(), $bulkWrite);
+        $this->assertEquals(1, $writeResult->getInsertedCount());
+
+        $commandResult = $this->database->renameCollection($this->getNamespace(), $renamedNamespace);
+        $this->assertCommandSucceeded($commandResult);
+        $this->assertCollectionCount($this->getNamespace(), 0);
+        $this->assertCollectionCount($renamedNamespace, 1);
```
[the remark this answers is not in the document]
tanlisu: Oh I used `$this->assertCollectionCount()` because I saw that the `testDrop()` function used it, but I realize now that function is testing dropping a database, not a collection. I'll change `testRenameCollection()` to use the existence assertions.

src/Operation/RenameCollection.php
```diff
@@ -0,0 +1,170 @@
...
+    public function execute(Server $server)
+    {
+        if (isset($this->options['writeConcern']) && ! server_supports_feature($server, self::$wireVersionForWriteConcern)) {
+            throw UnsupportedException::writeConcernNotSupported();
+        }
+
+        $inTransaction = isset($this->options['session']) && $this->options['session']->isInTransaction();
+        if ($inTransaction && isset($this->options['writeConcern'])) {
+            throw UnsupportedException::writeConcernNotSupportedInTransaction();
+        }
```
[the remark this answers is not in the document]
alcaeus: Defensive coding practices would suggest leaving this in. Creating a drivers ticket to check for this makes sense to me.

tests/FunctionalTestCase.php
```diff
@@ -147,6 +150,78 @@ protected function assertCollectionCount($namespace, $count): void
...
+     * Asserts that a collection with the given name exists on the server.
+     *
+     * If $databaseName is not specified, the current database will be used.
+     * An optional $callback may be provided, which should take a CollectionInfo
+     * argument as its first and only parameter. If a CollectionInfo matching
+     * the given name is found, it will be passed to the callback, which may
+     * perform additional assertions.
+     *
+     * @param string   $collectionName
+     * @param string   $databaseName
```
jmikola, line 193: Pending @alcaeus' input, this can be changed to either `?string` or `string|null`.

tests/Operation/RenameCollectionFunctionalTest.php
```diff
@@ -0,0 +1,168 @@
...
+            $this->getDatabaseName(),
+            $this->getCollectionName(),
+            $this->getDatabaseName(),
+            $this->toCollectionName
+        );
+        $commandResult = $operation->execute($server);
+    }
+
+    /**
+     * @depends testRenameExistingCollection
```
jmikola, line 121: Why is [`@depends`](https://phpunit.readthedocs.io/en/9.5/annotations.html#depends) used here? This doesn't seem like it would depend on `testRenameExistingCollection`.

tests/Operation/RenameCollectionFunctionalTest.php
```diff
@@ -0,0 +1,168 @@
...
+     */
+    public function testRenameNonexistentCollection(): void
+    {
+        $this->assertCollectionDoesNotExist($this->getNamespace());
```
jmikola, line 125: I think this is redundant and can be removed. The source collection is dropped by the parent `setUp()` method in `MongoDB\Tests\Operation\FunctionalTestCase`, which will run before each test.

If the collection actually existed and wasn't dropped, then we'd expect this test to fail anyway because the server would no longer throw a NamespaceNotFound(26) error.

tests/Operation/RenameCollectionFunctionalTest.php
```diff
@@ -0,0 +1,168 @@
...
+
+        $this->expectException(CommandException::class);
+        $this->expectExceptionCode(self::$errorCodeNamespaceNotFound);
+        $operation = new RenameCollection(
+            $this->getDatabaseName(),
+            $this->getCollectionName(),
+            $this->getDatabaseName(),
+            $this->toCollectionName,
+            ['writeConcern' => $this->createDefaultWriteConcern()]
```
jmikola, line 134: This looks like it was copied from `testDefaultWriteConcernIsOmitted()`, but it's not relevant to this test. I suggest removing it, since it won't be sent in the command anyway.

tests/Operation/RenameCollectionFunctionalTest.php
```diff
@@ -0,0 +1,168 @@
...
+        );
+        $commandResult = $operation->execute($server);
+
+        $this->assertCommandSucceeded($commandResult);
+        $this->assertCollectionDoesNotExist($this->getCollectionName());
+        $this->assertCollectionExists($this->toCollectionName);
+
+        $operation = new FindOne($this->getDatabaseName(), $this->toCollectionName, []);
+        $cursor = $operation->execute($server);
+        $this->assertSameDocument(['_id' => 1], $cursor);
```
jmikola, line 94:
```suggestion
        $this->assertSameDocument(['_id' => 1], $operation->execute($server));
```

Remember: `FindOne::execute()` returns a document, not a cursor.

tests/Operation/RenameCollectionFunctionalTest.php
```diff
@@ -0,0 +1,168 @@
...
+
+                $operation->execute($server);
+            },
+            function (array $event): void {
+                $this->assertObjectNotHasAttribute('writeConcern', $event['started']->getCommand());
+            }
+        );
+    }
+
+    public function testRenameExistingCollection(): void
```
jmikola, line 72:
```suggestion
    public function testRenameCollectionToNonexistentTarget(): void
```

A nonexistent source collection is an exception, and covered in a later test that expects an error. The successful path involves a collection that _does_ exist, so I don't think we need to clarify that. The distinguishing characteristic of this test is that the target does not exist.

tests/Operation/RenameCollectionFunctionalTest.php
```diff
@@ -0,0 +1,168 @@
...
+        $this->assertCommandSucceeded($commandResult);
+        $this->assertCollectionDoesNotExist($this->getCollectionName());
+        $this->assertCollectionExists($this->toCollectionName);
+
+        $operation = new FindOne($this->getDatabaseName(), $this->toCollectionName, []);
+        $cursor = $operation->execute($server);
+        $this->assertSameDocument(['_id' => 1], $cursor);
+    }
+
+    public function testRenameExistingCollectionExistingTarget(): void
```
jmikola, line 97:
```suggestion
    public function testRenameCollectionExistingTarget(): void
```

tests/Operation/RenameCollectionFunctionalTest.php
```diff
@@ -0,0 +1,168 @@
...
+
+        $this->expectException(CommandException::class);
+        $this->expectExceptionCode(self::$errorCodeNamespaceExists);
+        $operation = new RenameCollection(
+            $this->getDatabaseName(),
+            $this->getCollectionName(),
+            $this->getDatabaseName(),
+            $this->toCollectionName
+        );
+        $commandResult = $operation->execute($server);
```
jmikola, line 117:
```suggestion
        $operation->execute($server);
```

Even if the operation doesn't throw for some reason (and the test ultimately fails), this return value is never used. Consider what was done in the `testDefaultWriteConcernIsOmitted` and session tests.

tests/Operation/RenameCollectionFunctionalTest.php
```diff
@@ -0,0 +1,168 @@
...
+        $this->expectException(CommandException::class);
+        $this->expectExceptionCode(self::$errorCodeNamespaceNotFound);
+        $operation = new RenameCollection(
+            $this->getDatabaseName(),
+            $this->getCollectionName(),
+            $this->getDatabaseName(),
+            $this->toCollectionName,
+            ['writeConcern' => $this->createDefaultWriteConcern()]
+        );
+        $commandResult = $operation->execute($this->getPrimaryServer());
```
jmikola, line 136:
```suggestion
        $operation->execute($this->getPrimaryServer());
```

jmikola (commented):
Reviewed the code files but skipped the docs on this round. I'll come back to those on Friday and do a local build to check for any RST errors.

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -332,6 +333,50 @@ public function testFindWithinTransaction(): void
...
+
+        $document = $toCollection->findOne();
+        $this->assertSameDocument(['_id' => 1], $document);
+        $toCollection->drop();
+    }
+
+    public function testRenameToDifferentDatabase(): void
+    {
+        if ($this->isShardedCluster()) {
+            $this->markTestSkipped('Test does not apply on sharded clusters: need source and target databases to be on the same primary shard.');
```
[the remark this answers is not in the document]
tanlisu: I tried creating the target database explicitly as you suggested; the PHPUnit tests with sharded cluster seem to be working, but the evergreen sharded cluster test doesn't. Do you know what might be wrong or should I just mark the the test as skipped for now?

src/Operation/RenameCollection.php
```diff
@@ -0,0 +1,168 @@
...
+     *    renaming the collection.
+     *
+     * @param string $fromDatabaseName   Database name
+     * @param string $fromCollectionName Collection name
+     * @param string $toDatabaseName     New database name
+     * @param string $toCollectionName   New collection name
+     * @param array  $options            Command options
+     * @throws InvalidArgumentException for parameter/option parsing errors
+     */
+    public function __construct($fromDatabaseName, $fromCollectionName, $toDatabaseName, $toCollectionName, array $options = [])
```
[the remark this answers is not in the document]
alcaeus: Yes, any new API should be as strictly typed as possible with the current PHP version constraints (scalar types, nullable types, `void` allowed, `object`, property types, union types, `mixed` not allowed).

tests/FunctionalTestCase.php
```diff
@@ -147,6 +150,78 @@ protected function assertCollectionCount($namespace, $count): void
...
     }
 
+    /**
+     * Asserts that a collection with the given name does not exist on the
+     * server.
+     *
+     * If $databaseName is not specified, the current database will be used.
+     *
+     * @param string $collectionName
+     * @param string $databaseName
```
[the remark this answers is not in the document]
alcaeus: I believe PHPDoc understands `?string`, but since that syntax was only introduced with PHP 7.1, many people still prefer `string|null`. I have no preference for either since I prefer using an argument type directly and skipping the `@param` annotation since the type information is already in the signature.

src/Collection.php
```diff
@@ -1004,6 +1005,39 @@ public function mapReduce(JavascriptInterface $map, JavascriptInterface $reduce,
         return $operation->execute($server);
     }
 
+    /**
+     * Renames the collection.
+     *
+     * @see RenameCollection::__construct() for supported options
+     * @param string $toCollectionName New name of the collection
+     * @param string $toDatabaseName   New database name of the collection
```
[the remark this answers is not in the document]
alcaeus: In this case the parameter should be documented as nullable. As for the rest of the documentation, I'll defer to your judgement, it's better than mine in 99.99999% of all cases.

src/Operation/RenameCollection.php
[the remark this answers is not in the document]
jmikola: @tanlisu please apply these changes.

src/Collection.php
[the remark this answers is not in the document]
jmikola: @tanlisu Please change type to `?string` and apply the following suggested change to the description.

src/Database.php
```diff
@@ -470,6 +471,40 @@ public function modifyCollection($collectionName, array $collectionOptions, arra
         return $operation->execute($server);
     }
 
+    /**
+     * Rename a collection within this database.
+     *
+     * @see RenameCollection::__construct() for supported options
+     * @param string $fromCollectionName Collection name
+     * @param string $toCollectionName   New name of the collection
+     * @param string $toDatabaseName     New database name of the collection
```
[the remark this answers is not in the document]
jmikola: @tanlisu Please change type to ?string and apply the following suggested change to the description.

tests/FunctionalTestCase.php
jmikola: @tanlisu: Let's just delete both of these `@param` lines.

tests/Collection/CollectionFunctionalTest.php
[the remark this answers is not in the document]
jmikola: > PHPUnit tests with sharded cluster seem to be working, but the evergreen sharded cluster test doesn't

What do you mean by "PHPUnit tests with sharded cluster"? Were you running a sharded cluster locally?

docs/includes/apiargs-MongoDBCollection-method-rename-param.yaml
```diff
@@ -0,0 +1,28 @@
+arg_name: param
+name: $toCollectionName
+type: string
+description: |
+  The new name of the collection. If a full namespace is not
+  specified, the database of the original collection will be used. If the new
+  namespace specifies a different database, the command copies the collection
+  to the new database and drops the source collection.
```
jmikola, line 8: This text is leftover from when we had a single parameter for the collection name or namespace. I suggest replacing it with:

> The new name of the collection.

jmikola (commented):
Built the documentation locally and didn't see any syntax errors. Left one suggestion to reword, and then some other suggestions for outstanding comments on code docs.

docs/reference/method/MongoDBCollection-rename.txt
```diff
@@ -0,0 +1,79 @@
+=============================
+MongoDB\\Collection::rename()
+=============================
+
+.. versionadded:: 1.9
```
jmikola, line 5:
```suggestion
.. versionadded:: 1.10
```

docs/reference/method/MongoDBDatabase-renameCollection.txt
```diff
@@ -0,0 +1,79 @@
+=====================================
+MongoDB\\Database::renameCollection()
+=====================================
+
+.. versionadded:: 1.9
```
jmikola, line 5:
```suggestion
.. versionadded:: 1.10
```

tests/Collection/CollectionFunctionalTest.php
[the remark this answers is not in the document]
tanlisu: > Are you referring to the GitHub CI job that uses a sharded cluster (https://github.com/mongodb/mongo-php-library/runs/3125307713)?

Yes, that's what I meant. Got it!

tests/FunctionalTestCase.php
```diff
@@ -186,8 +186,6 @@ protected function assertCollectionDoesNotExist(string $collectionName, ?string
      * the given name is found, it will be passed to the callback, which may
      * perform additional assertions.
      *
-     * @param string   $collectionName
-     * @param string   $databaseName   Defaults to TestCase::getDatabaseName()
      * @param callable $callback
```
jmikola, line 189: Please remove the `@param` line for `$callback` as well. It adds nothing of value.

tests/FunctionalTestCase.php
```diff
@@ -147,6 +150,73 @@ protected function assertCollectionCount($namespace, $count): void
         $this->assertEquals($count, $document['n']);
     }
 
+    /**
+     * Asserts that a collection with the given name does not exist on the
+     * server.
+     *
+     * @param string $collectionName
+     * @param string $databaseName   Defaults to TestCase::getDatabaseName()
```
jmikola, line 158:
```suggestion
     * $databaseName defaults to TestCase::getDatabaseName() if unspecified.
```

tests/FunctionalTestCase.php
```diff
@@ -147,6 +150,73 @@ protected function assertCollectionCount($namespace, $count): void
...
+                break;
+            }
+        }
+
+        $this->assertNull($foundCollection, sprintf('Collection %s exists', $collectionName));
+    }
+
+    /**
+     * Asserts that a collection with the given name exists on the server.
+     *
```
jmikola, line 183:
```suggestion
     *
     * $databaseName defaults to TestCase::getDatabaseName() if unspecified.
```

jmikola (approved):
Tests appear to be passing, which confirms that the earlier issue was unrelated.

LGTM after a few doc block suggestions.

jmikola:
When squashing and merging this, I'd suggest the following description.

> PHPLIB-518: RenameCollection operation
>
> Adds Database::renameCollection() and Collection::rename() methods.
>
> Move assertCollectionDoesNotExist() and assertCollectionExists() to FunctionalTestCase

Most of the other commit messages shouldn't be relevant, and this summarizes the actual changes we might want to look back on down the line. You can also rename the PR itself to "PHPLIB-518: RenameCollection operation", which should suggest that title when squashing.

PR #844, opened 2021-07-13
PHPLIB-673: Fix libmongocrypt detection logic in test suite

tanlisu:
https://jira.mongodb.org/browse/PHPLIB-673

### DIFF
--- a/tests/FunctionalTestCase.php
+++ b/tests/FunctionalTestCase.php
@@ -406,7 +406,7 @@ abstract class FunctionalTestCase extends TestCase
             $this->markTestSkipped('Client Side Encryption only supported on FCV 4.2 or higher');
         }
 
-        if ($this->getModuleInfo('libmongocrypt') === 'disabled') {
+        if ($this->getModuleInfo('libmongocrypt crypto') === 'disabled') {
             $this->markTestSkipped('Client Side Encryption is not enabled in the MongoDB extension');
         }
     }


tests/FunctionalTestCase.php
```diff
@@ -406,7 +406,7 @@ protected function skipIfClientSideEncryptionIsNotSupported(): void
             $this->markTestSkipped('Client Side Encryption only supported on FCV 4.2 or higher');
         }
 
-        if ($this->getModuleInfo('libmongocrypt') === 'disabled') {
+        if ($this->getModuleInfo('libmongocrypt crypto') === 'disabled') {
```
jmikola, line 409: Did you test this when building the extension with the `--with-mongodb-client-side-encryption=no` option for `configure`?

The logic for `phpinfo()` output in the extension is in [`PHP_MINFO_FUNCTION(mongodb)`](https://github.com/mongodb/mongo-php-driver/blob/1.10.0/php_phongo.c#L3940). "libmongocrypt crypto" is only printed as "disabled" if libmongocrypt itself is available, but crypto is not. That is only possible if we're using libmongocrypt as a system library (assuming it can be compiled as such), since we require crypto when compiling the bundled version (per [CheckSSL.m4](https://github.com/mongodb/mongo-php-driver/blob/1.10.0/scripts/autotools/libmongocrypt/CheckSSL.m4)). But if libmongocrypt is not available _at all_ (i.e. driver configured with `--with-mongodb-client-side-encryption=no`), I don't think you'd even see "libmongocrypt crypto" in the output.

I just rebuilt the extension with `--with-mongodb-client-side-encryption=no` and can confirm that the original code is skipping as expected. I'm don't recall what was happening when I originally reported [PHPLIB-673](https://jira.mongodb.org/browse/PHPLIB-673) but it doesn't look like I can reproduce that.

I suggest we just close the ticket with no changes needed.

tanlisu: Ok, got it. I did test with `--with-mongodb-client-side-encryption=no` but just realized it should still be "libmongocrypt" and not "libmongocrypt crypto". When I build with `--with-mongodb-client-side-encryption=yes` the tests skip in other elsewhere which makes sense.

PR #847, opened 2021-07-26
PHPLIB-682: Add test for Database::dropCollection()

tanlisu:
https://jira.mongodb.org/browse/PHPLIB-682

### DIFF
--- a/tests/Database/DatabaseFunctionalTest.php
+++ b/tests/Database/DatabaseFunctionalTest.php
@@ -159,6 +159,19 @@ class DatabaseFunctionalTest extends FunctionalTestCase
         $this->assertCollectionCount($this->getNamespace(), 0);
     }
 
+    public function testDropCollection(): void
+    {
+        $bulkWrite = new BulkWrite();
+        $bulkWrite->insert(['x' => 1]);
+
+        $writeResult = $this->manager->executeBulkWrite($this->getNamespace(), $bulkWrite);
+        $this->assertEquals(1, $writeResult->getInsertedCount());
+
+        $commandResult = $this->database->dropCollection($this->getCollectionName());
+        $this->assertCommandSucceeded($commandResult);
+        $this->assertCollectionCount($this->getNamespace(), 0);
+    }
+
     public function testGetSelectsCollectionAndInheritsOptions(): void
     {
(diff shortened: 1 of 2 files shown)


tests/Database/DatabaseFunctionalTest.php
```diff
@@ -158,6 +158,19 @@ public function testDrop(): void
...
+    {
+        $bulkWrite = new BulkWrite();
+        $bulkWrite->insert(['x' => 1]);
+
+        $writeResult = $this->manager->executeBulkWrite($this->getNamespace(), $bulkWrite);
+        $this->assertEquals(1, $writeResult->getInsertedCount());
+
+        $commandResult = $this->database->dropCollection($this->getCollectionName());
+        $this->assertCommandSucceeded($commandResult);
+        $this->assertCollectionCount($this->getNamespace(), 0);
```
jmikola, line 171: Once #840 is merged, please change this use the `assertCollectionDoesNotExist()` method that was moved to the base class. Asserting that no collections exist in the namespace is not reliable. This is why DropCollectionFunctionalTest originally had its own `assertCollectionDoesNotExist` method.

I assume you might have copied this from `testDrop`. That case is different as we're dropping the entire database and should expect no collections to exist within it after the fact.

tanlisu: Ah my bad, I will fix that again

tanlisu: I noticed that `assertCollectionCount()` is still being used in the `testDrop()` of `CollectionFunctionalTest.php`. Should I create a new ticket to fix that?

https://github.com/mongodb/mongo-php-library/blob/722679fb633a405f2f127f0498627e755b57b757/tests/Collection/CollectionFunctionalTest.php#L267

jmikola: Ah, I didn't realize it was also being used there (dates back to https://github.com/mongodb/mongo-php-library/commit/c520df8acdf448e62794c77873cadb6c66702818#diff-1ab446f85499e8289fa4a8373d33f1aaadc5ffc63f5561dd6335342ef000b4d2R19). I originally assumed this was lifted from `DatabaseFunctionalTest::testDrop`.

Please do change that to `assertCollectionCount`.

jmikola (approved):

PR #854, opened 2021-08-11
PHPLIB-683 Test against Proxy as a Mongos

alcaeus:
PHPLIB-683

I've decided to vendor in the setup scripts for better local testing. For now this works, but I'd like to figure out a better alternative in the long-term to avoid having to duplicate changes made in drivers-evergreen-tools.


.evergreen/config.yml
```diff
@@ -199,6 +199,38 @@ functions:
...
+        working_dir: "src"
+        shell: bash
+        script: |-
+          ${PREPARE_SHELL}
+          export PROJECT=${PROJECT}
+          export SERVERLESS_DRIVERS_GROUP=${SERVERLESS_DRIVERS_GROUP}
+          export SERVERLESS_API_PUBLIC_KEY=${SERVERLESS_API_PUBLIC_KEY}
+          export SERVERLESS_API_PRIVATE_KEY=${SERVERLESS_API_PRIVATE_KEY}
+          export SERVERLESS_ATLAS_USER=${SERVERLESS_ATLAS_USER}
+          export SERVERLESS_ATLAS_PASSWORD=${SERVERLESS_ATLAS_PASSWORD}
```
jmikola, line 214: I assume these environment variables are defined in the Evergreen project config?

.evergreen/config.yml
```diff
@@ -631,3 +686,10 @@ buildvariants:
   run_on: rhel70
   tasks:
     - .versioned-api
+
+- matrix_name: "serverless"
+  matrix_spec: { "php-edge-versions": "latest-stable", "driver-versions": "latest-stable" }
+  display_name: "Serverless"
+  run_on: rhel70
```
jmikola, line 693: Didn't you mention possibly cutting off RHEL testing? Any reason not to use Debian or Ubuntu here?

CONTRIBUTING.md
```diff
@@ -34,7 +34,13 @@ The `phpunit.xml.dist` file is used as the default configuration file for the
...
 `MONGODB_URI` and `MONGODB_DATABASE` environment variables. You may customize
 this configuration by creating your own `phpunit.xml` file based on the
-`phpunit.xml.dist` file we provide.
+`phpunit.xml.dist` file we provide. To run the tests in serverless mode, set the
+`MONGODB_IS_SERVERLESS` environment variable to `on`.
+
+To run tests against a cluster that requires authentication, either include the
+credentials in the connection string given in the `MONGODB_URI` environment
+variable, or set the `MONGODB_USERNAME` and `MONGODB_PASSWORD` environment
+variables accordingly.
```
jmikola, line 43: I assume if credentials are expressed in both forms that the username/password will take precedence given how array options override the connection string. Do you want to mention that here?

I don't think we need to go so far as raising an error in that case (it'd be annoying to have to inspect the connection string to detect this).

tests/Collection/CrudSpecFunctionalTest.php
```diff
@@ -113,6 +123,32 @@ private function assertEquivalentCollections(Collection $expectedCollection, Col
...
 
+    private function checkServerlessRequirement(?string $serverless): void
+    {
+        switch ($serverless) {
+            case null:
+            case self::SERVERLESS_ALLOW:
+                return;
+
+            case self::SERVERLESS_FORBID:
+                if ($this->isServerless()) {
```
jmikola, line 134: I assume `checkServerlessRequirement` was intentionally duplicated in various spec test runners since it's more particular to the test runner, vs. `isServerless` which is more generic (like other methods we have to detect topologies or server versions).

tests/FunctionalTestCase.php
```diff
@@ -385,6 +395,16 @@ protected function isReplicaSet()
...
     }
 
+    /**
+     * Return whether serverless (i.e. proxy as mongos) is being utilized.
+     */
+    protected static function isServerless(): bool
+    {
+        $isServerless = getenv('MONGODB_IS_SERVERLESS');
+
+        return $isServerless !== false ? filter_var($isServerless, FILTER_VALIDATE_BOOLEAN) : false;
```
jmikola, line 405: Do we actually need all of the possible values detected by [`FILTER_VALIDATE_BOOLEAN`](https://www.php.net/manual/en/filter.filters.validate.php)? I don't see the harm in doing this, but just curious if it makes sense to do so vs. being very strict about only supporting particular values that the Evergreen config might specify.

tests/FunctionalTestCase.php
```diff
@@ -513,6 +533,26 @@ protected function skipIfTransactionsAreNotSupported(): void
         }
     }
 
+    private static function appendAuthenticationOptions(array $options): array
+    {
+        if (isset($options['username']) || isset($options['password'])) {
+            return $options;
```
jmikola, line 539: Perhaps it makes sense to raise an exception here, or maybe only if the options are later specified and would try to append themselves?

I reckon there isn't a use case where we'd actually want `MONGODB_USERNAME` or `MONGODB_PASSWORD` to override anything, apart from the previous edge case I mentioned about the connection string already having auth credentials.

tests/Collection/CrudSpecFunctionalTest.php
jmikola: I just realized that CrudSpecFunctionalTest apperas to be unique in being ony of the only test runners (outside of unified and perhaps GridFS) not in the SpecTests directory. That explains why this doesn't share the same method in the base class in SpecTests.

tests/SpecTests/RetryableReadsSpecTest.php
```diff
@@ -14,6 +14,7 @@
  * Retryable reads spec tests.
  *
  * @see https://github.com/mongodb/specifications/tree/master/source/retryable-reads
+ * @group serverless
```
jmikola, line 17: Do these tests actually pass? I assumed there might still be problems given [DRIVERS-1869](https://jira.mongodb.org/browse/DRIVERS-1869).

tests/SpecTests/FunctionalTestCase.php
```diff
@@ -137,8 +141,9 @@ protected function checkServerRequirements(array $runOn): void
             $minServerVersion = $req->minServerVersion ?? null;
             $maxServerVersion = $req->maxServerVersion ?? null;
             $topologies = $req->topology ?? null;
+            $serverlessMode = $req->serverless ?? null;
 
-            if ($this->isServerRequirementSatisifed($minServerVersion, $maxServerVersion, $topologies)) {
+            if ($this->isServerRequirementSatisifed($minServerVersion, $maxServerVersion, $topologies, $serverlessMode)) {
```
jmikola, line 146: Should serverless pre-empt any requirement checking for server version? Feel free to point me to some language in the spec test READMEs that say as much. I don't recall the requirements being mutually exclusive, though.

tests/SpecTests/TransactionsSpecTest.php
```diff
@@ -110,17 +110,43 @@ public static function assertCommandMatches(stdClass $expected, stdClass $actual
...
+     */
+    public function testTransactionsConvenientApi(stdClass $test, ?array $runOn = null, array $data, ?string $databaseName = null, ?string $collectionName = null): void
+    {
+        $this->runTransactionTest($test, $runOn, $data, $databaseName, $collectionName);
+    }
+
+    public function provideTransactionsConvenientApiTests(): array
+    {
+        return $this->provideTests('transactions-convenient-api');
+    }
```
jmikola, line 138: Were these methods necessary to split provisioning of all transaction tests? I assume convenient API tests weren't modified for serverless?

.evergreen/config.yml
alcaeus: That is correct, they are also available through the private design document for local use.

.evergreen/config.yml
alcaeus: None except that we have most on rhel70. I'm currently collecting the operating systems and versions actually used and adapting the build hosts to reflect reality. For the time being, I chose rhel70 based on the config I shamelessly copied and adapted.

tests/Collection/CrudSpecFunctionalTest.php
alcaeus: Yep - I opted against building something resilient or refactoring (we could theoretically use the unified test runner's `RunOnRequirement` class to handle this in all spec test runners), as those tests are migrated to the unified format in less time.

tests/FunctionalTestCase.php
alcaeus: The decision was mostly personal preference, since I can't remember which tools take which formats in their bool environment variables. That's why I opted for something that is as lenient as possible, especially when `on`, `yes`, and `1` all make sense for the behaviour.

tests/FunctionalTestCase.php
alcaeus: This method is called automatically, and only appends default options if the URI being given doesn't already contain them. This allows us to specifically test with invalid credentials on clusters with auth enabled, which would otherwise be impossible (either due to the exception we'd raise, or because we'd always inject the credentials from the environment).

tests/SpecTests/FunctionalTestCase.php
alcaeus: Theoretically, the check should pre-empt any other checks, as it's entirely based on the environment variable being set or not. That said, the unified test format currently prescribes that the requirements MUST be checked in the following order:
1. minServerVersion
2. maxServerVersion
3. topologies
4. serverless
5. serverParameters
6. auth

I don't think this order is efficient or even necessary, so I'm happy to have the spec updated to suggest a more efficient order without mandating it (to avoid unnecessary downstream changes in drivers)

CONTRIBUTING.md
alcaeus: I've added a note that env variables override URI credentials. Thanks for pointing it out, and I agree about not raising an error.

tests/SpecTests/RetryableReadsSpecTest.php
alcaeus: Yes, the pass in the latest [Evergreen Patch](https://spruce.mongodb.com/task/mongo_php_library_serverless__php_edge_versions~latest_stable_driver_versions~latest_stable_test_serverless_patch_60bd25a16ae3b78b08d03b74de05282d2591fd3f_6114cbd13627e074e76eb764_21_08_12_07_20_50/tests).
The last build fails on all non-serverless builds with the same test failure, but I think that is unrelated.

tests/SpecTests/FunctionalTestCase.php
jmikola: Reported https://jira.mongodb.org/browse/DRIVERS-1885

tests/FunctionalTestCase.php
alcaeus: Sorry, I should've clarified. Yes, this does not catch credentials sent through the connection string, which we currently don't do in tests. This ensures that any client created with URI options that set specific credentials (e.g. to test failing authentication) does not get correct credentials injected.

tests/SpecTests/FunctionalTestCase.php
alcaeus: I'd defer that work to a separate PR in an interest to get this merged.

PR #857, opened 2021-08-18
PHPLIB-707: Remove modifiers from command monitoring spec test

jmikola:
https://jira.mongodb.org/browse/PHPLIB-707

Synced with mongodb/specifications@b8928fa15752a9270e39ad2d2a14fd0b3eb0bfbc

### DIFF
--- a/tests/SpecTests/command-monitoring/find.json
+++ b/tests/SpecTests/command-monitoring/find.json
@@ -89,21 +89,19 @@
           "skip": {
             "$numberLong": "2"
           },
-          "modifiers": {
-            "$comment": "test",
-            "$hint": {
-              "_id": 1
-            },
-            "$max": {
-              "_id": 6
-            },
-            "$maxTimeMS": 6000,
-            "$min": {
-              "_id": 0
-            },
-            "$returnKey": false,
-            "$showDiskLoc": false
-          }
+          "comment": "test",
+          "hint": {
+            "_id": 1
+          },
+          "max": {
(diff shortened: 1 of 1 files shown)


tests/SpecTests/command-monitoring/find.json
```diff
@@ -89,21 +89,19 @@
...
+          },
+          "max": {
+            "_id": 6
+          },
+          "maxTimeMS": 6000,
+          "min": {
+            "_id": 0
+          },
+          "returnKey": false,
+          "showRecordId": false
```
jmikola, line 104: For context, this required no code changes in PHPLIB or PHPC since [`Query::__construct()`](https://www.php.net/manual/en/mongodb-driver-query.construct.php) accepts both. The documentation notes that these top-level options fall back to specific options within the `modifiers` option (itself a document) if unspecified. The actual logic for this sits in [`php_phongo_query_init`](https://github.com/mongodb/mongo-php-driver/blob/1.10.0/src/MongoDB/Query.c#L282) and is handled via the various macros combined with `else` statements. Each macro is an `if` condition, which allows us to chain them with `else` to implement the fallback behavior.

As for the spec test runner, it splits the `filter` parameter from `arguments` and sends all of the remaining fields through as the `$options` parameter for `find()`. The logic for that can be found in https://github.com/mongodb/mongo-php-library/blob/1.9.0/tests/SpecTests/Operation.php#L416.

PR #861, opened 2021-09-09
PHPLIB-651: Support $merge and $out executing on secondaries

jmikola:
https://jira.mongodb.org/browse/PHPLIB-651

POC for https://github.com/mongodb/specifications/pull/1062


alcaeus (commented):
LGTM pending approval of https://github.com/mongodb/specifications/pull/1062

src/Collection.php
```diff
@@ -225,11 +228,21 @@ public function aggregate(array $pipeline, array $options = [])
...
+        if (
+            $hasWriteStage && isset($options['readPreference']) &&
+            ! server_supports_feature($server, self::$wireVersionForSecondarySupportsWriteStage)
+        ) {
             $options['readPreference'] = new ReadPreference(ReadPreference::RP_PRIMARY);
-        }
 
-        $server = select_server($this->manager, $options);
+            if ($server->isSecondary()) {
+                $server = select_server($this->manager, $options);
```
jmikola, line 243: When I recently spoke to @kevinAlbs about this (re: https://github.com/mongodb/specifications/pull/1062#discussion_r710555018), I mentioned an alternative implementation that could bypass a second server selection attempt (i.e. `mongoc_client_select_server`) by instead iterating on [`Manager::getServers()`](https://www.php.net/manual/en/mongodb-driver-manager.getservers.php) to select a primary.

Thanks to the first server selection attempt, we can be guaranteed that the topology has been discovered. `getServers()` shouldn't return an empty array unless there really are no servers available. But I also don't think there's much benefit to that approach vs. simply invoking server selection a second time. The edge case is that if there really isn't a primary available, we'll block for `serverSelectionTimeoutMS` on this second attempt.

But if the root concern is potentially blocking the application for `2 * serverSelectionTimeoutMS`, let's consider how that might actually happen:

 * If the first attempt is looking for a secondary to execute an `$out`/`$merge` pipeline and the topology doesn't have a secondary at all, the default scenario will abort early thanks to `serverSelectionTryOnce` and abort the entire `aggregate()` method.
 * If `serverSelectionTryOnce=false` and a loop is permitted, server selection could theoretically discover a secondary just before `serverSelectionTimeoutMS` expires (e.g. at 29 seconds in) and return that server.
 * If we then discover the selected server is a pre-5.0 secondary, we'll re-enter server selection to look for a primary. Worst case, this could also take `serverSelectionTimeoutMS` to ultimately select a primary (e.g. wait through an election) or hit a server selection timeout.

That scenario seems quite rare, so I'm inclined to keep the code as-is rather and punt on any optimization until we actually have evidence that this causes problems down the line. IMO, the `aggregate()` helper is already very convoluted so anything we can do to avoid introducing more complexity is welcomed.

PR #876, opened 2021-12-01
PHPLIB-760: Force primary for $out/$merge if any servers are pre-5.0

jmikola:
https://jira.mongodb.org/browse/PHPLIB-760

### DIFF
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -87,9 +87,6 @@ class Collection
     /** @var integer */
     private static $wireVersionForReadConcernWithWriteStage = 8;
 
-    /** @var integer */
-    private static $wireVersionForSecondarySupportsWriteStage = 13;
-
     /** @var string */
     private $collectionName;
 
@@ -228,21 +225,9 @@ class Collection
             $options['readPreference'] = $this->readPreference;
         }
 
-        $server = select_server($this->manager, $options);
-
-        /* If a write stage is being used with a read preference (explicit or
-         * inherited), check that the wire version supports it. If not, force a
-         * primary read preference and select a new server if necessary. */
-        if (
-            $hasWriteStage && isset($options['readPreference']) &&
-            ! server_supports_feature($server, self::$wireVersionForSecondarySupportsWriteStage)
-        ) {
-            $options['readPreference'] = new ReadPreference(ReadPreference::RP_PRIMARY);
-
-            if ($server->isSecondary()) {
-                $server = select_server($this->manager, $options);
-            }
-        }
+        $server = $hasWriteStage
+            ? select_server_for_aggregate_write_stage($this->manager, $options)
+            : select_server($this->manager, $options);
 
         /* MongoDB 4.2 and later supports a read concern when an $out stage is
          * being used, but earlier versions do not.
--- a/src/functions.php
+++ b/src/functions.php
@@ -19,6 +19,7 @@ namespace MongoDB;
 
 use Exception;
 use MongoDB\BSON\Serializable;
+use MongoDB\Driver\Exception\RuntimeException as DriverRuntimeException;
 use MongoDB\Driver\Manager;
 use MongoDB\Driver\ReadPreference;
 use MongoDB\Driver\Server;
@@ -41,6 +42,32 @@ use function MongoDB\BSON\fromPHP;
 use function MongoDB\BSON\toPHP;
 use function reset;
 use function substr;
+
+/**
+ * Check whether all servers support executing a write stage on a secondary.
+ *
+ * @internal
+ * @param Server[] $servers
+ */
+function all_servers_support_write_stage_on_secondary(array $servers): bool
+{
+    /* Write stages on secondaries are technically supported by FCV 4.4, but the
+     * CRUD spec requires all 5.0+ servers since FCV is not tracked by SDAM. */
+    static $wireVersionForWriteStageOnSecondary = 13;
+
+    foreach ($servers as $server) {
+        // We can assume that load balancers only front 5.0+ servers
+        if ($server->getType() === Server::TYPE_LOAD_BALANCER) {
+            continue;
+        }
+
+        if (! server_supports_feature($server, $wireVersionForWriteStageOnSecondary)) {
+            return false;
+        }
+    }
+
+    return true;
+}
 
 /**
  * Applies a type map to a document.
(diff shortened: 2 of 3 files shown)


src/functions.php
```diff
@@ -42,6 +43,39 @@
...
+        try {
+            $manager->selectServer(new ReadPreference(ReadPreference::NEAREST));
+        } catch (DriverRuntimeException $e) {
+        }
+
+        $servers = $manager->getServers();
+    }
+
+    foreach ($servers as $server) {
+        if (! server_supports_feature($server, $wireVersionForWriteStageOnSecondary)) {
```
jmikola, line 71: For context, this just ensures that the server's minWireVersion is <= 13 and its maxWireVersion is >= 13.

src/Collection.php
```diff
@@ -228,22 +225,18 @@ public function aggregate(array $pipeline, array $options = [])
...
-        /* If a write stage is being used with a read preference (explicit or
-         * inherited), check that the wire version supports it. If not, force a
-         * primary read preference and select a new server if necessary. */
+        /* If a write stage is being used with a non-primary read preference and
+         * at least one server is pre-5.0, force a primary read preference. */
         if (
             $hasWriteStage && isset($options['readPreference']) &&
-            ! server_supports_feature($server, self::$wireVersionForSecondarySupportsWriteStage)
+            $options['readPreference']->getMode() !== ReadPreference::RP_PRIMARY &&
+            ! all_servers_support_write_stage_on_secondary($this->manager)
```
jmikola, line 233: Since the whole point of the spec change ([DRIVERS-1969](https://jira.mongodb.org/browse/DRIVERS-1969)) is to ensure we fall back to a primary read preference, we only need to bother with `all_servers_support_write_stage_on_secondary` if a non-primary read preference was specified (or inherited).

src/functions.php
```diff
@@ -42,6 +43,39 @@
...
+
+    $servers = $manager->getServers();
+
+    /* If no servers are available, assume that SDAM has not been initialized
+     * and make a single server selection attempt to do so now. We do not
+     * actually care about selecting a server so any error can be ignored. */
+    if (empty($servers)) {
+        try {
+            $manager->selectServer(new ReadPreference(ReadPreference::NEAREST));
+        } catch (DriverRuntimeException $e) {
```
jmikola, line 64: A server selection failure typically comes up as a ConnectionTimeoutException, but the extenion's RuntimeException is the base class and also suitable to work with here.

src/functions.php
```diff
@@ -42,6 +43,39 @@
...
+ * Check whether all servers support executing a write stage on a secondary.
+ *
+ * @internal
+ * @see https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#aggregation-pipelines-with-write-stages
+ */
+function all_servers_support_write_stage_on_secondary(Manager $manager): bool
+{
+    static $wireVersionForWriteStageOnSecondary = 13;
+
+    $servers = $manager->getServers();
```
jmikola, line 56: [`Manager::getServers()`](https://github.com/mongodb/mongo-php-driver/blob/1.11.1/src/MongoDB/Manager.c#L671) calls [`mongoc_client_get_server_descriptions`](http://mongoc.org/libmongoc/current/mongoc_client_get_server_descriptions.html), which doesn't initialize the topology. The PHP driver does model all of the various server types reported by libmongoc, so unknown (and other non-usable) servers will get reported here.

I don't believe we need to worry about a case where SDAM is initialized but there are no servers in the topology. The spec describes that as a [terminal condition](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#other-topologytypes), after which the driver would be left in an unusable state.

src/functions.php
kevinAlbs: IIUC I think this should also return true if `$server->getType() == Server::LOAD_BALANCED`. For load balanced, all backing servers are assumed to be >= 5.0.

src/functions.php
```diff
@@ -42,6 +43,39 @@
...
+
+    /* If no servers are available, assume that SDAM has not been initialized
+     * and make a single server selection attempt to do so now. We do not
+     * actually care about selecting a server so any error can be ignored. */
+    if (empty($servers)) {
+        try {
+            $manager->selectServer(new ReadPreference(ReadPreference::NEAREST));
```
kevinAlbs, line 63: IIUC the intent is here to initialize SDAM and do one full scan. `ReadPreference::NEAREST` is arbitrary.

Could this use the user's read preference, and return the selected matching server on success?

I think that could avoid this (rare) scenario:

- MongoDB cluster is: P5 S5 S4
- ReadPreference=Secondary and read preference tags are set to only match S4.
- S4 is unavailable initially. The initial scan keeps S4 as Unknown.
- The S4 becomes available shortly after.
- The next server selection selects waits for matching read preference tags and selects the S4.

kevinAlbs (commented):
Looks good overall! Thank you for the detailed comments. I noted a few things to consider.

src/functions.php
jmikola: Can do. The current language in the CRUD spec was a compromise coming out of https://github.com/mongodb/specifications/pull/1095#discussion_r744793693. My original intention was to have the driver check the `hello` response of the backing server (which we explicitly implemented in libmongoc for the benefit of wrapping drivers); however, that was problematic because the other specs (LB, SDAM) don't currently talk about an approach for doing so. I expect some future spec is going to have to deal with that (e.g. feature requiring 5.2+).

src/functions.php
jmikola: One concern would be if server selection fails and then we realize that all servers are 5.0+. In that case, the exception we ignored is actually what should be thrown. That said, I can just save the exception and rethrow it later in the function.

My original idea was to create a separate `select_server()` function specifically for `$out` and `$merge`, but I backpedaled to this boolean function to keep things simpler. I can try to rework this function by reverting it from a boolean check back to more complex server selection function.

src/functions.php
```diff
@@ -459,3 +486,49 @@ function select_server(Manager $manager, array $options): Server
...
+
+/**
+ * Performs server selection for an aggregate operation with a write stage. The
+ * $options parameter may be modified by reference if a primary read preference
+ * must be forced due to the existence of pre-5.0 servers in the topology.
+ *
+ * @internal
+ * @see https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#aggregation-pipelines-with-write-stages
+ */
+function select_server_for_aggregate_write_stage(Manager $manager, array &$options): Server
```
jmikola, line 498: @kevinAlbs: I changed `all_servers_support_write_stage_on_secondary` to no longer initialize the topology. It now just receives a list of servers directly.

This function is my attempt to incorporate your suggestion from https://github.com/mongodb/mongo-php-library/pull/876#discussion_r761158900.

src/functions.php
```diff
@@ -42,6 +43,32 @@
...
+{
+    /* Write stages on secondaries are technically supported by FCV 4.4, but the
+     * CRUD spec requires all 5.0+ servers since FCV is not tracked by SDAM. */
+    static $wireVersionForWriteStageOnSecondary = 13;
+
+    foreach ($servers as $server) {
+        // We can assume that load balancers only front 5.0+ servers
+        if ($server->getType() === Server::TYPE_LOAD_BALANCER) {
+            continue;
+        }
```
jmikola, line 62: I realize that if there's a load balancer it will be the only server in the topology, so this could also just `return true`. I could also have moved this logic above the call to `all_servers_support_write_stage_on_secondary`, but I was happier to contain this check within the function.

kevinAlbs (approved):
LGTM

src/functions.php
jmikola: I manually tested this approach with a mixed-version replica set:

 * `localhost:27044` (4.4 primary)
 * `localhost:27050` (5.0 secondary)

To create this replica set, I initialized a two-node replica set with 4.4.x and then took the secondary offline and restarted it with a 5.0.x binary. That was necessary to avoid the error described in [this community forum post](https://www.mongodb.com/community/forums/t/addind-replicaset-members-with-different-versions-4-4-6-and-5-0/118579/2), since a replica set initialized with 5.0.x does not allow adding older servers (due to minWireVersion being bumped).

From there, I ran the following script to insert some data fixtures and attempt execution of an `$out` pipeline on a secondary:

```
<?php

require_once 'vendor/autoload.php';

$client = new MongoDB\Client('mongodb://127.0.0.1:27044/?replicaSet=rs0');

$coll = $client->test->foo;
$coll->drop();
$coll->insertMany([['x' => 1], ['x' => 2], ['x' => 3]], ['writeConcern' => new MongoDB\Driver\WriteConcern(2)]);

$cursor = $coll->aggregate(
    [
        ['$match' => ['x' => ['$gt' => 1]]],
        ['$out' => 'bar'],
    ],
    ['readPreference' => new MongoDB\Driver\ReadPreference('secondary')]
);

echo "\nDocs in test.bar:\n";
foreach ($client->test->bar->find([]) as $doc) {
    printf("%s\n", json_encode($doc));
}
```

I also adding some debugging statements to confirm the server on which `aggregate` is executed. With this patch, the 4.4 primary is always selected. Before this patch, the 5.0 secondary would have been selected.

Next, I modified the read preference to include a non-matching tag set to coerce a server selection failure. Before this patch, a server selection exception would have been thrown. With this patch, the 4.4 primary is still selected as the fall back. This addresses the inadvertent BC break in the original implementation, which I alluded to in [DRIVERS-1969](https://jira.mongodb.org/browse/DRIVERS-1969).

jmikola:
Note: the test failure for PHP 7.1 is expected because it uses the PHPC master branch, which dropped support for 7.1 in [PHPC-1880](https://jira.mongodb.org/browse/PHPC-1880).

PR #877, opened 2021-12-04
Allow getting information of write errors

iwalkalone69:
Adding a function to get the array of write errors or an empty array if everything went ok on write operations results:
- BulkWriteResult
- DeleteResult
- InsertManyResult
- InsertOneResult
- UpdateResult
Currently we are able to know if inserts/updates/deletes are done but if all or some of them fails we do not know why. The purpose of this pull request is to be able to access to errors (MongoDB\Driver\WriteError) so we can debug them at least.

### DIFF
--- a/src/BulkWriteResult.php
+++ b/src/BulkWriteResult.php
@@ -189,4 +189,13 @@ class BulkWriteResult
     {
         return $this->isAcknowledged;
     }
+
+    /**
+     * Returns an array of WriteResult with errors ocurred during insert
+     * or empty if there was none
+     */
+    public function getWriteErrors()
+    {
+        return $this->writeResult->getWriteErrors();
+    }
 }
--- a/src/DeleteResult.php
+++ b/src/DeleteResult.php
@@ -67,4 +67,13 @@ class DeleteResult
     {
         return $this->isAcknowledged;
     }
+
+    /**
+     * Returns an array of WriteResult with errors ocurred during insert
+     * or empty if there was none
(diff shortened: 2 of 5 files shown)


jmikola:
IIRC, a [`MongoDB\Driver\WriteResult`](https://www.php.net/manual/en/class.mongodb-driver-writeresult.php) with write or write concern errors will only ever be attached to a [`MongoDB\Driver\Exception\BulkWriteException`](https://www.php.net/manual/en/class.mongodb-driver-exception-bulkwriteexception.php). That happens at the extension level (below this library).

Given that, I would expect that the methods introduced in this PR would always return an empty array. If I'm mistaken, could you share an example script where the library returns one of its result classes and there are actually one or more write errors on underlying WriteResult object?

iwalkalone69:
Yes, you're right. I think it is only useful for BulkWriteResult. If you find it too I can submit it with that only.

jmikola:
> I think it is only useful for BulkWriteResult. If you find it too I can submit it with that only.

I don't believe the WriteResult attached with a BulkWriteResult will ever contain a write or write concern error. If one or more errors are encountered (irrespective of the bulk write's `ordered` option), a BulkWriteException will be thrown by [`phongo_execute_bulk_write`](https://github.com/mongodb/mongo-php-driver/blob/1.11.1/php_phongo.c#L701), which is the common code path for all write operations. This will happen before the library constructs a BulkWriteResult.

PR #879, opened 2021-12-07
PHPLIB-724 : Korean translated versionof READEME , CONTRIBUTING

hjw705:
Upload README-KO.md and CONTRIBUTING-KO.md file,
which are Korean translated.

### DIFF
--- a/CONTRIBUTING-KO.md
+++ b/CONTRIBUTING-KO.md
@@ -1,0 +1,214 @@
+# MongoDB용 PHP 라이브러리에 기여
+
+## 리포지토리 초기화
+라이브러리에 기여하려는 개발자는 해당 라이브러리를 복제하고 [Composer](https://getcomposer.org/)를 사용하여 프로젝트 종속성을 초기화해야 합니다.
+
+```
+$ git clone https://github.com/mongodb/mongo-php-library.git
+$ cd mongo-php-library
+$ composer update
+```
+
+Composer는 프로젝트 종속성을 설치하는 것 외에도 다음을 확인합니다.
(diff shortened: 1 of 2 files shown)


hjw705:
Hello, I'm an undergraduate student in South Korea, Jian Han.
I was interested in MongoDB, while I was learning Database subject at school, and I wanted to participate in open source
contribution.

 I found that there wasn't Korean translated version of README file and CONTRIBUTING file.
I thought  more people can contribute together if there is a translated version. 

I hope  my commit can be checked and examined please.
Thank you.

@jmikola

jmikola:
Thanks for the contribution. I understand this would make the project more accessible to our Korean users, but I'm afraid we don't have the resources to maintain translations directly within the repository. I'd worry that these files would soon become out-of-date and having inaccurate information could be worse than having none at all (or  having users run the English documents through a translation tool).

That said, PHP.net actually has a system in place to support translations. There is presently no Korean translation, but that is something you can propose on the mailing list. [About the Manual: Translations](https://www.php.net/manual/en/about.translations.php) has more information for doing so.

The PHP.net manual also has a much larger audience than our GitHub repositories, so going that route would likely have a bigger impact for Korean users. If you wanted to focus on MongoDB specifically, the [MongoDB Driver](https://www.php.net/mongodb) section of the manual has pages on installing the driver and [using this library](https://www.php.net/manual/en/mongodb.tutorial.library.php), which could certainly be translated. Although contributing to the core PHP language docs would also be a noble cause with greater reach.

hjw705:
Thank you for your kind review. I understood what you meant. Thank you for your advice. 
I might find some other way to contribute to open source. :) Thank you again! 
It was a good experience.

PR #892, opened 2022-02-24
PHPLIB-717: Test serverless with load balancer

jmikola:
https://jira.mongodb.org/browse/PHPLIB-717


.evergreen/config.yml
```diff
@@ -494,10 +490,9 @@ tasks:
       tags: ["serverless"]
       commands:
         - func: "create serverless instance"
-          vars:
-            PROJECT: "mongo-php-library"
         - func: "run serverless tests"
-        - func: "delete serverless instance"
+          vars:
+            MONGODB_URI: "${SINGLE_ATLASPROXY_SERVERLESS_URI}"
```
jmikola, line 495: Like the "test-loadBalanced" task below, this uses the single-server URI exclusively. There is currently no provision to pass a second multi-server URI to the test suite.

Existing handling of single/multi-mongos URIs is handled internally by [FunctionalTestCase::getUri()](https://github.com/mongodb/mongo-php-library/blob/1.11.0/tests/FunctionalTestCase.php#L90), which modifies URI strings directly. That approach does not apply to the serverless and LB environments, since single and multi-server URIs may be entirely different.

 I've opened [PHPLIB-794](https://jira.mongodb.org/browse/PHPLIB-794) to track this task.

PR #896, opened 2022-03-04
PHPLIB-609: Allow CachingIterator to handle Iterators with non-unique…

aleksandr-rudo:
https://jira.mongodb.org/browse/PHPLIB-609

### DIFF
--- a/src/Model/CachingIterator.php
+++ b/src/Model/CachingIterator.php
@@ -40,8 +40,8 @@ use function reset;
  */
 class CachingIterator implements Countable, Iterator
 {
-    private const FIELD_KEY = 'key';
-    private const FIELD_VALUE = 'value';
+    private const FIELD_KEY = 0;
+    private const FIELD_VALUE = 1;
 
     /** @var array */
     private $items = [];
@@ -90,8 +90,8 @@ class CachingIterator implements Countable, Iterator
     #[ReturnTypeWillChange]
     public function current()
     {
-        /* Fetching current item */
         $currentItem = current($this->items);
+
         return $currentItem !== false ? $currentItem[self::FIELD_VALUE] : false;
     }
 
@@ -102,8 +102,8 @@ class CachingIterator implements Countable, Iterator
     #[ReturnTypeWillChange]
     public function key()
     {
-        /* Fetching current item */
         $currentItem = current($this->items);
+
         return $currentItem !== false ? $currentItem[self::FIELD_KEY] : null;
     }
 
@@ -172,7 +172,7 @@ class CachingIterator implements Countable, Iterator
             return;
         }
 
-        /* Storing a new item in the internal cache */
+        // Storing a new item in the internal cache
         $this->items[] = [
             self::FIELD_KEY => $this->iterator->key(),
             self::FIELD_VALUE => $this->iterator->current(),
--- a/src/Model/CachingIterator.php
+++ b/src/Model/CachingIterator.php
@@ -25,7 +25,6 @@ use Traversable;
 
 use function count;
 use function current;
-use function key;
 use function next;
 use function reset;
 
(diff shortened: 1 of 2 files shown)


src/Model/CachingIterator.php
```diff
@@ -40,6 +40,9 @@
  */
 class CachingIterator implements Countable, Iterator
 {
+    private const FIELD_KEY = 'key';
+    private const FIELD_VALUE = 'value';
```
jmikola, line 44: I'd suggest defining `FIELD_KEY` as `0` and `FIELD_VALUE` as `1`. This will allow PHP to use a packed array internally and avoid the overhead of storing key strings for each cached element.

Alternatively, we can do without the constants and just treat `$items` as a tuple assume the key and value are the first and second elements of each item, respectively. That would entail the following change in `storeCurrentItem()`:

```diff
- $this->items[$this->iterator->key()] = $this->iterator->current();
+ $this->items[] = [$this->iterator->key(), $this->iterator->current()];
```

I'll defer to you on whether you prefer to use constants (it's a small class so either approach is fine), but let's definitely avoid the strings.

src/Model/CachingIterator.php
```diff
@@ -87,7 +90,9 @@ public function count()
     #[ReturnTypeWillChange]
     public function current()
     {
-        return current($this->items);
+        /* Fetching current item */
```
jmikola, line 93: For consistency with other files in the project, single-line comments should use `//`. Alternatively, we can just remove these comments entirely as the subsequent line should be self-explanatory.

src/Model/CachingIterator.php
```diff
@@ -87,7 +90,9 @@ public function count()
...
+        $currentItem = current($this->items);
+        return $currentItem !== false ? $currentItem[self::FIELD_VALUE] : false;
```
jmikola, line 95: Coding standards will require a blank line before `return` statements. See: https://github.com/mongodb/mongo-php-library/runs/5471596028?check_suite_focus=true#step:9:1

This didn't show up earlier as I had to approve the workflow run for your first pull request; however, you might have caught this locally by running `vendor/bin/phpcs` (and `vendor/bin/phpcbf` could have fixed it for you).

jmikola (commented):
Additionally, please add a new test case to CachingIteratorTest that shows the class can now support iterators that return duplicate keys. The existing helper methods in that class (e.g. `getTraversable`) might not be usable for that purpose, but you can likely get by with creating a [generator](https://www.php.net/manual/en/language.generators.syntax.php). For example:

```php
$duplicateKeyGenerator = function() {
    yield 0 => 'a';
    yield 0 => 'b';
    yield 0 => 'c';
};
```

tests/Model/CachingIteratorTest.php
```diff
@@ -152,6 +152,12 @@ public function rewind(): void
         $this->assertCount(1, $iterator);
     }
 
+    public function testNonUniqueKeys(): void
```
jmikola, line 155:
```suggestion
    public function testCountNonUniqueKeys(): void
```

I'd also suggest creating `testIterationNonUniqueKeys()` based on `testIteration()`, which removes `$expectedKey` and does `$this->assertSame(0, $key);` during iteration. Since counting the CachingIterator immediately exhausts the internal iterator, the iteration test would cover the code path of the actual iteration methods (e.g. `next()`, `key()`).

jmikola (commented):
One suggestion to rename the count test and add `testIterationNonUniqueKeys`. I think we'll be good after that.

jmikola (approved):
LGTM, but please hold off on merging until I branch v1.12.

The one build failure is unrelated to this change and is due to upstream changes in MongoDB 6.0. I added a comment in [PHPLIB-810](https://jira.mongodb.org/browse/PHPLIB-810) to track it.

PR #898, opened 2022-03-11
PHPLIB-662: Unified test runner should error for unexpected operation arguments

aleksandr-rudo:
https://jira.mongodb.org/browse/PHPLIB-662


tests/UnifiedSpecTests/Operation.php
```diff
@@ -211,6 +211,7 @@ private function execute()
     private function executeForChangeStream(ChangeStream $changeStream)
     {
         $args = $this->prepareArguments();
+        Util::assertParamsBySchema(get_class($changeStream), $this->name, $args);
```
jmikola, line 214: Let's use the `::class` constant instead of `get_class` in all of these contexts.

```suggestion
        Util::assertParamsBySchema(ChangeStream::class, $this->name, $args);
```

I'll leave you to revise the other occurrences.

tests/MethodsParams.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+namespace MongoDB\Tests;
```
jmikola, line 3: This class is only relevant to the Unified Test Runner, so I don't think we need to have this in the top-level `MongoDB\Tests` namespace. That said, the class also has no methods on its own so I'd prefer to just move the array below to a `private static $operationArgs` array in the `MongoDB\Tests\UnifiedSpecTests\Util` class.

tests/UnifiedSpecTests/Util.php
```diff
@@ -33,6 +34,19 @@ public static function assertHasOnlyKeys($arrayOrObject, array $keys): void
         assertEmpty($diff, 'Unsupported keys: ' . implode(',', array_keys($diff)));
     }
 
+    /**
+     * @param string $executingObjectName
+     * @param string $operation
+     * @param array  $args
+     * @return void
+     */
```
jmikola, line 42: This can be removed. I only added it in the [PHPLIB-662 comment](https://jira.mongodb.org/browse/PHPLIB-662?focusedCommentId=4405627&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4405627) to explain the API a bit more.

```suggestion
```

tests/UnifiedSpecTests/Util.php
```diff
@@ -33,6 +34,19 @@ public static function assertHasOnlyKeys($arrayOrObject, array $keys): void
...
+    public static function assertParamsBySchema(string $executingObjectName, string $operation, array $args): void
+    {
+        if (isset(MethodsParams::$args[$executingObjectName][$operation])) {
```
jmikola, line 45: I realize the `isset()` helps since the `$args` array is incomplete, but ultimately I think this should be replaced with an assertion(s) that we have a list of parameter names for the operation:

 * `$args` has the `$executingObjectName` key
 * `$args[$executingObjectName]` is an array
 *  `$args[$executingObjectName]` has the `$operation` key
 * `$args[$executingObjectName][$operation]` is an array

Alternatively, you can use the [null coalescing operator](https://www.php.net/manual/en/language.operators.comparison.php#language.operators.comparison.coalesce). A single assertion that `$args[$executingObjectName][$operation] ?? null` is an array should have the same effect, and be more concise/performant.

tests/MethodsParams.php
```diff
@@ -0,0 +1,45 @@
...
+            'assertDifferentLsidOnLastTwoCommands' => ['client'],
+            'assertNumberConnectionsCheckedOut' => ['connections'],
+            'assertSessionDirty' => ['session'],
+            'assertSessionNotDirty' => ['session'],
+            'assertSessionPinned' => ['session'],
+            'assertSessionTransactionState' => ['session', 'state'],
+            'assertSessionUnpinned' => ['session'],
+            'failPoint' => ['client', 'failPoint', ''],
+            'targetedFailPoint' => ['session', 'failPoint' ],
+//            'loop' => [],
```
jmikola, line 35:
```suggestion
            'failPoint' => ['client', 'failPoint'],
            'targetedFailPoint' => ['session', 'failPoint'],
            'loop' => ['operations', 'storeErrorsAsEntity', 'storeFailuresAsEntity', 'storeSuccessesAsEntity', 'storeIterationsAsEntity'],
```

tests/UnifiedSpecTests/Util.php
```diff
@@ -26,13 +34,113 @@
...
+            'assertSessionNotDirty' => ['session'],
+            'assertSessionPinned' => ['session'],
+            'assertSessionTransactionState' => ['session', 'state'],
+            'assertSessionUnpinned' => ['session'],
+            'failPoint' => ['client', 'failPoint'],
+            'targetedFailPoint' => ['session', 'failPoint'],
+            'loop' => ['operations', 'storeErrorsAsEntity', 'storeFailuresAsEntity', 'storeSuccessesAsEntity', 'storeIterationsAsEntity'],
+        ],
+        Client::class => [
+            'createChangeStream' => ['pipeline', 'typeMap', 'session', 'fullDocument', 'resumeAfter', 'startAfter', 'startAtOperationTime', 'batchSize', 'collation', 'maxAwaitTimeMS'],
```
jmikola, line 59: `typeMap` can be removed from everything. That is a PHP-specific option and will never appear in any spec tests.

tests/UnifiedSpecTests/Util.php
```diff
@@ -26,13 +34,113 @@
...
+            'aggregate' => ['pipeline', 'typeMap', 'session', 'useCursor', 'allowDiskUse', 'batchSize', 'bypassDocumentValidation', 'collation', 'comment', 'explain', 'hint', 'let', 'maxAwaitTimeMS', 'maxTimeMS'],
+            'createChangeStream' => ['pipeline', 'typeMap', 'session', 'fullDocument', 'resumeAfter', 'startAfter', 'startAtOperationTime', 'batchSize', 'collation', 'maxAwaitTimeMS'],
+            'createCollection' => ['collection', 'typeMap', 'session', 'autoIndexId', 'capped', 'collation', 'expireAfterSeconds', 'flags', 'indexOptionDefaults', 'max', 'maxTimeMS', 'size', 'storageEngine', 'timeseries', 'validationAction', 'validationLevel', 'validator'],
+            'dropCollection' => ['collection', 'typeMap', 'session'],
+            'listCollectionNames' => ['authorizedCollections', 'filter', 'maxTimeMS', 'session'],
+            'listCollections' => ['authorizedCollections', 'filter', 'maxTimeMS', 'session'],
+            'runCommand' => [
+                'command', 'typeMap', 'session',
+                'commandName', // Argument is not used in PHPLib
+            ],
```
jmikola, line 73:
```suggestion
            // Note: commandName is not used by PHP
            'runCommand' => ['command', 'typeMap', 'session', 'commandName'],
```

tests/UnifiedSpecTests/Util.php
```diff
@@ -26,13 +34,113 @@
...
+        ],
+        Collection::class => [
+            'aggregate' => ['pipeline', 'typeMap', 'session', 'useCursor', 'allowDiskUse', 'batchSize', 'bypassDocumentValidation', 'collation', 'comment', 'explain', 'hint', 'let', 'maxAwaitTimeMS', 'maxTimeMS'],
+            'bulkWrite' => ['requests', 'session', 'ordered', 'bypassDocumentValidation'],
+            'createChangeStream' => ['pipeline', 'typeMap', 'session', 'fullDocument', 'resumeAfter', 'startAfter', 'startAtOperationTime', 'batchSize', 'collation', 'maxAwaitTimeMS'],
+            'createFindCursor' => ['filter', 'typeMap', 'session', 'allowDiskUse', 'allowPartialResults', 'batchSize', 'collation', 'comment', 'cursorType', 'hint', 'limit', 'max', 'maxAwaitTimeMS', 'maxScan', 'maxTimeMS', 'min', 'modifiers', 'noCursorTimeout', 'oplogReplay', 'projection', 'returnKey', 'showRecordId', 'skip', 'snapshot', 'sort'],
+            'createIndex' => [
+                'keys', 'commitQuorum', 'maxTimeMS', 'session',
+                'name', // Argument is not used in PHPLib
+            ],
```
jmikola, line 83:
```suggestion
            'createIndex' => ['keys', 'commitQuorum', 'maxTimeMS', 'name', 'session'],
```

I misspoke when I said that `name` was not used. It is a valid option and will be respected by PHP if set. When not set, the name is generated from the `keys` option.

This is documented in [`MongoDB\Collection::createIndex()`](https://docs.mongodb.com/php-library/current/reference/method/MongoDBCollection-createIndex/), which is also referenced by [`MongoDB\Collection::createIndexes()`](https://docs.mongodb.com/php-library/current/reference/method/MongoDBCollection-createIndexes/#indexes-parameter).

In the code, you'll find that both `createIndex()` and `createIndexes()` delegate to the CreateIndexes operation. That, in turn, uses the IndexInput class to validate index options. The specific code where `name` is generated if unset can be found [here](https://github.com/mongodb/mongo-php-library/blob/1.11.0/src/Model/IndexInput.php#L66).

Note: index creation is a rather unique case where we have a separate class to validate input arguments, in contrast to other operation classes that handle all of their validation in the constructor (including BulkWrite, which is quite complex).

tests/UnifiedSpecTests/Util.php
```diff
@@ -26,13 +34,113 @@
...
+        ],
+        Cursor::class => [
+            'close' => [],
+            'iterateUntilDocumentOrError' => [],
+        ],
+        Session::class => [
+            'abortTransaction' => [],
+            'commitTransaction' => [],
+            'endSession' => [],
+            'startTransaction' => [],
```
jmikola, line 116:
```suggestion
           'startTransaction' => ['maxCommitTimeMS', 'readConcern', 'readPreference', 'writeConcern'],
```

I realize this is unintuitive, but the [`startTransaction` operation](https://github.com/mongodb/mongo-php-library/blob/1.11.0/tests/UnifiedSpecTests/Operation.php#L614) passes all of its arguments directly to [`MongoDB\Driver\Session::startTransaction()`](https://www.php.net/manual/en/mongodb-driver-session.starttransaction.php) and thus supports the same options of _that_ method.

None of the unified spec tests currently use this argument; however, existing legacy tests do and those will eventually be converted by [DRIVERS-1709](https://jira.mongodb.org/browse/DRIVERS-1709). Since these arguments are already handled by the Operation class already in anticipation of those tests being ported, we should validate them now.

tests/UnifiedSpecTests/Util.php
```diff
@@ -26,13 +34,113 @@
...
+            'withTransaction' => ['callback', 'readConcern', 'writeConcern'],
```
jmikola, line 117:
```suggestion
            'withTransaction' => ['callback', 'maxCommitTimeMS', 'readConcern', 'readPreference', 'writeConcern'],
```

In the test runner's operation class, all arguments other than `callback` end up being passed as the `$transactionOptions` argument to `with_transaction()` (see: [here](https://github.com/mongodb/mongo-php-library/blob/1.11.0/tests/UnifiedSpecTests/Operation.php#L633)). In turn, [`with_transaction()`](https://github.com/mongodb/mongo-php-library/blob/1.11.0/src/functions.php#L429) passes those options as-is on to the [`WithTransaction`](https://github.com/mongodb/mongo-php-library/blob/1.11.0/src/Operation/WithTransaction.php) operation, where they are eventually passed on to `startSession()` in [`WithTransaction::execute()`](https://github.com/mongodb/mongo-php-library/blob/1.11.0/src/Operation/WithTransaction.php#L64).

Therefore, this operation should support the same arguments as `startTransaction` with the addition of `callback`.

tests/UnifiedSpecTests/Util.php
```diff
@@ -26,13 +34,113 @@
...
 
+    public static function assertArgumentsBySchema(string $executingObjectName, string $operation, array $args): void
+    {
+        assertArrayHasKey($executingObjectName, self::$args, 'Missing object "' . $executingObjectName . '" definition. The list of allowed objects, operation names and arguments should be defined at ' . self::class);
+        assertIsArray(self::$args[$executingObjectName], 'Object "' . $executingObjectName . '" is not correctly defined. The list of allowed objects, operation names and arguments should be defined at ' . self::class . '. ');
+        assertArrayHasKey($operation, self::$args[$executingObjectName], 'Missing operation "' . $operation . '" definition for object "' . $executingObjectName . '". The list of allowed objects, operation names and arguments should be defined at ' . self::class);
+        assertIsArray(self::$args[$executingObjectName][$operation], 'Operation "' . $operation . '" for object "' . $executingObjectName . '" is not correctly defined. The list of allowed objects, operation names and arguments should be defined at ' . self::class);
+        self::assertHasOnlyKeys($args, self::$args[$executingObjectName][$operation]);
```
jmikola, line 141:
```suggestion
        self::assertHasOnlyKeys($args, self::$args[$executingObjectName][$operation]);
```

Thinking about this some more in light of my [original comment](https://github.com/mongodb/mongo-php-library/pull/898#discussion_r824764788), I'd prefer to use a more concise approach.

`assertArgumentsBySchema()` is only going to be called from validate operation cases, so there really should not be a situation where `self::$args[$executingObjectName][$operation]` is not set. Nevertheless, we can avoid using the null-coalescing operator (i.e. `??`) as I originally proposed and invite a PHP error. That will provide the same sanity check without the overhead of  four additional assertions per operation.

jmikola (changes requested):

jmikola:
@aleksandr-rudo: I noticed in https://github.com/mongodb/mongo-php-library/pull/896 that you manually applied the same suggestions I left. In case you're not familiar with the feature, take a look at [applying suggested changes](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request#applying-suggested-changes) (that may save time for this PR).

You can batch them up so that they apply in a single commit, which will then be immediately tested by CI. If you need to continue development on the branch locally, you'll want to pull down changes from your remote fork first, though. And at merge time, we can always squash the commits and ignore the "Suggestions from code review" commit(s).

jmikola (commented):
@aleksandr-rudo: It looks like there are two more coding standard issues to fix. `vendor/bin/phpcbf` should be able to take care of those for you.

PR #900, opened 2022-03-16
PHPLIB-789: Snapshot query examples

jmikola:
https://jira.mongodb.org/browse/PHPLIB-789


tests/DocumentationExamplesTest.php
```diff
@@ -1575,6 +1575,103 @@ public function testCausalConsistency(): void
...
+            ['name' => 'Brian', 'color' => 'white', 'adoptable' => true],
+        ]);
+
+        $uriString = static::getUri(true);
+
+        ob_start();
+
+        // Start Snapshot Query Example 1
+        $catsCollection = $client->selectCollection('pets', 'cats');
+        $dogsCollection = $client->selectCollection('pets', 'dogs');
```
jmikola, line 1607: These are purposely redefined in order to demonstrate selecting a collection from the client object within the example. For context, the docs team parses this file for code between various start/end comments and uses that to render language-specific examples in the MongoDB manual.

tests/DocumentationExamplesTest.php
```diff
@@ -1575,6 +1575,103 @@ public function testCausalConsistency(): void
         ob_end_clean();
     }
 
+    public function testSnapshotQueries(): void
+    {
+        if (version_compare($this->getServerVersion(), '5.0.0', '<')) {
+            $this->markTestSkipped('Snapshot queries outside of transactions are not supported');
```
jmikola, line 1581: See: https://docs.mongodb.com/manual/reference/read-concern-snapshot/

tests/DocumentationExamplesTest.php
```diff
@@ -1575,6 +1575,103 @@ public function testCausalConsistency(): void
...
+
+        $adoptablePetsCount += $dogsCollection->aggregate(
+            [
+                ['$match' => ['adoptable' => true]],
+                ['$count' => 'adoptableDogsCount'],
+            ],
+            ['session' => $session]
+        )->toArray()[0]->adoptableDogsCount;
+
+        var_dump($adoptablePetsCount);
```
jmikola, line 1627: Various examples require printing output. Since we don't want that output to interfere with the test suite, we typically use [output buffering](https://www.php.net/manual/en/book.outcontrol.php) to capture and discard it.

tests/DocumentationExamplesTest.php
```diff
@@ -1575,6 +1575,103 @@ public function testCausalConsistency(): void
...
+        // End Snapshot Query Example 1
+
+        ob_end_clean();
+
+        $this->assertSame(3, $adoptablePetsCount);
+
+        $catsCollection->drop();
+        $dogsCollection->drop();
```
jmikola, line 1635: Since these examples use specific database and collection names, the cleanup logic in `tearDown()` does not apply and we manually clean up after each test.

tests/DocumentationExamplesTest.php
```diff
@@ -1575,6 +1575,103 @@ public function testCausalConsistency(): void
         ob_end_clean();
     }
 
+    public function testSnapshotQueries(): void
```
jmikola, line 1578: I observed SnapshotUnavailable(246) errors in CI builds. This PR will require work-arounds for [replica sets](https://github.com/mongodb/mongo-ruby-driver/commit/7c4117b58e3d12e237f7536f7521e18fc15f79ac#diff-7b02c2b4b34a58c688750366d0230014bb453e0381af394d736aa9d0a25538a5R374) and [sharded clusters](https://github.com/mongodb/mongo-ruby-driver/commit/7c4117b58e3d12e237f7536f7521e18fc15f79ac#diff-3f7d999c8369e37a728bdeb5eff37a5780d98403c143e6e4b64008d174e1bcdaR10), which we can adapt from [RUBY-2909](https://jira.mongodb.org/browse/RUBY-2909).

jmikola: Added private `waitForSnapshot()` and `preventStaleDbVersionError()` methods for replica sets and sharded clusters, respectively.

jmikola:
@aleksandr-rudo did you have any feedback/questions before I merge this?

tests/DocumentationExamplesTest.php
```diff
@@ -1751,4 +1867,39 @@ private function assertInventoryCount($count): void
     {
         $this->assertCollectionCount($this->getDatabaseName() . '.' . $this->getCollectionName(), $count);
     }
+
+    private function waitForSnapshot(string $databaseName, string $collectionName): void
```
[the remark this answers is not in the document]
jmikola: Note: this function was lifted from https://github.com/mongodb/mongo-ruby-driver/commit/7c4117b58e3d12e237f7536f7521e18fc15f79ac#diff-7b02c2b4b34a58c688750366d0230014bb453e0381af394d736aa9d0a25538a5R374 in the Ruby driver.

The function itself is not generally useful, since we're targeting a specific collection. The server-side error message is:

> Unable to read from a snapshot due to pending collection catalog changes; please retry the operation.

If a user were to encounter this, they would probably want to retry their original query. For our purposes here, ensuring _any_ snapshot query (even one with a different query filter) succeeds seems to do the trick to avoid an error in the documentation example that calls this method.

tests/DocumentationExamplesTest.php
```diff
@@ -1751,4 +1867,39 @@ private function assertInventoryCount($count): void
...
+
+                break;
+            } catch (CommandException $e) {
+                if ($e->getCode() === 246 /* SnapshotUnavailable */) {
+                    continue;
+                }
+
+                throw $e;
+            }
+        } while (microtime(true) < $startTime + 10);
```
jmikola, line 1894:
```suggestion
        /* Retry until a snapshot query succeeds or ten seconds elapse,
         * whichwever comes first.
         *
         * TODO: use hrtime() once the library requires PHP 7.3+ */
        $retryUntil = microtime(true) + 10;

        do {
            try {
                $collection->aggregate(
                    [['$match' => ['_id' => ['$exists' => true]]]],
                    ['session' => $session]
                );

                break;
            } catch (CommandException $e) {
                if ($e->getCode() === 246 /* SnapshotUnavailable */) {
                    continue;
                }

                throw $e;
            }
        } while (microtime(true) < $retryUntil);
```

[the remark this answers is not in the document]
jmikola: This is only called once, so I'm not inclined to parameterize the method; however, I do think this can be improved by removing the literal `10` here and changing the function to instead calculate the deadline up front (with a comment explaining the 10-second timeout).

PR #901, opened 2022-03-18
PHPLIB-808: Leave fullDocument change stream option unset by default

aleksandr-rudo:
https://jira.mongodb.org/browse/PHPLIB-808

### DIFF
--- a/src/Operation/Watch.php
+++ b/src/Operation/Watch.php
@@ -173,11 +173,10 @@ class Watch implements Executable, /* @internal */ CommandSubscriber
         }
 
         $options += [
-            'fullDocument' => self::FULL_DOCUMENT_DEFAULT,
             'readPreference' => new ReadPreference(ReadPreference::RP_PRIMARY),
         ];
 
-        if (! is_string($options['fullDocument'])) {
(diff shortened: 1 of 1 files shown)


src/Operation/Watch.php
```diff
@@ -173,11 +174,10 @@ public function __construct(Manager $manager, $databaseName, $collectionName, ar
         }
 
         $options += [
-            'fullDocument' => self::FULL_DOCUMENT_DEFAULT,
             'readPreference' => new ReadPreference(ReadPreference::RP_PRIMARY),
         ];
 
-        if (! is_string($options['fullDocument'])) {
+        if (array_key_exists('fullDocument', $options) && ! is_string($options['fullDocument'])) {
```
aleksandr-rudo, line 180: array_key_exists function is used to avoid case when array field "fullDocument" with NULL value is passed from input (there is already exists an incorrect NULL-value input data test in MongoDB\Tests\Operation\WatchTest:55)

jmikola (approved):

jmikola:
@aleksandr-rudo: This can be squash-merged. Please remember to delete the adjust the commit message to remove anything not relevant to the final commit.

PR #904, opened 2022-03-24
PHPLIB-658: optional new and upsert fields for findAndModify commands

aleksandr-rudo:


src/Operation/FindAndModify.php
```diff
@@ -151,7 +148,7 @@ public function __construct($databaseName, $collectionName, array $options)
             throw InvalidArgumentException::invalidType('"maxTimeMS" option', $options['maxTimeMS'], 'integer');
         }
 
-        if (! is_bool($options['new'])) {
+        if (array_key_exists('new', $options) && ! is_bool($options['new'])) {
```
aleksandr-rudo, line 151: The main idea of this check is to keep the same behavior of constructor verification stage. Previously there was an assumption that the array definitely contains field 'new', so the old check covered all values include NULL value.

src/Operation/FindAndModify.php
```diff
@@ -183,7 +180,7 @@ public function __construct($databaseName, $collectionName, array $options)
             throw InvalidArgumentException::invalidType('"writeConcern" option', $options['writeConcern'], WriteConcern::class);
         }
 
-        if (! is_bool($options['upsert'])) {
+        if (array_key_exists('upsert', $options) && ! is_bool($options['upsert'])) {
```
aleksandr-rudo, line 183: The main idea of this check is to keep the same behavior of constructor verification stage. Previously there was an assumption that the array definitely contains field 'new', so the old check covered all values include NULL value.

src/Operation/FindOneAndReplace.php
```diff
@@ -104,20 +105,16 @@ public function __construct($databaseName, $collectionName, $filter, $replacemen
...
-            'returnDocument' => self::RETURN_DOCUMENT_BEFORE,
-            'upsert' => false,
-        ];
-
         if (isset($options['projection']) && ! is_array($options['projection']) && ! is_object($options['projection'])) {
             throw InvalidArgumentException::invalidType('"projection" option', $options['projection'], 'array or object');
         }
 
-        if (! is_integer($options['returnDocument'])) {
+        if (array_key_exists('returnDocument', $options) && ! is_integer($options['returnDocument'])) {
```
aleksandr-rudo, line 112: The main idea of this check is to keep the same behavior of constructor verification stage. Previously there was an assumption that the array definitely contains field 'returnDocument', so the old check covered all values include NULL value.

src/Operation/FindOneAndUpdate.php
```diff
@@ -108,20 +109,16 @@ public function __construct($databaseName, $collectionName, $filter, $update, ar
...
-            'returnDocument' => self::RETURN_DOCUMENT_BEFORE,
-            'upsert' => false,
-        ];
-
         if (isset($options['projection']) && ! is_array($options['projection']) && ! is_object($options['projection'])) {
             throw InvalidArgumentException::invalidType('"projection" option', $options['projection'], 'array or object');
         }
 
-        if (! is_integer($options['returnDocument'])) {
+        if (array_key_exists('returnDocument', $options) && ! is_integer($options['returnDocument'])) {
```
aleksandr-rudo, line 116: The main idea of this check is to keep the same behavior of constructor verification stage. Previously there was an assumption that the array definitely contains field 'returnDocument', so the old check covered all values include NULL value.

src/Operation/FindAndModify.php
jmikola: Noted that this is the same approach used in https://github.com/mongodb/mongo-php-library/pull/901#discussion_r830150708.

jmikola (approved):

PR #907, opened 2022-03-30
PHPLIB-827: Use https links in documentation and comments

jmikola:
https://jira.mongodb.org/browse/PHPLIB-827

Also removes redundant `www` subdomains from some URLs (Apache license links intentionally left as-is).


docs/tutorial/collation.txt
```diff
@@ -66,7 +66,7 @@ Collation Parameters
    ]
 
 The only required parameter is ``locale``, which the server parses as an `ICU
-format locale ID <http://userguide.icu-project.org/locale>`_. For example, set
+format locale ID <https://userguide.icu-project.org/locale>`_. For example, set
```
jmikola, line 69:
```suggestion
format locale ID <https://unicode-org.github.io/icu/userguide/locale/>`_. For example, set
```

The original link, http://userguide.icu-project.org/locale, redirects to https://sites.google.com/site/icuprojectuserguide/locale, which in turns points to the above URL.

PR #921, opened 2022-04-22
PHPLIB-647: skipping CSFLE tests if mongocryptd is not available

aleksandr-rudo:
https://jira.mongodb.org/browse/PHPLIB-647

### DIFF
--- a/tests/FunctionalTestCase.php
+++ b/tests/FunctionalTestCase.php
@@ -30,7 +30,6 @@ use function getenv;
 use function implode;
 use function is_array;
 use function is_callable;
-use function is_executable;
 use function is_object;
 use function is_string;
 use function key;
@@ -44,10 +43,8 @@ use function preg_replace;
 use function sprintf;
 use function version_compare;
 
-use const DIRECTORY_SEPARATOR;
 use const FILTER_VALIDATE_BOOLEAN;
 use const INFO_MODULES;
-use const PATH_SEPARATOR;
 
 abstract class FunctionalTestCase extends TestCase
 {
--- a/tests/FunctionalTestCase.php
+++ b/tests/FunctionalTestCase.php
@@ -21,7 +21,6 @@ use stdClass;
 use UnexpectedValueException;
 
 use function array_merge;
-use function basename;
 use function call_user_func;
 use function count;
 use function current;
@@ -43,7 +42,6 @@ use function preg_match;
 use function preg_quote;
 use function preg_replace;
 use function sprintf;
-use function strpos;
 use function version_compare;
 
 use const DIRECTORY_SEPARATOR;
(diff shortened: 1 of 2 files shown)


tests/SpecTests/ClientSideEncryptionSpecTest.php
```diff
@@ -95,6 +95,9 @@ public function testClientSideEncryption(stdClass $test, ?array $runOn, array $d
             $this->markTestSkipped($test->skipReason);
         }
 
+        // Checking for mongocryptd availability
+        $this->skipIfMongocryptdIsUnavailableByTest($test);
```
aleksandr-rudo, line 99: The suggested in the source ticket function skipIfClientSideEncryptionIsNotSupported is not good enough for expected check because remote or specific location of mongocryptd can be defined in provided test or in some specific auto encryption options. That's the reason why each test function contains its personal check for mongocryptd.

tests/FunctionalTestCase.php
```diff
@@ -497,6 +502,33 @@ protected function skipIfClientSideEncryptionIsNotSupported(): void
...
+    {
+        $mongocryptdURI = $autoEncryptionOpts['extraOptions']['mongocryptdURI'] ?? null;
+
+        $this->skipIfMongocryptdIsUnavailable($mongocryptdURI);
+    }
+
+    protected function skipIfMongocryptdIsUnavailable(?string $mongocryptdURI = null): void
+    {
+        // Assuming that in specific cases with remote mongocryptd developer is responsible for mongocryptd to be available
+        if (! $mongocryptdURI || strpos($mongocryptdURI, 'mongodb://localhost') === 0) {
```
jmikola, line 525: Was there a specific reason to add this logic, such as your container configuration? Since the CSFLE spec tests aren't specifying custom URIs, I think it's preferable to just limit the logic to checking whether `mongocryptd` appears within one of the `PATH` directories. That would be sufficient for the needs of PHPLIB's test suite, and also limit the amount of code introduced by this PR.

tests/FunctionalTestCase.php
```diff
@@ -625,4 +657,31 @@ private function isFailCommandEnabled(): bool
...
+     * @return bool
+     */
+    private function isExecutablePathenv(string $filename): bool
+    {
+        if (is_executable($filename)) {
+            return true;
+        }
+
+        if ($filename !== basename($filename)) {
+            return false;
```
jmikola, line 674: What case is this covering?

Per my earlier comments, I think this would be better relocated to a private method as `ClientSideEncryptionSpecTest::skipIfMongocryptdIsUnavailable()`.

```
protected function skipIfMongocryptdIsUnavailable(): void
{
    $paths = explode(PATH_SEPARATOR, getenv("PATH"));

    foreach ($paths as $path) {
        if (is_executable($path . DIRECTORY_SEPARATOR . 'mongocryptd')) {
            return;
        }
    }

    $this->markTestSkipped('mongocryptd binary is not in PATH');
}
```

Then, that method can be called from `ClientSideEncryptionSpecTest::setUp()`. Even though not every test within the CSFLE test suite might actually require a running `mongocryptd`, I think that's a reasonable approach to avoid the errors from [PHPLIB-647](https://jira.mongodb.org/browse/PHPLIB-647) that we're trying to avoid here.

You can also consider using `markTestIncomplete` instead, since that's noisier than merely skipping the test. The goal here is the avoid local test failures while alerting the developer to the fact that they may be missing binaries -- so in that sense an incomplete test may be more likely to show up in PHPUnit logs and alert them to the issue.

jmikola (changes requested):

jmikola (approved):
Looks fine now. I must have submitted the previous review before you pushed `phpcbf` fixes.

PR #928, opened 2022-05-16
PHPLIB-843 and PHPLIB-856: clusteredIndex and changeStreamPreAndPostImages options

jmikola:
https://jira.mongodb.org/browse/PHPLIB-843
https://jira.mongodb.org/browse/PHPLIB-856

Also adds missing `createCollection` option for [PHPLIB-814](https://jira.mongodb.org/browse/PHPLIB-814), which was included in the recently synced spec tests.

### DIFF
--- a/tests/UnifiedSpecTests/Operation.php
+++ b/tests/UnifiedSpecTests/Operation.php
@@ -12,6 +12,8 @@ use MongoDB\Driver\Cursor;
 use MongoDB\Driver\Server;
 use MongoDB\Driver\Session;
 use MongoDB\GridFS\Bucket;
+use MongoDB\Model\CollectionInfo;
+use MongoDB\Model\DatabaseInfo;
 use MongoDB\Model\IndexInfo;
 use MongoDB\Operation\DatabaseCommand;
 use MongoDB\Operation\FindOneAndReplace;
@@ -23,6 +25,7 @@ use stdClass;
 use Throwable;
 
 use function array_diff_key;
+use function array_intersect_key;
 use function array_key_exists;
 use function array_map;
 use function current;
@@ -260,7 +263,12 @@ final class Operation
                 return iterator_to_array($client->listDatabaseNames($args));
 
             case 'listDatabases':
-                return iterator_to_array($client->listDatabases($args));
+                return array_map(
+                    function (DatabaseInfo $info) {
+                        return $info->__debugInfo();
+                    },
+                    iterator_to_array($client->listDatabases($args))
+                );
 
             default:
                 Assert::fail('Unsupported client operation: ' . $this->name);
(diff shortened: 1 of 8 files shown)


tests/UnifiedSpecTests/Operation.php
```diff
@@ -579,7 +592,28 @@ private function executeForDatabase(Database $database)
...
+                 * separately, so we must split the array after initially
+                 * filtering out the collection name.
+                 *
+                 * The typeMap option is intentionally omitted since it is
+                 * specific to PHPLIB and will never appear in spec tests. */
+                $options = array_diff_key($args, ['collection' => 1]);
+                $collectionOptions = array_diff_key($options, ['session' => 1, 'writeConcern' => 1]);
+                $options = array_intersect_key($options, ['session' => 1, 'writeConcern' => 1]);
+
+                return $database->modifyCollection($args['collection'], $collectionOptions, $options);
```
aleksandr-rudo, line 616: @jmikola As far as I traced this new method and according to Util.php update, which adds new option "changeStreamPreAndPostImages" for this "modifyCollection", I can't see the way this new option is used inside operating class (ModifyCollection.php):
https://github.com/jmikola/mongo-php-library/blob/phplib-843/src/Operation/ModifyCollection.php
And also it looks like this new option needs to be described in the doc file for modifyCollection method:
https://github.com/jmikola/mongo-php-library/blob/phplib-843/docs/includes/apiargs-MongoDBDatabase-method-modifyCollection-option.yaml

aleksandr-rudo: I have found the way this new option is passed, but still it needs to be described in the doc-file and in the "Supported options" section of ModifyCollection.php class.

jmikola: The options you're referring to above are those within the `$options` parameter, which is distinct from the `$collectionOptions` parameter.

PHPLIB has never validated or documented the individual options supported _within_ the `$collectionOptions` parameter. Our [API documentation](https://www.mongodb.com/docs/php-library/current/reference/method/MongoDBDatabase-modifyCollection/) just points users to the [`collMod` documentation](https://www.mongodb.com/docs/manual/reference/command/collMod/).

This is similar to what we do for `Collection::createIndexes()` and the [CreateIndexes](https://github.com/mongodb/mongo-php-library/blob/master/src/Operation/CreateIndexes.php) operation. PHPLIB does not document all of the supported index options and we instead direct users to the server manual from that method's [API reference](https://www.mongodb.com/docs/php-library/current/reference/method/MongoDBCollection-createIndexes/).

The `Collection::createIndex()` method is a different story. For better or worse, that method has historically taken _both_ index options and command options in the same argument, which requires the Collection method to split them up with `array_diff_key` and `array_intersect_key`. Thankfully, we did not make that same mistake when implementing `Database::modifyCollection()`.

I'll note that the [API reference for `createIndex()`](https://www.mongodb.com/docs/php-library/current/reference/method/MongoDBCollection-createIndex/) does happen to list _some_ index options supported by the server; however, comparing that content with [createIndexes](https://www.mongodb.com/docs/manual/reference/command/createIndexes/) in the server manual it appears our docs are outdated. I'm going to open a ticket to propose we rewrite this and remove documentation of these options and instead point users to the server manual for a canonical reference.

aleksandr-rudo (approved):

PR #931, opened 2022-05-17
PHPLIB-873: Restrict unified change stream tests to replica sets

jmikola:
https://jira.mongodb.org/browse/PHPLIB-873

This is a temporary solution to address flaky test failures on sharded clusters.

Synced with mongodb/specifications@f07351c0c61cd79e0e7a197f78a5cfcc028c2966


jmikola:
Merging without review, as this is a small change and CI reports that flaky test failures on sharded clusters are no more.

PR #948, opened 2022-07-05
PHPLIB-900: Spec test for discarding dirty implicit session incorrectly uses explicit session

aleksandr-rudo:
https://jira.mongodb.org/browse/PHPLIB-900


jmikola:
Test failure is unrelated and will be addressed by https://github.com/mongodb/mongo-php-library/pull/950

PR #949, opened 2022-07-06
PHPLIB-846: Improved change stream event visibility for C2C Replication

aleksandr-rudo:


tests/UnifiedSpecTests/Operation.php
```diff
@@ -491,7 +491,11 @@ function (IndexInfo $info) {
                 assertArrayHasKey('to', $args);
                 assertIsString($args['to']);
 
-                return $collection->rename($args['to']);
+                return $collection->rename(
+                    $args['to'],
+                    null,
+                    array_diff_key($args, ['to' => 1])
+                );
```
aleksandr-rudo, line 498: These changes were made to make able RenameCollection helper receive "dropTarget" option, which is needed for new test.

jmikola (changes requested):

PR #953, opened 2022-08-03
PHPLIB-933: Test on PHP 8.2

alcaeus:
PHPLIB-933

This includes changes to remove warnings due to new method signatures in PHPC.

### DIFF
--- a/tests/UnifiedSpecTests/Constraint/IsBsonTypeTest.php
+++ b/tests/UnifiedSpecTests/Constraint/IsBsonTypeTest.php
@@ -163,7 +163,7 @@ class IsBsonTypeTest extends TestCase
 // phpcs:disable Squiz.Classes.ClassFileName.NoMatch
 class SerializableArray implements Serializable
 {
-    public function bsonSerialize()
+    public function bsonSerialize(): array
     {
         return ['foo'];
     }
@@ -171,7 +171,7 @@ class SerializableArray implements Serializable
 
 class SerializableObject implements Serializable
 {
-    public function bsonSerialize()
+    public function bsonSerialize(): array
     {
         return ['x' => 1];
     }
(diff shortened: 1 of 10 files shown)


tests/UnifiedSpecTests/Constraint/IsBsonTypeTest.php
```diff
@@ -163,15 +163,15 @@ private function assertResult($expected, Constraint $constraint, $value, string
 // phpcs:disable Squiz.Classes.ClassFileName.NoMatch
 class SerializableArray implements Serializable
 {
-    public function bsonSerialize()
+    public function bsonSerialize(): array
```
jmikola, line 166: Any reason these implementations don't include `#[ReturnTypeWillChange]`?

tests/UnifiedSpecTests/Operation.php
```diff
@@ -295,7 +295,12 @@ private function executeForCollection(Collection $collection)
                 assertIsArray($args['requests']);
 
                 return $collection->bulkWrite(
-                    array_map('self::prepareBulkWriteRequest', $args['requests']),
```
jmikola, line 298: Was this syntax deprecated in PHP 8.2?

tests/UnifiedSpecTests/Constraint/IsBsonTypeTest.php
alcaeus: Since these are test classes where we're not bound by BC constraints, we can add the type right away.

tests/UnifiedSpecTests/Operation.php
alcaeus: Yes, partially supported callables were deprecated in PHP 8.2, including any callable patterns that include `self`, `parent`, or `static`. [This page](https://php.watch/versions/8.2/partially-supported-callable-deprecation) contains the full list of deprecated callable syntaxes.

jmikola (approved):

PR #954, opened 2022-08-04
PHPLIB-912: Fix changestream creation to avoid errors on sharded clusters

alcaeus:
PHPLIB-912

This is one possible way to fix the issue of certain tests sporadically failing on sharded clusters. My initial attempt to only pass a `startAtOperationTime` still caused failures, as the operation time of the previous command (in this case the `create` command to create the test collection) occurred within the same second and was thus captured by the change stream. To reduce sleep time to a minimum (as this applies to every test involving a change stream), we only sleep long enough to have the timestamp tick over to the next second.

This is something we may want to call out in the spec. I don't know if other drivers are facing the same issues (I'd expect they are), but it might be helpful to record this somewhere in the unified test format specification.

### DIFF
--- a/tests/UnifiedSpecTests/Operation.php
+++ b/tests/UnifiedSpecTests/Operation.php
@@ -2,8 +2,10 @@
 
 namespace MongoDB\Tests\UnifiedSpecTests;
 
+use DateTime;
 use Error;
 use MongoDB\BSON\Javascript;
+use MongoDB\BSON\Timestamp;
 use MongoDB\ChangeStream;
 use MongoDB\Client;
 use MongoDB\Collection;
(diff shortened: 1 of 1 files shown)


tests/UnifiedSpecTests/Operation.php
```diff
@@ -254,9 +258,12 @@ private function executeForClient(Client $client)
                 assertArrayHasKey('pipeline', $args);
                 assertIsArray($args['pipeline']);
 
+                $this->sleepUntilNextSecond();
```
jmikola, line 261: @alcaeus and I discussed this and he noted that this approach would require additional changes to event monitoring to ignore `startAtOperationTime` in the outgoing `aggregate` command.

An alternative approach would be to conditionally ignore extra `create` events in the change stream, and that's something we can likely do from the spec tests themselves. He's going to pursue that option next.

alcaeus:
Closing as we're fixing this in the specs repo, see https://github.com/mongodb/specifications/pull/1293.

PR #958, opened 2022-08-19
PHPLIB-715: Use flexible numeric comparisons in DocumentMatchConstraint

alcaeus:
PHPLIB-715

This unskips a mapReduce test that fails due to strict numerical comparison. To fix this, I applied the same logic used in the unified spec tests. This comes at the cost of removing placeholder functionality in DocumentsMatchConstraint; however this wasn't used in the constraint at all. The transaction spec tests make use of the `42` placeholder, but the transaction spec tests covers these in the command listener it uses to assert the commands, so it's safe to remove without breaking other tests.

### DIFF
--- a/tests/SpecTests/DocumentsMatchConstraint.php
+++ b/tests/SpecTests/DocumentsMatchConstraint.php
@@ -37,6 +37,8 @@ use function get_class;
 use function get_debug_type;
 use function in_array;
 use function is_array;
+use function is_float;
+use function is_int;
 use function is_object;
 use function is_scalar;
 use function method_exists;
@@ -58,9 +60,6 @@ class DocumentsMatchConstraint extends Constraint
 
     /** @var boolean */
     private $ignoreExtraKeysInEmbedded = false;
-
-    /** @var array */
-    private $placeholders = [];
 
     /**
      * TODO: This is not currently used, but was preserved from the design of
@@ -88,14 +87,12 @@ class DocumentsMatchConstraint extends Constraint
      * @param array|object $value
      * @param boolean      $ignoreExtraKeysInRoot     If true, ignore extra keys within the root document
      * @param boolean      $ignoreExtraKeysInEmbedded If true, ignore extra keys within embedded documents
-     * @param array        $placeholders              Placeholders for any value
      */
-    public function __construct($value, bool $ignoreExtraKeysInRoot = false, bool $ignoreExtraKeysInEmbedded = false, array $placeholders = [])
+    public function __construct($value, bool $ignoreExtraKeysInRoot = false, bool $ignoreExtraKeysInEmbedded = false)
     {
         $this->value = $this->prepareBSON($value, true, $this->sortKeys);
         $this->ignoreExtraKeysInRoot = $ignoreExtraKeysInRoot;
         $this->ignoreExtraKeysInEmbedded = $ignoreExtraKeysInEmbedded;
-        $this->placeholders = $placeholders;
         $this->comparatorFactory = Factory::getInstance();
     }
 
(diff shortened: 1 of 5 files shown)


tests/SpecTests/DocumentsMatchConstraintTest.php
```diff
@@ -146,13 +145,13 @@ public function errorMessageProvider()
                 ['foo' => ['foo' => 'bar', 'bar' => 'baz']],
             ],
             'Scalar value not equal' => [
-                'Field path "foo": Failed asserting that two values are equal.',
+                'Field path "foo": Failed asserting that two strings are equal.',
```
alcaeus, line 148: Error messages change due to us no longer using a direct comparison for scalar values, and instead leaving those checks to PHPUnit comparators.

tests/UnifiedSpecTests/Constraint/MatchesTest.php
```diff
@@ -23,6 +23,14 @@ public function testMatchesDocument(): void
         $this->assertResult(true, $c, ['y' => ['b' => 2, 'a' => 1], 'x' => 1], 'Root and embedded key order is not significant');
     }
 
+    public function testFlexibleNumericComparison(): void
```
alcaeus, line 26: Added these for the sake of completeness (and because I wanted to make sure that this works in the unified test runner).

tests/SpecTests/DocumentsMatchConstraint.php
```diff
@@ -37,6 +37,8 @@
 use function get_debug_type;
 use function in_array;
```
jmikola, line 38: Looks like you need to run `phpcbf` on this file.

tests/SpecTests/DocumentsMatchConstraint.php
```diff
@@ -421,6 +401,11 @@ private function doToString()
         return 'matches ' . $this->exporter()->export($this->value);
     }
 
+    private static function isNumeric($value): bool
+    {
+        return is_int($value) || is_float($value) || $value instanceof Int64;
```
jmikola, line 406: `prepareBSON` has logic to "Convert Int64 objects to integers on 64-bit platforms for compatibility reasons." Was that sufficient to allow flexible comparisons of Int64 objects with integers and floats? Mainly asking because I don't believe we have a test for this.

Int64 is conditionally used in DocumentsMatchConstraintTest only when the platform is 32-bit.

This probably makes no difference, but the fact that you reference the class here made me ask.

tests/SpecTests/Operation.php
```diff
@@ -206,6 +207,10 @@ public function assert(FunctionalTestCase $test, Context $context, bool $bubbleE
              * is not used (e.g. Command Monitoring spec). */
             if ($result instanceof Cursor) {
                 $result = $result->toArray();
+            } elseif ($result instanceof MapReduceResult) {
+                /* For mapReduce operations, we ignore the mapReduce metadata
+                 * and only return the result iterator for evaluation. */
+                $result = iterator_to_array($result->getIterator());
```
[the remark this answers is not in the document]
alcaeus: It doesn't. We could update the logic to call `iterator_to_array` on any `IteratorAggregate`, but `assertDocumentsMatch` compared the `MapReduceResult` instance to the expected object, causing a test failure.

tests/SpecTests/DocumentsMatchConstraint.php
alcaeus: Whoops, done. Thanks!

tests/SpecTests/DocumentsMatchConstraint.php
alcaeus: I shamelessly copied the logic from the unified spec tests where this was used for flexible numeric comparisons.

tests/SpecTests/Operation.php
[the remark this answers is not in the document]
jmikola: `ResultExpectation::assert()` had entirely different handling for those cases. `ASSERT_SAME_DOCUMENTS` invoked `TestCase::assertSameDocuments()`, which unpacks the Traversable with `iterator_to_array()`. `ASSERT_DOCUMENTS_MATCH` used `MongoDB\Tests\SpecTests\FunctionalTestCase::assertDocumentsMatch`.

This is fine. No sense adding significant refactoring for an outdated code path.

jmikola (approved):

PR #959, opened 2022-08-24
PHPLIB-599: Typing improvements

alcaeus:
This PR handles PHPLIB-600 and PHPLIB-602, adding parameter types and return types where applicable.

Rules for phpcs have been updated to require argument types; no changes were made to return types as most of the classes are non-final (something we may want to address for 2.0).

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -112,7 +112,6 @@
         </properties>
 
         <exclude name="SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingTraversableTypeHintSpecification" />
-        <exclude name="SlevomatCodingStandard.TypeHints.ParameterTypeHint.UselessAnnotation" />
     </rule>
 
     <rule ref="SlevomatCodingStandard.TypeHints.PropertyTypeHint">
@@ -126,7 +125,6 @@
         </properties>
 
         <exclude name="SlevomatCodingStandard.TypeHints.PropertyTypeHint.MissingTraversableTypeHintSpecification" />
-        <exclude name="SlevomatCodingStandard.TypeHints.PropertyTypeHint.UselessAnnotation" />
     </rule>
 
     <rule ref="SlevomatCodingStandard.TypeHints.ReturnTypeHint">
@@ -140,7 +138,6 @@
         </properties>
 
         <exclude name="SlevomatCodingStandard.TypeHints.ReturnTypeHint.MissingTraversableTypeHintSpecification" />
-        <exclude name="SlevomatCodingStandard.TypeHints.ReturnTypeHint.UselessAnnotation" />
     </rule>
 
 
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -133,8 +133,6 @@
 
     <rule ref="SlevomatCodingStandard.TypeHints.ReturnTypeHint">
         <properties>
-            <!-- Requires PHP 7.2 -->
-            <property name="enableObjectTypeHint" value="false" />
             <!-- Requires PHP 8.0 -->
             <property name="enableStaticTypeHint" value="false" />
             <!-- Requires PHP 8.0 -->
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -105,8 +105,6 @@
     <!-- **************************************************************************** -->
     <rule ref="SlevomatCodingStandard.TypeHints.ParameterTypeHint">
         <properties>
-            <!-- Requires PHP 7.4 -->
-            <property name="enableObjectTypeHint" value="false" />
             <!-- Requires PHP 8.0 -->
             <property name="enableMixedTypeHint" value="false" />
             <!-- Requires PHP 8.0 -->
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -162,12 +162,9 @@
     </rule>
 
 
-    <!-- *********************************************************************************** -->
-    <!-- Require native type hints for all parameters, properties, and return types in tests -->
-    <!-- *********************************************************************************** -->
-    <rule ref="SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingNativeTypeHint">
-        <exclude-pattern>src</exclude-pattern>
-    </rule>
+    <!-- *********************************************************** -->
+    <!-- Require native type hints for all code without a BC promise -->
+    <!-- *********************************************************** -->
     <rule ref="SlevomatCodingStandard.TypeHints.PropertyTypeHint.MissingNativeTypeHint">
         <exclude-pattern>src</exclude-pattern>
     </rule>
--- a/src/BulkWriteResult.php
+++ b/src/BulkWriteResult.php
@@ -35,8 +35,7 @@ class BulkWriteResult
     private $isAcknowledged;
 
     /**
-     * @param WriteResult $writeResult
-     * @param mixed[]     $insertedIds
+     * @param mixed[] $insertedIds
      */
     public function __construct(WriteResult $writeResult, array $insertedIds)
     {
--- a/src/BulkWriteResult.php
+++ b/src/BulkWriteResult.php
@@ -28,15 +28,12 @@ class BulkWriteResult
     /** @var WriteResult */
     private $writeResult;
 
-    /** @var mixed[] */
+    /** @var array */
     private $insertedIds;
 
     /** @var boolean */
     private $isAcknowledged;
 
-    /**
-     * @param mixed[] $insertedIds
-     */
     public function __construct(WriteResult $writeResult, array $insertedIds)
     {
         $this->writeResult = $writeResult;
@@ -89,7 +86,7 @@ class BulkWriteResult
      * field value. Any driver-generated ID will be a MongoDB\BSON\ObjectId
      * instance.
      *
-     * @return mixed[]
+     * @return array
      */
     public function getInsertedIds()
     {
@@ -164,7 +161,7 @@ class BulkWriteResult
      * This method should only be called if the write was acknowledged.
      *
      * @see BulkWriteResult::isAcknowledged()
-     * @return mixed[]
+     * @return array
      * @throws BadMethodCallException is the write result is unacknowledged
      */
     public function getUpsertedIds()
--- a/src/ChangeStream.php
+++ b/src/ChangeStream.php
@@ -224,7 +224,7 @@ class ChangeStream implements Iterator
      * @param boolean $incrementKey Increment $key if there is a current result
      * @throws ResumeTokenException
      */
-    private function onIteration($incrementKey)
+    private function onIteration(bool $incrementKey)
     {
         /* If the cursorId is 0, the server has invalidated the cursor and we
          * will never perform another getMore nor need to resume since any
--- a/src/ChangeStream.php
+++ b/src/ChangeStream.php
@@ -90,8 +90,6 @@ class ChangeStream implements Iterator
 
     /**
      * @internal
-     * @param ChangeStreamIterator $iterator
-     * @param callable             $resumeCallable
      */
     public function __construct(ChangeStreamIterator $iterator, callable $resumeCallable)
     {
@@ -194,8 +192,6 @@ class ChangeStream implements Iterator
      * Determines if an exception is a resumable error.
      *
      * @see https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst#resumable-error
-     * @param RuntimeException $exception
-     * @return boolean
      */
     private function isResumableError(RuntimeException $exception): bool
     {
@@ -251,8 +247,6 @@ class ChangeStream implements Iterator
 
     /**
      * Recreates the ChangeStreamIterator after a resumable server error.
-     *
-     * @return void
      */
     private function resume(): void
     {
@@ -265,7 +259,6 @@ class ChangeStream implements Iterator
     /**
      * Either resumes after a resumable error or re-throws the exception.
      *
-     * @param RuntimeException $exception
      * @throws RuntimeException
      */
     private function resumeOrThrow(RuntimeException $exception): void
--- a/src/ChangeStream.php
+++ b/src/ChangeStream.php
@@ -197,7 +197,7 @@ class ChangeStream implements Iterator
      * @param RuntimeException $exception
      * @return boolean
      */
-    private function isResumableError(RuntimeException $exception)
+    private function isResumableError(RuntimeException $exception): bool
     {
         if ($exception instanceof ConnectionException) {
             return true;
@@ -224,7 +224,7 @@ class ChangeStream implements Iterator
      * @param boolean $incrementKey Increment $key if there is a current result
      * @throws ResumeTokenException
      */
-    private function onIteration(bool $incrementKey)
+    private function onIteration(bool $incrementKey): void
     {
         /* If the cursorId is 0, the server has invalidated the cursor and we
          * will never perform another getMore nor need to resume since any
@@ -254,7 +254,7 @@ class ChangeStream implements Iterator
      *
      * @return void
      */
-    private function resume()
+    private function resume(): void
     {
         $this->iterator = call_user_func($this->resumeCallable, $this->getResumeToken(), $this->hasAdvanced);
         $this->iterator->rewind();
@@ -268,7 +268,7 @@ class ChangeStream implements Iterator
      * @param RuntimeException $exception
      * @throws RuntimeException
      */
-    private function resumeOrThrow(RuntimeException $exception)
+    private function resumeOrThrow(RuntimeException $exception): void
     {
         if ($this->isResumableError($exception)) {
             $this->resume();
--- a/src/Client.php
+++ b/src/Client.php
@@ -290,7 +290,6 @@ class Client
      * List databases.
      *
      * @see ListDatabases::__construct() for supported options
-     * @param array $options
      * @return DatabaseInfoIterator
      * @throws UnexpectedValueException if the command response was malformed
      * @throws InvalidArgumentException for parameter/option parsing errors
--- a/src/Client.php
+++ b/src/Client.php
@@ -116,7 +116,7 @@ class Client
 
         $driverOptions['driver'] = $this->mergeDriverInfo($driverOptions['driver'] ?? []);
 
-        $this->uri = (string) $uri;
+        $this->uri = $uri;
         $this->typeMap = $driverOptions['typeMap'] ?? null;
 
         unset($driverOptions['typeMap']);
--- a/src/Client.php
+++ b/src/Client.php
@@ -98,7 +98,7 @@ class Client
      * @throws DriverInvalidArgumentException for parameter/option parsing errors in the driver
      * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
      */
-    public function __construct($uri = 'mongodb://127.0.0.1/', array $uriOptions = [], array $driverOptions = [])
+    public function __construct(string $uri = 'mongodb://127.0.0.1/', array $uriOptions = [], array $driverOptions = [])
     {
         $driverOptions += ['typeMap' => self::$defaultTypeMap];
 
@@ -155,7 +155,7 @@ class Client
      * @param string $databaseName Name of the database to select
      * @return Database
      */
-    public function __get($databaseName)
+    public function __get(string $databaseName)
     {
         return $this->selectDatabase($databaseName);
     }
@@ -201,7 +201,7 @@ class Client
      * @throws InvalidArgumentException for parameter/option parsing errors
      * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
      */
-    public function dropDatabase($databaseName, array $options = [])
+    public function dropDatabase(string $databaseName, array $options = [])
     {
         if (! isset($options['typeMap'])) {
             $options['typeMap'] = $this->typeMap;
@@ -314,7 +314,7 @@ class Client
      * @return Collection
      * @throws InvalidArgumentException for parameter/option parsing errors
      */
-    public function selectCollection($databaseName, $collectionName, array $options = [])
+    public function selectCollection(string $databaseName, string $collectionName, array $options = [])
     {
         $options += ['typeMap' => $this->typeMap];
 
@@ -330,7 +330,7 @@ class Client
      * @return Database
      * @throws InvalidArgumentException for parameter/option parsing errors
      */
-    public function selectDatabase($databaseName, array $options = [])
+    public function selectDatabase(string $databaseName, array $options = [])
     {
         $options += ['typeMap' => $this->typeMap];
 
--- a/src/Command/ListDatabases.php
+++ b/src/Command/ListDatabases.php
@@ -99,7 +99,6 @@ class ListDatabases implements Executable
      * Execute the operation.
      *
      * @see Executable::execute()
-     * @param Server $server
      * @return array An array of database info structures
      * @throws UnexpectedValueException if the command response was malformed
      * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
@@ -119,8 +118,6 @@ class ListDatabases implements Executable
 
     /**
      * Create the listDatabases command.
-     *
-     * @return Command
      */
     private function createCommand(): Command
     {
--- a/src/Command/ListDatabases.php
+++ b/src/Command/ListDatabases.php
@@ -103,7 +103,7 @@ class ListDatabases implements Executable
      * @throws UnexpectedValueException if the command response was malformed
      * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
      */
-    public function execute(Server $server)
+    public function execute(Server $server): array
     {
         $cursor = $server->executeReadCommand('admin', $this->createCommand(), $this->createOptions());
         $cursor->setTypeMap(['root' => 'array', 'document' => 'array']);
--- a/src/Command/ListDatabases.php
+++ b/src/Command/ListDatabases.php
@@ -122,7 +122,7 @@ class ListDatabases implements Executable
      *
      * @return Command
      */
-    private function createCommand()
+    private function createCommand(): Command
     {
         $cmd = ['listDatabases' => 1];
 
@@ -148,7 +148,7 @@ class ListDatabases implements Executable
      * @see https://php.net/manual/en/mongodb-driver-server.executecommand.php
      * @return array
      */
-    private function createOptions()
+    private function createOptions(): array
     {
         $options = [];
 
--- a/src/Command/ListDatabases.php
+++ b/src/Command/ListDatabases.php
@@ -143,7 +143,6 @@ class ListDatabases implements Executable
      * the command be executed on the primary.
      *
      * @see https://php.net/manual/en/mongodb-driver-server.executecommand.php
-     * @return array
      */
     private function createOptions(): array
     {
--- a/src/Exception/InvalidArgumentException.php
+++ b/src/Exception/InvalidArgumentException.php
@@ -36,7 +36,7 @@ class InvalidArgumentException extends DriverInvalidArgumentException implements
      * @param string|string[] $expectedType Expected type
      * @return self
      */
-    public static function invalidType($name, $value, $expectedType)
+    public static function invalidType(string $name, $value, $expectedType)
     {
         if (is_array($expectedType)) {
             switch (count($expectedType)) {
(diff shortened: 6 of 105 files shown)


phpcs.xml.dist
```diff
@@ -105,8 +105,6 @@
     <!-- **************************************************************************** -->
     <rule ref="SlevomatCodingStandard.TypeHints.ParameterTypeHint">
         <properties>
-            <!-- Requires PHP 7.4 -->
-            <property name="enableObjectTypeHint" value="false" />
```
alcaeus, line 109: The comment was wrong - the `object` type was introduced in PHP 7.2.

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -32,28 +33,28 @@ class CollectionFunctionalTest extends FunctionalTestCase
...
         // TODO: Move to unit test once ManagerInterface can be mocked (PHPC-378)
         new Collection($this->manager, $this->getDatabaseName(), $collectionName);
     }
 
     public function provideInvalidDatabaseAndCollectionNames()
     {
         return [
-            [null],
-            [''],
+            [null, TypeError::class],
```
alcaeus, line 56: Adding the expected exception type to the dataset was necessary as `null` is now caught by PHP itself, while the empty string is caught by our logic.

tests/Model/ChangeStreamIteratorTest.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
```
alcaeus, line 3: This is necessary to disable type coercion in tests. Without this, the non-int test values `3.14` and `true` would be silently coerced to integers, which is not what we're expecting to test here. Alternatively, we can omit this `declare` and skip testing values that we know can be coerced.

phpcs.xml.dist
```diff
@@ -133,8 +131,6 @@
 
     <rule ref="SlevomatCodingStandard.TypeHints.ReturnTypeHint">
         <properties>
-            <!-- Requires PHP 7.2 -->
-            <property name="enableObjectTypeHint" value="false" />
```
jmikola, line 137: In this case, the comment above was always accurate (in contrast to the `ParameterTypeHint` rule above) and we're just removing this because the library requires 7.2+, correct?

phpcs.xml.dist
```diff
@@ -162,12 +158,9 @@
     </rule>
 
 
-    <!-- *********************************************************************************** -->
-    <!-- Require native type hints for all parameters, properties, and return types in tests -->
-    <!-- *********************************************************************************** -->
-    <rule ref="SlevomatCodingStandard.TypeHints.ParameterTypeHint.MissingNativeTypeHint">
-        <exclude-pattern>src</exclude-pattern>
-    </rule>
```
jmikola, line 170: In this case, we can remove `ParameterTypeHint` but `PropertyTypeHint` needs to stay because it's [PHP 7.4+](https://php.watch/versions/7.4/typed-properties), correct? Should we add a `<!-- Requires PHP 7.4 -->` comment below?

src/ChangeStream.php
```diff
@@ -197,7 +197,7 @@ public function valid()
      * @param RuntimeException $exception
      * @return boolean
      */
-    private function isResumableError(RuntimeException $exception)
+    private function isResumableError(RuntimeException $exception): bool
```
jmikola, line 200: Does it make sense to remove redundant doc block `@param` and `@return` lines that don't add any value? I think we'd just leave them in place if particular parameters have associated comments.

src/ChangeStream.php
```diff
@@ -254,7 +254,7 @@ private function onIteration($incrementKey)
      *
      * @return void
      */
-    private function resume()
+    private function resume(): void
```
jmikola, line 257: This is an example of where `@return void` above can simply be removed.

src/Client.php
```diff
@@ -155,7 +155,7 @@ public function __debugInfo()
      * @param string $databaseName Name of the database to select
      * @return Database
      */
-    public function __get($databaseName)
+    public function __get(string $databaseName)
```
jmikola, line 158: Noted that the [magic methods use a type hint](https://www.php.net/manual/en/language.oop5.overloading.php#object.get). I assume this might be problematic if they did not, unless it'd be permitted by type narrowing anyway.

src/Exception/InvalidArgumentException.php
```diff
@@ -36,7 +36,7 @@ class InvalidArgumentException extends DriverInvalidArgumentException implements
      * @param string|string[] $expectedType Expected type
      * @return self
      */
-    public static function invalidType($name, $value, $expectedType)
+    public static function invalidType(string $name, $value, $expectedType)
```
jmikola, line 39: Noted that we can't use union types for `$expectedType`, as that [requires PHP 8.0](https://php.watch/versions/8.0/union-types).

src/GridFS/Bucket.php
```diff
@@ -620,7 +620,7 @@ public function rename($id, $newFilename)
      * @throws StreamException if the file could not be uploaded
      * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
      */
-    public function uploadFromStream($filename, $source, array $options = [])
+    public function uploadFromStream(string $filename, $source, array $options = [])
```
jmikola, line 623: Noted that there's no `resource` type hint per https://wiki.php.net/rfc/resource_typehint.

src/GridFS/Bucket.php
```diff
@@ -695,7 +695,7 @@ private function getFilesNamespace()
      * @return stdClass
      * @throws InvalidArgumentException
      */
-    private function getRawFileDocumentForStream($stream)
+    private function getRawFileDocumentForStream($stream): stdClass
```
jmikola, line 698: I realize this is a `private` method so it's not really a concern here, but if we ever think about adding `stdClass` as a return type hint for a public method I think we should consider using `object` just to keep our options open. I'm thinking about `ClientEncryption::rewrapManyDataKey()` as I write this :)

src/Model/CollectionInfoCommandIterator.php
```diff
@@ -37,10 +37,7 @@ class CollectionInfoCommandIterator extends IteratorIterator implements Collecti
     /** @var string|null */
     private $databaseName;
 
-    /**
-     * @param string|null $databaseName
-     */
-    public function __construct(Traversable $iterator, $databaseName = null)
+    public function __construct(Traversable $iterator, ?string $databaseName = null)
```
jmikola, line 40: Noting that this was a case where you removed a redundant doc block.

src/Operation/ListIndexes.php
```diff
@@ -73,7 +73,7 @@ class ListIndexes implements Executable
      * @param array  $options        Command options
      * @throws InvalidArgumentException for parameter/option parsing errors
      */
-    public function __construct($databaseName, $collectionName, array $options = [])
+    public function __construct(string $databaseName, string $collectionName, array $options = [])
```
jmikola, line 76: It just occurred to me that after adding `string` type hints to `$databaseName` and `$collectionName` (among other parameters), there's an opportunity to clean up the explicit casting that often follows. For instance, the end of this constructor has:

```php
$this->collectionName = (string) $collectionName;
```

Maybe do a search for `(string)` or `(int)` casts throughout the library and see if any of those are now redundant.

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -32,28 +33,28 @@ class CollectionFunctionalTest extends FunctionalTestCase
...
+        $this->expectException($expectedExceptionClass);
         // TODO: Move to unit test once ManagerInterface can be mocked (PHPC-378)
         new Collection($this->manager, $databaseName, $this->getCollectionName());
     }
 
     /**
      * @dataProvider provideInvalidDatabaseAndCollectionNames
      */
-    public function testConstructorCollectionNameArgument($collectionName): void
+    public function testConstructorCollectionNameArgument($collectionName, string $expectedExceptionClass): void
```
jmikola, line 46: I just realized that the data provider here should be named `provideInvalidCollectionNames` instead. I don't recall if an older version of this test took a database name, but it certainly doesn't now.

tests/Database/DatabaseFunctionalTest.php
```diff
@@ -23,18 +24,18 @@ class DatabaseFunctionalTest extends FunctionalTestCase
...
         new Database($this->manager, $databaseName);
     }
 
     public function provideInvalidDatabaseNames()
     {
         return [
-            [null],
-            [''],
+            [null, TypeError::class],
+            ['', InvalidArgumentException::class],
```
jmikola, line 38: Would it make sense to move these `provideInvalidDatabaseNames` and `provideInvalidCollectionNames` data providers up to a base test class? That would remove the need to redefine them in all of these files. Could also punt this to a future code cleanup ticket (great for onboarding), as it's not really relevant to this PR.

tests/Model/ChangeStreamIteratorTest.php
jmikola: This explanation makes sense, but let's record it in a comment above `declare()`.

tests/Operation/DeleteTest.php
```diff
@@ -1,11 +1,12 @@
 <?php
 
+declare(strict_types=1);
```
jmikola, line 3: I assume this is the same reason as `ChangeStreamIteratorTest.php`, so this may warrant making a concise comment that you can copy-paste between files.

jmikola (commented):
Were most of these code changes manually done, or is this something `phpcbf` helped with after updating `phpcs.xml.dist`?

phpcs.xml.dist
alcaeus: That is correct, we just didn't touch this when dropping support for PHP < 7.2.

phpcs.xml.dist
alcaeus: The rules below not only require PHP 7.4 but would also cause BC breaks which is why they're disabled here. Using the native property type is controlled in the actual rule definition above - this section here silences the error even if that was enabled.

src/Client.php
alcaeus: That is correct - except that for method parameters type narrowing is not allowed, only widening ;)

src/GridFS/Bucket.php
alcaeus: Absolutely - I'd have to take another look (and will do so as part of the psalm integration), but this is one case where we may want to change it to `object` anyways.

tests/Collection/CollectionFunctionalTest.php
alcaeus: `provideInvalidDatabaseAndCollectionNames` is a single data provider for both database and collection names that is used in both `testConstructorDatabaseNameArgument` and `testConstructorCollectionNameArgument` - I suppose that's why the data provider name includes both "database" and "collection".

tests/Model/ChangeStreamIteratorTest.php
alcaeus: Done here and in the other test that uses this.

src/ChangeStream.php
alcaeus: I've enabled the `UselessAnnotation` checks for parameter, property, and return types. `phpcbf` removed a number of useless annotations and will ensure these are no longer added.

src/Operation/ListIndexes.php
alcaeus: Good point. I removed these using PhpStorm, which offers an inspection for unnecessary type casts.

alcaeus:
> Were most of these code changes manually done, or is this something `phpcbf` helped with after updating `phpcs.xml.dist`?

It was a mix of `phpcbf` and manual changes. Return types were added through manual labour as phpcs can't be configured to only enforce return types on private members or final classes.

src/GridFS/Bucket.php
jmikola: No objections if you'd like to just use `object` here.

tests/Database/DatabaseFunctionalTest.php
jmikola: Disregard, I misunderstood how these were being used.

tests/Collection/CollectionFunctionalTest.php
jmikola: Oh, you're right. I misunderstood how these were being used and thought they were appearing in various Operation tests instead of just the Database and Collection functional tests. All good here.

src/BulkWriteResult.php
```diff
@@ -35,8 +35,7 @@ class BulkWriteResult
     private $isAcknowledged;
 
     /**
-     * @param WriteResult $writeResult
-     * @param mixed[]     $insertedIds
+     * @param mixed[] $insertedIds
```
jmikola, line 38: I'm not sure if tooling left this behind, but it seems odd to only have _some_ parameters documented here. Is that intentional?

As-is, I don't think `mixed[]` adds any clarification to a generic `array` type hint, so maybe we just want to remove this entirely. Up to you.

src/Command/ListDatabases.php
```diff
@@ -99,7 +99,6 @@ public function __construct(array $options = [])
      * Execute the operation.
      *
      * @see Executable::execute()
-     * @param Server $server
      * @return array An array of database info structures
```
jmikola, line 102: Any reason not to use an `array` return type hint on `execute()` here? Correct me if I'm wrong, but I think that'd be permitted by [covariance](https://www.php.net/manual/en/language.oop5.variance.php).

If this is out of scope for this PR, feel free to ignore.

src/InsertManyResult.php
```diff
@@ -35,8 +35,7 @@ class InsertManyResult
     private $isAcknowledged;
 
     /**
-     * @param WriteResult $writeResult
-     * @param mixed[]     $insertedIds
+     * @param mixed[] $insertedIds
```
jmikola, line 38: Same comment as I left in BulkWriteResult if you'd like to remove this line, too.

src/InsertOneResult.php
```diff
@@ -35,8 +35,7 @@ class InsertOneResult
     private $isAcknowledged;
 
     /**
-     * @param WriteResult $writeResult
-     * @param mixed       $insertedId
+     * @param mixed $insertedId
```
jmikola, line 38: Feel free to remove this if you like.

src/MapReduceResult.php
```diff
@@ -79,7 +79,7 @@ public function getCounts()
      */
     public function getExecutionTimeMS()
     {
-        return (integer) $this->executionTimeMS;
+        return $this->executionTimeMS;
```
jmikola, line 82: The constructor will always set `$this->executionTimeMS` to an integer. Should we use an `int` return type on this method and remove the `@return` annotation?

Similarly, `getCounts()` and `getTiming()` can use an `array` return type and omit their `@return` annotations.

src/Model/BSONDocument.php
```diff
@@ -50,11 +50,9 @@ public function __clone()
      * by default.
      *
      * @see https://php.net/arrayobject.construct
-     * @param array   $input
-     * @param integer $flags
-     * @param string  $iteratorClass
+     * @param array $input
```
jmikola, line 53: This looks like it can be removed.

src/Operation/Aggregate.php
```diff
@@ -246,8 +246,8 @@ public function __construct($databaseName, $collectionName, array $pipeline, arr
             unset($options['batchSize']);
         }
 
-        $this->databaseName = (string) $databaseName;
-        $this->collectionName = isset($collectionName) ? (string) $collectionName : null;
+        $this->databaseName = $databaseName;
+        $this->collectionName = $collectionName ?? null;
```
jmikola, line 250: If the `$collectionName` param is now a nullable string, I think we can omit `??` here for the same effect. The ternary was only necessary to restrict the `(string)` cast for non-null values.

src/Operation/Aggregate.php
```diff
@@ -320,7 +318,7 @@ public function getCommandDocument(Server $server)
      *
      * @return array
      */
-    private function createCommandDocument()
+    private function createCommandDocument(): array
```
jmikola, line 321: `@return` can be removed. Should we do the same above with `getCommandDocument` (currently has a `@return` annotation but not actual return type hint)?

src/Operation/Count.php
```diff
@@ -183,7 +181,7 @@ public function getCommandDocument(Server $server)
      *
      * @return array
      */
-    private function createCommandDocument()
+    private function createCommandDocument(): array
```
jmikola, line 184: I noticed `getCommandDocument()` above also has no return type hint. If this is due to the Explainable interface not using the return type hint (likely a BC break even though we define the class as `@internal`), I think we'd still be permitted to use `array` in implementing classes per covariance rules. If so, you may want to just make this change en masse.

src/Operation/DatabaseCommand.php
```diff
@@ -92,7 +92,6 @@ public function __construct($databaseName, $command, array $options = [])
      * Execute the operation.
      *
      * @see Executable::execute()
-     * @param Server $server
      * @return Cursor
      */
     public function execute(Server $server)
```
jmikola, line 97: I know `Executable::execute()` doesn't define a return type hint (BC yada yada), but I imagine we can still use return type hints in the implementations per covariance if you like.

tests/SpecTests/Operation.php
```diff
@@ -562,9 +556,7 @@ private function executeForGridFSBucket(Bucket $bucket, Context $context)
     /**
      * Executes the session operation and return its result.
      *
-     * @param Session            $session
-     * @param FunctionalTestCase $test
-     * @param Context            $context Execution context
+     * @param Context $context Execution context
```
jmikola, line 559: Feel free to remove `* @param Context $context Execution context` everywhere if you don't think it adds value.

`@return mixed` can go as well if you don't think it's helpful.

tests/SpecTests/Operation.php
```diff
@@ -838,7 +827,6 @@ private function getResultAssertionTypeForDatabase()
     /**
      * Prepares a request element for a bulkWrite operation.
      *
-     * @param stdClass $request
      * @return array
```
jmikola, line 830: `@return array` is redundant. This may have just been missed by tooling, as I see it left behind for a lot of methods that already have an `array` return type hint.

tests/SpecTests/ResultExpectation.php
```diff
@@ -46,8 +46,7 @@ final class ResultExpectation
     private $assertionCallable;
 
     /**
-     * @param integer $assertionType
-     * @param mixed   $expectedValue
+     * @param mixed $expectedValue
```
jmikola, line 49: Not relevant. If you'd like to just search for `@param mixed` and delete most of those I won't object (no pun intended).

tests/SpecTests/FunctionalTestCase.php
```diff
@@ -301,10 +293,7 @@ private function isServerlessRequirementSatisfied(?string $serverlessMode): bool
     /**
      * Checks if server version and topology requirements are satifised.
      *
-     * @param string|null $minServerVersion
-     * @param string|null $maxServerVersion
-     * @param array|null  $topologies
-     * @return boolean
+     * @param array|null $topologies
```
jmikola, line 296: Seems redundant given `?array $topologies = null` below.

tests/SpecTests/FunctionalTestCase.php
```diff
@@ -229,8 +223,7 @@ protected function dropTestAndOutcomeCollections(array $testCollectionDropOption
     /**
      * Insert data fixtures into the test collection.
      *
-     * @param array       $documents
-     * @param string|null $collectionName
+     * @param array $documents
```
jmikola, line 226: Redundant. If you want to search for `@param array` and cross-reference with array type hints in method signatures, that might reveal other things you can remove.

tests/Operation/CreateIndexesFunctionalTest.php
```diff
@@ -223,9 +223,6 @@ public function testCommitQuorumUnsupported(): void
      * argument as its first and only parameter. If an IndexInfo matching the
      * given name is found, it will be passed to the callback, which may perform
      * additional assertions.
-     *
-     * @param string   $indexName
-     * @param callable $callback
```
jmikola, line 228: This was a good removal, as the `@param` should have used `callable|null`.

src/Operation/Watch.php
```diff
@@ -317,7 +317,6 @@ final public function commandSucceeded(CommandSucceededEvent $event): void
      * Execute the operation.
      *
      * @see Executable::execute()
-     * @param Server $server
      * @return ChangeStream
```
jmikola, line 320: Consider adding a `ChangeStream` return type to `execute()` and removing this.

src/Operation/Watch.php
```diff
@@ -264,7 +264,7 @@ public function __construct(Manager $manager, $databaseName, $collectionName, ar
 
         $this->manager = $manager;
         $this->databaseName = (string) $databaseName;
```
jmikola, line 266: I don't think this `(string)` cast is necessary given the `if ($databaseName === null)` conditional above that sets it to "admin" for cluster-wide change streams. Feel free to remove.

src/Operation/Watch.php
```diff
@@ -264,7 +264,7 @@ public function __construct(Manager $manager, $databaseName, $collectionName, ar
...
-        $this->collectionName = isset($collectionName) ? (string) $collectionName : null;
+        $this->collectionName = $collectionName ?? null;
```
jmikola, line 267: `?? null` can be removed here. The previous `(string)` cast and ternary was only used to avoid turning a null value into an empty string.

src/Operation/ListDatabaseNames.php
```diff
@@ -72,7 +72,6 @@ public function __construct(array $options = [])
      * Execute the operation.
      *
      * @see Executable::execute()
-     * @param Server $server
      * @return Iterator
```
jmikola, line 75: `execute()` here already uses an `Iterator` type hint. This will probably come up when searching for useless `@return` annotations that only include a single type. Unions might still be relevant if we use those anywhere.

jmikola (commented):
`getCommandDocument` and `execute()` methods in Operations can probably get their doc blocks stripped down and return types added in some places. I think covariance on the interface allows for that.

`@param` annotations with single types (e.g. array, mixed) that don't have a description (or a meaningful description) of the parameter can probably be removed. The single type is probably already in the method signature. Likewise for `|null` stuff.

Likewise for `@return` annotations.

Lastly, I understand if you'd like to avoid adding type hints to the Executable and Explainable interfaces, even though they're documented as `@internal`. Perhaps we should have a ticket to add type hints and schedule it for 2.0, though.

src/GridFS/Bucket.php
alcaeus: Changed this to `object` to avoid relying on `stdClass`.

src/BulkWriteResult.php
alcaeus: I replaced `mixed[]` with a generic `array`, you're right it doesn't make any sense.

src/Operation/Aggregate.php
alcaeus: Good point. Removed the unnecessary fallback to `null`.

src/Command/ListDatabases.php
alcaeus: Covariance would indeed allow us to add an `array` type here. However, backward compatibility promise won't allow us to do so, as anyone extending `ListDatabases` would be encountering fatal errors. Granted, it doesn't make much sense extending these classes, but they are all marked as part of the API and non-final, so we can't add return types to any public or protected methods.

src/InsertOneResult.php
alcaeus: Removed this and a number of other superfluous type annotations.

src/MapReduceResult.php
alcaeus: BC promise prevents us from adding a return type here unfortunately.

src/Model/BSONDocument.php
alcaeus: Removed this and a number of other superfluous type annotations.

src/Operation/Aggregate.php
alcaeus: Removed this and a number of other superfluous type annotations.

As for `getCommandDocument`, that method is part of the public API so the return type will have to wait until 2.0.

src/Operation/Count.php
alcaeus: The `Explainable` interface not having a return type is no problem, as covariance would allow us to add it. However, despite the interface being marked as `@internal`, the classes implementing the interface are not, and adding the return type to the interface would require adding it to the implementing classes as well. This again would constitute a BC break.

I'm not sure how the `@internal` is intended to be read: do we not expect users to implement the interface in their own classes, or do we treat the entire contract as internal for users not to rely on? If it's the latter, I would suggest fixing this in 2.0 by making the method `final` in implementing classes.

src/Operation/DatabaseCommand.php
alcaeus: Similar to the situation with `Explainable`, adding a return type here would constitute a BC break.

tests/SpecTests/Operation.php
alcaeus: Removed this and a number of other superfluous type annotations.

tests/SpecTests/Operation.php
alcaeus: Removed this and a number of other superfluous type annotations.

tests/SpecTests/ResultExpectation.php
alcaeus: Removed this and a number of other superfluous type annotations.

tests/SpecTests/FunctionalTestCase.php
alcaeus: Removed this and a number of other superfluous type annotations.

tests/SpecTests/FunctionalTestCase.php
alcaeus: Removed this and a number of other superfluous type annotations.

src/Operation/Watch.php
alcaeus: See above - it would constitute a BC break.

src/Operation/Watch.php
alcaeus: You are correct - for some reason the PhpStorm tooling still saw the cast as required and didn't remove it automatically. I removed this.

src/Operation/Watch.php
alcaeus: Good point - removed this.

src/Operation/ListDatabaseNames.php
alcaeus: This class is also protected by the BC promise, so return types will have to wait until 2.0.

alcaeus:
@jmikola you are correct - it looks like tooling did indeed leave most (if not all) `@return array` and `@param array` type annotations in place. I've removed them manually where applicable.

As for `@param mixed` and `@return mixed`, these are currently required to let the tooling know that the parameter expects any type (as opposed to "we don't know what it accepts) and returns a mixed value (as opposed to "we don't know what it returns") respectively. As such, removing them will cause complaints from the tooling. On the other hand, leaving them in place now will also help us when we drop PHP < 8 and can add the `mixed` pseudo type as an actual type.

Last but not least, most of our classes are neither `final` nor `@internal`, so the number of return types we can add is severely limited. In addition to that, the concept of tentative return types added in PHP 8.1 is only available to built-in and extension classes, but not to user land code. This means that we don't have a nice mechanism of telling people that the return type of a method will change in a future version. Making these classes `final` is something we should re-evaluate for 2.0.

src/Command/ListDatabases.php
jmikola: Oh good point about extension. I didn't consider that.

Shame on us for not making these things `final` by default. Somewhere, @Ocramius is silently shaking his head in disappointment.

src/Operation/Count.php
jmikola: I don't think there's any rules for `@internal`. It was a convention from the [old phpDocumentor days](https://pear.php.net/package/PhpDocumentor/docs/latest/phpDocumentor/tutorial_tags.internal.pkg.html).

PHPUnit goes a step further to clarify this with:

```
@internal This class is not covered by the backward compatibility promise for PHPUnit
```

The idea for PHPLIB was to allow us flexibility to ignore BC breaks on `@internal` functions/classes, but I dropped the ball by never considering BC breaks could still happen through the implementing classes. Granted, this was also before return type hints were usable for us.

This is all a good learning experience that should form the basis for a 2.0 ticket.

jmikola (approved):
Thanks for all the detailed work on this. I'll leave you to create some 2.0 tickets on your own time.

PR #962, opened 2022-08-30
PHPLIB-942: Fix pattern for matching phpinfo() values

jmikola:
https://jira.mongodb.org/browse/PHPLIB-942


jmikola:
Note: CI failures are due to [PHPLIB-930](https://jira.mongodb.org/browse/PHPLIB-930).

PR #961, opened 2022-08-29
PHPLIB-912, PHPLIB-914, PHPLIB-924: Sync change stream spec tests

jmikola:
https://jira.mongodb.org/browse/PHPLIB-912
https://jira.mongodb.org/browse/PHPLIB-914
https://jira.mongodb.org/browse/PHPLIB-924

This can serve as a POC for https://github.com/mongodb/specifications/pull/1293. It also includes two other PHPLIB tickets for previously updated change stream spec tests, neither of which required driver changes.

Tested locally using server 6.1.0-alpha-2444-gfb80d4b, which I grabbed from [download-mongodb.sh](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/download-mongodb.sh).

PR #963, opened 2022-08-31
PHPLIB-918: Add test that reads are not retried in a transaction

levon80999:
https://jira.mongodb.org/browse/PHPLIB-918

Synced with https://github.com/mongodb/specifications/commit/c5d1c778dba6439de72e6a3dbbaeab9de7539306

### DIFF
--- a/tests/UnifiedSpecTests/transactions/retryable-abort-handshake.json
+++ b/tests/UnifiedSpecTests/transactions/retryable-abort-handshake.json
@@ -1,0 +1,204 @@
+{
+  "description": "retryable abortTransaction on handshake errors",
+  "schemaVersion": "1.4",
+  "runOnRequirements": [
+    {
+      "minServerVersion": "4.2",
+      "topologies": [
+        "replicaset",
+        "sharded",
+        "load-balanced"
+      ],
+      "serverless": "forbid",
+      "auth": true
+    }
+  ],
+  "createEntities": [
+    {
+      "client": {
+        "id": "client0",
+        "useMultipleMongoses": false,
+        "observeEvents": [
+          "commandStartedEvent",
+          "connectionCheckOutStartedEvent"
+        ]
+      }
+    },
+    {
+      "database": {
+        "id": "database0",
+        "client": "client0",
+        "databaseName": "retryable-handshake-tests"
+      }
+    },
+    {
+      "collection": {
+        "id": "collection0",
+        "database": "database0",
+        "collectionName": "coll"
+      }
+    },
+    {
+      "session": {
+        "id": "session0",
+        "client": "client0"
+      }
+    },
+    {
+      "session": {
+        "id": "session1",
+        "client": "client0"
+      }
+    }
+  ],
+  "initialData": [
+    {
+      "collectionName": "coll",
+      "databaseName": "retryable-handshake-tests",
+      "documents": [
+        {
+          "_id": 1,
+          "x": 11
+        }
+      ]
+    }
+  ],
+  "tests": [
+    {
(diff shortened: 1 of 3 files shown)


tests/UnifiedSpecTests/transactions/retryable-abort-handshake.json
```diff
@@ -0,0 +1,204 @@
...
+          "_id": 1,
+          "x": 11
+        }
+      ]
+    }
+  ],
+  "tests": [
+    {
+      "description": "AbortTransaction succeeds after handshake network error",
+      "skipReason": "DRIVERS-2032: Pinned servers need to be checked if they are still selectable",
```
jmikola, line 69: I noted from the downstream changes in [DRIVERS-746](https://jira.mongodb.org/browse/DRIVERS-746) that both of the "retryable" tests are skipped, so it was not necessary to add these to `UnifiedSpecTest::$incompleteTests`.

jmikola (changes requested):
For consistency in the future, please use `PHPLIB-###: ` as a commit prefix instead of wrapping the issue in brackets. I revised the PR title, which should be sufficient to ensure the correct message is used when squash-merging.

The commit message is missing a comment indicating the [mongodb/specifications](https://github.com/mongodb/specifications) commit to which you've synced. When we discussed this earlier, I referenced the commits in https://github.com/mongodb/mongo-php-library/pull/961 as an example. Please add the following to your commit message:

```
Synced with mongodb/specifications@<full commit hash>
```

GitHub will then render that as a short link in its UI, as in https://github.com/mongodb/mongo-php-library/commit/f8eff8dd5b039ea34b1379907d76ca47e5113250.

Feel free to amend your commit and force-push to your PR branch.

tests/UnifiedSpecTests/transactions/retryable-abort-handshake.json
levon80999: I noted it too and that is why I didn't add new lines in UnifiedSpecTest::$incompleteTests. Also I double checked the [DRIVERS-2032](https://jira.mongodb.org/browse/DRIVERS-2032) ticket which is the parent ticket of [DRIVERS-746](https://jira.mongodb.org/browse/DRIVERS-746).

jmikola (approved):
CI failures are unrelated. I believe they are actually both fixed in master, but your PR branch is based on an older master commit. For example, the [tests/UnifiedSpecTests/change-streams](https://github.com/mongodb/mongo-php-library/tree/03735ac5841e18eecc4e77510c05960224152af7/tests/UnifiedSpecTests/change-streams) directory corresponding to this PR's commit doesn't show any recent modifications to the two failing change stream tests, which were recently updated in https://github.com/mongodb/mongo-php-library/commit/a292d0786a3fd31e3b455624df6da956db23ab46 and https://github.com/mongodb/mongo-php-library/commit/f8eff8dd5b039ea34b1379907d76ca47e5113250. For future reference, I suggest rebasing your PR branch on master regularly if you're already comfortable with force-pushing. It will limit conflicts (granted, that's not an issue here) and help avoid unrelated CI failures that might have been recently addressed.

Please just remove the full issue URL from the commit message when squash-merging, as I mentioned in https://github.com/mongodb/mongo-php-library/pull/967#issuecomment-1234570317.

jmikola:
@levon80999: FYI, the merge commit (https://github.com/mongodb/mongo-php-library/commit/4bc4607a185db5f52ea3a24dbe36cd25112d00e8) was missing the following:

```
Synced with mongodb/specifications/commit/c5d1c778dba6439de72e6a3dbbaeab9de7539306
```

PR #973, opened 2022-09-14
PHPLIB-610: Psalm Integration

alcaeus:
PHPLIB-610

This adds psalm to our CI pipeline to better check our code for typing errors. All existing code passes psalm checks on error level 3, while new code is expected to adhere to level 1. More information about error levels can be found in the [psalm documentation](https://psalm.dev/docs/running_psalm/error_levels/).


src/functions.php
```diff
@@ -505,8 +506,9 @@ function extract_read_preference_from_options(array $options): ?ReadPreference
 function select_server(Manager $manager, array $options): Server
 {
     $session = extract_session_from_options($options);
-    if ($session instanceof Session && $session->getServer() !== null) {
-        return $session->getServer();
+    $server = $session instanceof Session ? $session->getServer() : null;
+    if ($server !== null) {
+        return $server;
```
alcaeus, line 511: Interestingly enough, psalm was quite right to complain about this. Even if the first call to `getServer()` in the conditional returns a `Server` instance, it's not guaranteed that the second call in the `return` statement will return the same `Server` object, as the session could've been unpinned in between evaluating the conditional and executing the `return`. In that case, `getServer()` would return `null` which would be an invalid return value triggering a fatal error. (I'll leave the exercise of proving me wrong to the curious reader).

src/functions.php
```diff
@@ -561,5 +563,7 @@ function select_server_for_aggregate_write_stage(Manager $manager, array &$optio
         throw $serverSelectionError;
     }
 
+    assert($server instanceof Server);
```
alcaeus, line 566: Necessary as psalm trips over our fancy logic above. We can assume that if the original `select_server` call didn't return a server, `$serverSelectionError` would be set, so the `return` wouldn't be reached. This logic seems to be too advanced for psalm, so using this to silence the error.

src/Model/ChangeStreamIterator.php
```diff
@@ -143,6 +135,14 @@ public function current()
         return $this->isValid ? parent::current() : null;
     }
 
+    final public function getInnerIterator(): Cursor
```
alcaeus, line 138: Adding this method to reduce the number of assertions needed in the `ChangeStream` model class. Strictly speaking any instance of `ChangeStreamIterator` would return a cursor when calling `getInnerIterator()`, as that's enforced through the constructor. However, as the class isn't final and there are no covariance checks for constructors, a user could potentially sidestep this.

The `assert` call is to satisfy psalm, and when assertion checking is disabled a type error will be thrown. I believe this is better than letting a user run into an error about calling a method that doesn't exist. Let me know if you'd rather remove this method and instead use assertions in `ChangeStream`.

Note that this method can be removed when making the class final, which we should do in 2.0.

.github/workflows/coding-standards.yml
```diff
@@ -5,14 +5,16 @@ on:
     branches:
       - "v*.*"
       - "master"
+      - "feature/*"
```
stof, line 8: Do you really want to run the CI on PR *targetting* a feature branch ?

.github/workflows/coding-standards.yml
```diff
@@ -5,14 +5,16 @@ on:
...
   push:
     branches:
       - "v*.*"
       - "master"
+      - "feature/*"
```
stof, line 13: As you use such branches to open PRs, this will trigger 2 CI builds for each push (one for the push to the branch and one for the update to the PR).

composer.json
```diff
@@ -19,7 +19,8 @@
     "require-dev": {
         "squizlabs/php_codesniffer": "^3.6",
         "doctrine/coding-standard": "^9.0",
-        "symfony/phpunit-bridge": "^5.2"
+        "symfony/phpunit-bridge": "^5.2",
+        "vimeo/psalm": "^4.x-dev"
```
stof, line 23: is it expected that it does not allow using stable releases ?

.github/workflows/coding-standards.yml
alcaeus: Sure, if such a PR were to exist.

.github/workflows/coding-standards.yml
alcaeus: I push my branches to GitHub usually long before opening a pull request. This allows me to test a wider range of environments than I cover locally (before pushing) and ensuring things are stable and done before opening a pull request.

jmikola: IIRC, we have only ever used feature branches a handful of times when a large PR was ready but we weren't completely done with the JIRA epic (e.g. SDAM Monitoring in PHPC). In that case, we opened PRs against a feature branch from arbitrary branches on our forks and the feature branch was a waiting area before we could merge the functionality down the `master`.

I don't expect we will push directly to feature branches during development, so this shouldn't invite more CI builds than our typical workflow of PRs against `master` and release branches. @alcaeus: please correct me if I'm mistaken.

src/GridFS/CollectionWrapper.php
```diff
@@ -150,14 +152,17 @@ public function findFileByFilenameAndRevision(string $filename, int $revision):
...
-        return $this->filesCollection->findOne(
+        $file = $this->filesCollection->findOne(
             ['filename' => $filename],
             [
                 'skip' => $skip,
                 'sort' => ['uploadDate' => $sortOrder],
                 'typeMap' => ['root' => 'stdClass'],
             ]
         );
+        assert(is_object($file) || $file === null);
```
jmikola, line 163: I assume this is another case where psalm can't infer how the typeMap will be honored, correct?

src/GridFS/Exception/CorruptFileException.php
```diff
@@ -23,6 +23,14 @@
 
 class CorruptFileException extends RuntimeException
 {
+    /**
+     * Thrown when a chunk doesn't contain valid data.
+     */
+    public static function invalidChunkData(int $chunkIndex): self
```
jmikola, line 29: Noted that this is used by `ReadableStream::initBufferFromCurrentChunk()`.

src/GridFS/ReadableStream.php
```diff
@@ -250,16 +255,25 @@ private function initBufferFromCurrentChunk(): bool
...
         $currentChunk = $this->chunksIterator->current();
+        assert(is_object($currentChunk));
 
         if ($currentChunk->n !== $this->chunkOffset) {
             throw CorruptFileException::unexpectedIndex($currentChunk->n, $this->chunkOffset);
         }
 
+        if (! $currentChunk->data instanceof Binary) {
+            throw CorruptFileException::invalidChunkData($this->chunkOffset);
+        }
```
jmikola, line 275: This seems like a reasonable check that would not have been gracefully handled previously. It's quite possible someone could corrupt the `data` field in a chunk document since MongoDB doesn't enforce a schema on that collection.

src/GridFS/ReadableStream.php
```diff
@@ -48,13 +51,13 @@ class ReadableStream
     /** @var integer */
     private $chunkOffset = 0;
 
-    /** @var CursorInterface|null */
```
jmikola, line 51: This dates back to https://github.com/mongodb/mongo-php-library/commit/3c6b5f2999574d3e8d479195363a7ee241e1a584#diff-15b1c36af88c266bd8e54d1a2d3b409e6e29076f89c75f4dae0ec1c8538a787eR51 from [PHPLIB-598](https://jira.mongodb.org/browse/PHPLIB-598).

When the original IteratorIterator hint was replaced with CursorInterface, this became inconsistent with how we utilize `$chunksIterator` later in the class. CursorInterface provides a few methods (e.g. `setTypeMap()`) and extends Traversable, which wouldn't have guaranteed the iterator methods we later invoke.

[PHPC-1123](https://jira.mongodb.org/browse/PHPC-1123) added CursorInterface in 1.6.0 but the iterator methods didn't arrive until [PHPC-1691](https://jira.mongodb.org/browse/PHPC-1691) in 1.9.0 (when PHP 8.0 required them). And PHPLIB-598 was downstream work for PHPC-1691.

I suppose the original CursorInterface hint was just a mistake and PHPLIB-598 should have used Cursor. Pretty neat that we're just now catching this thanks to psalm.

src/GridFS/ReadableStream.php
```diff
@@ -48,13 +51,13 @@ class ReadableStream
...
+    /** @var Cursor|null */
     private $chunksIterator;
 
     /** @var CollectionWrapper */
     private $collectionWrapper;
 
-    /** @var float|integer */
+    /** @var integer */
```
jmikola, line 60: Good catch. Since we already asserted that the file length was an integer and `$expectedLastChunkSize` is calculated as the difference of `length` and `numChunks` (integer-cast result of `ceil()`), this should always have an integer type.

src/GridFS/ReadableStream.php
```diff
@@ -164,6 +167,8 @@ public function readBytes(int $length): string
             return '';
         }
 
+        assert($this->buffer !== null);
```
jmikola, line 170: Is this necessary to silence a psalm error?

If `initBufferFromCurrentChunk()` doesn't return `false`, `$this->buffer` is initialized from `Binary::getData()` and should never be null. I'd expect psalm to pick up on that.

src/GridFS/WritableStream.php
```diff
@@ -65,7 +66,7 @@ class WritableStream
     /** @var array */
     private $file;
 
-    /** @var resource */
+    /** @var HashContext|null */
```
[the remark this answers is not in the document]
jmikola: Ah, very convenient that we don't have to worry about PHP <7.2 anymore.

src/GridFS/WritableStream.php
```diff
@@ -262,7 +263,7 @@ private function fileCollectionInsert()
         $this->file['length'] = $this->length;
         $this->file['uploadDate'] = new UTCDateTime();
 
-        if (! $this->disableMD5) {
+        if (! $this->disableMD5 && $this->hashCtx) {
```
jmikola, line 266: Is this needed for psalm? The constructor always initializes `hashCtx` with `hash_init()` when `disableMD5` is `false`, so I'd expect this wouldn't be necessary unless psalm can't infer that `hash_init()` always returns a valid object.

That aside, should we use `!== null` or `instanceof` here instead of relying on a truthy cast?

src/Model/ChangeStreamIterator.php
```diff
@@ -109,7 +101,7 @@ final public function commandStarted(CommandStartedEvent $event): void
         }
 
         $this->batchPosition = 0;
-        $this->batchSize = null;
+        $this->batchSize = 0;
```
jmikola, line 104: Noted that this would have left `batchSize` null only if the command failed since `commandSucceeded()` sets it; however, there's no real need to do that since it violates the typing. Also, it has no functional impact on how this is used by `isAtEndOfBatch()`.

src/Model/ChangeStreamIterator.php
jmikola: Doesn't this only affect `ChangeStream::getCursorId()`? I don't see any other place where `getInnerIterator()` is called in the library apart from two instances in `WatchFunctionalTest`.

This is also interesting because ChangeStreamIterator extends IteratorIterator and enforces that its constructor is always a Cursor instance. I suppose psalm cannot infer that `IteratorIterator::getInnerIterator()` will always return an instance of _its_ constructor argument without additional templating.

No objection to leaving this method in place, but I think a doc block above that briefly explains why it's needed and that it can potentially be removed in 2.0 would be helpful. You can also make a ticket with a 2.0 fixVersion to remind us to remove this down the line.

src/Operation/Aggregate.php
```diff
@@ -295,7 +294,7 @@ public function execute(Server $server)
 
         $result = current($cursor->toArray());
 
-        if (! isset($result->result) || ! is_array($result->result)) {
+        if (! is_object($result) || ! isset($result->result) || ! is_array($result->result)) {
```
jmikola, line 297: Does psalm complain about using a property accessor on a potential non-object even if PHP would ignore it thanks to `isset()`?

src/Operation/DatabaseCommand.php
```diff
@@ -38,7 +38,7 @@ class DatabaseCommand implements Executable
     /** @var string */
     private $databaseName;
 
-    /** @var array|Command|object */
```
jmikola, line 41: This was originally added in https://github.com/mongodb/mongo-php-library/commit/4a764feef33ac21549055e84bd41e4b5820244ca#diff-75b1bd6340fef4dfe6e4d23cb317e77f242e2a7f8344ba6c911b80e9fe87f119R40, but there's no code path where `$command` would be anything other than a Command. I suppose that change was made based on the constructor signature. Good catch.

src/Operation/DropCollection.php
```diff
@@ -19,6 +19,7 @@
 
 use MongoDB\Driver\Command;
 use MongoDB\Driver\Exception\CommandException;
+use MongoDB\Driver\Exception\RuntimeException as DriverRuntimeException;
```
jmikola, line 22: Interesting that `phpcbf` didn't catch this. It's only referenced from a `@throws` annotation.

src/Operation/FindAndModify.php
```diff
@@ -253,7 +253,7 @@ public function execute(Server $server)
 
         $result = current($cursor->toArray());
 
-        return $result->value ?? null;
+        return is_object($result) ? $result->value ?? null : null;
```
jmikola, line 256: I know PHP doesn't require it, but this helps readability without having to remember operator precedence rules.

```suggestion
        return is_object($result) ? ($result->value ?? null) : null;
```

Alternatively, since this is one of many cases where psalm presumably doesn't like us using `->` on potential non-objects within an `isset()` context, perhaps there's a config option that we can use to ignore this? I don't feel strongly about it, so up to you if you're comfortable with the extra `is_object` calls.

src/functions.php
jmikola: Since `mongoc_client_session_get_server_id()` returns the `server_id` property of the `mongoc_client_session_t` struct, I'm not sure how you'd be able to interrupt this without some hacking approach like runtime debugging or `register_tick_function()`.

That said, avoiding a second call to `getServer()` seems entirely reasonable.

src/functions.php
jmikola: Works for me. Since this differs from some other instances where you needed to add `assert()` in that psalm fails to detect a potential assignment via `catch`, I wonder if they might be interested in a bug report to address this.

jmikola (approved):
A minor suggestion or two and some questions. Excellent work!

src/GridFS/Bucket.php
```diff
@@ -412,6 +413,7 @@ public function getFileIdForStream($stream)
          */
         $typeMap = ['root' => 'stdClass'] + $this->typeMap;
         $file = apply_type_map_to_document($file, $typeMap);
+        assert(is_object($file));
```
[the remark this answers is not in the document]
alcaeus: > I assume this is necessary because psalm only knows that `apply_type_map_to_document()` returns `array|object` and can't infer that the `root` type map will ensure it's a `stdClass` instance. I assume the assertion is enough to silence psalm's complaints.

Yes. The typeMap argument can be defined better (e.g. `array{?root: class-string<T>}`), which should then allow us to use `T` as a return type for the method. I haven't yet played around with it and how it would behave when the given typeMap doesn't contain a `root` element.

> This is the first instance of `assert()` in the library. Is this kosher because of the new handling for PHP 7+ code (see: [Expectations (PHP 7 only)](https://www.php.net/manual/en/function.assert.php#function.assert.expectations))?

Exactly. In development, the `assert` would trip if we're receiving anything other than an object, whilst in a typical production configuration it's entirely optimised away (i.e. doesn't generate an opcode) so has zero performance impact.

I've made a note to double-check the value of `zend.assertions` in our CI processes to ensure we're using development mode.

src/ChangeStream.php
```diff
@@ -250,7 +252,15 @@ private function onIteration(bool $incrementKey): void
...
     private function resume(): void
     {
+        if (! $this->resumeCallable) {
+            throw new BadMethodCallException('Cannot resume a closed change stream.');
+        }
+
         $this->iterator = call_user_func($this->resumeCallable, $this->getResumeToken(), $this->hasAdvanced);
+        if (! $this->iterator instanceof ChangeStreamIterator) {
+            throw new UnexpectedValueException('Expected change stream iterator from callable');
+        }
```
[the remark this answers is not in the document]
alcaeus: Very good point. I've added a new psalm type definition that is then referenced later on:
```php
/**
 * @psalm-type ResumeCallable = callable(array|object|null, bool): ChangeStreamIterator
 */

// ...

/** @var ResumeCallable|null */
private $resumeCallable;
```

This in fact removes the exception and has the added benefit of having psalm shout at us if we do indeed pass an invalid callable 👍

src/ChangeStream.php
```diff
@@ -250,7 +252,15 @@ private function onIteration(bool $incrementKey): void
      */
     private function resume(): void
     {
+        if (! $this->resumeCallable) {
+            throw new BadMethodCallException('Cannot resume a closed change stream.');
+        }
```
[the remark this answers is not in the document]
alcaeus: Yes - that has been my experience whenever adding strict static analysis: the errors are quite unlikely to happen, but you know that they'd be head scratchers if a user ever presented you with a bug report where they ran into the issue 😅

src/DeleteResult.php
```diff
@@ -43,7 +43,7 @@ public function __construct(WriteResult $writeResult)
      * This method should only be called if the write was acknowledged.
      *
      * @see DeleteResult::isAcknowledged()
-     * @return integer
+     * @return integer|null
```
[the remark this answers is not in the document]
alcaeus: Unfortunately, `WriteResult::getDeletedCount` may return `null` - like all other classes in `WriteResult`. This is due to the logic in [`PHONGO_WRITERESULT_RETURN_LONG_FROM_BSON_INT32`](https://github.com/mongodb/mongo-php-driver/blob/f0421fa1fe7437780adc0bb914b13c24e6d28cd0/src/MongoDB/WriteResult.c#L35): if the result document doesn't contain the key, the macro will not initialise the return value, thus returning `null`. This is one of those instances where we relied on a non-null return type. We can update the logic in 1.15 to return `0` if no value was found (which would make sense), which would also allow us to revert the change. As is, psalm complains about this (which is how I spotted the entire thing).

src/GridFS/CollectionWrapper.php
alcaeus: That is correct, yes.

src/GridFS/WritableStream.php
stof: I don't think Psalm can be smart enough to detect that you never mutate `disableMD5è and to keep a relation between type refinements for that (it might be able to do that with a readonly property when it knows it cannot be mutated, but I don't expect it to ever be able to handle that for mutable properties)

src/Operation/DropCollection.php
stof: I don't think phpcbf validates that `@throws` references a valid class. It probably cannot do it, as it works only on a single file IIRC.

.github/workflows/coding-standards.yml
stof: @jmikola the current PR is opened from a feature branch...

src/Operation/Aggregate.php
alcaeus: That is correct, here the corresponding psalm error:

ERROR: PossiblyInvalidPropertyFetch - src/Operation/Aggregate.php:297:22 - Cannot fetch property on possible non-object `$result` of type `array<array-key, mixed>` (see https://psalm.dev/114)
```
        if ( ! isset($result->result) || ! is_array($result->result)) {
```

I've asked to figure out whether this is a bug or intended behaviour. Either way, this issue goes away once we properly handle the type map in our methods.

src/Operation/FindAndModify.php
alcaeus: Good point. We could technically ignore the error and add it to the baseline, deferring it to later when we handle type maps correctly. For now I applied your suggestion to make the code more readable 👍

src/GridFS/Bucket.php
```diff
@@ -644,6 +646,7 @@ public function uploadFromStream(string $filename, $source, array $options = [])
     private function createPathForFile(object $file): string
     {
         if (! is_object($file->_id) || method_exists($file->_id, '__toString')) {
+            /** @psalm-suppress PossiblyInvalidCast */
```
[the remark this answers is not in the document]
alcaeus: The full error is:
> ERROR: PossiblyInvalidCast - src/GridFS/Bucket.php:649:28 - object cannot be cast to string (see https://psalm.dev/190)

I rewrote the conditional to avoid the error:
```php
if (is_array($file->_id) || (is_object($file->_id) && ! method_exists($file->_id, '__toString'))) {
    $id = toJSON(fromPHP(['_id' => $file->_id]));
} else {
    $id = (string) $file->_id;
}
```

I believe this covers all cases: arrays and objects without a `__toString` method are passed to `fromPHP`, while all other values are cast to string.

src/functions.php
alcaeus: I also checked whether this is expected behaviour based on a simplified example: https://psalm.dev/r/b1102ead61

psalm.xml
```diff
@@ -0,0 +1,16 @@
...
+    xmlns="https://getpsalm.org/schema/config"
+    xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
+>
+    <projectFiles>
+        <directory name="src" />
+        <ignoreFiles>
+            <directory name="vendor" />
+        </ignoreFiles>
+    </projectFiles>
+</psalm>
```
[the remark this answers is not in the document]
alcaeus: Good point, I didn't consider that. Psalm does indeed support a list file, so I moved configuration to `psalm.xml.dist` and added `psalm.xml` to `.gitignore` as suggested 👍

src/Model/ChangeStreamIterator.php
alcaeus: Yes and no. While psalm 

> Doesn't this only affect `ChangeStream::getCursorId()`? I don't see any other place where `getInnerIterator()` is called in the library apart from two instances in `WatchFunctionalTest`.

That is correct.

> This is also interesting because ChangeStreamIterator extends IteratorIterator and enforces that its constructor is always a Cursor instance. I suppose psalm cannot infer that `IteratorIterator::getInnerIterator()` will always return an instance of _its_ constructor argument without additional templating.

Yes and no. While this is certainly true within the context of `ChangeStreamIterator`, technically an extending class could skip calling the parent constructor and go straight to `IteratorIterator::__construct`: https://3v4l.org/iir32. In that case, our assumption that the inner iterator is always a cursor no longer holds true. Yes, it is very much theoretical, but then that's what psalm was made for.

> No objection to leaving this method in place, but I think a doc block above that briefly explains why it's needed and that it can potentially be removed in 2.0 would be helpful. You can also make a ticket with a 2.0 fixVersion to remind us to remove this down the line.

Added 👍

.github/workflows/coding-standards.yml
alcaeus: To summarise what we're currently testing and why it's useful:

PRs to any versioned branches (`v*.*`), the master branch and any feature branch (`feature/*`) should be tested, as we're using them to ensure we're only merging good code. The latter will not occur very often, but it could if we ever decide to build a larger feature in multiple PRs but don't want any logic in master before it's completely done.

Pushes to versioned branches and the master branch are important in the main repo. Pushes to feature branches will not occur very often here, however they do occur in forks (where we do our development work before creating a PR to the main repo). For quick PRs where we don't want to know if the checks are stable we can omit the `feature/` prefix to skip CI before making a PR.

Enabling builds on push is also done for the benefit of outside contributors: this allows them to see CI results before creating a pull request even when they don't have a good test environment set up locally.

src/GridFS/Bucket.php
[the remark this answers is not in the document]
jmikola: Oh, so this has nothing to do with an array. Psalm just had trouble understanding the inverted conditional. SGTM.

src/functions.php
jmikola: Nice that they have a web-based tool available. Is that example suitable to open a bug report? I assume this isn't desired behavior.

Beyond the scope of this PR to be sure, so no need to follow up here.

.github/workflows/coding-standards.yml
jmikola: > @jmikola the current PR is opened from a feature branch...

Totally missed that. @alcaeus and I talked about this today and I noted our workflows differ. I only recalled using feature branches on the mongodb repo when we wanted to merge down larger PRs without putting them into master. Andreas explained this is more for his fork, which inherits the workflow configuration.

He was last looking at whether some repo config options were available to conserve CI resources, so I'll defer to him on this.

src/DeleteResult.php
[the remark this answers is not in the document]
jmikola: Ah, I totally missed that when looking at the code locally. Confirmed the stubs _and_ PHP.net docs both correctly report `?int` return types. I opened [PHPC-2142](https://jira.mongodb.org/browse/PHPC-2142) to track this.

src/GridFS/ReadableStream.php
jmikola: Feel free to link the ticket here when you create it for cross-referencing purposes.

src/GridFS/WritableStream.php
jmikola: > any method call could reset it to null even with disableMD5 being false

Ah, and even though it doesn't happen in WritableStream the class isn't final so it theoretically _could_ happen in a child class. Similar explanation as some other questions -- I think I'm understanding the logic behind these warnings a bit better now. :+1:

jmikola (approved):

src/functions.php
alcaeus: https://github.com/vimeo/psalm/issues/8361 looks related and I added the snippet there. We'll see what comes of it.

composer.json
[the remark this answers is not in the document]
alcaeus: Filed PHPLIB-984 to track updating this.

PR #981, opened 2022-09-28
PHPLIB-987: Add more examples

alcaeus:
PHPLIB-987

This PR adds more examples for every-day use of the PHP library. New examples:
* bulk writes (this shows the syntax for each bulk write operation)
* `with_transaction` helper
* Aggregation pipeline
* Serialisation and Deserialisation using `MongoDB\BSON\Persistable`
* Deserialisation using `typeMap` and `MongoDB\BSON\Unserializable`

The first commit in the PR fixes an issue where running the example code from outside of the examples directory causes fatal errors.

### DIFF
--- a/examples/aggregate.php
+++ b/examples/aggregate.php
@@ -1,0 +1,59 @@
+<?php
+declare(strict_types=1);
+
+namespace MongoDB\Examples;
+
+use MongoDB\Client;
+
+use function assert;
+use function dirname;
+use function getenv;
+use function is_object;
+use function MongoDB\BSON\fromPHP;
+use function MongoDB\BSON\toRelaxedExtendedJSON;
+use function printf;
+use function rand;
+
+require dirname(__FILE__) . '/../vendor/autoload.php';
+
+function toJSON(object $document): string
+{
+    return toRelaxedExtendedJSON(fromPHP($document));
+}
+
+$client = new Client(getenv('MONGODB_URI') ?: 'mongodb://127.0.0.1/');
+
+$collection = $client->test->coll;
+$collection->drop();
+
+$documents = [];
+
+for ($i = 0; $i < 100; $i++) {
+    $documents[] = ['randomValue' => rand(0, 1000)];
+}
+
+$collection->insertMany($documents);
+
+$pipeline = [
+    [
+        '$group' => [
+            '_id' => null,
+            'totalCount' => ['$sum' => 1],
+            'evenCount' => [
+                '$sum' => ['$mod' => ['$randomValue', 2]],
+            ],
+            'oddCount' => [
+                '$sum' => ['$subtract' => [1, ['$mod' => ['$randomValue', 2]]]],
+            ],
+            'maxValue' => ['$max' => '$randomValue'],
+            'minValue' => ['$min' => '$randomValue'],
+        ],
+    ],
+];
+
+$cursor = $collection->aggregate($pipeline, ['batchSize' => 2]);
+
+foreach ($cursor as $document) {
+    assert(is_object($document));
+    printf("%s\n", toJSON($document));
+}
--- a/examples/aggregate.php
+++ b/examples/aggregate.php
@@ -51,7 +51,7 @@ $pipeline = [
     ],
 ];
 
-$cursor = $collection->aggregate($pipeline, ['batchSize' => 2]);
+$cursor = $collection->aggregate($pipeline);
 
 foreach ($cursor as $document) {
     assert(is_object($document));
--- a/examples/aggregate.php
+++ b/examples/aggregate.php
@@ -6,7 +6,6 @@ namespace MongoDB\Examples;
 use MongoDB\Client;
 
 use function assert;
-use function dirname;
 use function getenv;
 use function is_object;
 use function MongoDB\BSON\fromPHP;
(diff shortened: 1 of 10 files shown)


examples/aggregate.php
```diff
@@ -0,0 +1,59 @@
...
+            'oddCount' => [
+                '$sum' => ['$subtract' => [1, ['$mod' => ['$randomValue', 2]]]],
+            ],
+            'maxValue' => ['$max' => '$randomValue'],
+            'minValue' => ['$min' => '$randomValue'],
+        ],
+    ],
+];
+
+$cursor = $collection->aggregate($pipeline, ['batchSize' => 2]);
```
jmikola, line 54: Is `batchSize` relevant here (also in `bulk.php` and `with_transaction.php`)? This might be copied from the APM example, where I specifically did it to demonstrate `getMore` commands.

examples/aggregate.php
```diff
@@ -0,0 +1,59 @@
...
+use function assert;
+use function dirname;
+use function getenv;
+use function is_object;
+use function MongoDB\BSON\fromPHP;
+use function MongoDB\BSON\toRelaxedExtendedJSON;
+use function printf;
+use function rand;
+
+require dirname(__FILE__) . '/../vendor/autoload.php';
```
jmikola, line 17: `__DIR__` should be a suitable replacement for `dirname(__FILE__)` and spare us the additional `use function` import.

See: https://www.php.net/manual/en/language.constants.magic.php`

examples/persistable.php
```diff
@@ -0,0 +1,151 @@
...
+    {
+        return $this->name;
+    }
+
+    public function setName(string $name): void
+    {
+        $this->name = $name;
+    }
+
+    public function getEmails(): array
```
jmikola, line 51: Should this utilize a doc block like the `private $emails` member for Psalm to enforce the return type as a templated array, or is that unnecessary?

examples/persistable.php
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    public function addEmail(PersistableEmail $email): void
+    {
+        $this->emails[] = $email;
+    }
+
+    public function deleteEmail(PersistableEmail $email): void
+    {
+        $index = array_search($email, $this->emails, true);
```
jmikola, line 63: I noted that you used `$strict=true` here. Do you intend for this to only delete a specific PersistableEmail instance? This method doesn't appear to be used in the example script so it's hard to say what the intended use case is.

At a glance (and before I realized PersistableEmail was a document), I was going to point out that `addEmail` and `deleteEmail` seemed inconsistent, since `addEmail` appends unconditionally (and could add duplicates).

examples/persistable.php
```diff
@@ -0,0 +1,151 @@
...
+
+$collection = $client->test->coll;
+$collection->drop();
+
+$collection->insertOne($entry);
+
+$foundEntry = $collection->findOne([]);
+
+/** @psalm-suppress ForbiddenCode */
+var_dump($foundEntry);
```
jmikola, line 151: TIL https://psalm.dev/docs/running_psalm/issues/ForbiddenCode/

examples/typemap.php
```diff
@@ -0,0 +1,121 @@
...
+use UnexpectedValueException;
+
+use function dirname;
+use function getenv;
+use function is_array;
+use function var_dump;
+
+require dirname(__FILE__) . '/../vendor/autoload.php';
+
+class TypemapEntry implements Unserializable
```
jmikola, line 18: A minor nit, but perhaps you want to use `TypeMapEntry` and `TypeMapEmail` for consistency with `$typeMap`. I don't feel strongly either way, but just pointing it out in case you didn't give it any thought.

examples/with_transaction.php
```diff
@@ -0,0 +1,56 @@
...
+use function printf;
+
+require dirname(__FILE__) . '/../vendor/autoload.php';
+
+function toJSON(object $document): string
+{
+    return toRelaxedExtendedJSON(fromPHP($document));
+}
+
+$client = new Client(getenv('MONGODB_URI') ?: 'mongodb://127.0.0.1/');
```
jmikola, line 25: Would it be worth adding a comment here and in `changestream.php` to note that both features require replication? Maybe that's not necessary as the majority of users will likely just read/copy the code instead of execute them directly.

examples/persistable.php
alcaeus: Yeah, I attempted to make this look like a business object. I think I'll go back and make the properties public and remove the accessors.

examples/persistable.php
alcaeus: Same here - I was surprised when I saw the psalm error

examples/typemap.php
alcaeus: I can never remember the correct spelling. Changed to TypeMap.

examples/with_transaction.php
alcaeus: Good point, done.

examples/aggregate.php
alcaeus: Ah, yes. I've removed the batchSize option.

examples/aggregate.php
alcaeus: Done.

examples/bulk.php
```diff
@@ -71,7 +70,7 @@ function toJSON(object $document): string
     ]
 );
 
-$cursor = $collection->find([], ['batchSize' => 2]);
+$cursor = $collection->find([]);
```
jmikola, line 73: FWIW, `Collection::find()` can be called without arguments as `$filter` defaults to an empty array. This technically violates the CRUD spec but we snuck that past the goalie long ago. Up to you if you'd like to keep this as-is or change to `find()`.

Likewise in `with_transaction.php`.

jmikola (approved):

PR #983, opened 2022-09-29
PHPLIB-638: Ensure writeConcernError is reported for findAndModify

levon80999:
https://jira.mongodb.org/browse/PHPLIB-638

### DIFF
--- a/tests/Operation/FindAndModifyFunctionalTest.php
+++ b/tests/Operation/FindAndModifyFunctionalTest.php
@@ -3,6 +3,7 @@
 namespace MongoDB\Tests\Operation;
 
 use MongoDB\Driver\BulkWrite;
+use MongoDB\Driver\Exception\CommandException;
 use MongoDB\Driver\ReadPreference;
 use MongoDB\Driver\WriteConcern;
 use MongoDB\Exception\UnsupportedException;
@@ -88,6 +89,25 @@ class FindAndModifyFunctionalTest extends FunctionalTestCase
 
         $this->expectException(UnsupportedException::class);
         $this->expectExceptionMessage('Hint is not supported by the server executing this operation');
+
+        $operation->execute($this->getPrimaryServer());
+    }
+
+    public function testFindAndModifyReportedWriteConcernError(): void
+    {
+        if (($this->isShardedCluster() && ! $this->isShardedClusterUsingReplicasets()) || ! $this->isReplicaSet()) {
+            $this->markTestSkipped('Test only applies to replica sets');
+        }
+
+        $this->expectException(CommandException::class);
+        $this->expectExceptionCode(100);
+        $this->expectExceptionMessageMatches('/Write Concern error:/');
+
+        $operation = new FindAndModify(
+            $this->getDatabaseName(),
+            $this->getCollectionName(),
+            ['remove' => true, 'writeConcern' => new WriteConcern(50)]
+        );
 
         $operation->execute($this->getPrimaryServer());
     }


tests/Operation/FindAndModifyFunctionalTest.php
```diff
@@ -92,6 +93,23 @@ public function testHintOptionAndUnacknowledgedWriteConcernUnsupportedClientSide
         $operation->execute($this->getPrimaryServer());
     }
 
+    public function testFindAndModifyReportedWriteConcernError(): void
```
alcaeus, line 96: This test should be restricted to replica sets, as the exception message will differ on standalone servers (see failed test).

tests/Operation/FindAndModifyFunctionalTest.php
```diff
@@ -92,6 +93,23 @@ public function testHintOptionAndUnacknowledgedWriteConcernUnsupportedClientSide
...
+        try {
+            $operation = new FindAndModify(
+                $this->getDatabaseName(),
+                $this->getCollectionName(),
+                ['remove' => true, 'writeConcern' => new WriteConcern(50)]
+            );
+
+            $operation->execute($this->getPrimaryServer());
+        } catch (CommandException $e) {
+            $this->assertInstanceOf(CommandException::class, $e);
```
alcaeus, line 107: This check is unnecessary - any class that isn't a descendant of `CommandException` won't have been caught here.

At the same time, we can simplify this by using the `expectException`, `expectExceptionMessage`, and `expectExceptionCode` methods of PHPUnit.

alcaeus (changes requested):

tests/Operation/FindAndModifyFunctionalTest.php
```diff
@@ -95,19 +96,21 @@ public function testHintOptionAndUnacknowledgedWriteConcernUnsupportedClientSide
...
-                ['remove' => true, 'writeConcern' => new WriteConcern(50)]
-            );
-
-            $operation->execute($this->getPrimaryServer());
-        } catch (CommandException $e) {
-            $this->assertInstanceOf(CommandException::class, $e);
-            $this->assertStringContainsString('Write Concern error', $e->getMessage());
-            $this->assertSame(100, $e->getCode());
+        if (($this->isShardedCluster() && !$this->isShardedClusterUsingReplicasets()) || !$this->isReplicaSet()) {
+            $this->markTestSkipped(sprintf('Cannot find topology for primary of type "%d".', $this->getPrimaryServer()->getType()));
```
alcaeus, line 99:
```suggestion
            $this->markTestSkipped('Test only applies to replica sets');
```

tests/Operation/FindAndModifyFunctionalTest.php
levon80999: done

tests/Operation/FindAndModifyFunctionalTest.php
levon80999: I decide to use `expectExceptionMessageMatches` instead of `expectExceptionMessage` because I need to check that message contains `Write Concern Error` not all message. Is it ok ?

alcaeus (approved):

tests/Operation/FindAndModifyFunctionalTest.php
```diff
@@ -92,6 +93,25 @@ public function testHintOptionAndUnacknowledgedWriteConcernUnsupportedClientSide
...
     }
 
+    public function testFindAndModifyReportedWriteConcernError(): void
+    {
+        if (($this->isShardedCluster() && ! $this->isShardedClusterUsingReplicasets()) || ! $this->isReplicaSet()) {
+            $this->markTestSkipped('Test only applies to replica sets');
+        }
+
+        $this->expectException(CommandException::class);
+        $this->expectExceptionCode(100);
```
jmikola, line 103: Note: this is "UnsatisfiableWriteConcern" per https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.yml

jmikola: https://github.com/mongodb/mongo-php-library/pull/984

PR #985, opened 2022-10-04
PHPLIB-601: Document upcoming typing changes

alcaeus:
PHPLIB-601

I've added an `UPGRADE-1.14` document to inform users of changes they should be aware of when upgrading. Once changes for 2.0 become clearer (including a list of methods that will change their return types), we should add an `UPGRADE-2.0` document to inform users of those.

@jmikola I've renamed the previous upgrade guide in the docs to "Legacy driver upgrade guide", in case we want to add a more visible documentation of method signature changes once we add BC breaks in 2.0. I've skipped a list of changes for this PR, but can list method signature changes in the upgrade document as well. An example of this is the [upgrade document for doctrine/collections](https://github.com/doctrine/collections/blob/2.0.x/UPGRADE.md#doctrinecommoncollectionscollection) where interfaces were changed to add types.

We'll want to consider where to document upcoming return type changes for ext-mongodb. While users on PHP 8.1 and newer will be informed of tentative return types, users on previous version will not be aware of those changes. At the same time, as most users will consume the extension through this library (but may interact with extension classes directly), I'm not sure what the best place to document those changes are.

For now, I believe this document is sufficient to inform users of changes we've already made, as well as upcoming changes to internal classes.

### DIFF
--- a/UPGRADE-1.14.md
+++ b/UPGRADE-1.14.md
@@ -1,0 +1,23 @@
+# UPGRADE FROM 1.x to 1.14
+
+## Method signature changes
+
+### Parameter types
+
+Starting with 1.14, methods now declare types for their arguments. This will not
+cause BC breaks unless you've passed a type that was incompatible with the type
+documented in the PHPDoc comment. 
+
+### Return types
+
+Return types will be added in version 2.0. These types are documented in a
+PHPDoc comment, which will become the new return type. You can prepare for this
+change (which will trigger a BC break in any class you may extend) by adding the
+correct return type to your class at this time.
+
+## Internal classes
+
+Internal classes will become final where possible in a future release. At the
+same time, we will add return types to these internal classes. Note that
+internal classes are not covered by our backward compatibility promise, and you
+should not instantiate such classes directly.
--- a/UPGRADE-1.15.md
+++ b/UPGRADE-1.15.md
@@ -6,7 +6,8 @@
 
 Starting with 1.14, methods now declare types for their arguments. This will not
 cause BC breaks unless you've passed a type that was incompatible with the type
-documented in the PHPDoc comment. 
+documented in the PHPDoc comment. A list of changes can be found at the bottom
+of this document.
 
 ### Return types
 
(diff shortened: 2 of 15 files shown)


UPGRADE-1.14.md
```diff
@@ -0,0 +1,23 @@
+# UPGRADE FROM 1.x to 1.14
+
+## Method signature changes
```
jmikola, line 3: > I've skipped a list of changes for this PR, but can list method signature changes in the upgrade document as well. An example of this is the [upgrade document for doctrine/collections](https://github.com/doctrine/collections/blob/2.0.x/UPGRADE.md#doctrinecommoncollectionscollection) where interfaces were changed to add types.

That format looks good to me.

> We'll want to consider where to document upcoming return type changes for ext-mongodb. While users on PHP 8.1 and newer will be informed of tentative return types, users on previous version will not be aware of those changes.

Two thoughts here. Does it make sense to document all signature changes in the PHPC method changelogs? That seems like the expected place for it and would certainly be more accessible than a separate page in the PHP.net docs that might be easily overlooked.

I also wouldn't object to including PHPC API changes in this document, since those are going to arrive in the corresponding PHPC release (1.15 for PHPLIB 1.14).

alcaeus: > Does it make sense to document all signature changes in the PHPC method changelogs?

Yes. I've already added a note about tentative return types (see https://github.com/php/doc-en/commit/734bafeaf071b78b15d375f9af583befddd8c2a2), but calling out other changes (such as widening parameter types, etc.) should also be called out. I'll go through the changes and make sure that those are updated.

There aren't any other significant API changes for users, as we've only changed final classes in PHPC. However, it will become relevant for 2.0, at which point I'd definitely point out PHPC API changes in this upgrade document as well where it affects users.

UPGRADE-1.15.md
```diff
@@ -0,0 +1,65 @@
...
+internal classes are not covered by our backward compatibility promise, and you
+should not instantiate such classes directly.
+
+## Method signature changes by class
+
+### MongoDB\Client
+
+|                                                                                           1.13 | 1.15                                                                                  |
+|-----------------------------------------------------------------------------------------------:|:--------------------------------------------------------------------------------------|
+| `__construct($uri = 'mongodb://127.0.0.1', array $uriOptions = [], array $driverOptions = []`) | `__construct(?string $uri = null, array $uriOptions = [], array $driverOptions = [])` |
```
jmikola, line 32:
```suggestion
| `__construct($uri = 'mongodb://127.0.0.1', array $uriOptions = [], array $driverOptions = [])` | `__construct(?string $uri = null, array $uriOptions = [], array $driverOptions = [])` |
```

UPGRADE-1.15.md
```diff
@@ -0,0 +1,65 @@
...
+### Return types
+
+Return types will be added in version 2.0. These types are documented in a
+PHPDoc comment, which will become the new return type. You can prepare for this
+change (which will trigger a BC break in any class you may extend) by adding the
+correct return type to your class at this time.
+
+## Internal classes
+
+Internal classes will become final where possible in a future release. At the
```
jmikola, line 21: Maybe mention 
```suggestion
Internal classes (i.e. annotated with `@internal`) will become final where possible in a future release. At the
```

UPGRADE-1.15.md
```diff
@@ -0,0 +1,65 @@
+# UPGRADE FROM 1.x to 1.15
+
+## Method signature changes
+
+### Parameter types
+
+Starting with 1.14, methods now declare types for their arguments. This will not
```
jmikola, line 7:
```suggestion
Starting with 1.15, methods now declare types for their arguments. This will not
```

UPGRADE-1.15.md
```diff
@@ -0,0 +1,65 @@
...
+cause BC breaks unless you've passed a type that was incompatible with the type
+documented in the PHPDoc comment. A list of changes can be found at the bottom
```
jmikola, line 9:
```suggestion
previously documented in the PHPDoc comment. A list of changes can be found at the bottom
```

Feel free to ignore if this goes without saying, but I think this was the the intended message.

UPGRADE-1.15.md
```diff
@@ -0,0 +1,65 @@
...
+of this document.
+
+### Return types
+
+Return types will be added in version 2.0. These types are documented in a
+PHPDoc comment, which will become the new return type. You can prepare for this
```
jmikola, line 15:
```suggestion
PHPDoc comment, which will become the new return type. You can prepare for this
```
```suggestion
Return types will be added in version 2.0. These types are currently documented in a
PHPDoc comment and will eventually become a declared return type. You can prepare for this
```

docs/reference/method/MongoDBClient-dropDatabase.txt
```diff
@@ -19,7 +19,7 @@ Definition
 
    .. code-block:: php
 
-      function dropDatabase($databaseName, array $options []): array|object
```
jmikola, line 22: Good catch!

jmikola (commented):
Just a few suggestions you can apply and merge as you wish.

Take another look at the rendered Markdown file before merging, as that's how I caught some of the formatting mistakes.

jmikola (approved):

PR #986, opened 2022-10-04
PHPLIB-980: Make uri parameter for MongoDB\Client::__construct nullable

alcaeus:
PHPLIB-980

I decided to only make the `$uri` parameter nullable. While we could widen the types of both options parameters (constructors are excluded from contravariance checks), I believe this is contra-productive. The signature of `MongoDB\Manager` historically allowed `null`, but it's preferable to not allow `null` and instead default to an empty array. This is something I'd like to change in version 2.0 of the extension.

This also becomes a non-issue starting with PHP 8, as users can use named parameters if they want to only pass a single option:
```php
$client = new Client(driverOptions: ['foo' => 'bar']);
```

This makes using `null` as an easy default value easier.

For convenience and documentation purposes, I've also extracted the default URI into a public constant.

### DIFF
--- a/docs/reference/method/MongoDBClient__construct.txt
+++ b/docs/reference/method/MongoDBClient__construct.txt
@@ -19,7 +19,7 @@ Definition
 
    .. code-block:: php
 
-      function __construct($uri = 'mongodb://127.0.0.1/', array $uriOptions = [], array $driverOptions = [])
+      function __construct(?string $uri = null, array $uriOptions = [], array $driverOptions = [])
 
    This constructor has the following parameters:
 
--- a/src/Client.php
+++ b/src/Client.php
@@ -44,6 +44,8 @@ use function is_string;
 
 class Client
 {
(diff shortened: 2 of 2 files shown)


docs/reference/method/MongoDBClient__construct.txt
```diff
@@ -19,7 +19,7 @@ Definition
 
    .. code-block:: php
 
-      function __construct($uri = 'mongodb://127.0.0.1/', array $uriOptions = [], array $driverOptions = [])
+      function __construct(?string $uri = null, array $uriOptions = [], array $driverOptions = [])
```
jmikola, line 22: The `$uri` docs on https://www.mongodb.com/docs/php-library/current/reference/method/MongoDBClient__construct/ already state:

> Defaults to "mongodb://127.0.0.1:27017" if unspecified.

So there's no need to change anything else on this page. I also don't think there's any need to document the constant you introduced.

jmikola (approved):

PR #987, opened 2022-10-04
PHPLIB-989: Move persistable class documentation to tutorial

alcaeus:
PHPLIB-989

This moves the documentation on classes implementing `MongoDB\BSON\Persistable` from the BSON reference to a tutorial. It was already written as a tutorial, and linking it as a tutorial on the homepage will make it more visible to users.

I've also considered doing the same with the Type Map information, but considering that persistable classes cover typemap functionality as well, I'm not sure we'll want to. Any opinions @jmikola?


docs/index.txt
```diff
@@ -35,6 +35,8 @@ following pages should help you get started:
 
 - :doc:`/tutorial/gridfs`
 
+- :doc:`tutorial/persistable-classes`
```
jmikola, line 38: Shouldn't this be `/tutorial...` for consistency with the other links?

jmikola (commented):
> I've also considered doing the same with the Type Map information, but considering that persistable classes cover typemap functionality as well, I'm not sure we'll want to.

The [BSON](https://www.mongodb.com/docs/php-library/current/reference/bson/) reference page was always a bit of a dumping ground. I do think it'd make it makes sense to extract its sections out to other pages.

I'd suggest splitting [Type Maps](https://www.mongodb.com/docs/php-library/current/reference/bson/#type-maps) off to its own page or combining with the Persistable Classes tutorial page you're creating here.

[Emulating the Legacy Driver](https://www.mongodb.com/docs/php-library/current/reference/bson/#emulating-the-legacy-driver) can be moved to the [Upgrade Guide](https://www.mongodb.com/docs/php-library/current/upgrade/), which I believe you already plan to rename to "Legacy Driver Upgrade Guide" (or something to that effect). For that, I'd suggest renaming [BSON Type Classes](https://www.mongodb.com/docs/php-library/current/upgrade/#bson-type-classes) to "Type Classes" and placing the "Emulating" section parallel to that under a new, top-level "BSON" heading.

That would just leave the BSON reference page with two sections: Overview and BSON Classes. From there, I'd suggest renaming "BSON Classes" to just "Classes". Those remaining two sections would be appropriate for a reference page, and then you could also add "See Also" links to the newly created tutorial page(s) and relocated "Emulating the Legacy Driver" section.

docs/index.txt
alcaeus: That's correct, fixed.

alcaeus:
@jmikola great suggestions - I applied them as suggested.

docs/tutorial/persistable-classes.txt
```diff
@@ -0,0 +1,154 @@
+===================
+Persistable classes
```
jmikola, line 2: This can be "Type Maps and Persistable Classes" and I'd suggest renaming the file to `type-maps-persistable.txt` for SEO reasons, since that is used to derive the URI.

Alternatively, we can use another phrase like "Modeling BSON Data". That would set us up nicely to add additional sections down the line, such as a demonstration of the BSON interfaces (best use case there may be working with DateTime objects).

docs/tutorial/persistable-classes.txt
```diff
@@ -0,0 +1,154 @@
...
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+
+The driver's :php:`persistence specification <mongodb.persistence>` outlines how
```
jmikola, line 14: Currently, this page has only a single "Type Maps" section. I'd suggest adding a heading here:

```suggestion
Persistable Classes
-------------------

The driver's :php:`persistence specification <mongodb.persistence>` outlines how
```

Additionally, I propose moving "Type Maps" _above_ this section, since this section refers back to the previous (e.g. "without requiring the use of the ``typeMap`` option"). Users should read and understand how type maps work before jumping into Persistable classes.

jmikola (commented):
Let me know when the suggestions are applied and I'll check out the PR for a local docs build to inspect before merging. IIRC, you had some trouble with local docs builds.

docs/tutorial/persistable-classes.txt
alcaeus: Good idea, I renamed the file and changed the title to "Modeling BSON Data".

docs/tutorial/persistable-classes.txt
alcaeus: I did both: added a header for persistable classes and moved the type map documentation up before persistable classes.

jmikola (approved):
Content LGTM. I'll leave you to verify that RST build, since my Giza build environment is broken.

jmikola:
> since my Giza build environment is broken.

Just fixed it and was able to preview this PR. Everything looks good.

PR #988, opened 2022-10-05
PHPLIB-950: Run legacy CSFLE tests on serverless

alcaeus:
PHPLIB-950

Includes fixes required to get CSFLE running on Serverless in general:
* Needs mongocryptd or crypt_shared installed
* Needs secrets exported
* Needs KMS servers running

The download logic was shamelessly copied from the GO driver, as this isn't available through drivers-evergreen-tools. I've created DRIVERS-2459 to track this and will take a look when possible.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -212,6 +212,55 @@ functions:
     - command: expansions.update
       params:
         file: src/serverless-expansion.yml
+    - command: shell.exec
+      params:
+        shell: "bash"
+        script: |
+          ${PREPARE_SHELL}
+          
+          if [ -z "${SERVERLESS_MONGODB_VERSION}" ]; then
+            echo "expected SERVERLESS_MONGODB_VERSION to be set"
+            exit 1
+          fi
+          
+          # Download the enterprise server download for current platform to $MONGODB_BINARIES.
+          # This is required for tests that need mongocryptd.
+          # $MONGODB_BINARIES is added to the $PATH in fetch-source.
+          ${PYTHON3_BINARY} $DRIVERS_TOOLS/.evergreen/mongodl.py \
+              --component archive \
+              --version ${SERVERLESS_MONGODB_VERSION} \
+              --edition enterprise \
+              --out $MONGODB_BINARIES \
+              --strip-path-components 2
+          
+          # Download the crypt_shared dynamic library for the current platform.
+          ${PYTHON3_BINARY} $DRIVERS_TOOLS/.evergreen/mongodl.py \
+            --component crypt_shared \
+            --version ${SERVERLESS_MONGODB_VERSION} \
+            --edition enterprise \
+            --out . \
+            --only "**/mongo_crypt_v1.*" \
+            --strip-path-components 1
+          
+          # Find the crypt_shared library file in the current directory and set the CRYPT_SHARED_LIB_PATH to
+          # the path of that file. Only look for .so, .dll, or .dylib files to prevent matching any other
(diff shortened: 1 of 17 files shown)


.evergreen/config.yml
```diff
@@ -212,6 +212,55 @@ functions:
...
+          # This is required for tests that need mongocryptd.
+          # $MONGODB_BINARIES is added to the $PATH in fetch-source.
+          ${PYTHON3_BINARY} $DRIVERS_TOOLS/.evergreen/mongodl.py \
+              --component archive \
+              --version ${SERVERLESS_MONGODB_VERSION} \
+              --edition enterprise \
+              --out $MONGODB_BINARIES \
+              --strip-path-components 2
+          
+          # Download the crypt_shared dynamic library for the current platform.
```
jmikola, line 236: IIRC, this is handled by [run-orchestration.sh](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/2443440ca69bcdba825eb0b43f874006ae3b5bca/.evergreen/run-orchestration.sh#L105) in drivers-evergreen-tools for non-serverless tests. That appears to be exactly what [DRIVERS-2459](https://jira.mongodb.org/browse/DRIVERS-2459) is meant to address.

IMO, this should have been addressed as a prerequisite for [DRIVERS-2400](https://jira.mongodb.org/browse/DRIVERS-2400) before spawning language tickets and inviting drivers to sort this out on their own.

tests/SpecTests/client-side-encryption/tests/fle2-BypassQueryAnalysis.json
```diff
@@ -4,7 +4,8 @@
       "minServerVersion": "6.0.0",
       "topology": [
         "replicaset",
-        "sharded"
+        "sharded",
+        "load-balanced"
```
jmikola, line 8: None of the commits in this PR indicate the sync target for spec tests. According to [DRIVERS-2400](https://jira.mongodb.org/browse/DRIVERS-2400) this is https://github.com/mongodb/specifications/commit/59a07b718253457d380f1e69cd51ad417f36e575. Please add that and ensure it ends up in the final squash message.

jmikola (commented):
AFAIK, we don't run serverless tests for PR builds. Please schedule an Evergreen patch build that includes the serverless task(s) and attach it to this PR so we can confirm that all works before merging.

alcaeus:
> AFAIK, we don't run serverless tests for PR builds. Please schedule an Evergreen patch build that includes the serverless task(s) and attach it to this PR so we can confirm that all works before merging.

Sorry for not mentioning that above. I scheduled a patch build already: https://evergreen.mongodb.com/build/mongo_php_library_serverless__php_edge_versions~latest_stable_driver_versions~latest_stable_os~debian11_patch_8a327525dbb4a2a9bceac62bfc027e88309db756_633d6772a4cf4706e978aee3_22_10_05_11_16_03?redirect_spruce_users=true

Note that scheduling an additional task in a PR build has the advantage of making the required evergreen check fail if the additional task fails, which in turn prevents a merge of the PR. Only downside is that this has to be done every time after pushing changes, as that will trigger a new evergreen patch build discarding the previously selected tasks.

tests/SpecTests/client-side-encryption/tests/fle2-BypassQueryAnalysis.json
alcaeus: Ah, yes. Added to the last commit and will make sure it ends in the squash message.

PR #989, opened 2022-10-07
PHPLIB-909: Spec tests for expectedError.errorResponse assertions

jmikola:
https://jira.mongodb.org/browse/PHPLIB-909

This is a prototype for https://github.com/mongodb/specifications/pull/1316

### DIFF
--- a/tests/UnifiedSpecTests/ExpectedError.php
+++ b/tests/UnifiedSpecTests/ExpectedError.php
@@ -7,6 +7,7 @@ use MongoDB\Driver\Exception\CommandException;
 use MongoDB\Driver\Exception\ExecutionTimeoutException;
 use MongoDB\Driver\Exception\RuntimeException;
 use MongoDB\Driver\Exception\ServerException;
+use MongoDB\Tests\UnifiedSpecTests\Constraint\Matches;
 use PHPUnit\Framework\Assert;
 use stdClass;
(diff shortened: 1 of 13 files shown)


alcaeus:
Changes for 1.12 LGTM.

Possibly irrelevant since this is a prototype for the corresponding spec change, do we plan to implement the skipped schema versions before implementing 1.12?

jmikola:
> Possibly irrelevant since this is a prototype for the corresponding spec change, do we plan to implement the skipped schema versions before implementing 1.12?

I didn't intend on doing so, as there's quite a bit to cover and we likely can't fully the implementation without also implementing those specs (e.g. CSOT). In an ideal world, there would tests in the unified spec's `valid/` and `valid-failed` directories to provide coverage, but I'm not sure those exist and some syntax could still require a spec implementation to properly exercise.

I'll admit that this is a frustrating side effect of how we handle schema versions, but I don't have a better alternative to suggest at the moment.

alcaeus:
> I'll admit that this is a frustrating side effect of how we handle schema versions, but I don't have a better alternative to suggest at the moment.

No worries, I think that's sensible. FWIW, we can always expand our test runner to mark certain schema version as skipped to ensure we're not marking unsupported schema versions as supported.

alcaeus (approved):

Issue #589, opened 2018-11-01
Howto insert with document validation of type "long".


drupal-daffie:
How do I insert data in a collection when the collection has a validation on it with for a particular column of the type "long". This is used for 64 bit integers. When I try to a large integer the response that I get is that the document validation failed. With other special columns you create first object and insert that object as the column value (Decimal128). Using the Int64 class is not possible, because the construct method is private. Has anybody have an idea/example how to fix my problem. 

I am using: PHP 7.1 and version 1.4.2 of this project and MongoDB 3.6.

derickr:
Can you share:

- the code, as short as possible, and something I could run without modification
- the validation schema, perhaps also explaining how you've configured this
- the exact error message

Decimal128 should generally only be used for very specific use cases. 

Int64 is a class that is only used when reading data that is stored in MongoDB as a IntLong/64-bit integer on a 32-bit platform. Without an Int64 class wrapping the value, the data would end up being corrupt in that case. On 64-bit platforms, Int64 should never be generated for you when reading data.

drupal-daffie:
$collection = (new MongoDB\Client)->test->test;

$insertOneResult = $collection->insertOne([
    'big_integer' => 223372036854775807,
]);

The validation level is strict.
The collection test has the following validation:
{ 
    "$and" : [
        {
            "big_integer" : {
                "$type" : "long"
            }
        },
        {
            "big_integer" : {
                "$exists" : true
            }
        }
    ]
}

The error message that I get is: "MongoDB\Driver\Exception\BulkWriteException: Document failed validation"

Hopefully this is enough. And thank you in advance for your help.

drupal-daffie:
Yes, I am using query expression validation. I should have mentioned that.

Changing the validation from `$type => "long"` to `$type => ["int", "long"]` fixed my problem.
Thank you jmikola for your suggestion.

Integer values in the column are now stored as "Int32" for small integer values and as "Int64" for large integer values.
It was probably my [OCD](https://en.wikipedia.org/wiki/Obsessive%E2%80%93compulsive_disorder) side that wanted to store all integer values in that column as "Int64".

I would like to thank both derickr and jmikola for their fast responce to my problem and their very expert help/suggestions.
I will now close this issue. If someone still would like to add something, please do.

PR #991, opened 2022-10-10
PHPLIB-1001, PHPLIB-1010, PHPLIB-1011: Improvements to $$type operator

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1001
https://jira.mongodb.org/browse/PHPLIB-1010
https://jira.mongodb.org/browse/PHPLIB-1011

Also revises instantiation of deprecated BSON types and fixes an assertion message.

### DIFF
--- a/tests/UnifiedSpecTests/Constraint/IsBsonType.php
+++ b/tests/UnifiedSpecTests/Constraint/IsBsonType.php
@@ -38,8 +38,6 @@ use function is_string;
 use function range;
 use function sprintf;
 
-use const PHP_INT_SIZE;
-
 final class IsBsonType extends Constraint
 {
     use ConstraintTrait;
@@ -153,11 +151,7 @@ final class IsBsonType extends Constraint
                 return $other instanceof TimestampInterface;
 
             case 'long':
-                if (PHP_INT_SIZE == 4) {
-                    return $other instanceof Int64;
-                }
-
-                return is_int($other);
+                return is_int($other) || $other instanceof Int64;
 
             case 'decimal':
                 return $other instanceof Decimal128Interface;
@@ -169,7 +163,7 @@ final class IsBsonType extends Constraint
                 return $other instanceof MaxKeyInterface;
 
             case 'number':
-                return is_int($other) || (PHP_INT_SIZE == 4 && $other instanceof Int64) || is_float($other) || $other instanceof Decimal128Interface;
+                return is_int($other) || $other instanceof Int64 || is_float($other) || $other instanceof Decimal128Interface;
 
             default:
                 // This should already have been caught in the constructor
--- a/tests/UnifiedSpecTests/Constraint/IsBsonType.php
+++ b/tests/UnifiedSpecTests/Constraint/IsBsonType.php
@@ -67,6 +67,7 @@ final class IsBsonType extends Constraint
         'decimal',
         'minKey',
         'maxKey',
+        'number',
     ];
 
     /** @var string */
@@ -166,6 +167,9 @@ final class IsBsonType extends Constraint
 
             case 'maxKey':
                 return $other instanceof MaxKeyInterface;
+
+            case 'number':
+                return is_int($other) || (PHP_INT_SIZE == 4 && $other instanceof Int64) || is_float($other) || $other instanceof Decimal128Interface;
 
             default:
                 // This should already have been caught in the constructor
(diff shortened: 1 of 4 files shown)


tests/UnifiedSpecTests/Constraint/IsBsonType.php
```diff
@@ -167,6 +168,9 @@ private function doMatches($other): bool
             case 'maxKey':
                 return $other instanceof MaxKeyInterface;
 
+            case 'number':
+                return is_int($other) || (PHP_INT_SIZE == 4 && $other instanceof Int64) || is_float($other) || $other instanceof Decimal128Interface;
```
alcaeus, line 172: Out of curiosity, is there a reason for limiting the `Int64` check to 32-bit platforms? Technically we won't have an `Int64` instance on 64-bit platforms, but is there a practical reason for the check?

tests/UnifiedSpecTests/Constraint/IsBsonTypeTest.php
```diff
@@ -38,9 +38,10 @@ public function testConstraint($type, $value): void
 
     public function provideTypes()
     {
-        $undefined = toPHP(fromJSON('{ "undefined": {"$undefined": true} }'));
-        $symbol = toPHP(fromJSON('{ "symbol": {"$symbol": "test"} }'));
-        $dbPointer = toPHP(fromJSON('{ "dbPointer": {"$dbPointer": {"$ref": "phongo.test", "$id" : { "$oid" : "5a2e78accd485d55b405ac12" }  }} }'));
+        $undefined = toPHP(fromJSON('{ "x": {"$undefined": true} }'))->x;
+        $symbol = toPHP(fromJSON('{ "x": {"$symbol": "test"} }'))->x;
+        $dbPointer = toPHP(fromJSON('{ "x": {"$dbPointer": {"$ref": "db.coll", "$id" : { "$oid" : "5a2e78accd485d55b405ac12" }  }} }'))->x;
+        $long = PHP_INT_SIZE == 4 ? unserialize('C:18:"MongoDB\BSON\Int64":38:{a:1:{s:7:"integer";s:10:"4294967296";}}') : 4294967296;
```
alcaeus, line 44: Our tests make the best case for providing a constructor for `Int64`. Can't wait to eventually get rid of it.

tests/UnifiedSpecTests/Constraint/IsBsonTypeTest.php
```diff
@@ -89,10 +94,20 @@ public function testAnyOf(): void
         $c = IsBsonType::anyOf('double', 'int');
 
         $this->assertResult(true, $c, 1, 'int is double or int');
-        $this->assertResult(true, $c, 1.4, 'int is double or int');
+        $this->assertResult(true, $c, 1.4, 'double is double or int');
```
alcaeus, line 97: Good catch!

alcaeus (approved):

tests/UnifiedSpecTests/Constraint/IsBsonType.php
jmikola: This likely goes back to the legacy test format's [DocumentMatchConstraint](https://github.com/mongodb/mongo-php-library/blob/master/tests/SpecTests/DocumentsMatchConstraint.php), from which we derived IsBsonType.

I'll note that the corresponding [Matches constraint prepares 64-bit integers](https://github.com/mongodb/mongo-php-library/blob/ecdacff94528eb4bd27c5a053b34bbb2d457eb40/tests/UnifiedSpecTests/Constraint/Matches.php#L427) such that we should never have an Int64 type on a 64-bit platform; however, that doesn't mean IsBsonType needs to be this strict. This PR seems like a good opportunity to relax that logic, so I'll make the change.

Side note: PHPLIB-1001 only applies the the unified test runner, so DocumentMatchConstraint will continue to not support "number" in its `$$type` operator implementation.

tests/UnifiedSpecTests/Constraint/IsBsonTypeTest.php
jmikola: Remember that a constructor would provide users a foot-gun allowing them to force 32-bit values to serialize as 64-bit, only to have BSON decoding convert them back to basic integers. Previous discussions on this:

 * https://github.com/mongodb/mongo-php-library/issues/589
 * https://github.com/mongodb/mongo-php-driver/issues/525

Unless we implement some BSON codec API that allows users to coerce integer serialization (e.g. always use 64-bit BSON types), I think the current behavior is less likely to confuse folks while noting we can't please everyone.

jmikola (commented):
After I implemented [PHPLIB-1010](https://jira.mongodb.org/browse/PHPLIB-1010) to relax the Int64 logic I realized that the legacy `$$type` operator (a) did not have a default case for unsupported values and (b) never added support for an array of multiple types. There was also a separate bug where we used "boolean" instead of "bool" (actual type from the server operator).

The assertion method also had a `string` type hint (possibly added recently), so I imagine we'd have encountered an error attempting to cast an array to a string if we ever encountered a test using multiple types.

Anyway, this seemed like a good time to just refactor the entire legacy constraint to use IsBsonType, which added the necessary functionality and deleted a fair bit of code in the process (including some legacy code paths for PHPUnit < 6.5).

tests/UnifiedSpecTests/Constraint/IsBsonTypeTest.php
alcaeus: I wasn't suggesting we add a constructor - apologies if this was misunderstood. I was alluding to getting rid of Int64 if and when we drop support for 32-bit platforms. It's wishful thinking at this point, but I hope that x86 goes out of style sometime this century ^^

alcaeus (approved):
LGTM

PR #993, opened 2022-10-11
Add Github Actions workflow to build documentation

alcaeus:
This PR adds a build step to build documentation. This can be used in lieu of building documentation locally as python 2 is no longer available on many platforms. I was not able to make `giza make publish` exit with a non-zero exit code, even when using a wrong filename to link to a file. For now this means we'll have to verify the docs build manually, but it's better than not being able to run docs at all. The second commit fixes all current warnings.

The built documentation is uploaded as a build artefact to the Docs workflow, accessible through the actions tab.


docs/reference/bson.txt
```diff
@@ -14,8 +14,8 @@ Overview
 --------
 
 MongoDB stores data records as BSON documents. BSON is a binary representation
-of :term:`JSON` documents, though it contains more data types than JSON. For the
```
alcaeus, line 17: Apparently, this term isn't defined and the current docs don't render this in a special way, so I removed it.

docs/reference/method/MongoDBChangeStream-current.txt
```diff
@@ -103,4 +103,4 @@ See Also
 - :ref:`Tailable Cursor Iteration <php-tailable-cursor>`
 - :manual:`Change Streams </changeStreams>` documentation in the MongoDB manual
 - :manual:`Change Events </reference/change-events/>` documentation in the
-  MongoDB manual
\ No newline at end of file
+  MongoDB manual
```
alcaeus, line 106: This file (and the next) were lacking a trailing newline.

docs/reference/bson.txt
jmikola: Thanks. I've seen this error before but never traced it back to this line. I agree this serves no purpose.

It could be something that only builds correctly with the docs team's build tools, since those pull in additional resources used by the server manual (or _did_ if they no longer use Giza).

.github/workflows/docs.yml
```diff
@@ -0,0 +1,56 @@
...
+          repository: mongodb/docs-php-library
+          path: docs
+          fetch-depth: 2
+
+      - name: "Setup python"
+        uses: actions/setup-python@v4
+        with:
+          python-version: '2.7'
+
+      # The requirements file installs urllib3 with an incompatible version; we replace it with one that works
```
jmikola, line 39: Is this worth a PR to `mongodb/docs-tools`? Worst case it doesn't get merged.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: "Install giza"
+        run: |
+          pip install -r https://raw.githubusercontent.com/mongodb/docs-tools/master/giza/requirements.txt
+          pip install urllib3==1.25.2 
+
+      - name: "Sync documentation"
+        run: "rsync -a library/docs/ docs/source/"
```
jmikola, line 46: I suppose this may not be necessary since the CI environment is fresh, but my build script also did the following:

```
rm -r build/master
rsync -a --delete /path/to/mongo-php-library/docs/ /path/to/docs-php-library/source/
```

.github/workflows/docs.yml
```diff
@@ -0,0 +1,56 @@
...
+
+      - name: "Run Giza"
+        run: "giza make publish"
```
jmikola, line 49: @i80and: Any chance you know of a way to get `giza` to return a non-zero exit code on an erroneous docs build?

jmikola (approved):

.github/workflows/docs.yml
alcaeus: That is correct. Since we don't init the submodule, there are no outdated docs for us to worry about.

PR #994, opened 2022-10-14
PHPLIB-899: No longer use buildinfo command to detect Atlas Data Lake instances

alcaeus:
PHPLIB-899

This PR changes the detection logic for Atlas Data Lake to no longer use `buildInfo`, but rather rely on an externally set environment variable, as ADL no longer supports `buildInfo`. In addition, it changes the Evergreen build variant to use Ubuntu 18.04 as the built-in mongohoused didn't run on debian11.

Note that ADL tests will still fail due to CDRIVER-4502; this will be fixed in a separate pull request.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -192,7 +192,7 @@ functions:
     - command: shell.exec
       params:
         script: |
-          DRIVERS_TOOLS="${DRIVERS_TOOLS}" sh ${DRIVERS_TOOLS}/.evergreen/atlas_data_lake/build-mongohouse-local.sh
+          VARIANT=debian11 DRIVERS_TOOLS="${DRIVERS_TOOLS}" sh ${DRIVERS_TOOLS}/.evergreen/atlas_data_lake/build-mongohouse-local.sh
     - command: shell.exec
       params:
         background: true
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -192,7 +192,7 @@ functions:
     - command: shell.exec
       params:
         script: |
-          VARIANT=debian11 DRIVERS_TOOLS="${DRIVERS_TOOLS}" sh ${DRIVERS_TOOLS}/.evergreen/atlas_data_lake/build-mongohouse-local.sh
+          VARIANT=${VARIANT} DRIVERS_TOOLS="${DRIVERS_TOOLS}" sh ${DRIVERS_TOOLS}/.evergreen/atlas_data_lake/build-mongohouse-local.sh
     - command: shell.exec
       params:
         background: true
@@ -834,6 +834,8 @@ buildvariants:
   matrix_spec: { "php-edge-versions": "latest-stable", "driver-versions": "latest-stable" }
   display_name: "Atlas Data Lake"
   run_on: debian11
+  expansions:
+    VARIANT: debian11
   tasks:
     - name: "test-atlas-data-lake"
 


jmikola:
> the built-in mongohoused didn't run on debian11

Thanks for catching that. Are there any more details about the incompatibility that we can relay this to the ADL team?

/cc @craiggwilson in case he's still involved with that project

alcaeus:
> Thanks for catching that. Are there any more details about the incompatibility that we can relay this to the ADL team?

These are the relevant errors:
```
go: golang.org/x/text@v0.3.3: read tcp [2600:1f18:ae9:dd00:13c7:e76a:9dd9:b049]:56172->[2607:f8b0:4004:c09::8d]:443: read: connection reset by peer
go: golang.org/x/time@v0.0.0-20220210224613-90d013bbcef8: Get "https://proxy.golang.org/golang.org/x/time/@v/v0.0.0-20220210224613-90d013bbcef8.zip": read tcp [2600:1f18:ae9:dd00:13c7:e76a:9dd9:b049]:56172->[2607:f8b0:4004:c09::8d]:443: read: connection reset by peer
go: golang.org/x/tools@v0.1.12: Get "https://proxy.golang.org/golang.org/x/tools/@v/v0.1.12.zip": read tcp [2600:1f18:ae9:dd00:13c7:e76a:9dd9:b049]:56172->[2607:f8b0:4004:c09::8d]:443: read: connection reset by peer
go: golang.org/x/xerrors@v0.0.0-20220609144429-65e65417b02f: Get "https://proxy.golang.org/golang.org/x/xerrors/@v/v0.0.0-20220609144429-65e65417b02f.zip": read tcp [2600:1f18:ae9:dd00:13c7:e76a:9dd9:b049]:56172->[2607:f8b0:4004:c09::8d]:443: read: connection reset by peer
go version go1.18.5 linux/amd64
```

Trying to resolve this I came across this line in the Python driver's evergreen config: https://github.com/mongodb/mongo-python-driver/blob/3eb316ed3042d7b9690afbf04acb86b318b82658/.evergreen/config.yml#L314. Checking that build config, I didn't see any mention of debian and decided to take the easy way of testing on Ubuntu.

After re-evaluating this with a clear head, I realised that we were missing passing the `VARIANT` environment variable, which ensures that the golang config is correct. I've replaced the second commit in this PR with a simpler one that provides a variant and works just as well.

.evergreen/config.yml
```diff
@@ -192,7 +192,7 @@ functions:
     - command: shell.exec
       params:
         script: |
-          DRIVERS_TOOLS="${DRIVERS_TOOLS}" sh ${DRIVERS_TOOLS}/.evergreen/atlas_data_lake/build-mongohouse-local.sh
+          VARIANT=debian11 DRIVERS_TOOLS="${DRIVERS_TOOLS}" sh ${DRIVERS_TOOLS}/.evergreen/atlas_data_lake/build-mongohouse-local.sh
```
jmikola, line 195: I don't see `VARIANT` referenced in [build-mongohouse-local.sh](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/atlas_data_lake/build-mongohouse-local.sh). What is this used for?

If this is necessary, this approach seems fragile because we're two levels removed from the `run_on: debian11` in the `atlas-data-lake-test` matrix definition, which invokes the `test-atlas-data-lake` task that in turn calls this function.

tests/SpecTests/AtlasDataLakeSpecTest.php
```diff
@@ -237,13 +240,8 @@ public function testConnectwithSCRAMSHA256(): void
...
-        $cursor = $this->manager->executeCommand(
-            $this->getDatabaseName(),
-            new Command(['buildInfo' => 1])
-        );
-
-        $document = current($cursor->toArray());
+        $isDataLake = getenv('MONGODB_IS_DATA_LAKE');
 
-        return ! empty($document->dataLake);
+        return $isDataLake !== false ? filter_var($isDataLake, FILTER_VALIDATE_BOOLEAN) : false;
```
jmikola, line 245: Noted that this code was adapted from `MongoDB\Tests\FunctionalTestCase::isServerless()`.

tests/SpecTests/AtlasDataLakeSpecTest.php
```diff
@@ -237,13 +240,8 @@ public function testConnectwithSCRAMSHA256(): void
 
     private function isAtlasDataLake(): bool
     {
-        $cursor = $this->manager->executeCommand(
-            $this->getDatabaseName(),
-            new Command(['buildInfo' => 1])
-        );
-
-        $document = current($cursor->toArray());
+        $isDataLake = getenv('MONGODB_IS_DATA_LAKE');
```
jmikola, line 243: Please add a new section to the "Environment Variables" section in CONTRIBUTING.md:

```
The following environment variable is used for [Atlas Data Lake testing](https://github.com/mongodb/specifications/blob/master/source/atlas-data-lake-testing/tests/README.rst):

 * `MONGODB_IS_DATA_LAKE`: Specify a true boolean string
   (see: [`FILTER_VALIDATE_BOOLEAN`](https://www.php.net/manual/en/filter.filters.validate.php))
   if `MONGODB_URI` points to an Atlas Data Lake instance. Defaults to false.
```

I don't think we need additional prose (like we have for serverless) in the paragraphs above directly under "Testing".

alcaeus: Added, thank you for providing the wording!

.evergreen/config.yml
alcaeus: > I don't see `VARIANT` referenced in [build-mongohouse-local.sh](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/atlas_data_lake/build-mongohouse-local.sh). What is this used for?

Took me a while. The `VARIANT` environment variable is read in the [mongohouse build script](https://github.com/10gen/mongohouse/blob/9487b1aea0534c5d6f8923366b0f3dd9b269c5e3/cmd/buildscript/internal/config.go#L157). It is then used in the [`getDesiredTarget` function](https://github.com/10gen/mongohouse/blob/9487b1aea0534c5d6f8923366b0f3dd9b269c5e3/cmd/buildscript/internal/config.go#L1422). The value returned from there is later used to [download the correct MongoDB package](https://github.com/10gen/mongohouse/blob/9487b1aea0534c5d6f8923366b0f3dd9b269c5e3/cmd/buildscript/internal/config.go#L1481) and return the correct path to the binaries.

> If this is necessary, this approach seems fragile because we're two levels removed from the `run_on: debian11` in the `atlas-data-lake-test` matrix definition, which invokes the `test-atlas-data-lake` task that in turn calls this function.

That's a good point. I've changed this to add a `VARIANT` extension where we define the build variant, setting the environment variable from that expansion. That puts the definition of the build variant to where we also specify the OS the tests run on.

jmikola (approved):
Suggested comment on the `VARIANT` declaration but LGTM otherwise.

alcaeus:
When we originally created the build failure ticket, all we could see that the error appears when calling `buildInfo`, hence my original work to remove calling `buildInfo`. Only later did I realise that there is a secondary issue caused by a libmongoc bug that made all other commands fail as well. When I fixed that I didn't realise that this most likely also caused what we thought was an issue with the `buildInfo` command.

Now that I was able to test again, I was able to confirm that the `buildInfo` command works as expected and tests pass again with the libmongoc patch applied. I'll create a separate pull request that adds the variant.

PR #995, opened 2022-10-18
PHPLIB-986: Split keyAltName example and link key management docs

alcaeus:
PHPLIB-986

This PR splits the tutorial section on key alternate names in two parts: creating the key and using it. This is done to better explain the disconnect of creating an encryption key at one time (e.g. on deployment) and using it at a different time (i.e. when inserting data). While we could show ways how to serialise a `MongoDB\BSON\Binary` object and include that via the configuration, I believe the easier way is using an alternate name for this, as it's just a string as opposed to binary data.

In addition I've linked the server's [Encryption Key Management Documentation](https://www.mongodb.com/docs/manual/core/csfle/fundamentals/manage-keys/) as it contains more information about how to manage encryption keys, including examples how to dynamically choose an encryption key when inserting data.

### DIFF
--- a/docs/tutorial/client-side-encryption.txt
+++ b/docs/tutorial/client-side-encryption.txt
@@ -92,7 +92,7 @@ side.
    $database->dropCollection('coll'); // remove old data
 
    // This uses the key ID from the first example. The key ID could be read from
-   // configuration.
+   // a configuration file.
    $keyId = readDataKey();
 
    $database->createCollection('coll', [
@@ -147,7 +147,7 @@ encrypted fields.
    $client = new Client();
 
    // This uses the key ID from the first example. The key ID could be read from
-   // configuration.
+   // a configuration file.
    $keyId = readDataKey();
 
    $autoEncryptionOpts = [
@@ -210,7 +210,7 @@ encrypts and decrypts values in the document.
    $clientEncryption = $client->createClientEncryption($clientEncryptionOpts);
 
    // This uses the key ID from the first example. The key ID could be read from
-   // configuration.
+   // a configuration file.
    $keyId = readDataKey();
 
    $collection = $client->selectCollection('test', 'coll');
--- a/docs/tutorial/client-side-encryption.txt
+++ b/docs/tutorial/client-side-encryption.txt
@@ -202,6 +202,15 @@ specified when creating the key. The following example creates an encryption key
 with an alternative name, which could be done when deploying the application.
 The software then encrypts data by referencing the key by its alternative name.
 
+Creating the encryption key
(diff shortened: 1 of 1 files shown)


alcaeus:
After discussing this with @comandeo, we decided that it's better to restructure the tutorial to first show key creation and talk about key management. The current tutorials would then follow this first section, always referencing the key created in the first section. While this makes the examples less copy-pasteable, it does represent real-world usage better than the current examples.

alcaeus:
@jmikola thank you for your review, I applied your suggestions. I've extracted the key creation part to the beginning of the tutorial and updated the examples to account for this, so I'd appreciate another look.

docs/tutorial/client-side-encryption.txt
```diff
@@ -43,13 +87,13 @@ encrypted on insertion and decrypted when querying on the client side.
...
    $client = new Client();
-   $clientEncryption = $client->createClientEncryption($encryptionOpts);
 
    $database = $client->selectDatabase('test');
    $database->dropCollection('coll'); // remove old data
 
-   // Create new key in the key vault and store its ID for later use
-   $keyId = $clientEncryption->createDataKey('local');
+   // This uses the key ID from the first example. The key ID could be read from
+   // configuration.
```
jmikola, line 95:
```suggestion
   // a configuration file.
```

docs/tutorial/client-side-encryption.txt
```diff
@@ -101,18 +144,11 @@ encryption using the newly created key.
...
-       ],
-   ];
-
    $client = new Client();
-   $clientEncryption = $client->createClientEncryption($clientEncryptionOpts);
 
-   // Create new key in the key vault and store its ID for later use
-   $keyId = $clientEncryption->createDataKey('local');
+   // This uses the key ID from the first example. The key ID could be read from
+   // configuration.
```
jmikola, line 150:
```suggestion
   // a configuration file.
```

docs/tutorial/client-side-encryption.txt
```diff
@@ -173,8 +209,9 @@ explicitly encrypts and decrypts values in the document.
    $client = new Client();
    $clientEncryption = $client->createClientEncryption($clientEncryptionOpts);
 
-   // Create new key in the key vault and store its ID for later use
-   $keyId = $clientEncryption->createDataKey('local');
+   // This uses the key ID from the first example. The key ID could be read from
+   // configuration.
```
jmikola, line 213:
```suggestion
   // a configuration file.
```

jmikola (approved):
Comment revisions but LGTM otherwise.

PR #1002, opened 2022-11-14
PHPLIB-981: Add tests for examples

alcaeus:
PHPLIB-981

This PR adds tests for the example scripts. To ensure good coverage, the tests attempt to match against the output of the example scripts. This requires a change in the examples to no longer output to `STDERR`, as PHPUnit considers any output to `STDERR` as condition to fail a test and doesn't allow asserting against it.

When running the tests, I noticed that the tests fail on sharded clusters, because the command replies are different. This would complicate the output matchers by quite a lot, so I'm wondering if it's actually worth matching against the actual command output, or if we're happy asserting that the rough information is the same.

I will likely also change the examples for type maps and persistable classes to no longer use `var_dump`. Having xdebug installed (which is likely on dev machines) produces different output for `var_dump`, making the tests fail on developer machines.

### DIFF
--- a/examples/changestream.php
+++ b/examples/changestream.php
@@ -6,15 +6,12 @@ namespace MongoDB\Examples;
 use MongoDB\Client;
 
 use function assert;
-use function fprintf;
 use function getenv;
 use function is_object;
 use function MongoDB\BSON\fromPHP;
 use function MongoDB\BSON\toRelaxedExtendedJSON;
 use function printf;
 use function time;
-
-use const STDERR;
 
 require __DIR__ . '/../vendor/autoload.php';
 
@@ -53,7 +50,7 @@ while (true) {
     $changeStream->next();
 
     if (time() - $startTime > 3) {
-        fprintf(STDERR, "Aborting after 3 seconds...\n");
+        printf("Aborting after 3 seconds...\n");
         break;
     }
 }
--- a/examples/changestream.php
+++ b/examples/changestream.php
@@ -23,11 +23,11 @@ function toJSON(object $document): string
 // Change streams require a replica set or sharded cluster
 $client = new Client(getenv('MONGODB_URI') ?: 'mongodb://127.0.0.1/');
 
-$collection = $client->test->coll;
+$collection = $client->test->changestream;
 $collection->drop();
 
 // Create collection before starting change stream; this is required on MongoDB 3.6
-$client->test->createCollection('coll');
+$client->test->createCollection('changestream');
 
 $changeStream = $collection->watch();
 
--- a/examples/changestream.php
+++ b/examples/changestream.php
@@ -26,6 +26,9 @@ $client = new Client(getenv('MONGODB_URI') ?: 'mongodb://127.0.0.1/');
 $collection = $client->test->coll;
 $collection->drop();
 
+// Create collection before starting change stream; this is required on MongoDB 3.6
+$client->test->createCollection('coll');
+
 $changeStream = $collection->watch();
 
 $documents = [];
--- a/examples/command_logger.php
+++ b/examples/command_logger.php
@@ -10,15 +10,12 @@ use MongoDB\Driver\Monitoring\CommandSubscriber;
 use MongoDB\Driver\Monitoring\CommandSucceededEvent;
 
 use function assert;
-use function fprintf;
 use function get_class;
 use function getenv;
 use function is_object;
 use function MongoDB\BSON\fromPHP;
 use function MongoDB\BSON\toRelaxedExtendedJSON;
 use function printf;
-
-use const STDERR;
 
 require __DIR__ . '/../vendor/autoload.php';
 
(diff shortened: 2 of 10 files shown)


tests/ExamplesTest.php
```diff
@@ -0,0 +1,198 @@
...
+OUTPUT;
+
+        yield 'bulk' => [
+            'file' => __DIR__ . '/../examples/bulk.php',
+            'expectedOutput' => $expectedOutput,
+        ];
+
+        $expectedOutput = <<<'OUTPUT'
+drop command started
+command: { "drop" : "coll", "$db" : "test", "lsid" : { %s }%S }
```
alcaeus, line 46: This command output here and all subsequent command documents could be replaced with `{ %s }` if we're not interested in asserting the actual command results.

tests/ExamplesTest.php
```diff
@@ -0,0 +1,198 @@
...
+}
+OUTPUT;
+
+        yield 'typemap' => [
+            'file' => __DIR__ . '/../examples/typemap.php',
+            'expectedOutput' => $expectedOutput,
+        ];
+    }
+
+    public function testChangeStream(): void
```
alcaeus, line 157: This test (and `testWithTransaction`) below was extracted as it needs to check whether change streams are supported. I initially wanted to provide a closure in the data provider, but that causes errors due to closures not being serialisable (which is required to run tests in isolation). Instead of passing a string callable I opted for extracting the tests from the data provider and calling the original test method manually instead.

examples/changestream.php
```diff
@@ -53,7 +50,7 @@ function toJSON(object $document): string
     $changeStream->next();
 
     if (time() - $startTime > 3) {
-        fprintf(STDERR, "Aborting after 3 seconds...\n");
+        printf("Aborting after 3 seconds...\n");
```
alcaeus, line 53: Necessary to avoid PHPUnit considering the tests failed due to output, and to allow asserting against the full test output.

tests/ExamplesTest.php
```diff
@@ -0,0 +1,198 @@
...
+{ "_id" : { "_data" : "%s" }, "operationType" : "insert", "clusterTime" : { "$timestamp" : { "t" : %d, "i" : %d } }%S, "fullDocument" : { "_id" : { "$oid" : "%s" }, "x" : 6 }, "ns" : { "db" : "test", "coll" : "coll" }, "documentKey" : { "_id" : { "$oid" : "%s" } } }
+{ "_id" : { "_data" : "%s" }, "operationType" : "insert", "clusterTime" : { "$timestamp" : { "t" : %d, "i" : %d } }%S, "fullDocument" : { "_id" : { "$oid" : "%s" }, "x" : 7 }, "ns" : { "db" : "test", "coll" : "coll" }, "documentKey" : { "_id" : { "$oid" : "%s" } } }
+{ "_id" : { "_data" : "%s" }, "operationType" : "insert", "clusterTime" : { "$timestamp" : { "t" : %d, "i" : %d } }%S, "fullDocument" : { "_id" : { "$oid" : "%s" }, "x" : 8 }, "ns" : { "db" : "test", "coll" : "coll" }, "documentKey" : { "_id" : { "$oid" : "%s" } } }
+{ "_id" : { "_data" : "%s" }, "operationType" : "insert", "clusterTime" : { "$timestamp" : { "t" : %d, "i" : %d } }%S, "fullDocument" : { "_id" : { "$oid" : "%s" }, "x" : 9 }, "ns" : { "db" : "test", "coll" : "coll" }, "documentKey" : { "_id" : { "$oid" : "%s" } } }
```
jmikola, line 171: This seems fragile should the server introduce new response fields to change events. I'd suggest changing this to `{ %s }` assertions for each line. You can also lower the number of inserts in the example (maybe 5 instead of 10) if you want to make this more concise. There's really no reason it needs to emit this much output.

tests/ExamplesTest.php
```diff
@@ -11,6 +11,10 @@ public function setUp(): void
     {
         parent::setUp();
 
+        if ($this->isShardedCluster()) {
+            $this->markTestSkipped('Examples are not tested on sharded clusters.');
+        }
```
jmikola, line 16: Obligatory @amschwerin:

20160923_andy_sharding

I am a creature of habit: https://github.com/mongodb/specifications/pull/565#issuecomment-501395705

alcaeus: I tried hard to keep those tests in, but unfortunately some tests failed on sharing due to a server error ("no progress was made executing batch op after 5 rounds"). Since we're only dealing with examples here I opted against spending more time on testing against sharded clusters and gave up trying.

alcaeus:
Note: test failures against MongoDB latest tracked in PHPLIB-1044.

psalm-baseline.xml
```diff
@@ -1,5 +1,37 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="4.x-dev@5108834088c1d8cae3698df6ef6bf15fe6e76c53">
+<files psalm-version="4.30.0@d0bc6e25d89f649e4f36a534f330f8bb4643dd69">
+  <file src="examples/aggregate.php">
```
jmikola, line 3: I assume this files was auto-generated and there's no need to review.

src/Operation/BulkWrite.php
```diff
@@ -325,6 +326,8 @@ public function execute(Server $server)
             $type = key($operation);
             $args = current($operation);
 
+            assert(is_array($args));
```
jmikola, line 329: Is this necessary? Since `$operations` is private couldn't you also just add an `array` property type hint to it?

examples/command_logger.php
```diff
@@ -31,37 +28,37 @@ class CommandLogger implements CommandSubscriber
...
+        printf("\n");
     }
 }
 
 $client = new Client(getenv('MONGODB_URI') ?: 'mongodb://127.0.0.1/');
 
 $client->getManager()->addSubscriber(new CommandLogger());
 
-$collection = $client->test->coll;
+$collection = $client->test->commandlogger;
```
jmikola, line 61:
```suggestion
$collection = $client->test->command_logger;
```

examples/with_transaction.php
```diff
@@ -24,9 +24,12 @@ function toJSON(object $document): string
 // Transactions require a replica set (MongoDB >= 4.0) or sharded cluster (MongoDB >= 4.2)
 $client = new Client(getenv('MONGODB_URI') ?: 'mongodb://127.0.0.1/');
 
-$collection = $client->test->coll;
+$collection = $client->test->transaction;
```
jmikola, line 27:
```suggestion
$collection = $client->test->with_transaction;
```

examples/with_transaction.php
```diff
@@ -24,9 +24,12 @@ function toJSON(object $document): string
...
 $collection->drop();
 
+// Create collection outside of transaction; this is required when using MongoDB < 4.4
+$client->test->createCollection('transaction');
```
jmikola, line 31:
```suggestion
$client->test->createCollection('with_transaction');
```

jmikola (approved):
Some optional suggestions to make the test collections consistent with their filename. Everything else looks fine as-is.

psalm-baseline.xml
alcaeus: That's correct.

src/Operation/BulkWrite.php
alcaeus: I believe the error crept up due to a change in the baseline, so I've removed this `assert` call and updated the baseline instead. Fixing this would've required updating types in more than just this file, which is why it was initially deferred. We have follow-up tickets to fix issues like this one.

PR #1010, opened 2022-11-29
PHPLIB-1046: Fix snooty errors in documentation

alcaeus:
PHPLIB-1046

This PR fixes most of the errors reported when building docs with the Snooty tooling. A few errors are currently unfixable:
* `Unknown interpreted text role "php"`: This needs fixing on the tooling side
* `Error opening includes/....rst: No such file or directory`: This may need to be fixed in tooling or require local changes; waiting on feedback
* `Target not found: "phpmethod:phpmethod...."`: Waiting on feedback on how to fix

### DIFF
--- a/docs/reference/method/MongoDBCollection-explain.txt
+++ b/docs/reference/method/MongoDBCollection-explain.txt
@@ -50,18 +50,18 @@ Explainable Commands
 
 Explainable commands include, but are not limited to:
 
-- :phpclass:`MongoDB\\Operation\\Aggregate`
-- :phpclass:`MongoDB\\Operation\\Count`
-- :phpclass:`MongoDB\\Operation\\DeleteMany`
-- :phpclass:`MongoDB\\Operation\\DeleteOne`
-- :phpclass:`MongoDB\\Operation\\Distinct`
-- :phpclass:`MongoDB\\Operation\\Find`
-- :phpclass:`MongoDB\\Operation\\FindOne`
-- :phpclass:`MongoDB\\Operation\\FindOneAndDelete`
-- :phpclass:`MongoDB\\Operation\\FindOneAndReplace`
-- :phpclass:`MongoDB\\Operation\\FindOneAndUpdate`
-- :phpclass:`MongoDB\\Operation\\UpdateMany`
-- :phpclass:`MongoDB\\Operation\\UpdateOne`
+- `MongoDB\\Operation\\Aggregate`
+- `MongoDB\\Operation\\Count`
+- `MongoDB\\Operation\\DeleteMany`
+- `MongoDB\\Operation\\DeleteOne`
+- `MongoDB\\Operation\\Distinct`
+- `MongoDB\\Operation\\Find`
+- `MongoDB\\Operation\\FindOne`
+- `MongoDB\\Operation\\FindOneAndDelete`
+- `MongoDB\\Operation\\FindOneAndReplace`
+- `MongoDB\\Operation\\FindOneAndUpdate`
+- `MongoDB\\Operation\\UpdateMany`
+- `MongoDB\\Operation\\UpdateOne`
 
 Examples
 --------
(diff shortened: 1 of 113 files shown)


docs/reference/method/MongoDBModelCollectionInfo-getIdIndex.txt
```diff
@@ -49,7 +49,7 @@ Examples
 
    var_dump($info->getIdIndex());
 
-The output would then resemble::
+The output would then resemble:
```
jmikola, line 52: Is there a reason this doesn't have `.. code-block:: none` like most of the other occurrences?

docs/reference/method/MongoDBModelCollectionInfo-getInfo.txt
```diff
@@ -44,7 +44,7 @@ Examples
 
    var_dump($info->getInfo());
 
-The output would then resemble::
+The output would then resemble:
```
jmikola, line 47: Likewise no `.. code-block:: none`.

docs/reference/method/MongoDBModelCollectionInfo-getType.txt
```diff
@@ -40,7 +40,7 @@ Examples
 
    echo $info->getType();
 
-The output would then resemble::
+The output would then resemble:
```
jmikola, line 43: `.. code-block:: none`?

jmikola (commented):
Found a few cases where it looks like `.. code-block:: none` was omitted. I might have missed some others, so double-check if that's important.

Otherwise, LGTM.

alcaeus:
Good catch. I double-checked the diff and found another couple of instances where I removed a double colon but forgot to add the `code-block`.

schmalliso:
The :php: interpreted text role issue has been addressed and is now live, so those errors have disappeared from the current log, which I'll attach here for posterity. There is also now guidance in the parser output to suggest resolutions for some of the the "Target not found" issues, mostly it seems like there are missing `()` on the end of the method name.

[php-library-alcaeus-pr.log](https://github.com/mongodb/mongo-php-library/files/10124711/php-library-alcaeus-pr.log)

alcaeus:
Thanks @schmalliso for the update. The feature of suggesting an alternative target was a great help, especially with figuring out what to do with `:query:$set`, which I replaced with `:update:$set` (both including the correct backticks which wouldn't render well here). I hope that's correct, but I'll double-check the current documentation to be on the safe side. If current documentation is fine, I'd merge this PR as-is so we have the bulk of the errors gone.

docs/reference/method/MongoDBCollection-explain.txt
```diff
@@ -50,18 +50,18 @@ Explainable Commands
...
+- `MongoDB\\Operation\\DeleteMany`
+- `MongoDB\\Operation\\DeleteOne`
+- `MongoDB\\Operation\\Distinct`
+- `MongoDB\\Operation\\Find`
+- `MongoDB\\Operation\\FindOne`
+- `MongoDB\\Operation\\FindOneAndDelete`
+- `MongoDB\\Operation\\FindOneAndReplace`
+- `MongoDB\\Operation\\FindOneAndUpdate`
+- `MongoDB\\Operation\\UpdateMany`
+- `MongoDB\\Operation\\UpdateOne`
```
jmikola, line 64: I think you want double backticks here for RST monospace formatting. Not sure if the backslash escaping is still necessary, or if that was just needed for the old `:phpclass:` syntax.

```suggestion
- ``MongoDB\\Operation\\Aggregate``
- ``MongoDB\\Operation\\Count``
- ``MongoDB\\Operation\\DeleteMany``
- ``MongoDB\\Operation\\DeleteOne``
- ``MongoDB\\Operation\\Distinct``
- ``MongoDB\\Operation\\Find``
- ``MongoDB\\Operation\\FindOne``
- ``MongoDB\\Operation\\FindOneAndDelete``
- ``MongoDB\\Operation\\FindOneAndReplace``
- ``MongoDB\\Operation\\FindOneAndUpdate``
- ``MongoDB\\Operation\\UpdateMany``
- ``MongoDB\\Operation\\UpdateOne``
```

PR #1012, opened 2022-12-06
PHPLIB-909: Add more error response tests

alcaeus:
This is a follow-up to PHPLIB-909 that removes the 4 previously skipped tests now that `WriteResult` exposes the errors as sent from the server. Since there may be multiple replies, I decided to only check the first one - another option would be to pass the test if any error response matches, but I'm not sure if that's feasible.

### DIFF
--- a/tests/UnifiedSpecTests/ExpectedError.php
+++ b/tests/UnifiedSpecTests/ExpectedError.php
@@ -7,6 +7,7 @@ use MongoDB\Driver\Exception\CommandException;
 use MongoDB\Driver\Exception\ExecutionTimeoutException;
 use MongoDB\Driver\Exception\RuntimeException;
 use MongoDB\Driver\Exception\ServerException;
+use MongoDB\Driver\Exception\WriteException;
 use MongoDB\Tests\UnifiedSpecTests\Constraint\Matches;
 use PHPUnit\Framework\Assert;
 use stdClass;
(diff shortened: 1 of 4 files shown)


tests/UnifiedSpecTests/ExpectedError.php
```diff
@@ -166,8 +170,15 @@ public function assert(?Throwable $e = null): void
...
-            assertInstanceOf(CommandException::class, $e);
-            assertThat($e->getResultDocument(), $this->matchesResultDocument, 'CommandException result document matches');
+            assertThat($e, logicalOr(isInstanceOf(CommandException::class), isInstanceOf(WriteException::class)));
+
+            if ($e instanceof CommandException) {
+                assertThat($e->getResultDocument(), $this->matchesResultDocument, 'CommandException result document matches');
+            } elseif ($e instanceof WriteException) {
+                $writeErrors = $e->getWriteResult()->getErrorReplies();
+                assertCount(1, $writeErrors);
+                assertThat($writeErrors[0], $this->matchesResultDocument, 'WriteException result document matches');
```
jmikola, line 180: This isn't addressed in the spec (see: [expectedError](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#expectederror)) but this is the correct approach. I intentionally only issued to single commands in the corresponding write command tests for this reason.

jmikola (approved):
LGTM if CI is green.

jmikola:
> Since there may be multiple replies, I decided to only check the first one - another option would be to pass the test if any error response matches, but I'm not sure if that's feasible.

I was just reminded of this PR. This assumption is consistent with what was done in libmongoc. See: https://github.com/mongodb/mongo-c-driver/pull/1144#discussion_r1021038593 and my comment in [DRIVERS-1547](https://jira.mongodb.org/browse/DRIVERS-1547?focusedCommentId=4981812&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4981812).

PR #1020, opened 2022-12-23
Ship `.gitattributes` file to shrink the dist package

OskarStark:
This is considered a best practice AFAIK

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,13 @@
+.* export-ignore
+*.md export-ignore
+tests export-ignore
+docs export-ignore
+examples export-ignore
+mongo-orchestration export-ignore
+tools export-ignore
+Makefile export-ignore
+phpcs.xml.dist export-ignore
+phpunit.evergreen.xml export-ignore
+phpunit.xml.dist export-ignore
+psalm.xml.dist export-ignore
+psalm-baseline.xml export-ignore


.gitattributes
```diff
@@ -0,0 +1,13 @@
+.* export-ignore
+*.md export-ignore
+tests export-ignore
+docs export-ignore
+examples export-ignore
+mongo-orchestration export-ignore
```
jmikola, line 6: [This directory](https://github.com/mongodb/mongo-php-library/tree/master/mongo-orchestration) hasn't been updated in over two years and I think it's obsolete since stopped using Travis CI and migrated to GitHub Actions (including the mongo-orchestration action in [drivers-evergreen-tools](https://github.com/mongodb-labs/drivers-evergreen-tools/)). I also found no references to this directory in the project.

@alcaeus: Do you concur that we should remove it entirely?

.gitattributes
```diff
@@ -0,0 +1,13 @@
...
+tools export-ignore
+Makefile export-ignore
```
jmikola, line 8: This file might just be a candidate for removal, per https://github.com/mongodb/mongo-php-library/pull/1021#discussion_r1057405814.

jmikola, line 8: This file might just be a candidate for removal, per https://github.com/mongodb/mongo-php-library/pull/1021#discussion_r1057405814.

.gitattributes
```diff
@@ -0,0 +1,13 @@
+.* export-ignore
+*.md export-ignore
+tests export-ignore
+docs export-ignore
+examples export-ignore
+mongo-orchestration export-ignore
+tools export-ignore
```
jmikola, line 7: I think it makes sense to include `examples/` and `tests/` in exports since they're intended for our users. @alcaeus: thoughts?

OskarStark: I would prefer to refer to the documentation in this case. If someone really needs this, one can hop over to github or use prefer source composer option.

We save a lot of traffic when thinking about CI ✌🏻

It's up to you 🤓

jmikola: I'm admittedly in the minority of users that frequently jumps through the `vendor/` directory to reference code. The impact on CI and users specifying `--prefer-source` are both good points. I'm on board with stripping all of this from the git archive in that case.

.gitattributes
jmikola: Opened [PHPLIB-1058](https://jira.mongodb.org/browse/PHPLIB-1058) to track this.

jmikola (approved):
I think this is good to merge as-is. The `mongo-orchestration` and `Makefile` lines can be addressed later in separate PRs as needed.

.gitattributes
OskarStark: Thank you 👍

OskarStark:
> I think this is good to merge as-is. The mongo-orchestration and Makefile lines can be addressed later in separate PRs as needed.

I will open PRs when back from vacation 👍

PR #1021, opened 2022-12-23
Place `.PHONY` in front of the commands

OskarStark:


Makefile
```diff
@@ -1,7 +1,6 @@
-.PHONY: composer test
```
jmikola, line 1: AFAIK, there's no functional difference between enumerating all phony targets here or doing so individually. One benefit to the individual approach you're suggesting here is that the attribute is closer to the target definition and we'd be less likely to forget to update this line when adding/removing a phony target down the line. Is that correct?

---

I'll defer to @alcaeus to chime in here when he returns from PTO, but I actually think this entire `Makefile` can just be removed. The `test` target isn't accurate since we actually use `vendor/bin/simple-phpunit` (as mentioned in `CONTRIBUTING.md`).

OskarStark: There is no technical difference, that's right.

However I prefer to have the PHONY next to the target, it makes a cleaner diff and if you copy paste a target, it's more easy to not forget it.

If you prefer to remove the Makefile @alcaeus, I will provide a PR

OskarStark:
I was lazy and used the `Update branch` button in the Github UI 😄

alcaeus:
Closing as the Makefile can definitely be removed. I'm running sanity checks on the mongo-orchestration folder (as drivers-evergreen-tools *can* use files from the project directory) before creating a PR to drop this as part of PHPLIB-1058.

PR #1032, opened 2023-01-10
Remove `<?php` in code examples

OskarStark:
Not sure if this works, but in the Symfony documentation we configured the syntax highlighting on a global level, so there is no need to add `<?php`, which also reduces some noise.
As you told me, there is some mongodb overall documentation stuff thingy, it might be a) not possible or b) it should be kept.

In this case, just close this PR, thanks.

### DIFF
--- a/docs/reference/method/MongoDBChangeStream-current.txt
+++ b/docs/reference/method/MongoDBChangeStream-current.txt
@@ -40,8 +40,6 @@ This example reports events while iterating a change stream.
 
 .. code-block:: php
 
-   <?php
-
    $uri = 'mongodb://rs1.example.com,rs2.example.com/?replicaSet=myReplicaSet';
 
    $collection = (new MongoDB\Client($uri))->test->inventory;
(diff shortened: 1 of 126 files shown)


docs/reference/method/MongoDBChangeStream-current.txt
```diff
@@ -40,8 +40,6 @@ This example reports events while iterating a change stream.
 
 .. code-block:: php
 
-   <?php
```
jmikola, line 43: @alcaeus: You'll need to confirm with docs if this is required for their syntax highlighter.

alcaeus: I've compared the current docs (left) and the build attached to this PR (right) side-by-side:
Docs comparison

At least the current version of the docs no longer does syntax highlighting without PHP tags, so we can't merge this for the time being. Maybe the move to snooty may allow us to remove these tags (@schmalliso would be able to comment on this), but for the time being we'll have to close this PR.

OskarStark: Thanks for the feedback 👍

schmalliso:
@alcaeus I believe that the new tooling doesn't require the `<?php` to recognize that it's valid PHP, based on the two PHP blocks on https://www.mongodb.com/docs/manual/indexes/#create-an-index.

PR #1033, opened 2023-01-10
Use `actions/cache@v3`

OskarStark:
Based on https://github.com/actions/cache/blob/main/RELEASES.md I cannot find any BC breaks


alcaeus (approved):

alcaeus:
Thanks @OskarStark!

PR #1041, opened 2023-01-17
[Documentation] Remove the words `easily` and `simply`

OskarStark:
Follows
* #1039 

In Symfony we try to be kind to newcomers, which means we try to avoid words like "simply", "easily" etc., because it depends on the knowledge of the reader and it can therefore be offensive.

### DIFF
--- a/docs/tutorial/tailable-cursor.txt
+++ b/docs/tutorial/tailable-cursor.txt
@@ -98,10 +98,10 @@ preceding example to use the Iterator methods directly:
    Calling ``$cursor->next()`` after the ``while`` loop naturally ends would
    throw an exception, since all results on the cursor have been exhausted.
 
-The purpose of this example is simply to demonstrate the functional equivalence
-between ``foreach`` and manual iteration with PHP's :php:`Iterator <iterator>`
-API. For normal cursors, there is little reason to manually iterate results
-instead of a concise ``foreach`` loop.
+The purpose of this example is to demonstrate the functional equivalencebetween
+``foreach`` and manual iteration with PHP's :php:`Iterator <iterator>`API.
+For normal cursors, there is little reason to manually iterate results instead
+of a concise ``foreach`` loop.
 
 Iterating a Tailable Cursor
 ---------------------------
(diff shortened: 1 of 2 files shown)


docs/tutorial/tailable-cursor.txt
```diff
@@ -98,10 +98,10 @@ preceding example to use the Iterator methods directly:
    Calling ``$cursor->next()`` after the ``while`` loop naturally ends would
    throw an exception, since all results on the cursor have been exhausted.
 
-The purpose of this example is simply to demonstrate the functional equivalence
-between ``foreach`` and manual iteration with PHP's :php:`Iterator <iterator>`
-API. For normal cursors, there is little reason to manually iterate results
-instead of a concise ``foreach`` loop.
+The purpose of this example is to demonstrate the functional equivalencebetween
```
OskarStark, line 101:
```suggestion
The purpose of this example is to demonstrate the functional equivalence between
```

GromNaN (approved):

docs/upgrade.txt
```diff
@@ -118,8 +118,8 @@ problematic:
   original BSON type.
 
 - Numerically-indexed PHP arrays would be serialized as BSON documents if there
-  was a gap in their key sequence. Such gaps were easily caused by unsetting a
-  key to remove an element and forgetting to numerically reindex the array.
+  was a gap in their key sequence. Such gaps were caused by unsetting a key to
```
jmikola, line 121: No objection to removing "easily" here, as the sentence is just as clear without it, but I'd argue the word match here is a false positive. This language wasn't assuming that something should be "easy" for users to understand or do; rather, it was stating that a common (and valid) operation in PHP could produce an unintentional outcome in BSON serialization.

jmikola (approved):
LGTM. Coincidentally, I was just reading https://css-tricks.com/words-avoid-educational-writing/ earlier today so this was fresh on my mind.

docs/upgrade.txt
OskarStark: I agree 👍🏻

OskarStark:
Thanks for the link 👌🏻

PR #1043, opened 2023-01-18
Skip unrelated builds for documentation

alcaeus:
With this change, GitHub Actions will no longer run the `coding-standards`,`static-analysis`, and `tests` workflows when only documentation was changed. At the same time, the `docs` workflow is only run if files in the docs folder were changed.

I'll caution that this may cause issues with branch protection: since we require the `tests` workflow to be stable in order for a PR to be mergeable, this may prove to be unfeasible due to us having to override branch protection rules for such pull requests.

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -6,11 +6,15 @@ on:
       - "v*.*"
       - "master"
       - "feature/*"
+    paths-ignore:
+      - "docs/**"
   push:
     branches:
       - "v*.*"
       - "master"
       - "feature/*"
+    paths-ignore:
+      - "docs/**"
 
 jobs:
   phpcs:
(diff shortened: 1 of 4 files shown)


.github/workflows/coding-standards.yml
```diff
@@ -6,11 +6,15 @@ on:
       - "v*.*"
       - "master"
       - "feature/*"
+    paths-ignore:
+      - "docs/**"
```
alcaeus, line 10: I decided to only skip builds if we're only changing documentation: there are too many variables that affect the "should this require a test run" to be feasible (e.g. changes in orchestration, tests, config files, sources, tests, etc.). Listing all the directories in `paths` doesn't seem sensible, and skipping builds for pull requests/commits that ONLY change documentation should be sufficient for the time being.

jmikola:
@alcaeus: I noticed this was never merged up to master. I'm doing that now while merging https://github.com/mongodb/mongo-php-library/pull/1053 up from v1.13.

PR #1045, opened 2023-01-18
PHPLIB-1037: Don't error when creating context for sharded clusters using mongodb+srv connection string

alcaeus:
PHPLIB-1037

The current logic makes some of the atlas tests fail as they are using a `mongodb+srv` URI. Since none of those tests currently use `useMultipleMongoses`, we remove the assertions that cause the failures to allow testing this scenario. I'm currently investigating how other drivers handle this to then correctly generate URIs with a single and with multiple mongoses based on the discovered servers.

### DIFF
--- a/tests/UnifiedSpecTests/UnifiedTestRunner.php
+++ b/tests/UnifiedSpecTests/UnifiedTestRunner.php
@@ -35,8 +35,6 @@ use function PHPUnit\Framework\assertIsArray;
 use function PHPUnit\Framework\assertIsString;
 use function PHPUnit\Framework\assertNotEmpty;
 use function PHPUnit\Framework\assertNotFalse;
-use function PHPUnit\Framework\assertStringContainsString;
-use function PHPUnit\Framework\assertStringStartsWith;
 use function preg_match;
 use function preg_replace;
 use function sprintf;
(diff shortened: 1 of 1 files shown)


tests/UnifiedSpecTests/UnifiedTestRunner.php
```diff
@@ -531,14 +528,15 @@ private function createContext(): Context
...
-            assertStringStartsWith('mongodb://', $multiMongosUri);
-            assertStringContainsString(',', parse_url($multiMongosUri, PHP_URL_HOST));
-
-            $singleMongosUri = self::removeMultipleHosts($multiMongosUri);
+            if (strpos($multiMongosUri, 'mongodb+srv://') === 0) {
+                /* TODO: If an SRV URI is provided, we can consider connecting and
+                 * checking the topology for multiple mongoses and then selecting a
+                 * single mongos to reconstruct a single mongos URI; however, that
+                 * may omit necessary URI options provided by TXT records. */
+                $singleMongosUri = $multiMongosUri;
```
jmikola, line 536: @alcaeus: How does this address [PHPLIB-1037](https://jira.mongodb.org/browse/PHPLIB-1037)? I'd expect this could introduce problems for tests that specify `useMultipleMongoses: false`, since they'd now run against a multi-mongos URI. Is that not the case?

alcaeus: On atlas, we're not running any tests with useMultipleMongoses of any value. That said, the assertion makes a number of tests fail as it's testing a sharded cluster with a mongodb+srv connection string. Removing the assertions fixes those tests without introducing any other test failures.

PR #1050, opened 2023-03-20
PHPLIB-913: Database::createEncryptedCollection() helper

jmikola:
https://jira.mongodb.org/browse/PHPLIB-913


tests/SpecTests/ClientSideEncryption/Prose21_AutomaticDataEncryptionKeysTest.php
```diff
@@ -0,0 +1,163 @@
...
+use function version_compare;
+
+/**
+ * Prose test 21: Automatic Data Encryption Keys
+ *
+ * @see https://github.com/mongodb/specifications/tree/master/source/client-side-encryption/tests#automatic-data-encryption-keys
+ * @group csfle
+ * @group serverless
+ */
+class Prose21_AutomaticDataEncryptionKeysTest extends FunctionalTestCase
```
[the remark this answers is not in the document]
alcaeus: I like the wording as you have it, let's suppress the checks here 👍

src/Database.php
```diff
@@ -300,6 +304,59 @@ public function createCollection(string $collectionName, array $options = [])
...
+     * If any error is encountered while creating data keys or invoking
+     * createCollection(), a CreateEncryptedCollectionException will be thrown.
+     * The original exception and modified "encryptedFields" option can be
+     * accessed via the getPrevious() and getEncryptedFields() methods,
+     * respectively.
+     *
+     * @see CreateCollection::__construct() for supported options
+     * @return array A tuple consisting of the createCollection() result and modified "encryptedFields" option
+     * @throws InvalidArgumentException for parameter/option parsing errors
+     * @throws CreateEncryptedCollectionException for any errors creating data keys or invoking createCollection()
```
alcaeus, line 324: This class is not included, adding a `use` statement will fix the psalm error.

src/Database.php
```diff
@@ -300,6 +304,59 @@ public function createCollection(string $collectionName, array $options = [])
...
+     */
+    public function createEncryptedCollection(string $collectionName, ClientEncryption $clientEncryption, string $kmsProvider, ?array $masterKey, array $options): array
+    {
+        if (! isset($options['encryptedFields']) || ! is_array($options['encryptedFields']) && ! is_object($options['encryptedFields'])) {
+            throw InvalidArgumentException::invalidType('"encryptedFields" option', $options['encryptedFields'] ?? null, ['array', 'object']);
+        }
+
+        $encryptedFields = (array) recursive_copy($options['encryptedFields']);
```
alcaeus, line 332: The following type would cover the `fields` property of the `encryptedFields` option:

```suggestion
        /** @var array{fields: list<array{path: string, bsonType: string, keyId: ?Binary}|object{path: string, bsonType: string, keyId: ?Binary}>} */
        $encryptedFields = (array) recursive_copy($options['encryptedFields']);
```

This could potentially be simplified by extracting the array shape and object shapes to separate [`@psalm-type`](https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-type) annotations and referencing them here, in which case I'd suggest also using `@psalm-var` instead of `@var`.

To make this available to end users, we can also include the type in the type annotation for the method parameter as well. However, I'd rather wait until we use psalm v5 for this, as it changes the way array and list shapes are handled: https://psalm.dev/docs/annotating_code/type_syntax/array_types/#unsealed-array-and-list-shapes

src/Database.php
```diff
@@ -300,6 +304,59 @@ public function createCollection(string $collectionName, array $options = [])
...
+                    }
+                }
+
+                $options['encryptedFields'] = $encryptedFields;
+            }
+
+            $result = $this->createCollection($collectionName, $options);
+
+            return [$result, $encryptedFields];
+        } catch (Exception $e) {
```
alcaeus, line 355: Should be a FQCN which would then be simplified by a `use` statement by `phpcbf`.

tests/SpecTests/ClientSideEncryption/FunctionalTestCase.php
```diff
@@ -0,0 +1,130 @@
...
+        $collection->drop();
+
+        if (empty($keyVaultData)) {
+            return;
+        }
+
+        $collection->insertMany($keyVaultData);
+    }
+
+    private function createInt64(string $value): Int64
```
alcaeus, line 72: Unrelated: I can't wait to finally drop support for 32-bit platforms so we can stop messing around with stuff like this...

src/Database.php
```diff
@@ -300,6 +304,59 @@ public function createCollection(string $collectionName, array $options = [])
...
+                        $field->keyId = $clientEncryption->createDataKey(...$createDataKeyArgs);
+                    }
+                }
+
+                $options['encryptedFields'] = $encryptedFields;
+            }
+
+            $result = $this->createCollection($collectionName, $options);
+
+            return [$result, $encryptedFields];
```
alcaeus, line 354: Is returning `$encryptedFields` the reason why we're copying the option earlier? What's the advantage of this approach over modifying `$options['encryptedFields']` directly and returning that?

src/Database.php
jmikola: The structure for `encryptedFields` is defined in [CSFLE: Terms](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#terms). It includes a bit more than what you have in the `@var` annotation, but I'm curious if that's even required. Should we only tell psalm about types to the extent that we're accessing data?

If so, a more concise form would be:

```
/** @var array{fields: list<array{keyId: ?Binary}|object{keyId: ?Binary}>} */
```

I'm also curious how `@psalm-type` works. The docs show it defined in a class-level doc block. Can it also be defined in a method doc block and apply to a more limited scope? Is there some higher scope (beyond a class) where you can define a type? I ask because this type isn't really specific to the Database class or this method, since it's shared between two helper methods and might also apply within the CreateCollection operation class.

A separate idea I wanted to explore in this PR (apart from revising the prose test organization) was to move the create/drop helpers with their CSFLE logic into separate operation classes. I'm not entirely satisfied with the current implementation, as validation of CreateCollection's options are all deferred until _after_ we generate data keys. I think there's room for improvement there, which would allow us to avoid needing to wrap InvalidArgumentExceptions. Also, refactoring the helpers into operations would reduce duplication between the drop helpers.

src/Database.php
jmikola: If `$options['encryptedFields']` is an object, we risk modifying the user's input argument. This is admittedly an edge case, but we technically allow `array|object` for "document" options such as this and I didn't want to impose a restriction.

On a related note, I'm not a fan of needing separate code paths for modifying `keyId`, but the field document (i.e. value of `$options['encryptedFields'][]`) can also be `array|object`. This has started me thinking on some utility APIs that might make it more pleasant to work with arrays and stdClass objects interchangeably.

There's another edge case I intentionally didn't explore, which is a `MongoDB\BSON\Serializable` instance being used for these options. Handling that would require eagerly "serializing" the input in the helper method before we inspect and modify `$options['encryptedFields']`. I don't think it's worth handling that (very unlikely to occur in practice) but I think it may deserve a comment to acknowledge the edge case.

tests/SpecTests/ClientSideEncryption/Prose21_AutomaticDataEncryptionKeysTest.php
```diff
@@ -0,0 +1,163 @@
...
+        );
+    }
+
+    /**
+     * @see https://github.com/mongodb/specifications/blob/bc37892f360cab9df4082922384e0f4d4233f6d3/source/client-side-encryption/tests/README.rst#case-3-invalid-keyid
+     * @dataProvider provideKmsProviderAndMasterKey
+     */
+    public function testCase3_InvalidKeyId(string $kmsProvider, ?array $masterKey): void
+    {
+        $this->expectException(CommandException::class);
```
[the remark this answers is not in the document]
jmikola: This was the result of the missing `use` statements in Database.php. Thanks for catching it!

tests/SpecTests/ClientSideEncryption/Prose21_AutomaticDataEncryptionKeysTest.php
```diff
@@ -0,0 +1,163 @@
...
+
+        $this->assertCommandSucceeded($result);
+        $this->assertInstanceOf(Binary::class, $encryptedFields['fields'][0]['keyId'] ?? null);
+
+        $this->expectException(BulkWriteException::class);
+        $this->expectExceptionMessage('Document failed validation');
+        $this->database->selectCollection($this->getCollectionName())->insertOne(['ssn' => '123-45-6789']);
+    }
+
+    public static function provideKmsProviderAndMasterKey(): iterable
```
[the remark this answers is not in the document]
jmikola: Fixed. I added a `use` statement to avoid using a FQCN within the file (consistent with other tests).

src/Database.php
alcaeus: Ah, yes, objects would be passed by reference and thus changed. That makes sense 👍

src/Database.php
alcaeus: > I'm also curious how `@psalm-type` works. The docs show it defined in a class-level doc block. Can it also be defined in a method doc block and apply to a more limited scope?

I'm not sure if they can be defined on a level lower than the class, I'd have to test that. As for a different scope, psalm offers the [`@psalm-import-type` annotation](https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-import-type) to allow importing types being declared somewhere else. Since the annotation syntax seems to reference the class name where the type was defined on, I believe we may want to declare types only on the class level to allow importing them somewhere else if necessary.

To bring an example, let's say you were to extract this logic into an operation class, in which case you'd want to declare the type in the operation class as that's where you're handling it. We could then import the type using `@psalm-import-type` here and reference it again as if it were declared in the class.

jmikola: Operation refactoring was completed. I'll resolve this.

src/Database.php
```diff
@@ -268,36 +277,63 @@ public function createCollection(string $collectionName, array $options = [])
...
+        if (! isset($options['writeConcern']) && ! is_in_transaction($options)) {
+            $options['writeConcern'] = $this->writeConcern;
         }
 
-        return $result;
+        $operation = new CreateEncryptedCollection($this->databaseName, $collectionName, $options);
+        $server = select_server($this->manager, $options);
+
+        try {
+            $operation->createDataKeys($clientEncryption, $kmsProvider, $masterKey, $encryptedFields);
```
jmikola, line 330: I originally implemented _two_ operation classes:

 * CreateCollectionForQueryableEncryption for `createCollection()` with an `encryptedFields` option, which composed several CreateCollection operations and one CreateIndexes operation.
 * CreateEncryptedCollection for this helper, which composed a CreateCollectionForQueryableEncryption.

In that approach, `CreateEncryptedCollection::execute()` created data keys before calling `CreateCollectionForQueryableEncryption::execute()` and contained the logic for CreateEncryptedCollectionException. The downside was updating the options for the inner-most CreateCollection operation when `encryptedFields` changed.

I also wasn't too happy with the naming for both operations. It made sense for CreateEncryptedCollection to match the helper method, but it was only responsible for creating data keys. CreateCollectionForQueryableEncryption was actually creating an encrypted collection.

----

I then combined both operations into CreateEncryptedCollection and moved the key generation logic to a separate `createDataKeys()` method to be called prior to `execute()`. This made it easier to reconstruct the inner CreateCollection operation.

Using a pass-by-reference parameter to expose modifications to `encryptedFields` allowed me to avoid dealing with CreateEncryptedCollectionException in the operation class entirely. I briefly explored returning `encryptedFields` _and_ throwing the exception when needed, and it made both `createDataKeys()` and the helper's `try/catch` more complicated.

src/Database.php
```diff
@@ -268,36 +277,63 @@ public function createCollection(string $collectionName, array $options = [])
...
+            $result = $operation->execute($server);
+
+            return [$result, $encryptedFields];
+        } catch (Throwable $e) {
```
jmikola, line 334: I considered limiting this to `\Exception`, but that would have required silencing a phpcs warning. I think it's unlikely we'd ever get an internal [PHP error](https://www.php.net/error) here, but there's no harm in wrapping it.

This is unlike [UnifiedSpecTest.php](https://github.com/mongodb/mongo-php-library/blob/446985d7d5fc1bc297b2394e5405bfb44f0b09fc/tests/UnifiedSpecTests/UnifiedSpecTest.php#L244), which has a specific reason to avoid catching errors.

src/Database.php
```diff
@@ -268,36 +277,63 @@ public function createCollection(string $collectionName, array $options = [])
...
+            throw new CreateEncryptedCollectionException($e, $encryptedFields ?? []);
```
jmikola, line 335: `$encryptedFields ?? []` satisfies two psalm warnings:

 * PossiblyUndefinedVariable: Possibly undefined variable $encryptedFields defined in try block (see https://psalm.dev/018)
 * PossiblyNullArgument: Argument 2 of CreateEncryptedCollectionException::__construct cannot be null, possibly null value provided (see https://psalm.dev/078)

In practice, I don't think that's possible since it is assigned in the first line of `createDataKeys()` and we've already verified that the option exists in the operation constructor.

I'm open to any feedback, but this seemed like a harmless, defensive fix.

src/Operation/CreateEncryptedCollection.php
```diff
@@ -0,0 +1,167 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace MongoDB\Operation;
+
+// phpcs:disable SlevomatCodingStandard.Namespaces.UnusedUses.UnusedUse
+use MongoDB\BSON\Binary;
```
jmikola, line 21: phpcs does not parse `@psalm-var` annotations, and using `@var` instead would produce a difference error:

> Invalid inline documentation comment format `@var array{fields: list<array{keyId: ?Binary}|object{keyId: ?Binary}>}`, expected `@var type $variable`

src/Operation/CreateEncryptedCollection.php
```diff
@@ -0,0 +1,167 @@
...
+    private $databaseName;
+
+    /** @var string */
+    private $collectionName;
+
+    /** @var array */
+    private $options;
+
+    /**
+     * @see CreateCollection::__construct() for supported options
```
jmikola, line 67: I can't tell you how great it felt to write an operation constructor doc block with option documentation copypasta.

src/Operation/CreateEncryptedCollection.php
```diff
@@ -0,0 +1,167 @@
...
+            $masterKey !== null ? ['masterKey' => $masterKey] : [],
+        ];
+
+        foreach ($encryptedFields['fields'] as $i => $field) {
+            // Skip invalid types and defer to the server to raise an error
+            if (! is_array($field) && ! is_object($field)) {
+                continue;
+            }
+
+            $field = (array) $field;
```
jmikola, line 156: The earliest version of this logic relied on `recursive_copy()` to ensure we didn't inadvertently modify an object still referenced by the caller. I eventually realized that array casting here would allow me to avoid that risk _and_ eliminate an extra code path I had for `$field` being an object.

src/Operation/CreateEncryptedCollection.php
```diff
@@ -0,0 +1,167 @@
...
+
+            if (array_key_exists('keyId', $field) && $field['keyId'] === null) {
+                $field['keyId'] = $clientEncryption->createDataKey(...$createDataKeyArgs);
+                $encryptedFields['fields'][$i] = $field;
+            }
+        }
+
+        $this->options['encryptedFields'] = $encryptedFields;
```
jmikola, line 164: Technically, this is only necessary if one or more data keys were created in the preceding `foreach`; however, the optimization didn't seem necessary (this is hardly a hot code path).

tests/Operation/CreateEncryptedCollectionFunctionalTest.php
```diff
@@ -0,0 +1,146 @@
...
+use const PATH_SEPARATOR;
+
+class CreateEncryptedCollectionFunctionalTest extends FunctionalTestCase
+{
+    public const LOCAL_MASTERKEY = 'Mng0NCt4ZHVUYUJCa1kxNkVyNUR1QURhZ2h2UzR2d2RrZzh0cFBwM3R6NmdWMDFBMUN3YkQ5aXRRMkhGRGdQV09wOGVNYUMxT2k3NjZKelhaQmRCZGJkTXVyZG9uSjFk';
+
+    /** @var ClientEncryption */
+    private $clientEncryption;
+
+    public function setUp(): void
```
jmikola, line 28: Most of this is lifted from the CSFLE prose tests. Copying this over reminded me of our previous discussion about using traits (vs. piling more onto the base FunctionalTestCase class), but that's something for a future PR.

alcaeus (approved):
LGTM - feel free to apply or ignore the suggested patches at your own discretion.

I like the way `CreateEncryptedCollection` is structured, and that people don't necessarily need to call `$db->createEncryptedCollection` to create an encrypted collection but can rely on `createCollection` to do the right thing. 🎉

jmikola:
Coding standard CI failure is due to [PHPLIB-1106](https://jira.mongodb.org/browse/PHPLIB-1106) and will be addressed by https://github.com/mongodb/mongo-php-library/pull/1058.

The CI build failures for `mongodb-latest` are due to [PHPLIB-1088](https://jira.mongodb.org/browse/PHPLIB-1088) and will be addressed in a subsequent PR.

PR #1054, opened 2023-04-03
Update docs for encryptedFields create/drop collection option

jmikola:
The MongoDB manual reference is more helpful than the CSFLE specification. Additionally, clarify how the option is used by the drop collection helpers.

This is a general docs improvement that came up while iterating on #1050.


alcaeus (approved):
LGTM

PR #1058, opened 2023-04-12
PHPLIB-1106: Upgrade doctrine/coding-standard to 11.1.0

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1106

This PR is based on the process from https://github.com/mongodb/mongo-php-library/commit/903e101d2e498d432a56b5fa97aeb709a57d036e (previous upgrade).

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -14,27 +14,13 @@
     <file>tests</file>
     <file>tools</file>
 
+    <!-- Target minimum supported PHP version -->
+    <config name="php_version" value="70200"/>
+
     <!-- ****************************************** -->
     <!-- Import rules from doctrine/coding-standard -->
     <!-- ****************************************** -->
     <rule ref="Doctrine">
-        <!-- ********************************************** -->
-        <!-- Exclude sniffs that require newer PHP versions -->
-        <!-- ********************************************** -->
-
-        <!-- Requires PHP 7.3 -->
-        <exclude name="SlevomatCodingStandard.Functions.RequireTrailingCommaInCall.MissingTrailingComma" />
-
-        <!-- Requires PHP 7.4 -->
-        <exclude name="SlevomatCodingStandard.ControlStructures.RequireNullCoalesceEqualOperator" />
-
-        <!-- Requires PHP 8.0 -->
-        <exclude name="SlevomatCodingStandard.Classes.ModernClassNameReference.ClassNameReferencedViaFunctionCall" />
-        <exclude name="SlevomatCodingStandard.Classes.RequireConstructorPropertyPromotion.RequiredConstructorPropertyPromotion" />
-        <exclude name="SlevomatCodingStandard.Exceptions.RequireNonCapturingCatch.NonCapturingCatchRequired" />
-        <exclude name="SlevomatCodingStandard.TypeHints.UnionTypeHintFormat.DisallowedShortNullable" />
-
-
         <!-- *********************************** -->
         <!-- Exclude sniffs that cause BC breaks -->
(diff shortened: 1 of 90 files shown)


phpcs.xml.dist
```diff
@@ -22,11 +22,17 @@
         <!-- Exclude sniffs that require newer PHP versions -->
         <!-- ********************************************** -->
 
+        <!-- Requires PHP 7.3 -->
+        <exclude name="SlevomatCodingStandard.Functions.RequireTrailingCommaInCall.MissingTrailingComma" />
```
jmikola, line 26: Out of curiosity, how does doctrine/coding-standard utilize these sniffs if it claims compatibility with PHP 7.2+ (see: [composer.json](https://github.com/doctrine/coding-standard/blob/11.1.0/composer.json#L25)). Does this have something to do with the version-specific patch files?

src/Model/BSONIterator.php
```diff
@@ -128,9 +128,7 @@ public function rewind()
         $this->advance();
     }
 
-    /**
-     * @see https://php.net/iterator.valid
-     */
+    /** @see https://php.net/iterator.valid */
```
jmikola, line 131: These fixes came from SlevomatCodingStandard.Commenting.RequireOneLineDocComment.

I don't mind them applying to `src/`. I think they make the most sense in `tests/` where we had a lot of `@dataProvider` annotations taking up multiple lines.

src/Model/ChangeStreamIterator.php
```diff
@@ -199,6 +197,7 @@ public function next(): void
 
         try {
             parent::next();
+
```
jmikola, line 200: From SlevomatCodingStandard.Classes.ParentCallSpacing.

I don't feel strongly about this and it was a small change affecting only this file.

alcaeus (approved):
LGTM, thanks!

phpcs.xml.dist
alcaeus: I'd have to run some tests to figure this one out, but it is possible that specifying a PHP version in the phpcs configuration (like `<config name="php_version" value="70200"/>`) would fix this. IIRC it depends on the sniff actively checking for it though, so it may not work for every sniff. That said, I'm fine excluding such sniffs manually until we upgrade to the required PHP version.

jmikola: Ah, that makes sense. I do see a PHP version check in [RequireTrailingCommaInCallSniff](https://github.com/slevomat/coding-standard/blob/8.10.0/SlevomatCodingStandard/Sniffs/Functions/RequireTrailingCommaInCallSniff.php#L47).

Per our earlier Slack conversation where you said: 

> I can only imagine it’s an issue due to us running phpcs on PHP 7.4 (e.g. an older version of phpdoc-parser that doesn’t support object shapes). You can try adding `<config name="php_version" value="70400"/>` to phpcs.xml.dist and then running phpcs on a newer PHP version to see if that fixes the issue

I misunderstood this and thought we were running `phpcs` with syntax rules for 7.4, which isn't the case. I now see you were just referring to the GitHub Action running `phpcs` through PHP 7.4.

I'll play around with this PR some more. I think adding `<config name="php_version" value="70200"/>` (and keeping that in sync with our minimum supported version) would be best if that allows us to remove the "Exclude sniffs that require newer PHP versions" block here.

alcaeus: Yes, I believe I confused this with psalm where the `phpVersion` config does something else. Adding the `php_version` configuration and keeping it in sync with our minimum required version sounds sensible to me.

PR #1064, opened 2023-04-18
PHPLIB-878: Spec tests for queryable encryption range indexes

jmikola:
https://jira.mongodb.org/browse/PHPLIB-878
https://jira.mongodb.org/browse/PHPLIB-1110

Depends on https://github.com/mongodb/mongo-php-driver/pull/1409


tests/SpecTests/ClientSideEncryption/Prose22_RangeExplicitEncryptionTest.php
```diff
@@ -0,0 +1,479 @@
...
+        }
+
+        /* Read the encryptedFields file directly into BSON to preserve typing
+         * for 64-bit integers. This means that DropEncryptedCollection and
+         * CreateEncryptedCollection will be unable to inspect the option for
+         * metadata collection names, but that's not necessary for the test. */
+        $encryptedFields = Document::fromJSON(file_get_contents(__DIR__ . '/../client-side-encryption/etc/data/range-encryptedFields-' . $type . '.json'));
+
+        $database = $this->encryptedClient->selectDatabase($this->getDatabaseName());
+        $database->dropCollection('explicit_encryption', ['encryptedFields' => $encryptedFields]);
```
GromNaN, line 88: Why do you need to set the encryptedFields when dropping the collection?

jmikola: This is mentioned in "Test Setup" section of [22. Range Explicit Encryption](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/tests/README.rst#range-explicit-encryption) but it's not explained in detail.

In addition to the main collection, queryable encryption uses additional metadata collections. The driver's `createCollection()` helper makes those when an `encryptedFields` option is provided. If an `encryptedFields` option is passed to the drop helper, it will additionally drop those metadata collections. The notable difference is that `encryptedFields` is actually an option on the `create` command, but not `drop`. The option here is entirely used client-side to tell PHPLIB to clean up additional collections. [Drop Collection Helper](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#drop-collection-helper) in the CSFLE spec has more detail on that.

jmikola:
Note: test failures on mongodb-latest are due to [PHPLIB-1071](https://jira.mongodb.org/browse/PHPLIB-1071) and [PHPLIB-1115](https://jira.mongodb.org/browse/PHPLIB-1115).

tests/SpecTests/ClientSideEncryption/Prose22_RangeExplicitEncryptionTest.php
```diff
@@ -0,0 +1,479 @@
...
+    public function setUp(): void
+    {
+        parent::setUp();
+
+        if ($this->isStandalone() || ($this->isShardedCluster() && ! $this->isShardedClusterUsingReplicasets())) {
+            $this->markTestSkipped('Range explicit encryption tests require replica sets');
+        }
+
+        if (version_compare($this->getServerVersion(), '7.0.0', '<')) {
+            $this->markTestSkipped('Range explicit encryption tests require MongoDB 7.0 or later');
```
GromNaN, line 45: We should create a helper `skipIfServerVersionNotSupported($version, $operator, $message)`. I create a ticket?

tests/SpecTests/ClientSideEncryption/Prose22_RangeExplicitEncryptionTest.php
```diff
@@ -0,0 +1,479 @@
...
+            'autoEncryption' => $autoEncryptionOpts,
+            /* libmongocrypt caches results from listCollections. Use a new
+             * client in each test to ensure its encryptedFields is applied. */
+            'disableClientPersistence' => true,
+        ]);
+    }
+
+    public function setUpWithTypeAndRangeOpts(string $type, array $rangeOpts): void
+    {
+        if ($type === 'DecimalNoPrecision' || $type === 'DecimalPrecision') {
```
GromNaN, line 77: With private consts, you can be sure that there is no typo in this strings.

tests/SpecTests/ClientSideEncryption/Prose22_RangeExplicitEncryptionTest.php
```diff
@@ -0,0 +1,479 @@
...
+            ['_id' => 0, $fieldName => $this->clientEncryption->encrypt($cast(0), $encryptOpts)],
+            ['_id' => 1, $fieldName => $this->clientEncryption->encrypt($cast(6), $encryptOpts)],
+            ['_id' => 2, $fieldName => $this->clientEncryption->encrypt($cast(30), $encryptOpts)],
+            ['_id' => 3, $fieldName => $this->clientEncryption->encrypt($cast(200), $encryptOpts)],
+        ]);
+    }
+
+    public function tearDown(): void
+    {
+        $this->collection = null;
```
GromNaN, line 112: We don't free the `Collection` instance in other test cases. Did you hit an issue?

tests/SpecTests/Operation.php
```diff
@@ -723,7 +723,7 @@ private function getResultAssertionTypeForCollection()
                     return ResultExpectation::ASSERT_NOTHING;
                 }
 
-                return ResultExpectation::ASSERT_SAME_DOCUMENTS;
+                return ResultExpectation::ASSERT_DOCUMENTS_MATCH;
```
GromNaN, line 726: If I understand correctly, you're making the assertion less strict, what's the reason?

tests/SpecTests/ClientSideEncryption/Prose22_RangeExplicitEncryptionTest.php
jmikola: SGTM. It can go on the base FunctionalTestCase class.

tests/SpecTests/ClientSideEncryption/Prose22_RangeExplicitEncryptionTest.php
jmikola: Will do. This also made me realize that we can replace most of the existing `private static` class properties with private constants. Opened [PHPLIB-1152](https://jira.mongodb.org/browse/PHPLIB-1152) to track that.

tests/SpecTests/Operation.php
jmikola: This is explained in the commit message for https://github.com/mongodb/mongo-php-library/pull/1064/commits/42bce4ea0e4c42b30ee7b8017ec76c537adae6d2. To quote:

> Allow extra fields in aggregate results
>
> Most operations that use ASSERT_SAME_DOCUMENT and ASSERT_SAME_DOCUMENTS should technically use ASSERT_MATCHES_DOCUMENT and ASSERT_DOCUMENTS_MATCH, respectively; however, this is the minimum change needed to satisfy new CSFLE spec tests. Related to dc1bb17.

[`assertSameDocuments()`](https://github.com/mongodb/mongo-php-library/blob/1.15.0/tests/TestCase.php#L95) does not permit extra fields in the actual value, while [`assertDocumentsMatch()`](https://github.com/mongodb/mongo-php-library/blob/1.15.0/tests/SpecTests/FunctionalTestCase.php#L91) does (see: [DocumentsMatchConstraint](https://github.com/mongodb/mongo-php-library/blob/1.15.0/tests/SpecTests/DocumentsMatchConstraint.php)).

The CSFLE legacy test runner inherits rules from the Transactions legacy test runner, which in turn inherits document matching rules from the [CRUD v1 runner](https://github.com/mongodb/specifications/blob/master/source/crud/tests/README.rst#evaluating-matches) -- which is where extra fields are permitted, among other things.

Per the commit message above, _more_ operations should be using the same logic here but I wanted to be conservative and avoid making extra changes to the legacy test runner (looking forward to the day we can just delete this code after porting everything to the unified format). No objections if you'd like to create a ticket to look into this at some point.

tests/SpecTests/ClientSideEncryption/Prose22_RangeExplicitEncryptionTest.php
GromNaN: https://jira.mongodb.org/browse/PHPLIB-1153

GromNaN (approved):

tests/SpecTests/ClientSideEncryption/Prose22_RangeExplicitEncryptionTest.php
jmikola: Since `encryptedClient` is created with `disableClientPersistence`, I originally did this to ensure that libmongoc's `mongoc_client_t` would be destroyed at the end of the test. This has been done before, but usually for session objects to ensure they are returned to the internal pool.

I just tested this locally and can confirm that without this, the Manager objects and `mongoc_client_t` instances accumulate and are not freed until the end of the test _suite_ (not just the test _case_). I opened [PHPLIB-1154](https://jira.mongodb.org/browse/PHPLIB-1154) to track this.

In the meantime, I've added a comment to explain this and removed the unsetting of `key1Id`, which wasn't necessary.

jmikola:
There is one unrelated [task failure](https://spruce.mongodb.com/task/mongo_php_library_test_mongodb_versions__php_edge_versions~latest_stable_mongodb_versions~latest_driver_versions~latest_stable_os~debian11_test_sharded_cluster_patch_d95e7623390b8f657755f6601859d4649612a40d_64775cb00305b9cba85d368f_23_05_31_14_41_53/tests?execution=0&sortBy=STATUS&sortDir=ASC). I opened [PHPLIb-1155](https://jira.mongodb.org/browse/PHPLIB-1155) to track that.

PR #1071, opened 2023-05-17
PHPLIB-693: Remove early returns for failed tests from tearDown methods

GromNaN:
Fix [PHPLIB-693](https://jira.mongodb.org/browse/PHPLIB-693)

Ensure `parent::tearDown` is always called. The call to `disableFailPoints` in `FunctionalTestCase::tearDown` was not necessary since fail points are not used in the modified test classes.

### DIFF
--- a/tests/DocumentationExamplesTest.php
+++ b/tests/DocumentationExamplesTest.php
@@ -40,11 +40,9 @@ class DocumentationExamplesTest extends FunctionalTestCase
 
     public function tearDown(): void
     {
-        if ($this->hasFailed()) {
-            return;
+        if (! $this->hasFailed()) {
+            $this->dropCollection();
         }
-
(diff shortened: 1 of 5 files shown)


alcaeus (approved):
LGTM. Test failures are unrelated.

tests/DocumentationExamplesTest.php
```diff
@@ -40,12 +40,10 @@ public function setUp(): void
 
     public function tearDown(): void
     {
-        if ($this->hasFailed()) {
-            return;
+        if (! $this->hasFailed()) {
+            $this->dropCollection();
```
jmikola, line 44: Noted that this will likely change in https://github.com/mongodb/mongo-php-library/pull/1075 but it's fine for this PR.

PR #1072, opened 2023-05-17
PHPLIB-892: Skip CSFLE automatic encryption tests for non-enterprise servers

GromNaN:
Fixes [PHPLIB-892](https://jira.mongodb.org/browse/PHPLIB-892).

According to the [doc](https://www.mongodb.com/docs/manual/core/csfle/#introduction), automatic CSFLE is only supported by Enterprise (and Atlas) servers.


alcaeus (approved):
LGTM

tests/DocumentationExamplesTest.php
```diff
@@ -1821,9 +1821,7 @@ public function testQueryableEncryption(): void
             $this->markTestSkipped('Explicit encryption tests require MongoDB 7.0 or later');
         }
 
-        if (! $this->isEnterprise()) {
```
jmikola, line 1824: In [PHPLIB-892](https://jira.mongodb.org/browse/PHPLIB-892?focusedCommentId=5447736&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-5447736) you wrote:

> I'll run the `DocumentationExamplesTest::testQueryableEncryption()` with `crypt_shared` and/or `mongocryptd` available with the community server. If it passes, let's remove the `isEnterprise()` method and change the logic in `DocumentationExamplesTest::testQueryableEncryption()` to instead check if either `crypt_shared` or `mongocryptd` is available.

`skipIfClientSideEncryptionIsNotSupported()` doesn't check for either of those things. Did something change since you wrote that comment?

Also, in an earlier comment on that issue I pointed out that ``DocumentationExamplesTest::testQueryableEncryption()`` did require `crypt_shared` and/or `mongocryptd` (since it uses queryable encryption) but the prose test did not (since it explicitly encrypts). You could run the prose tests w/o `crypt_shared` and `mongocryptd` to verify that.

GromNaN:
The skipped tests effectively fail when both are disabled.

> MongoDB\Driver\Exception\ConnectionTimeoutException : mongocryptd error: No suitable servers found: `serverselectiontimeoutms` timed out: [connection refused calling hello on 'localhost:27020']:

tests/FunctionalTestCase.php
```diff
@@ -571,6 +560,40 @@ protected function skipIfTransactionsAreNotSupported(): void
         }
     }
 
+    /**
+     * The Automatic Encryption Shared Library is installed with MongoDB Enterprise or Atlas (version 6.0 and later).
```
jmikola, line 564: This comment is going to be outdated once we start testing `crypt_shared` with older server versions going back to 4.2. See: [DRIVERS-2465](https://jira.mongodb.org/browse/DRIVERS-2465) and [DOCSP-29387](https://jira.mongodb.org/browse/DOCSP-29387) for more context.

I think both `@see` links are sufficient here and in `isMongocryptdAvailable()` below, if you'd like to remove the comment there as well.

Since I originally reported [PHPLIB-892](https://jira.mongodb.org/browse/PHPLIB-892), I've learned that MongoDB Enterprise (at least the server binary) is not relevant to whether the driver can perform automatic encryption, so I see no reason to mention it in a comment intended to be read by driver developers.

tests/UnifiedSpecTests/UnifiedTestRunner.php
```diff
@@ -357,30 +353,6 @@ private function isClientSideEncryptionSupported(): bool
         return static::isCryptSharedLibAvailable() || static::isMongocryptdAvailable();
     }
 
-    private static function isCryptSharedLibAvailable(): bool
```
jmikola, line 360: @eramongodb, @kevinAlbs: The `csfle` [runOnRequirement](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#runonrequirement) dates back to https://github.com/mongodb/specifications/commit/0f65908778188a8d6eeb22571c7791a2b1674234 and [DRIVERS-2017](https://jira.mongodb.org/browse/DRIVERS-2017). Reading the existing language in the unified spec, there's no indication that `crypt_shared` and/or `mongocryptd` are required. 

> If true, the tests MUST only run if the driver and server support Client-Side Field Level Encryption. A server supports CSFLE if it is version 4.2.0 or higher. If false, tests MUST only run if CSFLE is not enabled. If this field is omitted, there is no CSFLE requirement.

The PHPLIB unified test runner currently checks three things:

 * Server FCV >= 4.2
 * PHPC is compiled with libmongocrypt (not just libmongoc)
 * At least one of `crypt_shared` and `mongocryptd` is available

I realize that the only unified CSFLE tests at the moment are for key management, which isn't an enterprise feature. But ignoring that for a moment and assuming legacy tests will eventually be ported to the unified format, was it your intention that the `csfle` requirement also ensure that queryable encryption be supported?

If yes, then I'd like to clarify this in the unified test spec. If no, then I think we should remove the `crypt_shared`/`mongocryptd` check from PHPLIB's unified test runner (and maybe still clarify that in the unified spec).

tests/SpecTests/ClientSideEncryptionSpecTest.php
```diff
@@ -2001,28 +1995,4 @@ private function prepareCorpusData(string $fieldName, stdClass $data, ClientEncr
 
         return $data->allowed ? $returnData : $data;
     }
-
```
jmikola, line 2004: @eramongodb, @kevinAlbs: On a related note, our legacy CSFLE spec test runner also requires at least one of `crypt_shared` and `mongocryptd` to be available, but I don't see that explicitly mentioned in [Use as integration tests](https://github.com/mongodb/specifications/tree/master/source/client-side-encryption/tests#use-as-integration-tests). I understand why they _would_ be required since various tests due cover automatic encryption.

Am I just missing the text, or is it only implied?

tests/UnifiedSpecTests/UnifiedTestRunner.php
kevinAlbs: > At least one of `crypt_shared` and `mongocryptd` is available

I expect `crypt_shared` or `mongocryptd` would be required once the unified test format supports automatic encryption. I expect it is not necessary for the current unified test format. I suggest keeping the check for `crypt_shared` or `mongocryptd`, as I expect it will be needed in the future.

> was it your intention that the `csfle` requirement also ensure that queryable encryption be supported?

I think yes. Both "Client-Side Field Level Encryption" and "Queryable Encryption" require libmongocrypt. As of DRIVERS-2454, "In-Use Encryption" can be used to refer to both features.

tests/SpecTests/ClientSideEncryptionSpecTest.php
kevinAlbs: I think it is implied, but the test README may benefit from edits:

> Do the following before running spec tests:
>
> - If available for the platform under test, obtain a crypt_shared binary and place it in a location accessible to the tests. Refer to: Using crypt_shared
> - Start the mongocryptd process.

This may benefit from edits. I expect mongocryptd does not need to be started, only available on the PATH for drivers to spawn. If testing with crypt_shared, I expect mongocryptd does not need to be available.

tests/UnifiedSpecTests/UnifiedTestRunner.php
jmikola: Thanks. I opened https://github.com/mongodb/specifications/pull/1439 and [DRIVERS-2661](https://jira.mongodb.org/browse/DRIVERS-2661) to update the unified/legacy test docs.

tests/DocumentationExamplesTest.php
jmikola: @GromNaN: In light of https://github.com/mongodb/specifications/pull/1439, I would be in favor of adding a check for `crypt_shared` and/or `mongocryptd` to the base `skipIfClientSideEncryptionIsNotSupported()` method.

A call to `skipIfClientSideEncryptionIsNotSupported()` could then be used here (after the topology and 7.0+ version check). I'm aware the version check in `skipIfClientSideEncryptionIsNotSupported()` would be redundant, but that's OK.

You could then also remove the explicit crypt_shared/mongocryptd check in `ClientSideEncryptionSpecTest::setUp()`.

With that done, `skipIfClientSideEncryptionIsNotSupported()` would become consistent with `UnifiedTestRunner::isClientSideEncryptionSupported()` and all of our tests would perform the same checks.

GromNaN: > I'm aware the version check in skipIfClientSideEncryptionIsNotSupported() would be redundant, but that's OK.

There is a slight difference between the server version check and the feature compatibility version check in `skipIfClientSideEncryptionIsNotSupported`. So this is not redundant.

GromNaN: I followed your instructions.

tests/UnifiedSpecTests/UnifiedTestRunner.php
eramongodb: The `csfle` requirement was added as an (overly?) simple all-or-nothing check for whether anything related to Client Side Encryption is required by the test, which implicitly requires - or at least, was assumed to require - either `mongocryptd` (the only consideration back then) or `crypt_shared` (a new option since the wording was added), since most CSE functions cannot fulfill their specified behavior without them, including the Key management API. I support the clarification of wording that documents the `mongocryptd`/`crypt_shared` requirement; submitted an approving review for https://github.com/mongodb/specifications/pull/1439 accordingly.

tests/DocumentationExamplesTest.php
jmikola: Aye, technically different checks. I said "redundant" as our test suite is never run on mixed-version or upgraded replica sets, so FCV will always equate the server version.

tests/UnifiedSpecTests/UnifiedTestRunner.php
```diff
@@ -354,31 +352,7 @@ private function isClientSideEncryptionSupported(): bool
...
-        $paths = explode(PATH_SEPARATOR, getenv("PATH"));
-
-        foreach ($paths as $path) {
-            if (is_executable($path . DIRECTORY_SEPARATOR . 'mongocryptd')) {
-                return true;
-            }
-        }
-
-        return false;
+        return FunctionalTestCase::isCryptSharedLibAvailable() || FunctionalTestCase::isMongocryptdAvailable();
```
jmikola, line 355: This is one thing I had in mind when creating [PHPLIB-1170](https://jira.mongodb.org/browse/PHPLIB-1170).

jmikola (approved):

PR #1073, opened 2023-05-19
PHPLIB-1056 Improve type coverage on option arrays

GromNaN:
Fix [PHPLIB-1056](https://jira.mongodb.org/browse/PHPLIB-1056)

I think I'm going a bit to far by documenting array shapes. This is very useful for code completion. This is very verbose and contagious, and leads to new phpstan errors like this:

```
ERROR: DocblockTypeContradiction - src/Operation/Aggregate.php:209:15 - Docblock-defined type bool for $options['useCursor'] is always bool (see https://psalm.dev/155)
        if (! is_bool($options['useCursor'])) {
```

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -196,8 +196,8 @@ class Client
      * Drop a database.
      *
      * @see DropDatabase::__construct() for supported options
-     * @param string $databaseName Database name
-     * @param array  $options      Additional options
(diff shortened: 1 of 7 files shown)


GromNaN:
Closing as this is not what was expected.

PR #1075, opened 2023-05-19
PHPLIB-816: Cleanup databases and collections in DocumentationExamplesTest

GromNaN:
Fix [PHPLIB-816](https://jira.mongodb.org/browse/PHPLIB-816)

- Modify the tests that use a single collection to use "inventory"
- List extra databases required by transaction tests to drop them in `tearDown`


tests/DocumentationExamplesTest.php
```diff
@@ -1194,11 +1208,10 @@ public function testRunCommand_example_1(): void
     public function testRunCommand_example_2(): void
     {
         $db = new Database($this->manager, $this->getDatabaseName());
-        $db->dropCollection('restaurants');
-        $db->createCollection('restaurants');
+        $db->createCollection('inventory');
 
         // Start runCommand Example 2
-        $cursor = $db->command(['collStats' => 'restaurants']);
+        $cursor = $db->command(['collStats' => 'inventory']);
```
jmikola, line 1214: The content between `// Start` and `// End` should not be changed, as this is parsed by the docs team for inclusion in the MongoDB server manual. Examples are expected to be consistent across languages. 

For example, see the language selection widget in the right column on https://www.mongodb.com/docs/manual/tutorial/insert-documents/.

On a somewhat related note: "runCommand Example 2" will probably be removed when we implement [PHPLIB-1065](https://jira.mongodb.org/browse/PHPLIB-1065), but that's outside the scope of this PR.

tests/DocumentationExamplesTest.php
```diff
@@ -28,9 +30,14 @@
  * @see https://jira.mongodb.org/browse/DRIVERS-356
  * @see https://jira.mongodb.org/browse/DRIVERS-488
  * @see https://jira.mongodb.org/browse/DRIVERS-547
+ *
+ * Note: Example should use the "inventory" collection in the default database unless they need multiple databases.
```
jmikola, line 34: Some thoughts:

The initial `dropCollection()` during `setUp()` is mainly for the benefit of ensuring clean test execution after a previous failure. This correlates with `tearDown()` NOP-ing on a failure, which is more for the benefit of local testing so we inspect a collection after running a failed test in isolation.

Since the test methods in this class are using various databases _and_ collections, I think it makes sense to:

 * Remove the the overridden `getCollectionName()` method in this class
 * Remove the unconditional `dropCollection()` calls in `setUp()` and `tearDown()`.
 * Ensure nothing else in this test is relying on the default collection name

In its place:

 * Allow each test to register any number of databases or individual collections that it will use. For a custom database (as with the transaction examples) you can drop it entirely as you're already doing in this PR. For a collection in the default database, I think it'd make sense to just target the collection directly (since other tests in the suite are sharing the default database).
 * To handle "setUp" behavior, optionally have the registration method drop things or handle that in a second method that you call immediately after. If the dropping is done in a separate method, then you can re-use that in `tearDown()`.
 * The main `setUp()` and `tearDown()` methods should also be responsible for clearing the database/collection registry.

This idea is somewhat related to what I ended up doing with `setUpWithTypeAndRangeOpts()` in [mongodb/mongo-php-library#1064](https://github.com/mongodb/mongo-php-library/pull/1064/files#diff-86fa81787b2fd04d6e360e683890362020e60ca3b3b796570b19aea9ca9ec09bR75), as I needed to customize setup behavior based on the data provider's input.

tests/DocumentationExamplesTest.php
```diff
@@ -1788,6 +1809,7 @@ public function testWithTransactionExample(): void
         $this->skipIfTransactionsAreNotSupported();
 
         $uriString = static::getUri(true);
+        $this->dbNames = ['mydb1', 'mydb2'];
```
jmikola, line 1812: Per my earlier suggestion about registering databases/collections and dropping them before a test, I do think it's wise to keep that logic _after_ stuff like `skipIfTransactionsAreNotSupported()` to avoid unnecessary work.

Based on where you placed `$this->dbNames = [...]` here I assume you may have already had that in mind.

tests/DocumentationExamplesTest.php
GromNaN: Hum, ok. I suspected those examples to be automatically parsed, but not that much integrated into the doc. I'll implement a more generic mechanism to drop databases or collection invoked by `createCollection` or `dropCollection`.

tests/FunctionalTestCase.php
```diff
@@ -64,6 +68,10 @@ public function setUp(): void
 
     public function tearDown(): void
     {
+        if (! $this->hasFailed()) {
+            $this->cleanupCollections();
```
GromNaN, line 72: Overriding the `tearDown` method in child classes is no longer necessary: the collection will be automagically dropped when the test calls `createCollection` or `dropCollection`.

tests/DocumentationExamplesTest.php
```diff
@@ -31,24 +31,9 @@
...
-        if (! $this->hasFailed()) {
-            $this->dropCollection();
-        }
-
-        parent::tearDown();
-    }
-
     public function testExample_1_2(): void
     {
+        $this->dropCollection(null, 'inventory');
```
GromNaN, line 36: It is more verbose to explicitly define the name of the collection in each test case, but this is how I can handle the variety of names used in the documentation.

tests/FunctionalTestCase.php
```diff
@@ -252,33 +260,65 @@ public static function getModuleInfo(string $row): ?string
...
     /**
-     * Drops the test collection with the specified options.
-     *
-     * If the "writeConcern" option is not specified but is supported by the
-     * server, a majority write concern will be used. This is helpful for tests
-     * using transactions or secondary reads.
+     * Drops the test collection and ensures it is dropped again during tearDown().
      */
-    protected function dropCollection(array $options = []): void
+    protected function dropCollection(?string $databaseName = null, ?string $collectionName = null): void
```
GromNaN, line 289: I am tempted to reverse `$databaseName` & `$collectionName`. Both are optional, but `$collectionName` is more often modified than `$databaseName`.

tests/FunctionalTestCase.php
```diff
@@ -252,33 +260,65 @@ public static function getModuleInfo(string $row): ?string
...
+
+        $operation = new DropCollection($databaseName, $collectionName, $options);
         $operation->execute($this->getPrimaryServer());
+        $this->collectionsToCleanup[$databaseName][$collectionName] = true;
+    }
+
+    protected function cleanupCollections(): void
+    {
+        foreach ($this->collectionsToCleanup as $databaseName => $collectionNames) {
+            // The default database is never dropped.
```
GromNaN, line 309: I have noticed that when I delete all the collections in a database, the database is no longer listed. Could we delete the default database directly?

tests/FunctionalTestCase.php
jmikola: This seems fine, as I can't think of any places in the library where both a database and collection are accepted and we use that reverse order. Seeing an explicit `null` in the calls will also serve as a reminder that we're using the default database.

tests/FunctionalTestCase.php
jmikola: I think this comment predates our [Slack conversation](https://mongodb.slack.com/archives/C026UA4GF0B/p1684745839221829). Feel free to resolve if so.

tests/FunctionalTestCase.php
```diff
@@ -252,33 +260,58 @@ public static function getModuleInfo(string $row): ?string
     }
 
     /**
-     * Creates the test collection with the specified options.
+     * Creates the test collection with the specified options  and ensures it is dropped again during tearDown().
```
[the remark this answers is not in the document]
GromNaN: In PHPStorm, the limit is 120 by default. I think this is a good limit. We have a lot of lines of code or links in the comments that exceed 80.

tests/Operation/RenameCollectionFunctionalTest.php
```diff
@@ -3,7 +3,6 @@
 namespace MongoDB\Tests\Operation;
 
 use MongoDB\Driver\Exception\CommandException;
-use MongoDB\Operation\DropCollection;
```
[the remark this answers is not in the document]
GromNaN: I modified this one to remove the tearDown override, so I'll use `createCollection` and `dropCollection` where it makes sense:

- FindFunctionalTest
- ExplainFunctionalTest
- ListIndexesFunctionalTest
- MapReduceFunctionalTest
- ChangeStreamIteratorTest

tests/DocumentationExamplesTest.php
```diff
@@ -1291,8 +1293,8 @@ public function testTransactions_intro_example_1(): void
         $client = static::createTestClient();
 
         /* The WC is required: https://mongodb.com/docs/manual/core/transactions/#transactions-and-locks */
```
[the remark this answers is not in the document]
GromNaN: I don't think we need this comment at all once it's on `createCollection` and `dropCollection`.

tests/DocumentationExamplesTest.php
```diff
@@ -1471,12 +1473,12 @@ public function testTransactions_retry_example_3(): void
...
```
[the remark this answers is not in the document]
GromNaN: Removed the comment here. The method that is called has this explanations.

tests/DocumentationExamplesTest.php
```diff
@@ -1471,12 +1473,12 @@ public function testTransactions_retry_example_3(): void
...
-        $client->hr->dropCollection('employees', ['writeConcern' => new WriteConcern('majority')]);
-        $client->reporting->dropCollection('events', ['writeConcern' => new WriteConcern('majority')]);
+        $this->dropCollection('hr', 'employees');
+        $this->dropCollection('reporting', 'events');
 
         /* Collections need to be created before a transaction starts */
-        $client->hr->createCollection('employees', ['writeConcern' => new WriteConcern('majority')]);
-        $client->reporting->createCollection('events', ['writeConcern' => new WriteConcern('majority')]);
+        $this->createCollection('hr', 'employees');
+        $this->createCollection('reporting', 'events');
```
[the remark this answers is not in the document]
GromNaN: Applied.

tests/DocumentationExamplesTest.php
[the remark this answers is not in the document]
jmikola: SGTM. Feel free to remove all instances of this in the file.

tests/FunctionalTestCase.php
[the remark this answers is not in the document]
jmikola: OK to leave as-is. I opened [PHPLIB-1138](https://jira.mongodb.org/browse/PHPLIB-1138) to look into increasing the limit to 120 (and perhaps re-wrapping existing comments).

tests/DocumentationExamplesTest.php
```diff
@@ -1865,6 +1865,8 @@ public function testQueryableEncryption(): void
             $this->markTestSkipped('Automatic encryption requires MongoDB Enterprise');
         }
 
+        $this->dropCollection();
```
jmikola, line 1868: I just remembered that the entire database is dropped below:

```php
$client->selectDatabase($databaseName)->drop(['writeConcern' => new WriteConcern(WriteConcern::MAJORITY)]);
```

I think we should probably remove `$this->dropCollection();` here.

The multi-line comment below explains that the database is dropped because "setUp() only drops the collection under test", which is no longer true. I'd suggest revising that comment to say:

> Create a client without auto encryption. Drop existing data in both the keyvault and database under test. The latter is necessary to clean up any internal collections for queryable encryption.

Note: this highlights an issue I may have mentioned in a previous comment about your create and drop helpers not handling encrypted collections. We don't need to deal with this now. You can consider making a new PHPLIB ticket to revisit that in the future (and just mark it as related to PHPLIB-816). The solution for that will probably entail creating a `MongoDB\Collection` object to use _its_ create/drop helpers instead using the specific operation classes as you do now.

----

For additional context: this CSFLE test is unique among other tests in the file in that it is _not_ parsed into the server manual. It was only added as a reference for the docs team in [PHPLIB-863](https://jira.mongodb.org/browse/PHPLIB-863).

tests/Operation/RenameCollectionFunctionalTest.php
```diff
@@ -25,18 +24,8 @@ public function setUp(): void
...
-    public function tearDown(): void
-    {
-        if (! $this->hasFailed()) {
-            $operation = new DropCollection($this->getDatabaseName(), $this->toCollectionName);
-            $operation->execute($this->getPrimaryServer());
-        }
-
-        parent::tearDown();
+        $this->dropCollection($this->getDatabaseName(), $this->toCollectionName);
+        $this->dropCollection($this->getDatabaseName(), $this->getCollectionName());
```
jmikola, line 28: I think I missed this earlier, but I noticed you're passing explicit arguments here instead of `null, null`. I realize it's more verbose but I actually think it makes these lines more readable.

If you agree, feel free to apply this everywhere and remove the "default on null" logic. If you prefer the "default on null" logic, then I'd suggest revising this to make it consistent with other tests.

jmikola (approved):
Two minor suggestions, which you can address on your own.

tests/Operation/RenameCollectionFunctionalTest.php
GromNaN: I removed the default `null`. That's better for readability.

tests/DocumentationExamplesTest.php
GromNaN: The `dropCollection` is here to ensure the collection is dropper on teadDown. I'll add a comment to explain the duplicate with the database drop.

This test cannot run until https://github.com/mongodb/mongo-php-library/pull/1064

jmikola: > The dropCollection is here to ensure the collection is dropper on teadDown. I'll add a comment to explain the duplicate with the database drop.

Ah, that makes sense. Since the metadata collections are going to be left behind anyway (per the second-to-last paragraph in my last response), I don't think it's worth addressing this now. You should make a note of it in the follow-up issue, though (and cross-reference this comment thread for context).

GromNaN: I updated `dropCollection` and `cleanupCollections` so they pass the options to `Collection::drop($options)`. This effectively drops the metadata collections.

tests/DocumentationExamplesTest.php
```diff
@@ -1585,17 +1570,19 @@ public function testSnapshotQueries(): void
             $this->markTestSkipped('Snapshot read concern is only supported with replicasets');
         }
 
+        $this->dropCollection('pets', 'cats');
```
jmikola, line 1573: Since `dropCollection()` now returns a Collection object, you could remove the three sequential `dropCollection()` calls here and have them replace the original `selectCollection()` calls below:

```diff
-$catsCollection = $client->selectCollection('pets', 'cats');
-$catsCollection->drop();
+$catsCollection = $this->dropCollection('pets', 'cats');
 $catsCollection->insertMany([
```

Doing so will mean that `$client = static::createTestClient();` is no longer needed so early in the function, so I'd suggest moving that further down as well:

```diff
+$client = static::createTestClient();

 ob_start();

 // Start Snapshot Query Example 1
```

Also, the three `Collection::drop()` calls outside of start/end example blocks can be removed, since you're handling that in the new `tearDown()` logic.

```diff
 $this->assertSame(3, $adoptablePetsCount);

-$catsCollection->drop();
-$dogsCollection->drop();
```

```diff
 $this->assertSame(1, $totalDailySales);

-$salesCollection->drop();
```

jmikola, line 1573: Since `dropCollection()` now returns a Collection object, you could remove the three sequential `dropCollection()` calls here and have them replace the original `selectCollection()` calls below:

```diff
-$catsCollection = $client->selectCollection('pets', 'cats');
-$catsCollection->drop();
+$catsCollection = $this->dropCollection('pets', 'cats');
 $catsCollection->insertMany([
```

Doing so will mean that `$client = static::createTestClient();` is no longer needed so early in the function, so I'd suggest moving that further down as well:

```diff
+$client = static::createTestClient();

 ob_start();

 // Start Snapshot Query Example 1
```

Also, the three `Collection::drop()` calls outside of start/end example blocks can be removed, since you're handling that in the new `tearDown()` logic.

```diff
 $this->assertSame(3, $adoptablePetsCount);

-$catsCollection->drop();
-$dogsCollection->drop();
```

```diff
 $this->assertSame(1, $totalDailySales);

-$salesCollection->drop();
```

tests/DocumentationExamplesTest.php
```diff
@@ -1870,13 +1858,13 @@ public function testQueryableEncryption(): void
...
         $namespace = $this->getNamespace();
 
-        /* Create a client without auto encryption. Drop existing data in both
-         * the keyvault and database under test. The latter is necessary since
-         * setUp() only drops the collection under test, which will leave behind
-         * internal collections for queryable encryption. */
+        /* Create a client without auto encryption. Drop existing data in both the
+         * keyvault and the collection under test, the "encryptedFields" option
+         * is necessary to clean up any internal collections for queryable
+         * encryption, assuming they use their default names */
```
jmikola, line 1864: Good comment :+1:

tests/Model/IndexInfoFunctionalTest.php
```diff
@@ -14,17 +14,8 @@ public function setUp(): void
     {
         parent::setUp();
 
+        $this->dropCollection($this->getDatabaseName(), $this->getCollectionName());
```
jmikola, line 17: This would be consistent with what you did above in ChangeStreamIteratorTest.

```suggestion
        $this->collection = $this->dropCollection($this->getDatabaseName(), $this->getCollectionName());
```

Note: I couldn't remove the following line in the suggestion since it wasn't modified in the PR.

tests/Collection/CrudSpecFunctionalTest.php
```diff
@@ -48,8 +48,7 @@ public function setUp(): void
     {
         parent::setUp();
 
-        $this->expectedCollection = new Collection($this->manager, $this->getDatabaseName(), $this->getCollectionName() . '.expected');
-        $this->expectedCollection->drop();
+        $this->expectedCollection = $this->dropCollection($this->getDatabaseName(), $this->getCollectionName() . '.expected');
```
jmikola, line 51: Noted that `$this->collection` isn't handled here because you already do so in the base Collection\FunctionalTestCase class.

tests/Collection/FunctionalTestCase.php
```diff
@@ -19,15 +19,6 @@ public function setUp(): void
...
-    }
-
-    public function tearDown(): void
-    {
-        if (! $this->hasFailed()) {
-            $this->dropCollection();
-        }
-
-        parent::tearDown();
+        $this->dropCollection($this->getDatabaseName(), $this->getCollectionName());
```
jmikola, line 22:
```suggestion
        $this->collection = $this->dropCollection($this->getDatabaseName(), $this->getCollectionName());
```

Along with removing the line above (can't suggest since it's outside of the PR diff).

I saw you do this in ChangeStreamIteratorTest and elsewhere.

tests/FunctionalTestCase.php
```diff
@@ -252,33 +259,55 @@ public static function getModuleInfo(string $row): ?string
...
+     * A majority write concern is applied by default to ensure that the
+     * transaction can acquire the required locks.
+     * See: https://www.mongodb.com/docs/manual/core/transactions/#transactions-and-operations
      */
-    protected function createCollection(array $options = []): void
+    protected function createCollection(string $databaseName, string $collectionName, array $options = []): Collection
     {
-        $options += ['writeConcern' => new WriteConcern(WriteConcern::MAJORITY)];
+        if (isset($options['encryptedFields'])) {
+            throw new InvalidArgumentException('The "encryptedFields" option is not supported by createCollection(). Time to refactor!');
```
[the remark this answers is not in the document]
GromNaN: I'd like to use `Collection::create`. https://jira.mongodb.org/browse/PHPLIB-1145

tests/FunctionalTestCase.php
```diff
@@ -252,33 +260,62 @@ public static function getModuleInfo(string $row): ?string
...
+     * transaction can acquire the required locks.
+     * See: https://www.mongodb.com/docs/manual/core/transactions/#transactions-and-operations
+     *
+     * @param array $options CreateCollection options
      */
-    protected function createCollection(array $options = []): void
+    protected function createCollection(string $databaseName, string $collectionName, array $options = []): Collection
     {
-        $options += ['writeConcern' => new WriteConcern(WriteConcern::MAJORITY)];
+        // @see https://jira.mongodb.org/browse/PHPLIB-1145
```
jmikola, line 275: I asked this in another PR earlier today, but should this be `See: ` since we're not in a doc block? Fine either way as it's a matter of personal preference.

I suppose this makes it easier to grep for `@see` in the project.

jmikola (approved):
One very minor question but LGTM.

Great work on this :tada:

tests/FunctionalTestCase.php
GromNaN: I notices the `See: ` in the phpdoc of this method. I think it's better to stick on 1 way to write comments.

PR #1078, opened 2023-05-23
PHPLIB-881: Consolidate validation of aggregation pipeline options

GromNaN:
Fix PHPLIB-881

Leverage `is_pipeline` to validate `$pipeline` argument and `"pipeline"` option. Since this function returns a boolean, it does not provide context about the error. This change allows to factorise the code and to go further in the validation, but the message in the exception is less precise.

### DIFF
--- a/src/Operation/Aggregate.php
+++ b/src/Operation/Aggregate.php
@@ -31,7 +31,6 @@ use MongoDB\Exception\UnexpectedValueException;
 use MongoDB\Exception\UnsupportedException;
 use stdClass;
 
-use function array_is_list;
 use function current;
 use function is_array;
 use function is_bool;
@@ -40,7 +39,7 @@ use function is_object;
 use function is_string;
 use function MongoDB\create_field_path_type_map;
 use function MongoDB\is_last_pipeline_operator_write;
-use function sprintf;
+use function MongoDB\is_pipeline;
 
 /**
  * Operation for the aggregate command.
@@ -132,20 +131,14 @@ class Aggregate implements Executable, Explainable
      *
      * @param string      $databaseName   Database name
      * @param string|null $collectionName Collection name
-     * @param array       $pipeline       List of pipeline operations
+     * @param array       $pipeline       Aggregation pipeline
      * @param array       $options        Command options
      * @throws InvalidArgumentException for parameter/option parsing errors
      */
     public function __construct(string $databaseName, ?string $collectionName, array $pipeline, array $options = [])
     {
-        if (! array_is_list($pipeline)) {
-            throw new InvalidArgumentException('$pipeline is not a list');
-        }
-
-        foreach ($pipeline as $i => $operation) {
-            if (! is_array($operation) && ! is_object($operation)) {
-                throw InvalidArgumentException::invalidType(sprintf('$pipeline[%d]', $i), $operation, 'array or object');
-            }
+        if (! is_pipeline($pipeline, true /* allowEmpty */)) {
+            throw new InvalidArgumentException('$pipeline is not a valid aggregation pipeline');
         }
 
         $options += ['useCursor' => true];
--- a/src/Operation/CreateCollection.php
+++ b/src/Operation/CreateCollection.php
@@ -24,15 +24,13 @@ use MongoDB\Driver\Session;
 use MongoDB\Driver\WriteConcern;
 use MongoDB\Exception\InvalidArgumentException;
 
-use function array_is_list;
-use function assert;
 use function current;
 use function is_array;
 use function is_bool;
 use function is_integer;
 use function is_object;
 use function is_string;
-use function sprintf;
+use function MongoDB\is_pipeline;
 use function trigger_error;
 
 use const E_USER_DEPRECATED;
@@ -242,18 +240,8 @@ class CreateCollection implements Executable
             trigger_error('The "autoIndexId" option is deprecated and will be removed in a future release', E_USER_DEPRECATED);
         }
 
-        if (isset($options['pipeline'])) {
-            $pipeline = $options['pipeline'];
-            assert(is_array($pipeline));
-            if (! array_is_list($pipeline)) {
-                throw new InvalidArgumentException('The "pipeline" option is not a list');
-            }
-
-            foreach ($options['pipeline'] as $i => $operation) {
-                if (! is_array($operation) && ! is_object($operation)) {
-                    throw InvalidArgumentException::invalidType(sprintf('$options["pipeline"][%d]', $i), $operation, 'array or object');
-                }
-            }
+        if (isset($options['pipeline']) && ! is_pipeline($options['pipeline'], true /* allowEmpty */)) {
+            throw new InvalidArgumentException('"pipeline" option is not a valid aggregation pipeline');
         }
 
         $this->databaseName = $databaseName;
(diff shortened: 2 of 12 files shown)


src/Operation/Aggregate.php
```diff
@@ -138,18 +138,9 @@ class Aggregate implements Executable, Explainable
...
-                throw new InvalidArgumentException(sprintf('$pipeline is not a list (unexpected index: "%s")', $i));
-            }
-
-            if (! is_array($operation) && ! is_object($operation)) {
-                throw InvalidArgumentException::invalidType(sprintf('$pipeline[%d]', $i), $operation, 'array or object');
-            }
-
-            $expectedIndex += 1;
+        // Empty pipelines are allowed
+        if ($pipeline !== [] && ! is_pipeline($pipeline)) {
```
GromNaN, line 142: `is_pipeline` returns `false` for an empty list. There is a test that assert an empty pipeline is valid `AggregateFunctionalTest::testEmptyPipelineReturnsAllDocuments`.

I wonder if empty list is allowed elsewhere and if the constraint should be relaxed in `is_pipeline` to be explicitly stricter when necessary.

src/Operation/Aggregate.php
```diff
@@ -138,18 +138,9 @@ class Aggregate implements Executable, Explainable
...
+            throw new InvalidArgumentException('$pipeline is not a valid list of pipeline operations');
```
alcaeus, line 143:
```suggestion
            throw new InvalidArgumentException('$pipeline is not a valid aggregation pipeline');
```

src/Operation/CreateCollection.php
```diff
@@ -241,20 +241,8 @@ public function __construct(string $databaseName, string $collectionName, array
...
-                }
-
-                if (! is_array($operation) && ! is_object($operation)) {
-                    throw InvalidArgumentException::invalidType(sprintf('$options["pipeline"][%d]', $i), $operation, 'array or object');
-                }
-
-                $expectedIndex += 1;
-            }
+        if (isset($options['pipeline']) && ! is_pipeline($options['pipeline'])) {
+            throw new InvalidArgumentException('"pipeline" option is not a valid list of pipeline operations');
```
alcaeus, line 245:
```suggestion
            throw new InvalidArgumentException('"pipeline" option is not a valid aggregation pipeline');
```

alcaeus (commented):
Suggested rewording the exception message for clarity, as we generally should refer to this as an "aggregation pipeline" as it is usually called in the documentation.

src/Operation/Aggregate.php
alcaeus: I'm on the fence about this. In general I'd say that `is_pipeline` should return `true` for anything that is a valid aggregation pipeline, which includes an empty array. In those cases where we want to forbid using an empty pipeline, we should explicitly guard against a non-empty array. Thoughts @jmikola?

src/Operation/Aggregate.php
alcaeus: Yes, I think that would make sense.

GromNaN: Done.

src/Operation/BulkWrite.php
```diff
@@ -229,7 +229,7 @@ public function __construct(string $databaseName, string $collectionName, array
                         throw InvalidArgumentException::invalidType(sprintf('$operations[%d]["%s"][1]', $i, $type), $args[1], 'array or object');
                     }
 
-                    if (! is_first_key_operator($args[1]) && ! is_pipeline($args[1])) {
+                    if ($args[1] === [] || ! is_first_key_operator($args[1]) && ! is_pipeline($args[1])) {
```
GromNaN, line 232: I need to check the spec to see if an empty array is accepted in the following cases:
- BulkWrite UPDATE_MANY and UPDATE_ONE
- FindOneAndUpdate
- ReplaceOne
- Update, UpdateMany, UpdateOne

GromNaN: So I'll change the behavior for `ReplaceOne` that is currently rejecting empty array.

alcaeus: Should be fine - just make sure it is covered by a test.

FWIW, the shell behaves the same way:
```
test> // Replace with empty replacement
test> db.test.replaceOne({_id: 1}, {})
{
  acknowledged: true,
  insertedId: null,
  matchedCount: 1,
  modifiedCount: 1,
  upsertedCount: 0
}
test> // Update with empty document
test> db.test.updateOne({_id: 1}, {})
MongoInvalidArgumentError: Update document requires atomic operators
test> // Update with empty pipeline
test> db.test.updateOne({_id: 1}, [])
MongoInvalidArgumentError: Update document requires atomic operators
```

GromNaN: Thanks for checking, I'll adopt this reflex.

Test case already added: https://github.com/mongodb/mongo-php-library/pull/1078/files#diff-6e13b3e2ae70bee5981233671f8a994c0111a3cf84e2f958e723bdd08c9c83d2R45

tests/Operation/ReplaceOneTest.php
```diff
@@ -31,29 +31,31 @@ public function testConstructorReplacementArgument($replacement): void
...
         return $this->wrapValuesForDataProvider([
+            [],
             ['y' => 1],
             (object) ['y' => 1],
             new BSONDocument(['y' => 1]),
         ]);
     }
 
-    public function provideUpdateDocuments()
+    public function provideInvalidReplacementDocuments()
```
jmikola, line 52: Note: this is going to conflict with https://github.com/mongodb/mongo-php-library/pull/1077. The changes in that PR are more involved, so I'd suggest waiting on that to be merged.

Consideration of PackedArray instances for pipeline values is also going to impact this PR.

GromNaN:
PR rebased.

> Consideration of PackedArray instances for pipeline values is also going to impact this PR.

The dumb comparison `$pipeline === []` is not possible with `PackedArray`. So I added a, argument to `is_pipeline` to specify the return value in case of empty list. This also simplified conditions in existing `if` statements.

src/Operation/Aggregate.php
jmikola: Note that `is_pipeline()` was only created for validation of `update` and `findAndModify` command options (not `aggregate`). The original prohibition for empty arrays dates back to https://github.com/mongodb/mongo-php-library/commit/2628c098e3bfc015d7e2d59d1437cf1fed046e12#diff-e093a0fd8f57549d81ad2871e6ca3872ecaef99d0525118e26c0501bdd0c1d92R241 from [PHPLIB-418](https://jira.mongodb.org/browse/PHPLIB-418).

I don't recall why that was done. I recently confirmed in [DRIVERS-2638](https://jira.mongodb.org/browse/DRIVERS-2638) that empty pipelines are accepted for updates.

alcaeus (approved):
LGTM. While I'm not a fan of the comments indicating an option name, I'm fine with having them since we can't use named arguments for the purpose.

alcaeus:
> I copied the style introduced by @jmikola and [discussed in review](https://github.com/mongodb/mongo-php-library/pull/1098/files#r1223980183). Could be reverted.

Ah, I missed that one. No worries then, perfectly fine with it 👍

PR #1080, opened 2023-05-25
PHPLIB-802: Send readConcern but not writeConcern to explain commands

GromNaN:
Fix [PHPLIB-802](https://jira.mongodb.org/browse/PHPLIB-802)

`ReadConcern` is relevant for query plan evaluation, not `WriteConcern` ([comment](https://jira.mongodb.org/browse/SERVER-28678?focusedCommentId=3457279&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-3457279))

This should be covered by unit tests PHPLIB-1144.

### DIFF
--- a/src/Operation/Aggregate.php
+++ b/src/Operation/Aggregate.php
@@ -309,7 +309,14 @@ class Aggregate implements Executable, Explainable
      */
     public function getCommandDocument(Server $server)
     {
-        return $this->createCommandDocument();
+        $cmd = $this->createCommandDocument();
+
+        // Read concern can change the query plan
+        if (isset($this->options['readConcern'])) {
+            $cmd['readConcern'] = $this->options['readConcern'];
+        }
+
+        return $cmd;
     }
 
     /**
--- a/src/Operation/Delete.php
+++ b/src/Operation/Delete.php
@@ -178,13 +178,7 @@ class Delete implements Executable, Explainable
      */
     public function getCommandDocument(Server $server)
     {
-        $cmd = ['delete' => $this->collectionName, 'deletes' => [['q' => $this->filter] + $this->createDeleteOptions()]];
-
-        if (isset($this->options['writeConcern'])) {
-            $cmd['writeConcern'] = $this->options['writeConcern'];
-        }
-
-        return $cmd;
+        return ['delete' => $this->collectionName, 'deletes' => [['q' => $this->filter] + $this->createDeleteOptions()]];
     }
 
     /**
--- a/src/Operation/Find.php
+++ b/src/Operation/Find.php
@@ -331,7 +331,14 @@ class Find implements Executable, Explainable
      */
     public function getCommandDocument(Server $server)
     {
-        return $this->createCommandDocument();
+        $cmd = $this->createCommandDocument();
+
+        // Read concern can change the query plan
+        if (isset($this->options['readConcern'])) {
+            $cmd['readConcern'] = $this->options['readConcern'];
+        }
+
+        return $cmd;
     }
 
     /**
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -286,10 +286,11 @@
       <code>$this-&gt;options['typeMap']</code>
       <code>$this-&gt;options['typeMap']</code>
     </MixedArgument>
-    <MixedAssignment occurrences="5">
+    <MixedAssignment occurrences="6">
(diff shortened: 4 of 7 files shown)


src/Operation/Delete.php
```diff
@@ -180,10 +180,6 @@ public function getCommandDocument(Server $server)
     {
         $cmd = ['delete' => $this->collectionName, 'deletes' => [['q' => $this->filter] + $this->createDeleteOptions()]];
 
-        if (isset($this->options['writeConcern'])) {
```
GromNaN, line 183: I'm not sure if `readConcern` is relevant for delete operations.

alcaeus: The `Delete` operation doesn't take a `readConcern` option, as the command doesn't support it either. If that was a typo and you meant `writeConcern`, then yes, it is relevant for delete operations as it is for all other writing commands, but it is not relevant when used in `explain`.

GromNaN: Ah well, yes. That was not a typo. I didn't even think to check if the operation supported readConcern. So I'm doing the right thing by deleting these lines of code.

alcaeus: Yes, that's correct.

src/Operation/Aggregate.php
```diff
@@ -322,7 +322,7 @@ private function createCommandDocument(): array
             'pipeline' => $this->pipeline,
         ];
 
-        foreach (['allowDiskUse', 'bypassDocumentValidation', 'comment', 'explain', 'maxTimeMS'] as $option) {
+        foreach (['allowDiskUse', 'bypassDocumentValidation', 'comment', 'explain', 'maxTimeMS', 'readConcern'] as $option) {
```
jmikola, line 325: `readConcern` should _not_ be added directly to the command document. The option is already [handled in `executeCommand()`](https://github.com/mongodb/mongo-php-library/blob/1.15.0/src/Operation/Aggregate.php#L372) along with other options for the execute method.

That code path allows for sensible inheritance of client-level options. I'd have to double-check libmongoc internals to confirm, but I think directly adding it in a command document here might result in duplicate BSON fields being sent to the server. I know that's possible if you were to manually inject a session ID in the `lsid` field.

Instead, I think you should add this option in `getCommandDocument()`. You can use the array union operator to do so and just add `['readConcern' => ...]` to `$this->createCommandDocument()` before returning. A more verbose approach with an assigned variable is fine as well.

----

Likewise for other Explainable commands where you modified the value used to construct the Command object instead of the `Explainable::getCommandDocument()` implementation.

src/Operation/Find.php
```diff
@@ -349,6 +349,9 @@ private function createCommandDocument(): array
         // maxAwaitTimeMS is a Query level option so should not be considered here
         unset($options['maxAwaitTimeMS']);
 
+        // writeConcern is not relevant to evaluating the query plan
+        unset($options['writeConcern']);
```
jmikola, line 353: The Find operation has never documented a `writeConcern` option. Moreover, `createQueryOptions()` never creates a `writeConcern` key in its return array. This can be removed.

jmikola (commented):
I noted the discussion in [SERVER-28678](https://jira.mongodb.org/browse/SERVER-28678) that `readConcern` might be relevant but `writeConcern` is not. Feel free to amend the original title/description of PHPLIB-802 to accurately reflect the change.

src/Operation/Aggregate.php
GromNaN: I understand better how it works now.
I opted for the more verbose way to add the `readConcern` option when available: its easier to read and [better for code analysis](https://localheinz.com/articles/2023/03/18/avoiding-one-liners-in-php/).

src/Operation/FindAndModify.php
```diff
@@ -264,7 +264,14 @@ public function execute(Server $server)
      */
     public function getCommandDocument(Server $server)
     {
-        return $this->createCommandDocument();
+        $cmd = $this->createCommandDocument();
+
+        // Read concern can change the query plan
```
jmikola, line 269: [findAndModify](https://www.mongodb.com/docs/manual/reference/command/findAndModify/) doesn't support a `readConcern` option. This should be removed.

Easy mistake to make as PHPLIB-802 was originally focused on findAndModify.

jmikola (approved):
LGTM after removing the change to FindAndModify.php.

jmikola (approved):

PR #1081, opened 2023-05-26
PHPLIB-1135: Remove var_dump from tests, use print_r instead

GromNaN:
`var_dump` output is modified by [xdebug](https://xdebug.org/docs/develop), what causes tests to fail during debug.

I think it's easier to read this way.

There other calls to `var_dump` in `DocumentationExamplesTest`, but they are hidden and not tested.

### DIFF
--- a/examples/persistable.php
+++ b/examples/persistable.php
@@ -105,5 +105,4 @@ $collection->insertOne($entry);
 
 $foundEntry = $collection->findOne([]);
 
-/** @psalm-suppress ForbiddenCode */
 print_r($foundEntry);
--- a/examples/persistable.php
+++ b/examples/persistable.php
@@ -10,7 +10,7 @@ use MongoDB\Model\BSONArray;
 use UnexpectedValueException;
 
 use function getenv;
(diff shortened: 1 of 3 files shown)


examples/persistable.php
```diff
@@ -106,4 +106,4 @@ public function bsonUnserialize(array $data): void
 $foundEntry = $collection->findOne([]);
 
 /** @psalm-suppress ForbiddenCode */
```
jmikola, line 108: ForbiddenCode suppressions can be removed as only `var_dump` and `shell_exec` are included by default: https://psalm.dev/docs/running_psalm/configuration/#forbiddenfunctions

I tested this locally and confirmed.

jmikola (approved):
LGTM with removed `psalm-suppress` annotations.

examples/persistable.php
GromNaN: Done. This further impacts my lines added < lines deleted ratio. 😂

alcaeus (approved):
LGTM 👍

PR #1082, opened 2023-05-26
PHPLIB-475: Use ReadPreference string constant consistently in the code and the doc

GromNaN:
Fix PHPLIB-475
Caused by PHPC-1489
Block PHPC-1021

The `ReadPreference::RP_*` constants are deprecated and replaced by the string constants for readability.

- Replace `int` constants by their equivalent `string` constants
- Use `$readPreference->getModeString()` instead of `$readPreference->getMode()` to compare string values.

I used [rector](https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md#renameclassconstfetchrector) to automate the refactoring. Library users can do the same by copying the rule in `rector.php`. I have an issue to change the constructor argument from the string `"primary"` to the constant `MongoDB\Driver\ReadPreference::PRIMARY` https://github.com/rectorphp/rector/issues/7956. More rector automation expected in PHPLIB-1141.

```php
<?php // rector.php

use Rector\Config\RectorConfig;
use Rector\Renaming\Rector\ClassConstFetch\RenameClassConstFetchRector;
use Rector\Renaming\ValueObject\RenameClassConstFetch;

return static function (RectorConfig $rectorConfig): void {
    $rectorConfig->paths([
        __DIR__ . '/examples',
        __DIR__ . '/src',
        __DIR__ . '/tests',
        __DIR__ . '/tools',
    ]);

    $rectorConfig->ruleWithConfiguration(RenameClassConstFetchRector::class, [
        new RenameClassConstFetch(MongoDB\Driver\ReadPreference::class, 'RP_PRIMARY', 'PRIMARY'),
        new RenameClassConstFetch(MongoDB\Driver\ReadPreference::class, 'RP_PRIMARY_PREFERRED', 'PRIMARY_PREFERRED'),
        new RenameClassConstFetch(MongoDB\Driver\ReadPreference::class, 'RP_SECONDARY', 'SECONDARY'),
        new RenameClassConstFetch(MongoDB\Driver\ReadPreference::class, 'RP_SECONDARY_PREFERRED', 'SECONDARY_PREFERRED'),
        new RenameClassConstFetch(MongoDB\Driver\ReadPreference::class, 'RP_NEAREST', 'NEAREST'),
    ]);
};
```


tests/ClientTest.php
```diff
@@ -95,7 +95,7 @@ public function testSelectCollectionInheritsOptions(): void
         $this->assertInstanceOf(ReadConcern::class, $debug['readConcern']);
         $this->assertSame(ReadConcern::LOCAL, $debug['readConcern']->getLevel());
         $this->assertInstanceOf(ReadPreference::class, $debug['readPreference']);
-        $this->assertSame(ReadPreference::RP_SECONDARY_PREFERRED, $debug['readPreference']->getMode());
+        $this->assertSame(ReadPreference::SECONDARY_PREFERRED, $debug['readPreference']->getMode());
```
GromNaN, line 98: This assert fails. The value is the int `6`, not the string `secondaryPreferred`. 
Looking at the [driver test](https://github.com/mongodb/mongo-php-driver/blob/master/tests/manager/bug0851-001.phpt#L42), the value of `Manager::getReadPreference()` should be the one sent in options?

GromNaN: image

The inner value of `$debug['readPreference']->mode` is `secondaryPreferred`, but the method `getMode()` returns `6`.

GromNaN: I could have used `get_object_vars` to access the inner property, but I'll compare `ReadPreference` objects instead.

tests/ClientTest.php
```diff
@@ -94,8 +94,7 @@ public function testSelectCollectionInheritsOptions(): void
 
         $this->assertInstanceOf(ReadConcern::class, $debug['readConcern']);
         $this->assertSame(ReadConcern::LOCAL, $debug['readConcern']->getLevel());
```
GromNaN, line 96: I can go a step further and refactor the `ReadConcern` and `WriteConcern` assertions. Maybe in an other PR.

src/functions.php
```diff
@@ -578,7 +578,7 @@ function select_server_for_aggregate_write_stage(Manager $manager, array &$optio
 
     /* If there is either no read preference or a primary read preference, there
      * is no special server selection logic to apply. */
-    if ($readPreference === null || $readPreference->getMode() === ReadPreference::RP_PRIMARY) {
+    if ($readPreference === null || $readPreference->getMode() === ReadPreference::PRIMARY) {
```
GromNaN, line 581: Thanks Psalm for the trigger. Getting the primary mode integer without using the constant could be done like this, but I feel this is not optimal.
```suggestion
    static $primaryMode;
    $primaryMode = $primaryMode ?? (new ReadPreference(ReadPreference::PRIMARY))->getMode();
    
    if ($readPreference === null || $readPreference->getMode() === $primaryMode) {
```

I notice this function is not covered by tests.

docs/reference/method/MongoDBClient-selectCollection.txt
```diff
@@ -72,7 +72,7 @@ with a custom read preference:
        'test',
        'users',
        [
-           'readPreference' => new MongoDB\Driver\ReadPreference('primaryPreferred'),
+           'readPreference' => new MongoDB\Driver\ReadPreference(MongoDB\Driver\ReadPreference::PRIMARY_PREFERRED),
```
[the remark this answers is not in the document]
jmikola: I don't feel strongly about this as the strings values are often used as-is in connection strings (similar to "majority" for write concerns). This may makes the example code harder to read in the docs UI since there's no wrapping, but we already have that problem for method signatures.

That said, the UI issue is a separate issue we'll need to address down the line (such as guidelines on wrapping and splitting method declarations across multiple lines for readability).

rector.php
```diff
@@ -0,0 +1,29 @@
...
+        __DIR__ . '/examples',
+        __DIR__ . '/src',
+        __DIR__ . '/tests',
+        __DIR__ . '/tools',
+        __FILE__,
+    ]);
+
+    // ReadPreference::RP_* constants are deprecated in favor of the string constants
+    $rectorConfig->ruleWithConfiguration(RenameClassConstFetchRector::class, [
+        new RenameClassConstFetch('MongoDB\Driver\ReadPreference', 'RP_PRIMARY', 'PRIMARY'),
```
jmikola, line 18: I haven't used rector before, but are class names typically string literals or should you prefer something like `MongoDB\Driver\ReadPreference::class`? I assume what you have here may be preferable to avoid any dependency on the extension itself.

rector.php
```diff
@@ -0,0 +1,29 @@
...
+use Rector\Renaming\Rector\ClassConstFetch\RenameClassConstFetchRector;
+use Rector\Renaming\ValueObject\RenameClassConstFetch;
+
+return static function (RectorConfig $rectorConfig): void {
+    $rectorConfig->paths([
+        __DIR__ . '/examples',
+        __DIR__ . '/src',
+        __DIR__ . '/tests',
+        __DIR__ . '/tools',
+        __FILE__,
```
jmikola, line 13: What's the logic for including the configuration file itself?

tests/SpecTests/TransactionsSpecTest.php
```diff
@@ -324,7 +324,7 @@ private static function killAllSessions(): void
         }
 
         $manager = static::createTestManager();
-        $primary = $manager->selectServer(new ReadPreference('primary'));
+        $primary = $manager->selectServer(new ReadPreference(ReadPreference::PRIMARY));
```
jmikola, line 327:
```suggestion
        $primary = $manager->selectServer();
```

Since PHPC 1.11, [`selectServer()`](https://www.php.net/manual/en/mongodb-driver-manager.selectserver.php#refsect1-mongodb-driver-manager.selectserver-changelog) defaults to a primary read preference. Feel free to refactor any calls like this to simply use no arguments.

Related issue is [PHPC-1937](https://jira.mongodb.org/browse/PHPC-1937). I realize this is unrelated to the PR, so you can consider leaving this as-is and just making these changes in a separate PR if you prefer. Up to you if you'd like to track that in a separate PHPLIB ticket that references PHPC-1937.

Additionally, this may be something you can implement with refactor.

tests/SpecTests/PrimaryStepDownSpecTest.php
```diff
@@ -292,7 +292,7 @@ private function waitForPrimaryReelection(): void
 
             return;
         } catch (DriverException $e) {
-            $this->client->getManager()->selectServer(new ReadPreference('primary'));
+            $this->client->getManager()->selectServer(new ReadPreference(ReadPreference::PRIMARY));
```
jmikola, line 295:
```suggestion
            $this->client->getManager()->selectServer();
```

tests/SpecTests/PrimaryStepDownSpecTest.php
```diff
@@ -267,12 +267,12 @@ private function dropAndRecreateCollection(): void
 
     private function getTotalConnectionsCreated(?Server $server = null)
     {
-        $server = $server ?: $this->client->getManager()->selectServer(new ReadPreference('primary'));
+        $server = $server ?: $this->client->getManager()->selectServer(new ReadPreference(ReadPreference::PRIMARY));
```
jmikola, line 270:
```suggestion
        $server = $server ?: $this->client->getManager()->selectServer();
```

tests/SpecTests/PrimaryStepDownSpecTest.php
```diff
@@ -208,7 +208,7 @@ public function testGetMoreIteration(): void
         $totalConnectionsCreated = $this->getTotalConnectionsCreated();
 
         // Send a {replSetStepDown: 5, force: true} command to the current primary and verify that the command succeeded
-        $primary = $this->client->getManager()->selectServer(new ReadPreference(ReadPreference::RP_PRIMARY));
+        $primary = $this->client->getManager()->selectServer(new ReadPreference(ReadPreference::PRIMARY));
```
jmikola, line 211:
```suggestion
        $primary = $this->client->getManager()->selectServer();
```

tests/Operation/WatchFunctionalTest.php
```diff
@@ -165,7 +165,7 @@ public function testNextResumesAfterConnectionException(): void
          * a socket timeout that is less than the change stream's maxAwaitTimeMS
          * option. */
         $manager = static::createTestManager(null, ['socketTimeoutMS' => 50]);
-        $primaryServer = $manager->selectServer(new ReadPreference(ReadPreference::RP_PRIMARY));
+        $primaryServer = $manager->selectServer(new ReadPreference(ReadPreference::PRIMARY));
```
jmikola, line 168:
```suggestion
        $primaryServer = $manager->selectServer();
```

tests/Operation/FindAndModifyFunctionalTest.php
```diff
@@ -115,7 +115,7 @@ public function provideUpdatePipelines(): array
     public function testManagerReadConcernIsOmitted(): void
     {
         $manager = static::createTestManager(null, ['readConcernLevel' => 'majority']);
-        $server = $manager->selectServer(new ReadPreference(ReadPreference::RP_PRIMARY));
+        $server = $manager->selectServer(new ReadPreference(ReadPreference::PRIMARY));
```
jmikola, line 118:
```suggestion
        $server = $manager->selectServer();
```

tests/FunctionalTestCase.php
```diff
@@ -611,7 +611,7 @@ private static function getUriWithoutMultipleMongoses(): string
         }
 
         $manager = static::createTestManager($uri);
-        if ($manager->selectServer(new ReadPreference(ReadPreference::RP_PRIMARY))->getType() !== Server::TYPE_MONGOS) {
+        if ($manager->selectServer(new ReadPreference(ReadPreference::PRIMARY))->getType() !== Server::TYPE_MONGOS) {
```
jmikola, line 614:
```suggestion
        if ($manager->selectServer()->getType() !== Server::TYPE_MONGOS) {
```

tests/FunctionalTestCase.php
```diff
@@ -305,15 +305,15 @@ protected function getFeatureCompatibilityVersion(?ReadPreference $readPreferenc
 
     protected function getPrimaryServer()
     {
-        return $this->manager->selectServer(new ReadPreference(ReadPreference::RP_PRIMARY));
+        return $this->manager->selectServer(new ReadPreference(ReadPreference::PRIMARY));
```
jmikola, line 308:
```suggestion
        return $this->manager->selectServer();
```

composer.json
```diff
@@ -18,8 +18,9 @@
         "symfony/polyfill-php80": "^1.27"
     },
     "require-dev": {
-        "squizlabs/php_codesniffer": "^3.7",
         "doctrine/coding-standard": "^11.1",
+        "rector/rector": "^0.16.0",
```
jmikola, line 22:
```suggestion
        "rector/rector": "^0.16",
```

Is there any significance of the trailing `.0`? I don't recall offhand if `^` has different rules for pre-1.0 versions or not (where this might instead prohibit 0.17+).

rector.php
```diff
@@ -0,0 +1,29 @@
...
+    ]);
+
+    // ReadPreference::RP_* constants are deprecated in favor of the string constants
```
jmikola, line 16: Consider adding a reference to https://jira.mongodb.org/browse/PHPC-1489 and optionally https://jira.mongodb.org/browse/PHPC-1021.

jmikola (changes requested):

rector.php
GromNaN: Code updated. That's better for code analysis, and the C extension is not required: `::class` constants are resolved at compile time.

tests/SpecTests/TransactionsSpecTest.php
GromNaN: The fact that "primary" is the default value is not mentioned in the stub. It makes sense to me now.
I updated many other places in tests.

> Additionally, this may be something you can implement with refactor.

I'm not sure Rector can do that. Search/replace was good enough.

docs/reference/method/MongoDBClient-selectCollection.txt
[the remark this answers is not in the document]
GromNaN: The other way was done in https://github.com/mongodb/mongo-php-library/commit/dea346e48c8b8b44ae193371b3e225a3a979274e#diff-f631b3b3beb6be1d2034cda96ac452537b5827b20a04c64b45c86179ec7c29b8R75

This would be more concise with class imports (`use` statements), and that's how developers write their code in the end. But that's not how the doc is written.

composer.json
GromNaN: Before 1.0, that's more safe to stay with `^0.16.0` which is equivalent to `>= 0.16.0 < 0.17.0`
https://getcomposer.org/doc/articles/versions.md#caret-version-range-
> For pre-1.0 versions it also acts with safety in mind and treats ^0.3 as >=0.3.0 <0.4.0 and ^0.0.3 as >=0.0.3 <0.0.4.

docs/reference/method/MongoDBClient-selectCollection.txt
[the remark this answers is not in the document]
jmikola: Aye, we tend to avoid `use` statements in the docs since they're often snippets and not complete PHP files. The flip side of that is needing to refer to things by their FQCNs all the time.

If you're OK to I think that's best. I think the users that care will find the constants and use them, but readability is in the best interest of our docs audience.

tests/SpecTests/TransactionsSpecTest.php
jmikola: > The fact that "primary" is the default value is not mentioned in the stub.

I reckon that's because we'd need to provide a constructed ReadPreference object as the default value, and that's not supported AFAIK. Instead, we just have `null` as a default and some supporting documentation.

composer.json
jmikola: Aye, that's what I was thinking of. Thanks for confirming.

rector.php
```diff
@@ -0,0 +1,33 @@
...
+return static function (RectorConfig $rectorConfig): void {
+    $rectorConfig->paths([
+        __DIR__ . '/examples',
+        __DIR__ . '/src',
+        __DIR__ . '/tests',
+        __DIR__ . '/tools',
+    ]);
+
+    /**
+     * ReadPreference::RP_* constants are deprecated in favor of the string constants
```
jmikola, line 16: Thinking about this some more. I know the change was helpful for implementing the PR, but does this make sense as a permanent change in the Rector configuration?

The constants are technically only interchangeable in the ReadPreference constructor, and I don't see anything here that limits the replacement to that context. Hypothetically, if someone else was using the original constants with an integer type hint, applying this rule could yield invalid code, no?

jmikola (approved):
One question about whether we should remove the rector rule. I defer to your decision there.

Reverting the RST docs to use string literals SGTM.

Provisional LGTM with those two items addressed.

rector.php
GromNaN: I'll track the rector config in this PR description and remove rector from this PR.

docs/reference/method/MongoDBClient-selectCollection.txt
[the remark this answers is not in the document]
GromNaN: Reverted.

PR #1083, opened 2023-05-29
PHPLIB-1115: Fix waitForSnapshot() for MongoDB 7.0+

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1115

Using find instead of aggregate should ensure documents are available in the calling test case on 7.0+ servers.

### DIFF
--- a/tests/DocumentationExamplesTest.php
+++ b/tests/DocumentationExamplesTest.php
@@ -19,6 +19,7 @@ use function in_array;
 use function microtime;
 use function ob_end_clean;
 use function ob_start;
+use function usleep;
 use function var_dump;
 use function version_compare;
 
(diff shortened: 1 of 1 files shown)


jmikola:
The initial patch didn't address the test failure so I've moved this back to draft while I work on it further.

jmikola:
Note: test failures on mongodb-latest are due to [PHPLIB-1071](https://jira.mongodb.org/browse/PHPLIB-1071).

tests/DocumentationExamplesTest.php
```diff
@@ -1973,27 +1974,21 @@ private function waitForSnapshot(string $databaseName, string $collectionName):
         $collection = new Collection($this->manager, $databaseName, $collectionName);
         $session = $this->manager->startSession(['snapshot' => true]);
 
-        /* Retry until a snapshot query succeeds or ten seconds elapse,
-         * whichwever comes first.
-         *
-         * TODO: use hrtime() once the library requires PHP 7.3+ */
```
GromNaN, line 1979: This todo makes sense.
You can also remove the `microtime` calls and count 10000 iterations with 1ms sleep.

GromNaN (approved):

tests/DocumentationExamplesTest.php
jmikola: > count 10000 iterations with 1ms sleep.

Doing so wouldn't consider the time spent executing the FindOne operation.

> This todo makes sense

I don't recall my original intention behind adding a TODO for `hrtime()` (it wasn't discussed in https://github.com/mongodb/mongo-php-library/pull/900), but I gave this some thought when implementing this PR and couldn't see any reason to use it:

 * `microtime()` isn't deprecated and we don't need the precision offered by `hrtime()`
 * The variable return types for `hrtime()` seemed more complicated to work with
 * A monotonic timer is nice, but not necessary for a rarely used test method where we also don't need to worry about system clock changes

Side note: I did review the [7.3 polyfill](https://github.com/symfony/polyfill/blob/v1.27.0/src/Php73/Php73.php#L29) while working on this PR.

Having said all that, we do have many uses of `microtime()` in the test suite:

 * DocumentationExamplesTest
 * FindFunctionalTest
 * WatchFunctionalTest
 * UnifiedSpecTests/EventCollector and Loop (these should remain as-is)

The first three instances all pertain to calculating durations (actual timestamp isn't relevant) so I wouldn't object if you want to create a ticket to integrate `hrtime()` there.

PR #1084, opened 2023-05-30
PHPLIB-1071: Do not create ECC collection and check wire version for QEv2 

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1071

### DIFF
--- a/tests/SpecTests/FunctionalTestCase.php
+++ b/tests/SpecTests/FunctionalTestCase.php
@@ -16,6 +16,7 @@ use function in_array;
 use function json_encode;
 use function MongoDB\BSON\fromJSON;
 use function MongoDB\BSON\toPHP;
+use function property_exists;
 use function sprintf;
 use function version_compare;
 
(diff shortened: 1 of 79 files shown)


GromNaN (approved):

tests/SpecTests/FunctionalTestCase.php
```diff
@@ -95,6 +96,28 @@ public static function assertDocumentsMatch($expectedDocument, $actualDocument,
...
+    protected static function assertCommandOmittedFields(stdClass $expected, stdClass $actual): void
+    {
+        foreach ($expected as $key => $value) {
+            if ($value === null) {
+                static::assertObjectNotHasAttribute($key, $actual);
+                unset($expected->{$key});
+                continue;
+            }
+
+            if ($value instanceof stdClass && property_exists($actual, $key) && $actual->{$key} instanceof stdClass) {
```
jmikola, line 115: Note: the previous code for this function was:

```php
if ($value instanceof stdClass) {
    static::assertObjectHasAttribute($key, $actual);
    static::assertInstanceOf(stdClass::class, $actual->{$key});
    static::assertCommandOmittedFields($value, $actual->{$key});
}
```

This caused multiple [test failures](https://spruce.mongodb.com/version/64760f3e61837d63d6c75fe6/tasks) in the `getMore.json` test, which asserts that the `getMore` field of the outgoing command matches `{ "$$type": "long" }`. The actual value in that case is an integer, which was problematic for `assertInstanceOf()` above.

Rather than rewrite `assertCommandOmittedFields()` to skip special operations (i.e. keys starting with "$$"), which is only defined in the [CSFLE legacy test format](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/tests/README.rst#spec-test-format), I opted to just NOP and avoid recursion. This keeps `assertCommandOmittedFields()` focused on it's original purpose of only asserting omitted fields and allows the caller to handle everything else (likely via `assertDocumentsMatch()`).

PR #1087, opened 2023-05-30
PHPLIB-1117: Remove @api annotation

GromNaN:
Fix PHPLIB-1117

All classes are public API by default, unless marked with `@internal`.

These classes had none of these 2 annotations `@api` or `@internal` and are part of the public api to be consistent with other `*Result` classes.
- `BulkWriteResult`
- `Client`
- `Collection`
- `Database`
- `DeleteResult`
- `InsertManyResult`
- `InsertOneResult`
- `UpdateResult`

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -1,0 +1,23 @@
+<?php
+
+use Rector\Config\RectorConfig;
+use Rector\DeadCode\Rector\ClassLike\RemoveAnnotationRector;
+
+return static function (RectorConfig $rectorConfig): void {
+    $rectorConfig->paths([
+        __DIR__ . '/examples',
+        __DIR__ . '/src',
+        __DIR__ . '/tests',
+        __DIR__ . '/tools',
+    ]);
+
+    /**
+     * All classes are public API by default, unless marked with @internal.
+     */
+    $rectorConfig->ruleWithConfiguration(RemoveAnnotationRector::class, ['api']);
+
+    // define sets of rules
+    //    $rectorConfig->sets([
+    //        LevelSetList::UP_TO_PHP_72
+    //    ]);
+};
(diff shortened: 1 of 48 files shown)


src/ChangeStream.php
```diff
@@ -33,11 +33,10 @@
 /**
  * Iterator for a change stream.
  *
- * @psalm-type ResumeCallable = callable(array|object|null, bool): ChangeStreamIterator
- *
- * @api
  * @see \MongoDB\Collection::watch()
  * @see https://mongodb.com/docs/manual/reference/method/db.watch/#mongodb-method-db.watch
+ *
+ * @psalm-type ResumeCallable = callable(array|object|null, bool): ChangeStreamIterator
```
GromNaN, line 39: CS fix: Incorrect order of annotations groups.

rector.php
```diff
@@ -0,0 +1,23 @@
...
+        __DIR__ . '/examples',
+        __DIR__ . '/src',
+        __DIR__ . '/tests',
+        __DIR__ . '/tools',
+    ]);
+
+    /**
+     * All classes are public API by default, unless marked with @internal.
+     */
+    $rectorConfig->ruleWithConfiguration(RemoveAnnotationRector::class, ['api']);
```
jmikola, line 17: I don't think PHPLIB does this, but is there any harm of removing `@api` from method annotations as well?

jmikola (approved):
One question about whether we should target function doc blocks as well but overall changes LGTM.

rector.php
GromNaN: The purpose of leaving these rules in place in the configuration file is to impose them on new developments or pending branches. If, for example, a developer forgets or ignores that `@api` annotations are no longer used, this will be corrected automatically before the code review.

PR #1088, opened 2023-05-30
PHPLIB-1141: Setup rector

GromNaN:
Fix PHPLIB-1141

The library uses [rector](https://getrector.com/) to refactor the code for new features.
To run automatic refactoring, use the `rector` command:

```
$ vendor/bin/rector
```

Job merged with psalm: https://github.com/mongodb/mongo-php-library/actions/runs/5134018025/jobs/9237409213?pr=1088

The rules need to be adjusted due to false positive like this:
- The 3rd argument is correctly defined in ext-mongodb stub.
```
src/Operation/Delete.php:164

    ---------- begin diff ----------
@@ @@
         $bulk = new Bulk($this->createBulkWriteOptions());
         $bulk->delete($this->filter, $this->createDeleteOptions());

-        $writeResult = $server->executeBulkWrite($this->databaseName . '.' . $this->collectionName, $bulk, $this->createExecuteOptions());
+        $writeResult = $server->executeBulkWrite($this->databaseName . '.' . $this->collectionName, $bulk);

         return new DeleteResult($writeResult);
     }
    ----------- end diff -----------
```

- The variable is always initialized when accessed.
```
tests/SpecTests/ClientSideEncryptionSpecTest.php:308

    ---------- begin diff ----------
@@ @@
      */
     public function testClientSideEncryption(stdClass $test, ?array $runOn, array $data, ?stdClass $encryptedFields = null, ?array $keyVaultData = null, ?stdClass $jsonSchema = null, ?string $databaseName = null, ?string $collectionName = null): void
     {
+        $context = null;
         if (isset(self::$incompleteTests[$this->dataDescription()])) {
             $this->markTestIncomplete(self::$incompleteTests[$this->dataDescription()]);
         }
@@ @@

     private function encryptCorpusValue(string $fieldName, stdClass $data, ClientEncryption $clientEncryption)
     {
+        $encrypted = null;
         $encryptionOptions = [
             'algorithm' => $data->algo === 'rand' ? ClientEncryption::AEAD_AES_256_CBC_HMAC_SHA_512_RANDOM : ClientEncryption::AEAD_AES_256_CBC_HMAC_SHA_512_DETERMINISTIC,
         ];
    ----------- end diff -----------
```


examples/aggregate.php
```diff
@@ -28,7 +28,7 @@ function toJSON(object $document): string
 $documents = [];
 
 for ($i = 0; $i < 100; $i++) {
-    $documents[] = ['randomValue' => rand(0, 1000)];
+    $documents[] = ['randomValue' => random_int(0, 1000)];
```
GromNaN, line 31: https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md#randomfunctionrector

tests/GridFS/UnusableStream.php
```diff
@@ -20,7 +20,7 @@ public static function register($protocol = 'unusable'): void
             stream_wrapper_unregister($protocol);
         }
 
-        stream_wrapper_register($protocol, static::class, STREAM_IS_URL);
+        stream_wrapper_register($protocol, self::class, STREAM_IS_URL);
```
GromNaN, line 23: `static` is useless for a `final` class. `self` can be resolved at compile time.

tests/UnifiedSpecTests/Constraint/Matches.php
```diff
@@ -368,6 +368,7 @@ private function doToString()
         return 'matches ' . $this->exporter()->export($this->value);
     }
 
+    /** @psalm-return never-return */
```
GromNaN, line 371: Helps to detect code paths and avoid false-positive undefined variable access.

examples/aggregate.php
jmikola: This seems a bit overzealous, as we don't actually need a cryptographically secure number here. But it's also just an example file so I don't feel strongly about it.

rector.php
```diff
@@ -0,0 +1,28 @@
...
+    ]);
+
+    $rectorConfig->sets([LevelSetList::UP_TO_PHP_72]);
+
+    $rectorConfig->skip([
+        // Falsely detect unassigned variables in code paths stopped by PHPUnit\Framework\Assert::markTestSkipped()
+        AddDefaultValueForUndefinedVariableRector::class => [
+            __DIR__ . '/tests/',
+        ],
+        // @see https://github.com/phpstan/phpstan-src/pull/2429
```
jmikola, line 23: Would `See: ` make more sense here since we're not in an actual doc block?

tests/GridFS/BucketFunctionalTest.php
```diff
@@ -549,7 +549,7 @@ public function testOpenDownloadStreamByNameShouldRequireFilenameAndRevisionToEx
         $this->bucket->uploadFromStream('filename', $this->createStream('bar'));
 
         $this->expectException(FileNotFoundException::class);
-        $this->bucket->openDownloadStream($filename, ['revision' => $revision]);
+        $this->bucket->openDownloadStream($filename);
```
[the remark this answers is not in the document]
jmikola: Looks like you caught an outstanding bug. This should be `openDownloadStreamByName()`.

No objections if you want to fix that here or in a separate PR. I'll defer to you if you'd like to create a JIRA ticket, but IMO it's not necessary in this case.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -1107,8 +1107,7 @@ public function testResumeTokenNotFoundDoesNotAdvanceKey(): void
         try {
             $changeStream->next();
             $this->fail('Exception for missing resume token was not thrown');
-        } catch (ResumeTokenException $e) {
-        } catch (ServerException $e) {
+        } catch (ResumeTokenException | ServerException $e) {
```
jmikola, line 1110: Ah, I missed that this was available since 7.1 (long before general union types in 8.0): https://wiki.php.net/rfc/multiple-catch

jmikola:
> The variable is always initialized when accessed (in ClientSideEncryptionSpecTest)

Are those two examples due to the tooling not realizing that `fail()` and `markTestIncomplete()` throw?

.github/workflows/static-analysis.yml
```diff
@@ -65,3 +65,6 @@ jobs:
 
       - name: "Run Psalm"
         run: "vendor/bin/psalm --show-info=false --stats --output-format=github --threads=$(nproc)"
+
+      - name: "Run Rector"
+        run: "vendor/bin/rector --ansi --dry-run"
```
jmikola, line 70: Noted that `--dry-run` returns a non-zero exit code if changes are detected, which makes it suitable for CI.

tests/GridFS/BucketFunctionalTest.php
jmikola, line 552:
```suggestion
        $this->bucket->openDownloadStreamByName($filename, ['revision' => $revision]);
```

examples/aggregate.php
GromNaN: That's the new "best practice". Using it proves that the doc have been updated 😉

tests/GridFS/BucketFunctionalTest.php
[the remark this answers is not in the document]
GromNaN: Fixed.

GromNaN:
> Are those two examples due to the tooling not realizing that `fail()` and `markTestIncomplete()` throw?

PHPUnit [9.6 had the annotation](https://github.com/sebastianbergmann/phpunit/blob/9.6.8/src/Framework/Assert.php#L2840) `@psalm-return never-return` that was [then replaced](https://github.com/sebastianbergmann/phpunit/commit/1fdf56250baa241d1d5bf826e3ca87d21bf89299) by `:never` native return type.

People seems to have the opposite behavior: https://github.com/phpstan/phpstan-phpunit/issues/52

PR #1090, opened 2023-05-31
PHPLIB-1063: Import writeConcernError in unified tests

GromNaN:
Fix PHPLIB-1063

Import test file: https://github.com/mongodb/specifications/blob/master/source/command-logging-and-monitoring/tests/monitoring/writeConcernError.json

I checked that the test effectively fails when I modify any value in the test file.

### DIFF
--- a/tests/UnifiedSpecTests/command-monitoring/writeConcernError.json
+++ b/tests/UnifiedSpecTests/command-monitoring/writeConcernError.json
@@ -1,0 +1,155 @@
+{
+  "description": "writeConcernError",
+  "schemaVersion": "1.4",
+  "runOnRequirements": [
+    {
+      "minServerVersion": "4.1.0",
+      "topologies": [
+        "replicaset"
+      ],
+      "serverless": "forbid"
+    }
+  ],
+  "createEntities": [
+    {
+      "client": {
+        "id": "client",
+        "observeEvents": [
+          "commandStartedEvent",
+          "commandSucceededEvent",
+          "commandFailedEvent"
+        ]
+      }
+    },
+    {
+      "database": {
+        "id": "database",
+        "client": "client",
+        "databaseName": "command-monitoring-tests"
+      }
+    },
+    {
+      "collection": {
+        "id": "collection",
+        "database": "database",
+        "collectionName": "test"
+      }
+    }
+  ],
+  "initialData": [
+    {
+      "collectionName": "test",
+      "databaseName": "command-monitoring-tests",
+      "documents": [
+        {
+          "_id": 1,
+          "x": 11
+        }
+      ]
+    }
+  ],
+  "tests": [
+    {
+      "description": "A retryable write with write concern errors publishes success event",
+      "operations": [
+        {
+          "name": "failPoint",
+          "object": "testRunner",
+          "arguments": {
+            "client": "client",
+            "failPoint": {
+              "configureFailPoint": "failCommand",
+              "mode": {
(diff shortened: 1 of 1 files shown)


tests/UnifiedSpecTests/command-monitoring/writeConcernError.json
```diff
@@ -0,0 +1,155 @@
+{
+  "description": "writeConcernError",
+  "schemaVersion": "1.12",
```
GromNaN, line 3: The schema version is downgraded from `1.13` to `1.12`: this file uses none of the new fields of `1.13`: `observeLogMessages`, `logComponent`, `logSeverityLevel`, `expectedLogMessagesForClient`, `expectLogMessages`.

jmikola: This should be reverted to 1.13 for now. We don't modify the JSON test files when syncing them to the specs repo. Some other drivers, like libmongoc, do that, but IMO it becomes problematic as we're then unable to differentiate driver-specific changes from upstream changes.

Correcting this would be better done upstream. That would entail the overhead of a DRIVERS ticket, but you could avoid spawning language tickets by selecting "Driver Changes: Not Needed" like I did in [DRIVERS-2641](https://jira.mongodb.org/browse/DRIVERS-2641).

@alcaeus and I were actually discussing this last week based on a comment I left in https://github.com/mongodb/specifications/pull/1423#discussion_r1200688693 (another spec test using a higher schema version than required). He'd like to explore adding a CI step to the specs repo to catch this and I assume the implementation would also entail revising existing spec test files. He'll likely create a DRIVERS ticket to track that when he's back from PTO, so I think we can just wait on that (and avoid fixing this test file upstream for now).

tests/UnifiedSpecTests/command-monitoring/writeConcernError.json
```diff
@@ -0,0 +1,155 @@
+{
```
jmikola, line 1: When syncing spec tests, please include a reference to the specs commit. For this PR, it should be:

```
Synced with mongodb/specifications@18932c00f9e2f8be66b48c930cbbc428499db668
```

This makes it easier to jump to the specs repo when viewing PHPLIB's git history.

jmikola (approved):
LGTM with the commit message modified and change to `schemaVersion` reverted.

tests/UnifiedSpecTests/command-monitoring/writeConcernError.json
GromNaN: I cannot revert the schemaVersion without updating `MAX_SCHEMA_VERSION`. Which is even worse if the new fields are not implemented in the test runner.
https://github.com/mongodb/mongo-php-library/blob/f6551edd60a368f019bb7cd1dda4fe78a5e7597b/tests/UnifiedSpecTests/UnifiedTestRunner.php#L64-L66

jmikola: Noted. I think it's best to get this addressed upstream first, then. You can follow the process I used in [DRIVERS-2641](https://jira.mongodb.org/browse/DRIVERS-2641) to avoid extra language tickets.

GromNaN: New ticket DRIVERS-2644 and PR https://github.com/mongodb/specifications/pull/1431

PR #1092, opened 2023-05-31
PHPLIB-1128: Move generate_index_name() to private method within IndexInput

GromNaN:
Fix PHPLIB-1128

`generate_index_name()` was originally made a utility function in [bc65629](https://github.com/mongodb/mongo-php-library/commit/bc65629c772a53c57b4f5debad6aae5b76c9a01c) because it was shared by both the CreateIndexes (via IndexInput) and Count operations. It was later removed from Count in [PHPLIB-285](https://jira.mongodb.org/browse/PHPLIB-285).

### DIFF
--- a/tests/Model/IndexInputTest.php
+++ b/tests/Model/IndexInputTest.php
@@ -2,8 +2,10 @@
 
 namespace MongoDB\Tests\Model;
 
+use MongoDB\BSON\Document;
 use MongoDB\BSON\Serializable;
 use MongoDB\Exception\InvalidArgumentException;
+use MongoDB\Model\BSONDocument;
 use MongoDB\Model\IndexInput;
 use MongoDB\Tests\TestCase;
 use stdClass;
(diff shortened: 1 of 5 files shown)


tests/Model/IndexInputTest.php
```diff
@@ -40,16 +42,22 @@ public function testConstructorShouldRequireNameToBeString(): void
...
     }
 
-    public function provideExpectedNameAndKey()
+    public function provideExpectedNameAndKey(): array
     {
         return [
             ['x_1', ['x' => 1]],
+            ['x_1', (object) ['x' => 1]],
+            ['x_1', new BSONDocument(['x' => 1])],
+            ['x_1', Document::fromPHP(['x' => 1])],
```
jmikola, line 60: I noted that the original tests for the function were more comprehensive (the four possible type casts are applied to each index specification), but I don't think there's any significant benefit to doing that. Testing a basic index spec with the non-array types seems sufficient.

tests/Model/IndexInputTest.php
```diff
@@ -40,16 +42,22 @@ public function testConstructorShouldRequireNameToBeString(): void
         new IndexInput(['key' => ['x' => 1], 'name' => 1]);
     }
 
-    /** @dataProvider provideExpectedNameAndKey */
-    public function testNameGeneration($expectedName, array $key): void
+    /**
+     * @dataProvider provideExpectedNameAndKey
+     * @param array|object $key
+     */
+    public function testNameGeneration($expectedName, $key): void
```
jmikola, line 49:
```suggestion
    public function testNameGeneration(string $expectedName, $key): void
```

Might as well add this while we're changing this line.

jmikola (commented):
Please check the two Psalm errors in `functions.php`. I'm not sure why those were caught in the PR but not locally when you regenerated the baseline.

GromNaN:
I'm fighting with psalm. baseline is not the same for my local PHP 8.2 and the CI PHP 7.4
```
Error: src/functions.php:86:12: MixedInferredReturnType: Could not verify return type 'array<array-key, mixed>|object' for MongoDB\apply_type_map_to_document (see https://psalm.dev/047)
Error: src/functions.php:95:12: MixedReturnStatement: Could not infer a return type (see https://psalm.dev/138)
Error: Process completed with exit code 2.
```

PR #1097, opened 2023-06-08
PHPLIB-1143: Add search index operations to Collection class

GromNaN:
Fix PHPLIB-1143

Spec files sync from https://github.com/mongodb/specifications/commit/267a54d1bb9af894d05d3dd1a6bda0fa53e20a3f
Previously introduced by https://github.com/mongodb/specifications/commit/ae35908e5eadc6cda2d8f5c4e33dada25ba8fa98

- [x] Create methods `createSearchIndex`, `createSearchIndexes`, `dropSearchIndex`, `updateSearchIndex`, `listSearchIndexes`
- [x] Create operation classes
- [x] Add unit tests on new Operation classes
- [x] Add functional tests for error using unified test json.
- [x] Add functional test that requires a `MONGODB_URI` variable on Atlas. Currently only M10+ are supported.


tests/UnifiedSpecTests/index-management/createSearchIndexes.json
```diff
@@ -0,0 +1,169 @@
...
+        "replicaset",
+        "load-balanced",
+        "sharded"
+      ],
+      "serverless": "forbid"
+    }
+  ],
+  "tests": [
+    {
+      "description": "empty index definition array",
```
GromNaN, line 42: Some tests fail because the the exception is thrown by PHP before sending the operation to the server.

src/Collection.php
```diff
@@ -943,6 +1014,22 @@ public function listIndexes(array $options = [])
...
 
+    /**
+     * Returns information for all indexes for the collection.
+     *
+     * @see ListSearchIndexes::__construct() for supported aggregation and list options
+     * @return IndexInfoIterator
+     * @throws InvalidArgumentException for parameter/option parsing errors
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function listSearchIndexes(string $name, array $aggregationOptions = [], array $listIndexOptions = [])
```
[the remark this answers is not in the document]
GromNaN: I merged option arrays as proposed by the [drivers spec](https://github.com/mongodb/specifications/blob/master/source/index-management/index-management.rst#search-index-management-helper-options) and discussed with @alcaeus 

> Drivers MAY combine the aggregation options with any future listSearchIndexes stage options, if that is idiomatic for a driver's language.

tests/UnifiedSpecTests/index-management/createSearchIndexes.json
jmikola: It may be worth asking the DBX spec test authors whether they considered using `isClientError` here. But assuming that's unspecified and drivers are free to validate this before sending a command or rely on the server to report an error, I'm still not sure why this would fail.

IIRC, [`isError: true`](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#expectederror) in the test format shouldn't distinguish about the type of error. Server-side and client-side errors should both satisfy it. If that's not the case, perhaps there's a bug with out test runner implementation.

src/Operation/DropIndexes.php
```diff
@@ -76,8 +76,6 @@ class DropIndexes implements Executable
      */
     public function __construct(string $databaseName, string $collectionName, string $indexName, array $options = [])
     {
-        $indexName = $indexName;
```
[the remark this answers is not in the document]
jmikola: For additional context, this dates back to https://github.com/mongodb/mongo-php-library/commit/1b29a6d71e5511ea46b596461442ca94069bdf42#diff-4171c1485fb13dfe08523475f1293bb5755904638332d6cf1086050b8c2ab40aR80.

GromNaN (commented):
I think I've figured out how spec works: all the Atlas Search index command get a server error because they are not available on local servers. The spec checks the command that is sent to the server with `expectedEvents`. Therefore, I don't validate some inputs client-side to let the server return an error. This is more future-proof as some options that are invalid in 7.0 could be allowed in a future release.

tests/Collection/SearchIndexFunctionalTest.php
```diff
@@ -0,0 +1,70 @@
...
+    public function testIndexLifecycle(): void
+    {
+        $atlasUri = getenv('MONGODB_ATLAS_URI');
+
+        if (! $atlasUri) {
+            $this->markTestSkipped('MONGODB_ATLAS_URI environment variable is not set');
+        }
+
+        $this->manager = static::createTestManager($atlasUri);
+
```
GromNaN, line 27: I'd like to check the server requirement (Atlas + version >= 7.0) or skip the test.

tests/UnifiedSpecTests/index-management/createSearchIndexes.json
GromNaN: > I'm still not sure why this would fail.

Because it's not Atlas, the command is not supported by non-Atlas server. The actual assertion is in `expectEvents`.

tests/Operation/ListSearchIndexesTest.php
```diff
@@ -0,0 +1,35 @@
...
+
+namespace MongoDB\Tests\Operation;
+
+use MongoDB\Exception\InvalidArgumentException;
+use MongoDB\Operation\ListSearchIndexes;
+
+class ListSearchIndexesTest extends TestCase
+{
+    /** @dataProvider provideInvalidStringValues */
+    public function testConstructorExpectNameFilterToBeAString($name): void
```
jmikola, line 11: If you end up changing how `name` is passed, I suppose you could fold this into `testConstructorOptionTypeChecks()`.

tests/Operation/UpdateSearchIndexTest.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace MongoDB\Tests\Operation;
+
+use MongoDB\Exception\InvalidArgumentException;
+use MongoDB\Operation\UpdateSearchIndex;
+
+class UpdateSearchIndexTest extends TestCase
+{
+    public function testConstructorRequiresANotEmptyIndexName(): void
```
jmikola, line 10: If you end up taking the suggestion for DropSearchIndexTest, this could be changed to `testConstructorIndexNameMustNotBeEmpty()`.

tests/UnifiedSpecTests/Util.php
```diff
@@ -88,6 +88,8 @@ final class Util
             'createChangeStream' => ['pipeline', 'session', 'fullDocument', 'fullDocumentBeforeChange', 'resumeAfter', 'startAfter', 'startAtOperationTime', 'batchSize', 'collation', 'maxAwaitTimeMS', 'comment', 'showExpandedEvents'],
             'createFindCursor' => ['filter', 'session', 'allowDiskUse', 'allowPartialResults', 'batchSize', 'collation', 'comment', 'cursorType', 'hint', 'limit', 'max', 'maxAwaitTimeMS', 'maxScan', 'maxTimeMS', 'min', 'modifiers', 'noCursorTimeout', 'oplogReplay', 'projection', 'returnKey', 'showRecordId', 'skip', 'snapshot', 'sort'],
             'createIndex' => ['keys', 'comment', 'commitQuorum', 'maxTimeMS', 'name', 'session', 'unique'],
+            'createSearchIndex' => ['model'],
+            'createSearchIndexes' => ['models'],
```
jmikola, line 92: Since this is probably your first interaction with this array, I'm curious if you have any thoughts on it. It dates back to https://github.com/mongodb/mongo-php-library/commit/2a81aafc96beada8cc58a18381817b34c1497395 for [PHPLIB-662](https://jira.mongodb.org/browse/PHPLIB-662) but I'm not entirely convinced it's more useful than annoying.

src/Operation/CreateSearchIndexes.php
```diff
@@ -0,0 +1,114 @@
...
+    {
+        if (! array_is_list($indexes)) {
+            throw new InvalidArgumentException('$indexes is not a list');
+        }
+
+        foreach ($indexes as $i => $index) {
+            if (is_object($index)) {
+                $index = (array) $index;
+            } elseif (! is_array($index)) {
+                throw InvalidArgumentException::invalidType(sprintf('$index[%d]', $i), $index, 'array');
```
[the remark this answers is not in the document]
alcaeus: Note that you can also specify an array as the type argument:
```suggestion
                throw InvalidArgumentException::invalidType(sprintf('$index[%d]', $i), $index, ['object', 'array']);
```

tests/Collection/SearchIndexFunctionalTest.php
```diff
@@ -0,0 +1,70 @@
...
+        do {
+            sleep(1);
+            $result = $collection->listSearchIndexes($name, ['typeMap' => ['root' => 'array', 'document' => 'array']]);
+            $this->assertInstanceOf(Cursor::class, $result);
+            $result->rewind();
+
+            if ($count++ > 90) {
+                $this->fail('Search index was not deleted');
+            }
+        } while ($result->valid());
```
GromNaN, line 96: I can't use `$result->count()` here because there is an exception when `$result instanceof MongoDB\Model\CachingIterator` and the list of indexes is empty. `MongoDB\Driver\Exception\RuntimeException : Cannot advance a completed or failed cursor.`

tests/UnifiedSpecTests/Util.php
GromNaN: IMO the schema validation of does not cover which fields are allowed for each operation. But this should be validated in the spec repository, not here.

GromNaN (commented):
Thank you @jmikola for the huge review. I made a lot of changes according to your comments.

tests/UnifiedSpecTests/Util.php
jmikola: [Incorporate referenced entity operations into the schema version](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#incorporate-referenced-entity-operations-into-the-schema-version) in the unified test format spec explains that in more detail. The problem with validating operations in the schema is that we'd have to bump schema versions much more frequently.

tests/UnifiedSpecTests/index-management/createSearchIndexes.json
jmikola: Ah, is this something that will ultimately be covered by adding the end-to-end testing? I suppose we'll need some new syntax to skip these tests on non-Atlas servers, then.

tests/UnifiedSpecTests/Operation.php
```diff
@@ -557,6 +558,29 @@ function (IndexInfo $info) {
...
+            case 'updateSearchIndex':
+                assertArrayHasKey('name', $args);
+                assertArrayHasKey('definition', $args);
+                assertIsString($args['name']);
+                assertInstanceOf(stdClass::class, $args['definition']);
+
+                return $collection->updateSearchIndex($args['name'], $args['definition']);
+
+            case 'listSearchIndexes':
+                return $collection->listSearchIndexes($args['name'] ?? null, document_to_array($args['aggregationOptions'] ?? []));
```
jmikola, line 582: I think you'd be fine to use an `(array)` cast here, as is done for the `createDataKey` and `rewrapManyDataKey` cases in `executeForClientEncryption()`. Alternatively, you can consider  changing those to use `document_to_array()`.

tests/Collection/SearchIndexFunctionalTest.php
jmikola: That seems like a bug in CachingIterator. Would you mind opening a ticket and looking into it? Here's a short script to reproduce it locally (change the URI accordingly):

```php
<?php

require __DIR__ . '/vendor/autoload.php';

$client = new MongoDB\Client('mongodb://localhost:27070/?replicaSet=rs0');
$collection = $client->test->foo;
$collection->drop();

$it = new MongoDB\Model\CachingIterator($collection->find());

var_dump($it);
var_dump($it->count());
```

GromNaN: Fix https://github.com/mongodb/mongo-php-library/pull/1118

tests/UnifiedSpecTests/index-management/createSearchIndexes.json
GromNaN: If we can run the tests on Atlas from Github Actions, this tests on failing commands could be removed from the spec repository.

src/Collection.php
```diff
@@ -392,6 +396,55 @@ public function createIndexes(array $indexes, array $options = [])
...
+     * @see https://www.mongodb.com/docs/manual/reference/command/createSearchIndexes/
+     * @see https://mongodb.com/docs/manual/reference/method/db.collection.createSearchIndex/
+     * @param array[] $indexes List of search index specifications
+     * @param array   $options Command options
+     * @return string[] The names of the created search indexes
+     * @throws UnsupportedException if options are not supported by the selected server
+     * @throws InvalidArgumentException for parameter/option parsing errors
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function createSearchIndexes(array $indexes, array $options = []): array
```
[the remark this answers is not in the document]
GromNaN: I tried to pass `timeoutMS: 1`, either the operation is super-fast, or the option is ignored.

src/Collection.php
```diff
@@ -390,6 +394,56 @@ public function createIndexes(array $indexes, array $options = [])
...
+     * @return string The name of the created search index
+     * @throws UnsupportedException if options are not supported by the selected server
+     * @throws InvalidArgumentException for parameter/option parsing errors
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function createSearchIndex(array $index, array $options = []): string
+    {
+        $names = $this->createSearchIndexes([$index], $options);
+
+        return current($names);
```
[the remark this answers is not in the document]
GromNaN: I removed the `$result->indexesCreated ?? []`, I checked the server behavior and `indexesCreated` is always an array which size is the number of created indexes.

src/Collection.php
```diff
@@ -390,6 +394,56 @@ public function createIndexes(array $indexes, array $options = [])
...
+    }
+
+    /**
+     * Create one or more Atlas Search indexes for the collection.
+     * Only available when used against a 7.0+ Atlas cluster.
+     *
+     * Each element in the $indexes array must have a "name" and a "definition" document.
```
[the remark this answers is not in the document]
GromNaN: Comment modified.

src/Collection.php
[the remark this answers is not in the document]
jmikola: The PHP driver still doesn't support client-side operation timeouts (see: [PHPC-1760](https://jira.mongodb.org/browse/PHPC-1760)), so the option is ignored.

Anyway, you're adding `$options` here, so that will definitely cover us when support is added down the line. It does sound like the spec's omission of options was an oversight, though.

src/Operation/CreateSearchIndexes.php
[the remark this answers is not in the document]
jmikola: The array handling dates back to https://github.com/mongodb/mongo-php-library/commit/b2cd2866c64690603896b788cd56e45ff4cc69a4 and was bundled in with some CSFLE changes for [PHPLIB-511](https://jira.mongodb.org/browse/PHPLIB-511).

I'll defer to you guys if we'd like a ticket to formally remove this. The few times I used it myself was because @alcaeus reminded me it was available in a code review -- but writing list of types as a string myself was never a pain point.

src/Operation/DropSearchIndex.php
```diff
@@ -0,0 +1,91 @@
...
+     * @see Executable::execute()
+     * @return array|object Command result document
+     * @throws UnsupportedException if write concern is used and unsupported
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function execute(Server $server)
+    {
+        $cursor = $server->executeCommand($this->databaseName, $this->createCommand());
+
+        return current($cursor->toArray());
```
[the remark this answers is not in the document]
jmikola: Offhand, I don't recall if [`dropIndexes`https://www.mongodb.com/docs/manual/reference/command/dropIndexes/] returns any special fields, but the cluster/operation time could be relevant for folks using change streams or [causal consistency](https://github.com/mongodb/specifications/blob/master/source/causal-consistency/causal-consistency.rst). Emphasis on _could_ here, as I doubt this is used very much at all in practice.

This would mainly be for API consistency, as I don't believe we have any operation helpers in the core classes (e.g. Collection) with a `void` return type. A separate concern might be that we wouldn't be able to change the signature later if needed (IIRC this conflicts with overrides in child classes).

Note: an annoyance here is that DropIndexes supports a `typeMap` option purely to provide control the returned command result. I suppose we'd also support that here for consistency.

src/Collection.php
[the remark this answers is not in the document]
jmikola: Just had a separate thought. I wonder if `comment` is supported on these commands, since they are directed a mongod/mongos. It's not mentioned in the [syntax document](https://docs.google.com/document/d/1rCU7vLmMEn-87SuSZ3xnnSpnkwoWpekyOl15NPIqTJk/edit), but could be another case of the spec forgetting about a general option when defining the API.

src/Operation/CreateSearchIndexes.php
```diff
@@ -0,0 +1,104 @@
...
+     * @see Executable::execute()
+     * @return string[] The names of the created indexes
+     * @throws UnsupportedException if write concern is used and unsupported
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function execute(Server $server): array
+    {
+        $cmd = [
+            'createSearchIndexes' => $this->collectionName,
+            'indexes' => $this->indexes,
```
jmikola, line 92: Note: CreateIndexes supports a `comment` option, which it would inject here.

src/Operation/DropSearchIndex.php
```diff
@@ -0,0 +1,78 @@
...
+     *
+     * @see Executable::execute()
+     * @throws UnsupportedException if write concern is used and unsupported
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function execute(Server $server): void
+    {
+        $cmd = [
+            'dropSearchIndex' => $this->collectionName,
+            'name' => $this->name,
```
jmikola, line 73: Note: DropIndexes supports a `comment` option, which it would inject here.

src/Operation/ListSearchIndexes.php
```diff
@@ -0,0 +1,104 @@
...
+    {
+        $cursor = $this->aggregate->execute($server);
+
+        return new CachingIterator($cursor);
+    }
+
+    private function createAggregate(): Aggregate
+    {
+        $pipeline = [
+            ['$listSearchIndexes' => (object) $this->listSearchIndexesOptions],
```
jmikola, line 99: Not an action item but just FYI: I saw that the [design document](https://docs.google.com/document/d/1rCU7vLmMEn-87SuSZ3xnnSpnkwoWpekyOl15NPIqTJk/edit) mentioned a `listSearchIndexes` command. I asked why the spec wasn't using that in [this #dbx-devs thread](https://mongodb.slack.com/archives/C72LB5RPV/p1687803718544879]). The server command predated the aggregation stage, and will likely be removed before 7.0 is released.

This will leave us with just the pipeline stage as the API for listing search indexes.

tests/Collection/SearchIndexFunctionalTest.php
```diff
@@ -0,0 +1,107 @@
...
+class SearchIndexFunctionalTest extends FunctionalTestCase
+{
+    private const ATLAS_TLD = '/\.(mongodb\.net|mongodb-dev\.net)/';
+
+    /**
+     * Connect to a MongoDB Atlas Server.
+     */
+    public static function getUri($allowMultipleMongoses = false): string
+    {
+        $atlasUri = getenv('MONGODB_ATLAS_URI');
```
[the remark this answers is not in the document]
GromNaN: That's a good point. I thought that only these tests would run on atlas, but thinking about the way the other tests are built, I understand better that it's necessary to be able to run all the tests on all the environments while skipping when an incompatibility is detected.

So I reverted to `MONGODB_URI` but with the Atlas host check.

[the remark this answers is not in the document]
jmikola: Just to throw another example at you, in the PR for AWS Lambda testing, the setup scripts assign the Atlas connection string to the `MONGODB_URI` variable (see: [mongodb-labs/drivers-evergreen-tools#325|https://github.com/mongodb-labs/drivers-evergreen-tools/pull/325/files#diff-bbc90d2c444efad446a3b080f0b6116258a7dc45366140be98ca8b891934c646R121-R122]).

One exception to this is testing for sharded clusters and load balancers, where we sometimes need separate connection strings to the same server deployment. For example, the load balancer spec defines two variables for LBs backing one or multiple mongoses ([Load Balancer Configuration](https://github.com/mongodb/specifications/blob/master/source/load-balancers/tests/README.rst#load-balancer-configuration)).

GromNaN:
Updated to use `is_document` from #1117

src/Collection.php
```diff
@@ -941,6 +1023,27 @@ public function listIndexes(array $options = [])
...
+    /**
+     * Returns information for all Atlas Search indexes for the collection.
+     * Only available when used against a 7.0+ Atlas cluster.
+     *
+     * @return Iterator&Countable
+     * @throws InvalidArgumentException for parameter/option parsing errors
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     * @see ListSearchIndexes::__construct() for supported aggregation and list options
+     */
+    public function listSearchIndexes(?string $name = null, array $options = []): Iterator
```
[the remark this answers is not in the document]
GromNaN: Sure, I've harmonized `Collection::listSearchIndexes($options)` with `Database::listCollections($options)` and `Collection::listIndexes($options)`.

Do you want to report this signature change to the spec owner?

src/Collection.php
```diff
@@ -390,6 +394,56 @@ public function createIndexes(array $indexes, array $options = [])
...
+     * @see https://www.mongodb.com/docs/manual/reference/command/createSearchIndexes/
+     * @see https://mongodb.com/docs/manual/reference/method/db.collection.createSearchIndex/
+     * @param array{name?: string, definition: array|object} $index   Atlas Search index specification
+     * @param array                                          $options Command options
+     * @return string The name of the created search index
+     * @throws UnsupportedException if options are not supported by the selected server
+     * @throws InvalidArgumentException for parameter/option parsing errors
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function createSearchIndex(array $index, array $options = []): string
```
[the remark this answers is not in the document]
GromNaN: I updated the signature to `createSearchIndex(array|object $definition, array $options = [])` with `name` as an option.

src/Collection.php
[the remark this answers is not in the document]
jmikola: > Do you want to report this signature change to the spec owner?

I think it makes sense to follow up with the spec authors about this. The current signature in the spec is:

```
listSearchIndexes(name: Optional<String>, aggregationOptions: Optional<AggregationOptions>, listIndexOptions: Optional<ListSearchIndexOptions>): Cursor<Document>;
```

ListSearchIndexOptions isn't defined in the spec, but I see no reason for `name` to be broken out as a positional argument. If the database enumeration spec is relevant here (see: [Filters](https://github.com/mongodb/specifications/blob/master/source/enumerate-databases.rst#filters)), I should mention that there's no signature defined in that spec. It just loosely talks about various options for the `listDatabases` command.

Having said that, moving `$name` to `$options` in PHPLIB should not require any changes in the spec. The specs are seldom prescriptive about _how_ to solicit options, so the only important bit here is that `name` is still an option.

src/Collection.php
```diff
@@ -554,6 +619,31 @@ public function dropIndexes(array $options = [])
...
+    public function dropSearchIndex(string $name, array $options = []): void
+    {
+        $operation = new DropSearchIndex($this->databaseName, $this->collectionName, $name);
+        $server = select_server($this->manager, $options);
+
+        try {
+            $operation->execute($server);
+        } catch (CommandException $e) {
+            // Suppress namespace not found errors for idempotency
+            if ($e->getCode() !== 26) {
```
GromNaN, line 641: A spec e2e test expect this error to be ignored. The error code `26` comes from [mongo source](https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.yml#L64)

GromNaN (commented):
Implemented e2e tests from https://github.com/mongodb/specifications/pull/1442

src/Operation/DropSearchIndex.php
GromNaN: Added with a loop just below.

tests/Collection/SearchIndexFunctionalTest.php
```diff
@@ -0,0 +1,70 @@
...
+use MongoDB\Driver\Cursor;
+use MongoDB\Tests\FunctionalTestCase;
+
+use function bin2hex;
+use function getenv;
+use function random_bytes;
+use function sleep;
+
+/**
+ * Functional tests for the Atlas Search index management.
```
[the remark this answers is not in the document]
GromNaN: Updated with the [spec](https://github.com/mongodb/specifications/pull/1442) and the [node example](https://github.com/mongodb/node-mongodb-native/pull/3736).

src/Operation/DropSearchIndex.php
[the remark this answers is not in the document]
GromNaN: > the cluster/operation time could be relevant for folks using change streams or [causal consistency](https://github.com/mongodb/specifications/blob/master/source/causal-consistency/causal-consistency.rst).

I think monitoring features of PHPC a made made for this. I really don't see the point.

src/Collection.php
GromNaN: [NamespaceNotFound Errors](https://github.com/mongodb/specifications/blob/78554de6a78fc996fa8d1ef9733b123a29ad8309/source/index-management/index-management.rst#namespacenotfound-errors) spec

GromNaN:
https://github.com/mongodb/specifications/pull/1442 have been merged and the spec tests files are up to date.

Last thing is to run tests in CI against Atlas [PHPLIB-1168](https://jira.mongodb.org/browse/PHPLIB-1168).

tests/FunctionalTestCase.php
```diff
@@ -519,6 +522,25 @@ protected function skipIfTransactionsAreNotSupported(): void
         }
     }
 
+    protected function isEnterprise(): bool
```
GromNaN, line 525: I recovered this function removed by [#1072](https://github.com/mongodb/mongo-php-library/pull/1072/files#diff-93b2fb11cb160b931a627169b6f5cb9c5d12b940e5a7ff1aec827f39abf1fc11L400-L412)

tests/SpecTests/SearchIndexSpecTest.php
```diff
@@ -0,0 +1,190 @@
...
+        $this->expectNotToPerformAssertions();
+    }
+
+    /**
+     * Randomize the collection name to avoid duplicate index names when running tests concurrently.
+     * Search index operations are asynchronous and can take up to a few minutes.
+     */
+    protected function getCollectionName(): string
+    {
+        return sprintf('%s.%s', parent::getCollectionName(), bin2hex(random_bytes(5)));
```
jmikola, line 153: I assume this may address my previous concern with randomized index names. Feel free to resolve both of these comments if so.

tests/SpecTests/SearchIndexSpecTest.php
```diff
@@ -0,0 +1,190 @@
...
+        $name = 'test-search-index';
+        $mapping = ['mappings' => ['dynamic' => false]];
+
+        $createdName = $collection->createSearchIndex(
+            $mapping,
+            ['name' => $name],
+        );
+        $this->assertSame($name, $createdName);
+
+        $this->waitForIndexes($collection, fn ($indexes) => $this->allIndexesAreQueryable($indexes));
```
jmikola, line 122: Should this also assert that an index was created? This seem related to the comment I left on the third prose test.

tests/UnifiedSpecTests/Operation.php
```diff
@@ -541,6 +541,35 @@ private function executeForCollection(Collection $collection)
...
                 );
 
+            case 'createSearchIndex':
+                $options = [];
+                if (isset($args['model']->name)) {
+                    assertIsString($args['model']->name);
+                    $options['name'] = $args['model']->name;
+                }
+
+                return $collection->createSearchIndex($args['model']->definition, $options);
```
jmikola, line 551: Should this operation assert that `definition` is a stdClass instance? That would be consistent with what you do below for `updateSearchIndex`:

```php
assertInstanceOf(stdClass::class, $args['definition']);
```

tests/UnifiedSpecTests/Operation.php
```diff
@@ -541,6 +541,35 @@ private function executeForCollection(Collection $collection)
...
+            case 'updateSearchIndex':
+                assertArrayHasKey('name', $args);
+                assertArrayHasKey('definition', $args);
+                assertIsString($args['name']);
+                assertInstanceOf(stdClass::class, $args['definition']);
+
+                return $collection->updateSearchIndex($args['name'], $args['definition']);
+
+            case 'listSearchIndexes':
+                return $collection->listSearchIndexes($args + (array) ($args['aggregationOptions'] ?? []));
```
jmikola, line 571: I realize how this works, but it may be more correct to unset `aggregationOptions` after merging it into `$args` to avoid sending more than is necessary to the underlying operation. For example:

```php
$args += (array) ($args['aggregationOptions'] ?? []);
unset($args['aggregationOptions']);
```

tests/Model/SearchIndexInputTest.php
```diff
@@ -0,0 +1,48 @@
...
+class SearchIndexInputTest extends TestCase
+{
+    public function testConstructorIndexDefinitionMustBeDefined(): void
+    {
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage('Required "definition" document is missing from search index specification');
+        new SearchIndexInput([]);
+    }
+
+    public function testConstructorIndexDefinitionMustBeADocument(): void
```
jmikola, line 19: I assume this was based on `testConstructorShouldRequireKeyToBeArrayOrObject` from `IndexInputTest`. Would it make more sense to rewrite both to use the base TestCase's `provideInvalidDocumentValues()` data provider for completeness?

tests/FunctionalTestCase.php
```diff
@@ -51,6 +52,8 @@
 
 abstract class FunctionalTestCase extends TestCase
 {
+    private const ATLAS_TLD = '/\.(mongodb\.net|mongodb-dev\.net)/';
```
jmikola, line 55: UnifiedTestRunner has historically checked in a URI contained "mongodb.net" to detect an Atlas server. This pattern seems more robust, so I'd suggest making a similar change there. You can optionally address this in a separate PR, but I wouldn't object to doing it now.

This would require copying the `isAtlas()` method since UnifiedTestRunner does not extend this class.

src/Operation/UpdateSearchIndex.php
```diff
@@ -0,0 +1,92 @@
+<?php
+/*
+ * Copyright 2015-present MongoDB, Inc.
```
jmikola, line 3: New copyright blocks should use the current year. I'll defer to you to update other occurrences.

```suggestion
 * Copyright 2023-present MongoDB, Inc.
```

src/Operation/UpdateSearchIndex.php
```diff
@@ -0,0 +1,92 @@
...
+            'updateSearchIndex' => $this->collectionName,
+            'name' => $this->name,
+            'definition' => $this->definition,
+        ];
+
+        foreach (['comment'] as $option) {
+            if (isset($this->options[$option])) {
+                $cmd[$option] = $this->options[$option];
+            }
+        }
```
jmikola, line 88: Apologies if a previous comment of mine resulted in you using `foreach` here. If so, feel free to ignore this comment with extreme prejudice.

If there are no other options to support in the foreseeable future, consider unwrapping this:

```suggestion
        if (isset($this->options['comment'])) {
            $cmd['comment'] = $this->options['comment'];
        }
```

src/Operation/DropSearchIndex.php
```diff
@@ -0,0 +1,82 @@
...
+        ];
+
+        foreach (['comment'] as $option) {
+            if (isset($this->options[$option])) {
+                $cmd[$option] = $this->options[$option];
+            }
+        }
```
jmikola, line 78: Related to my suggestion in UpdateSearchIndex.

```suggestion
        if (isset($this->options['comment'])) {
            $cmd['comment'] = $this->options['comment'];
        }
```

src/Operation/CreateSearchIndexes.php
jmikola: I just came across this previously resolved comment after leaving suggestions to unwrap the loops for `comment` options in other files. Feel free to do so here as well, or just keep everything as-is and decline the suggestions accordingly.

Also, you can resolve this thread again after reading.

src/Operation/CreateSearchIndexes.php
```diff
@@ -0,0 +1,103 @@
...
+        if (! array_is_list($indexes)) {
+            throw new InvalidArgumentException('$indexes is not a list');
+        }
+
+        foreach ($indexes as $i => $index) {
+            if (! is_document($index)) {
+                throw InvalidArgumentException::expectedDocumentType(sprintf('$indexes[%d]', $i), $index);
+            }
+
+            $this->indexes[] = new SearchIndexInput((array) $index);
```
jmikola, line 67: This is documented as accepting documents for `$index`, but you perform an array cast here. What happens if a user passes a document in a form other than array or stdClass here (e.g. `MongoDB\BSON\Document` or a Serializable instance)?

For this reason, I think it'd make sense to require `$indexes` be a list of arrays. That would be consistent to how CreateIndexes works today (it uses `! is_array($index)` for validation). That behavior is also covered by a `testConstructorRequiresIndexSpecificationsToBeAnArray()` unit test.

src/Model/SearchIndexInput.php
```diff
@@ -0,0 +1,73 @@
...
+     */
+    public function __construct(array $index)
+    {
+        if (! isset($index['definition'])) {
+            throw new InvalidArgumentException('Required "definition" document is missing from search index specification');
+        }
+
+        if (! is_document($index['definition'])) {
+            throw InvalidArgumentException::expectedDocumentType('"definition" option', $index['definition']);
+        }
```
jmikola, line 53: If PHPLIB never works with this field directly and it is only serialized directly to BSON by PHPC, I'm fine using `is_document()` here. This is contrast to my suggestion to require array types for index specifications in CreateSearchIndexes.

src/Collection.php
```diff
@@ -812,6 +901,24 @@ public function listIndexes(array $options = [])
         return $operation->execute(select_server($this->manager, $options));
     }
 
+    /**
+     * Returns information for all Atlas Search indexes for the collection.
+     * Only available when used against a 7.0+ Atlas cluster.
```
jmikola, line 906: I realize this PR doesn't introduce any new documentation. We typically add code and docs in the same PR for PHPLIB, but I don't want to hold this up for that. I've added a `documentation-needed` label to [PHPLIB-1143](https://jira.mongodb.org/browse/PHPLIB-1143) so we can follow up with that later.

src/Collection.php
jmikola: I think this should be handled in the operation class. See DropCollection and its `ERROR_CODE_NAMESPACE_NOT_FOUND` constant for prior art.

The benefit of doing so is that other libraries that use the operation directly (admittedly a hypothetical) can also benefit from the behavior. I took this a step further with the CSFLE collection management operation classes introduced earlier this year.

src/Collection.php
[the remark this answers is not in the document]
jmikola: @GromNaN are we OK to resolve this thread?

tests/UnifiedSpecTests/UnifiedSpecTest.php
```diff
@@ -273,6 +273,25 @@ public function provideFailingTests()
         yield from $this->provideTests(__DIR__ . '/valid-fail/*.json');
     }
 
+    /** @dataProvider provideIndexManagementTests */
+    public function testIndexManagement(UnifiedTestCase $test): void
+    {
+        if (self::isAtlas()) {
+            self::markTestSkipped('Search Indexes tests must run on a non-atlas cluster');
```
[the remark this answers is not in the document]
GromNaN: We can revisit this code after [DRIVERS-2698](https://jira.mongodb.org/browse/DRIVERS-2698). This need to be checked somewhere to help a developer that would run the test suite on the wrong server environment.

tests/SpecTests/SearchIndexSpecTest.php
```diff
@@ -0,0 +1,190 @@
...
+    public function setUp(): void
+    {
+        if (! self::isAtlas()) {
+            self::markTestSkipped('Search Indexes are only supported on MongoDB Atlas');
+        }
+
+        parent::setUp();
+    }
+
+    public function testCreateAndListSearchIndexes(): void
```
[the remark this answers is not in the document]
GromNaN: Added both the test title and link.

tests/SpecTests/SearchIndexSpecTest.php
```diff
@@ -0,0 +1,190 @@
...
+ * @group atlas
+ */
+class SearchIndexSpecTest extends FunctionalTestCase
+{
+    private const WAIT_TIMEOUT = 300; // 5 minutes
+
+    public function setUp(): void
+    {
+        if (! self::isAtlas()) {
+            self::markTestSkipped('Search Indexes are only supported on MongoDB Atlas');
```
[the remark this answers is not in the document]
GromNaN: Added the version check.

tests/SpecTests/SearchIndexSpecTest.php
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+ * Functional tests for the Atlas Search index management.
+ *
+ * @see https://github.com/mongodb/specifications/blob/master/source/index-management/index-management.rst
+ * @group atlas
+ */
+class SearchIndexSpecTest extends FunctionalTestCase
+{
+    private const WAIT_TIMEOUT = 300; // 5 minutes
```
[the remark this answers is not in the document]
GromNaN: Updated.

src/Operation/UpdateSearchIndex.php
GromNaN: Thanks to the foreach I could avoid the psalm warning. 
```
psalm: MixedAssignment: Unable to determine the type of this assignment
```

src/Operation/CreateSearchIndexes.php
GromNaN: I added the restriction to accept only arrays.

src/Model/SearchIndexInput.php
GromNaN: I removed `is_document` here to accept only arrays.

src/Collection.php
GromNaN: Thanks.

tests/UnifiedSpecTests/Operation.php
GromNaN: Added

tests/SpecTests/SearchIndexSpecTest.php
GromNaN: Yes, that randomize the collection name for the whole class.

src/Operation/DropSearchIndex.php
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /**
+     * Execute the operation.
+     *
+     * @see Executable::execute()
+     * @throws UnsupportedException if write concern is used and unsupported
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function execute(Server $server): void
```
jmikola, line 70: One thing I noticed: other drop operations (e.g. DropCollection) return the command result, which can always be ignored by the caller. Did we previously discuss doing that for these operations instead of using a `void` return type?

I know there are some cases where an application might want to access other fields in a command response, such as cluster time, but that might not apply to these Atlas-specific operations.

I'll also note that the other operations support a `typeMap` option specifically to apply to the returned command result; however, I doubt many users (if any at all) make use of that and I wouldn't push for it here.

src/Operation/DropSearchIndex.php
```diff
@@ -0,0 +1,90 @@
...
+        ];
+
+        if (isset($this->options['comment'])) {
+            $cmd['comment'] = $this->options['comment'];
+        }
+
+        try {
+            $server->executeCommand($this->databaseName, new Command($cmd));
+        } catch (CommandException $e) {
+            // Drop operations is idempotent. The server may return an error if the collection does not exist.
```
jmikola, line 84:
```suggestion
            // Drop operations are idempotent. The server may return an error if the collection does not exist.
```

src/Operation/UpdateSearchIndex.php
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /**
+     * Execute the operation.
+     *
+     * @see Executable::execute()
+     * @throws UnsupportedException if write concern is used and unsupported
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function execute(Server $server): void
```
jmikola, line 76: Related to my earlier question: should we return the raw server response here?

tests/SpecTests/SearchIndexSpecTest.php
```diff
@@ -0,0 +1,218 @@
...
+{
+    private const WAIT_TIMEOUT_SEC = 300;
+
+    public function setUp(): void
+    {
+        if (! self::isAtlas()) {
+            self::markTestSkipped('Search Indexes are only supported on MongoDB Atlas 7.0+');
+        }
+
+        parent::setUp();
```
jmikola, line 37: Is the parent method called in the middle because `isAtlas()` does not depend on `$this->manager`, while `skipIfServerVersion()` does? Just want to confirm, as tests typically just lead with the parent method.

tests/SpecTests/SearchIndexSpecTest.php
```diff
@@ -0,0 +1,190 @@
...
+
+        [$index] = $this->waitForIndexes($collection, fn ($indexes) => $this->allIndexesAreQueryable($indexes));
+
+        $this->assertSame($name, $index->name);
+
+        // Convert to JSON to compare nested associative arrays and nested objects
+        $this->assertJsonStringEqualsJsonString(
+            json_encode($mapping, JSON_THROW_ON_ERROR),
+            json_encode($index->latestDefinition, JSON_THROW_ON_ERROR),
+        );
```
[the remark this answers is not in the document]
jmikola: @GromNaN: did you miss this comment?

tests/SpecTests/SearchIndexSpecTest.php
```diff
@@ -0,0 +1,190 @@
...
+        $createdNames = $collection->createSearchIndexes([
+            ['name' => $names[0], 'definition' => $mapping],
+            ['name' => $names[1], 'definition' => $mapping],
+        ]);
+        $this->assertSame($names, $createdNames);
+
+        $indexes = $this->waitForIndexes($collection, fn ($indexes) => $this->allIndexesAreQueryable($indexes));
+
+        foreach ($names as $key => $name) {
+            $index = $indexes[$key];
```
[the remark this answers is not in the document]
jmikola: @GromNaN: see above.

tests/SpecTests/SearchIndexSpecTest.php
```diff
@@ -0,0 +1,218 @@
...
+        if (count($indexes) === 0) {
+            return false;
+        }
+
+        foreach ($indexes as $index) {
+            if (! $index->queryable) {
+                return false;
+            }
+
+            if (! $index->status === 'READY') {
```
jmikola, line 211: Should we prefer a constant here instead of a string literal?

Is there any reference for this special value in the documentation/spec?

tests/UnifiedSpecTests/Operation.php
jmikola: @GromNaN: see above

src/Collection.php
```diff
@@ -812,6 +893,24 @@ public function listIndexes(array $options = [])
...
+     *
+     * @param array{name?: string} $options Command options
+     * @return Countable&Iterator<array{id: string, name: string, status: string, queryable: bool, latestDefinition: array}>
```
GromNaN, line 901: The return type depends on the `typeMap` option. I have to relax the return type.

src/Collection.php
GromNaN: Starting the example: https://github.com/mongodb/mongo-php-library/pull/1147

tests/Operation/CreateSearchIndexesTest.php
```diff
@@ -0,0 +1,52 @@
...
+    public function testConstructorIndexDefinitionMustBeAnArray($definition): void
+    {
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage('Expected "definition" option to have type "document"');
+        new CreateSearchIndexes($this->getDatabaseName(), $this->getCollectionName(), [['definition' => $definition]], []);
+    }
+
+    public function provideInvalidIndexSpecificationTypes(): array
+    {
+        return $this->wrapValuesForDataProvider($this->getInvalidDocumentValues());
```
jmikola, line 50:
```suggestion
        return $this->wrapValuesForDataProvider($this->getInvalidArrayValues());
```

I think this can be updated now that you're using `is_array()` to validate the list elements. The above code is consistent with `provideInvalidIndexSpecificationTypes()` in `CreateIndexesTest.php`.

tests/SpecTests/SearchIndexSpecTest.php
```diff
@@ -0,0 +1,204 @@
...
+            $mapping,
+            ['name' => $name, 'comment' => 'Index creation test'],
+        );
+        $this->assertSame($name, $createdName);
+
+        $indexes = $this->waitForIndexes($collection, fn ($indexes) => $this->allIndexesAreQueryable($indexes));
+
+        $this->assertCount(1, $indexes);
+        $this->assertSame($name, $indexes[0]->name);
+        $this->assertDocumentsMatch($mapping, $indexes[0]->latestDefinition);
```
jmikola, line 61: Apologies for not specifying, but I was suggesting `assertMatchesDocument()` or `assertSameDocument()` from the base TestCase class instead of comparing JSON strings here.

`assertDocumentsMatch()` is unique to the legacy spec test runner (note that its DocumentsMatchConstraint class differs from the unified test runner's Matches constraint), and I'd be hesitant to move the legacy test runner logic into the main test suite, as the ultimate goal is to delete all of that code once each spec can be ported over to the unified format.

Is it feasible to use one of the two existing TestCase methods here? `assertMatchesDocument()` allows extra fields in the actual document, while `assertSameDocument()` does not -- so you can decide which is most appropriate. If so, you can then revert the introduction of DocumentsMatchConstraint to FunctionalTestCase.

tests/FunctionalTestCase.php
```diff
@@ -201,6 +205,21 @@ protected function assertSameObjectId($expectedObjectId, $actualObjectId): void
...
 
+    /**
+     * Asserts that two given documents match.
+     *
+     * Extra keys in the actual value's document(s) will be ignored.
+     *
+     * @param array|object $expectedDocument
+     * @param array|object $actualDocument
+     */
+    public static function assertDocumentsMatch($expectedDocument, $actualDocument, string $message = ''): void
```
jmikola, line 216: I explained in another thread why I'd rather not introduce legacy spec test classes into the main test suite, but even if this was to stay this would be a method to add on the base TestCase class since it doesn't depend on a server connection.

The base TestCase is also where the existing BSON assertions are.

jmikola (approved):
I suggest reverting the introduction of `assertDocumentsMatch` to the base FunctionalTestCast class and sticking with one of the existing BSON assertions in the base TestCase class.

LGTM after that.

src/Operation/DropSearchIndex.php
jmikola: SGTM. If we were implementing the other operations today, I'd be convinced that a `void` return type would be the way to go. This comment was prompted by me thinking of prior art, but that's not an excuse to perpetuate a bad design.

tests/Operation/CreateSearchIndexesTest.php
GromNaN: Fixed.

jmikola (approved):

PR #1096, opened 2023-06-07
PHPLIB-1122: Additional support for BSON objects

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1122

Adds support for BSON objects with:

 * MapReduce `$out` parameter
 * CreateEncryptedCollection and DropEncryptedCollection `encryptedFields` option

Adds test coverage for BulkWrite operations.

PR #1098, opened 2023-06-08
PHPLIB-1129: Improve pipeline checks for replace operations

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1129

Consolidates data providers in base TestCase and FunctionalTestCase classes.

Change is_pipeline() to conditionally consider empty arrays as pipelines. This was needed for improvements to replace/update parameter validation, and will prove useful for consolidated pipeline validation in [PHPLIB-881](https://jira.mongodb.org/browse/PHPLIB-881).

Prohibit update documents and pipelines (both empty and non-empty) for replacement operations.

Prohibit replacement documents and empty pipeines for update operations. Empty pipelines must be prohibited because libmongoc does not support them for update commands (they're indistinguishable from empty replacement documents). For consistency, we also prohibit them for FindOneAndUpdate even though PHPLIB could work around that.

Prohibit update pipelines for replace operations (both empty and non-empty). Previously ReplaceOne only prohibited non-empty pipelines and BulkWrite replaceOne and FindOneAndReplace has no validation. A notable exception is empty arrays, which are treated as replacement documents for BC.

Add tests for valid update/replacement arguments for BulkWrite replaceOne, updateMany, and updateOne operations. Also adds more comprehensive tests for valid $update arguments for UpdateMany and UpdateOne operations (update documents and non-empty pipelines).

----

This depends on https://github.com/mongodb/mongo-php-library/pull/1096 and can be rebased after that is merged.

### DIFF
--- a/src/Operation/BulkWrite.php
+++ b/src/Operation/BulkWrite.php
@@ -191,8 +191,17 @@ class BulkWrite implements Executable
                         throw InvalidArgumentException::invalidType(sprintf('$operations[%d]["%s"][1]', $i, $type), $args[1], 'array or object');
                     }
 
+                    // Treat empty arrays as replacement documents for BC
+                    if ($args[1] === []) {
+                        $args[1] = (object) $args[1];
+                    }
+
                     if (is_first_key_operator($args[1])) {
                         throw new InvalidArgumentException(sprintf('First key in $operations[%d]["%s"][1] is an update operator', $i, $type));
+                    }
+
+                    if (is_pipeline($args[1], true /* allowEmpty */)) {
+                        throw new InvalidArgumentException(sprintf('$operations[%d]["%s"][1] is an update pipeline', $i, $type));
                     }
 
                     if (! isset($args[2])) {
@@ -229,7 +238,7 @@ class BulkWrite implements Executable
                     }
 
                     if (! is_first_key_operator($args[1]) && ! is_pipeline($args[1])) {
-                        throw new InvalidArgumentException(sprintf('First key in $operations[%d]["%s"][1] is neither an update operator nor a pipeline', $i, $type));
+                        throw new InvalidArgumentException(sprintf('Expected update operator(s) or non-empty pipeline for $operations[%d]["%s"][1]', $i, $type));
                     }
 
                     if (! isset($args[2])) {
--- a/src/Operation/FindAndModify.php
+++ b/src/Operation/FindAndModify.php
@@ -295,6 +295,14 @@ class FindAndModify implements Executable, Explainable
         if (isset($this->options['update'])) {
             /** @psalm-var array|object */
             $update = $this->options['update'];
+            /* A non-empty pipeline will encode as a BSON array, so leave it
+             * as-is. Cast anything else to an object since a BSON document is
+             * likely expected. This includes empty arrays, which historically
+             * can be used to represent empty replacement documents.
+             *
+             * This also allows an empty pipeline expressed as a PackedArray or
+             * Serializable to still encode as a BSON array, since the object
+             * cast will have no effect. */
             $cmd['update'] = is_pipeline($update) ? $update : (object) $update;
         }
 
--- a/src/functions.php
+++ b/src/functions.php
@@ -213,7 +213,22 @@ function is_first_key_operator($document): bool
 }
 
 /**
- * Returns whether an update specification is a valid aggregation pipeline.
+ * Returns whether the argument is a valid aggregation or update pipeline.
+ *
+ * This is primarily used for validating arguments for update and replace
+ * operations, but can also be used for validating an aggregation pipeline.
+ *
+ * The $allowEmpty parameter can be used to control whether an empty array
+ * should be considered a valid pipeline. Empty arrays are generally valid for
+ * an aggregation pipeline, but the things are more complicated for update
+ * pipelines.
+ *
+ * Update operations must prohibit empty pipelines, since libmongoc may encode
+ * an empty pipeline array as an empty replacement document when writing an
+ * update command (arrays and documents have the same bson_t representation).
+ * For consistency, findOneAndUpdate should also prohibit empty pipelines.
+ * Replace operations (e.g. replaceOne, findOneAndReplace) should reject empty
+ * and non-empty pipelines alike, since neither is a replacement document.
  *
  * Note: this method may propagate an InvalidArgumentException from
  * document_or_array() if a Serializable object within the pipeline array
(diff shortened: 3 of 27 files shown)


src/Operation/BulkWrite.php
```diff
@@ -191,10 +191,19 @@ public function __construct(string $databaseName, string $collectionName, array
...
+                    // Treat empty arrays as replacement documents for BC
+                    if ($args[1] === []) {
+                        $args[1] = (object) $args[1];
+                    }
+
                     if (is_first_key_operator($args[1])) {
                         throw new InvalidArgumentException(sprintf('First key in $operations[%d]["%s"][1] is an update operator', $i, $type));
                     }
 
+                    if (is_pipeline($args[1], true /* allowEmpty */)) {
```
GromNaN, line 203: I don't need such comment with phpstorm. It may be replaced by named argument with PHP 8 named argument when the lib is upgraged.

image

src/Operation/BulkWrite.php
```diff
@@ -191,10 +191,19 @@ public function __construct(string $databaseName, string $collectionName, array
                         throw InvalidArgumentException::invalidType(sprintf('$operations[%d]["%s"][1]', $i, $type), $args[1], 'array or object');
                     }
 
+                    // Treat empty arrays as replacement documents for BC
+                    if ($args[1] === []) {
+                        $args[1] = (object) $args[1];
```
GromNaN, line 196: I don't get why you need to convert the empty array into an empty object. Replacement documents can be array, isn't it?

src/functions.php
```diff
@@ -213,7 +213,22 @@ function is_first_key_operator($document): bool
...
+ * should be considered a valid pipeline. Empty arrays are generally valid for
+ * an aggregation pipeline, but the things are more complicated for update
+ * pipelines.
+ *
+ * Update operations must prohibit empty pipelines, since libmongoc may encode
+ * an empty pipeline array as an empty replacement document when writing an
+ * update command (arrays and documents have the same bson_t representation).
+ * For consistency, findOneAndUpdate should also prohibit empty pipelines.
+ * Replace operations (e.g. replaceOne, findOneAndReplace) should reject empty
+ * and non-empty pipelines alike, since neither is a replacement document.
```
GromNaN, line 231: Thanks for this insider comment.

GromNaN (approved):
LGTM. My PR added the same argument to `is_pipeline` for this same issue; I'll rebase it once this one get merged. https://github.com/mongodb/mongo-php-library/pull/1078#issuecomment-1572712603

src/Operation/BulkWrite.php
jmikola: Interesting. How does PhpStorm decide to show the second parameter name but not the first (`$args[1]`)? It looks like it shows the first parameter for `sprintf()` on the second line.

src/Operation/BulkWrite.php
jmikola: Without this, `is_pipeline()` with `allowEmpty: true` is going to consider the argument a pipeline and throw an exception (since pipelines are only acceptable for update operations).

Since `is_pipeline()` historically returned `false` for empty arrays, an empty array was always accepted for ReplaceOne and would be encoded as a BSON object in the `update` command (per libmongoc's internal logic discussed in [CDRIVER-4658](https://jira.mongodb.org/browse/CDRIVER-4658)).

Since this PR is adding logic to reject empty pipelines for _both_ update and replace operations, we need to explicitly allow empty arrays for BC.

BulkWrite `replaceOne` operations historically has never used `is_pipeline()`, so pipelines could have snuck through there. Empty PHP arrays would still be encoded as a replacement document, though (same libmongoc logic).

FindOneAndReplace also never used `is_pipeline()` during constructor validation. The only relevant code path was `FindAndModify::createCommandDocument()` conditionally casting the `update` option to an object if `is_pipeline()` returned false (see: [this change](https://github.com/mongodb/mongo-php-library/commit/2628c098e3bfc015d7e2d59d1437cf1fed046e12#diff-efc373289c930c390bc98d4dbbfb00041c37eda8f5689514b6941697f916d770R265-R269) from [PHPLIB-418](https://jira.mongodb.org/browse/PHPLIB-418)). I actually think there was a bug there, as an empty array would have been encoded as a BSON array in the `findAndModify` command -- and thus treated as an empty pipeline (i.e. NOP) in MongoDB 4.2+.

By prohibiting empty pipelines for _both_ updates and replacements, we resolve the last issue with FindOneAndReplace and also enable users to express empty replacement documents as `[]` like they already could for ReplaceOne and BulkWrite `replaceOne` (where they previously might have had to pass `(object) []` explicitly).

src/Operation/BulkWrite.php
GromNaN: The hint is not shown when a variable is passed, it's shown for static values only.

src/Operation/BulkWrite.php
jmikola: > I actually think there was a bug there, as an empty array would have been encoded as a BSON array in the findAndModify command -- and thus treated as an empty pipeline (i.e. NOP) in MongoDB 4.2+.

Correction. Since FindOneAndReplace previously never validated its `$replacement` parameter with `is_pipeline()`, an empty BSON array would still have correctly encoded as a BSON document due to this logic: https://github.com/mongodb/mongo-php-library/blob/3a681a3b2f2c0ebac227a3b86bb9057d0e6eb8f8/src/Operation/FindAndModify.php#L295-L299

`is_pipeline()` returns false and the parameter gets cast to an object.

So there was no bug.

Also, I can confirm that the cast in `FindAndModify::createCommandDocument()` is still necessary. If we execute FindAndModify directly (as is done in the test suite), then we rely on that code to ensure an empty array encodes properly.

Technically, we could also change the constructor logic in FindOneAndReplace to cast all `$replacement` arrays to objects `is_pipeline($replacement)` is false, but it's that just adds additional overhead and wouldn't cover the case where FindAndModify is executed directly. The empty array check is cheap and plays nicely with the subsequent validation that rejects any pipeline (empty and non-empty alike).

src/Operation/FindAndModify.php
```diff
@@ -295,6 +295,14 @@ private function createCommandDocument(): array
...
             /** @psalm-var array|object */
             $update = $this->options['update'];
+            /* A non-empty pipeline will encode as a BSON array, so leave it
+             * as-is. Cast anything else to an object since a BSON document is
+             * likely expected. This includes empty arrays, which historically
+             * can be used to represent empty replacement documents.
+             *
+             * This also allows an empty pipeline expressed as a PackedArray or
+             * Serializable to still encode as a BSON array, since the object
+             * cast will have no effect. */
```
jmikola, line 305: I wanted to add a comment explaining this conditional cast after diving into it in https://github.com/mongodb/mongo-php-library/pull/1098#discussion_r1224601406.

tests/Operation/FindAndModifyFunctionalTest.php
```diff
@@ -75,38 +74,16 @@ function (array $event) use ($expectedUpdate): void {
...
-        ];
-    }
-
-    public function provideUpdateDocuments(): array
-    {
-        $expected = (object) ['$set' => (object) ['x' => 1]];
-
+        /* Note: this expected value differs from UpdateFunctionalTest because
+         * FindAndModify is not affected by libmongoc's pipeline detection for
+         * update commands (see: CDRIVER-4658). */
```
jmikola, line 81: I previously added this test case to UpdateFunctionalTest so we had a record of [CDRIVER-4658](https://jira.mongodb.org/browse/CDRIVER-4658). FindAndModify behaves correctly (unaffected by libmongoc's code path for `update` commands) and I wanted to make a record of that in the test suite.

PR #1099, opened 2023-06-09
PHPLIB-1152: replace `private static` with `private const`

GromNaN:
Fix PHPLIB-1152

`Client::$version` is the only one left because it is initialised from composer install.

### DIFF
--- a/src/ChangeStream.php
+++ b/src/ChangeStream.php
@@ -42,15 +42,11 @@ class ChangeStream implements Iterator
 {
     /**
      * @deprecated 1.4
-     * @todo Remove this in 2.0 (see: PHPLIB-360)
+     * @todo make this constant private in 2.0 (see: PHPLIB-360)
      */
     public const CURSOR_NOT_FOUND = 43;
 
-    /** @var int */
-    private static $cursorNotFound = 43;
-
-    /** @var int[] */
-    private static $resumableErrorCodes = [
+    private const RESUMABLE_ERROR_CODES = [
         6, // HostUnreachable
         7, // HostNotFound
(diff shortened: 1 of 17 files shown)


src/ChangeStream.php
```diff
@@ -42,15 +42,11 @@ class ChangeStream implements Iterator
 {
     /**
      * @deprecated 1.4
-     * @todo Remove this in 2.0 (see: PHPLIB-360)
+     * @todo make this constant private in 2.0 (see: PHPLIB-360)
```
jmikola, line 45: Ah, I understand why this was changed. You couldn't convert `private static $cursorNotFound` to a `private const` because of this existing definition.

I just revised [PHPLIB-360](https://jira.mongodb.org/browse/PHPLIB-360) accordingly to capture this change.

src/Operation/Delete.php
```diff
@@ -43,8 +43,7 @@
  */
 class Delete implements Executable, Explainable
 {
-    /** @var integer */
-    private static $wireVersionForHint = 9;
+    private const WIRE_VERSION_FOR_HINT = 5;
```
jmikola, line 46:
```suggestion
    private const WIRE_VERSION_FOR_HINT = 9;
```

Not sure how this was changed from 9 to 5, but the correct wire version is definitely 9. See: https://github.com/mongodb/specifications/blob/master/source/wireversion-featurelist.rst

jmikola (approved):
LGTM once `Delete::WIRE_VERSION_FOR_HINT` is corrected.

alcaeus (approved):
Nice one!

PR #1100, opened 2023-06-12
PHPLIB-722: Ignore writeConcern option for read-only aggregations

GromNaN:
Fix PHPLIB-722

This does not change anything functionally, but cleanup `$options` in the constructor instead of picking the `writeOption` conditionally in `executeCommand()`. There is no impact on `getCommandDocument()` which already ignores `writeConcern`.

### DIFF
--- a/tests/Collection/CollectionFunctionalTest.php
+++ b/tests/Collection/CollectionFunctionalTest.php
@@ -7,7 +7,6 @@ use MongoDB\BSON\Javascript;
 use MongoDB\Collection;
 use MongoDB\Database;
 use MongoDB\Driver\BulkWrite;
-use MongoDB\Driver\Exception\CommandException;
 use MongoDB\Driver\ReadConcern;
 use MongoDB\Driver\ReadPreference;
 use MongoDB\Driver\WriteConcern;
@@ -457,6 +456,7 @@ class CollectionFunctionalTest extends FunctionalTestCase
                 }, 'r',
             ],
 
+            /* Disabled, as write aggregations are not supported in transactions
             'read-write aggregate' => [
                 function ($collection, $session, $options = []): void {
                     $collection->aggregate(
@@ -468,6 +468,7 @@ class CollectionFunctionalTest extends FunctionalTestCase
                     );
                 }, 'rw',
             ],
+            */
 
             'bulkWrite insertOne' => [
                 function ($collection, $session, $options = []): void {
@@ -748,16 +749,7 @@ class CollectionFunctionalTest extends FunctionalTestCase
 
         (new CommandObserver())->observe(
             function () use ($method, $collection, $session): void {
-                try {
-                    call_user_func($method, $collection, $session);
-                } catch (CommandException $exception) {
-                    // Write aggregates are not supported in transactions
-                    if ($exception->getResultDocument()->codeName === 'OperationNotSupportedInTransaction') {
-                        $this->markTestSkipped('OperationNotSupportedInTransaction: ' . $exception->getMessage());
-                    }
-
-                    throw $exception;
-                }
+                call_user_func($method, $collection, $session);
             },
             function (array $event): void {
                 $this->assertObjectNotHasAttribute('writeConcern', $event['started']->getCommand());
--- a/tests/Collection/CollectionFunctionalTest.php
+++ b/tests/Collection/CollectionFunctionalTest.php
@@ -7,6 +7,7 @@ use MongoDB\BSON\Javascript;
 use MongoDB\Collection;
 use MongoDB\Database;
 use MongoDB\Driver\BulkWrite;
+use MongoDB\Driver\Exception\CommandException;
 use MongoDB\Driver\ReadConcern;
 use MongoDB\Driver\ReadPreference;
 use MongoDB\Driver\WriteConcern;
@@ -21,7 +22,7 @@ use function array_filter;
 use function call_user_func;
 use function is_scalar;
 use function json_encode;
-use function strchr;
+use function str_contains;
 use function usort;
 use function version_compare;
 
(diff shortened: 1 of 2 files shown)


GromNaN:
Failing test:
```
1) MongoDB\Tests\Collection\CollectionFunctionalTest::testMethodInTransactionWithWriteConcernOption with data set #0 (Closure Object (...), 'rw')
Failed asserting that exception of type "MongoDB\Exception\UnsupportedException" is thrown.
```

Ignoring the `writeConcern` in the constructor skips the exception thrown in [`Aggregate::execute()`](https://github.com/mongodb/mongo-php-library/blob/master/src/Operation/Aggregate.php#L271)

jmikola:
`collectionMethodClosures()` includes the following `aggregate()` call, which is classified as a read and write:

```
[
    function ($collection, $session, $options = []): void {
        $collection->aggregate(
            [['$match' => ['_id' => ['$lt' => 3]]]],
            ['session' => $session] + $options
        );
    }, 'rw',
],
```

Can you try changing "rw" there to "r" and creating a second copy with an actual write stage (e.g. `$out`) with "rw" changed to "w"? That might overcome the unsetting of `writeConcern` and allow the actual exception to be triggered.

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -698,32 +711,28 @@ function($collection, $session, $options = []) {
...
-                    return true;
-                }
+                return str_contains($rw[1], 'w');
             }
         );
     }
 
     /** @dataProvider collectionMethodClosures */
-    public function testMethodDoesNotInheritReadWriteConcernInTranasaction(Closure $method): void
+    public function testMethodDoesNotInheritReadWriteConcernInTransaction(Closure $method): void
```
GromNaN, line 735: Fixed a typo

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -739,7 +748,16 @@ public function testMethodDoesNotInheritReadWriteConcernInTranasaction(Closure $
 
         (new CommandObserver())->observe(
             function () use ($method, $collection, $session): void {
-                call_user_func($method, $collection, $session);
+                try {
+                    call_user_func($method, $collection, $session);
+                } catch (CommandException $exception) {
+                    // Write aggregates are not supported in transactions
+                    if ($exception->getResultDocument()->codeName === 'OperationNotSupportedInTransaction') {
+                        $this->markTestSkipped('OperationNotSupportedInTransaction: ' . $exception->getMessage());
```
GromNaN, line 756: `$out` and `$merge` operators aren't supported in transaction. They throw a server error. I have a doubt on the best solution to exclude the new "read-write aggregate" method:
1. This solution: skip the test in case of exception. But unlike tests skipped due to a temporary incompatiblity with the running environment, this test will never pass.
2. Ignores the error and still run the assertions on command options. Even if that does not make sense to test a failing command. That could hide other errors and make debugging more difficult.
3. exclude this case from the test cases by adding a `collectionTransactionMethodClosures` data provider and add a new char `t` in `$rw` to exclude methods incompatible with transactions.

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -447,16 +448,28 @@ public function testMapReduce(): void
     public function collectionMethodClosures()
     {
         return [
-            [
+            'read-only aggregate' => [
```
GromNaN, line 451: Debugging is simplified when the data provider uses names.

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -698,32 +711,28 @@ function($collection, $session, $options = []) {
...
-    public function collectionReadMethodClosures()
+    public function collectionReadMethodClosures(): array
     {
         return array_filter(
             $this->collectionMethodClosures(),
             function ($rw) {
-                if (strchr($rw[1], 'r') !== false) {
-                    return true;
-                }
+                return str_contains($rw[1], 'r');
```
GromNaN, line 719: Small refactoring to use the PHP 8 function, provided by the polyfill for older versions.

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -447,16 +448,28 @@ public function testMapReduce(): void
...
     {
         return [
-            [
+            'read-only aggregate' => [
                 function ($collection, $session, $options = []): void {
                     $collection->aggregate(
                         [['$match' => ['_id' => ['$lt' => 3]]]],
                         ['session' => $session] + $options
                     );
+                }, 'r',
```
GromNaN, line 457: This pipeline is actually read-only. I added a write pipeline as @jmikola recommended.

tests/Collection/CollectionFunctionalTest.php
jmikola: > 1. This solution: skip the test in case of exception. But unlike tests skipped due to a temporary incompatiblity with the running environment, this test will never pass.
>
> 2. Ignores the error and still run the assertions on command options. Even if that does not make sense to test a failing command. That could hide other errors and make debugging more difficult.

I don't think conditionally skipping the test makes sense, given that the test only cares about asserting that `readConcern` and `writeConcern` are omitted on the outgoing command.

Catching and ignoring [`OperationNotSupportedInTransaction`](https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.yml#LL312C57-L312C57) seems preferable. For consistency, I would suggest defining a private constant for its numeric code (263).

That said, I don't think that's a permanent solution. Read on.

> 3. exclude this case from the test cases by adding a collectionTransactionMethodClosures data provider and add a new char t in $rw to exclude methods incompatible with transactions.

I don't think this is necessary. `collectionMethodClosures()` already comments out various methods that cannot be run within a transaction. I was going to suggest removing all of those comments after adding logic to ignore OperationNotSupportedInTransaction, but then I realized that would only fix `testMethodDoesNotInheritReadWriteConcernInTransaction`.

`testMethodInTransactionWithWriteConcernOption` and `testMethodInTransactionWithReadConcernOption` use derivative data providers that filter for strings containing "w" and "r", respectively. Those tests would never work with the unsupported operations because we never added logic to check for active transactions and throw UnsupportedException.

Aggregate is a special case because pipelines are generally supported in transactions and only write operators are unsupported.

----

With all of that in mind, I think the most sensible solution is removing the pipeline with a write stage and just leaving a comment above read-only pipeline to explain that we're intentionally not testing a pipeline with a write stage since those are unsupported.

That would mean we have no test coverage for the following logic near the top of `Aggregate::execute()`:

```php
if (isset($this->options['writeConcern'])) {
   throw UnsupportedException::writeConcernNotSupportedInTransaction();
}
```

If you'd like to cover that, I'd suggest doing one of the following:

 * Create a second data provider for `testMethodDoesNotInheritReadWriteConcernInTransaction()` that only includes the write pipeline. The test only takes one parameter so there'd be no reason to include "rw" in the data provider. This would still require modifying the test to catch and ignore OperationNotSupportedInTransaction errors.
 * Create a separate test similar to `testMethodDoesNotInheritReadWriteConcernInTransaction()` that tests Aggregate directly with the write pipeline. A doc block for that test would be a good opportunity to explain why we couldn't do this with the common data provider.

Alternatively, just remove the write pipeline and leave this untested. I'm not too concerned about doing so since we know this code path will lead to an exception anyway.

GromNaN: You're right, there's no point in testing behavior on a command that's going to fail anyway.

jmikola (approved):

PR #893, opened 2022-02-25
PHPLIB-795, PHPLIB-796: Remove references to pre-3.6 servers

jmikola:
https://jira.mongodb.org/browse/PHPLIB-795
https://jira.mongodb.org/browse/PHPLIB-796

Patch build: https://spruce.mongodb.com/version/6219441961837d53eb3910d1/tasks

PR #1102, opened 2023-06-12
PHPLIB-1162: Remove unused argument Explainable::getCommandDocument(Server $server)

GromNaN:
Fix PHPLIB-1162

### DIFF
--- a/src/Operation/Explainable.php
+++ b/src/Operation/Explainable.php
@@ -17,8 +17,6 @@
 
 namespace MongoDB\Operation;
 
-use MongoDB\Driver\Server;
-
 /**
  * Explainable interface for explainable operations (aggregate, count, distinct,
  * find, findAndModify, delete, and update).
@@ -32,5 +30,5 @@ interface Explainable extends Executable
      *
      * @return array
      */
-    public function getCommandDocument(Server $server);
+    public function getCommandDocument();
 }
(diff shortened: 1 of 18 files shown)


src/Operation/Explainable.php
```diff
@@ -32,5 +32,5 @@ interface Explainable extends Executable
      *
      * @return array
      */
-    public function getCommandDocument(Server $server);
```
GromNaN, line 35: This interface is `@internal`, we can change it without breaking use code.

alcaeus (commented):
Code changes LGTM. I'm a little bit on the fence about whether we should make this change.

FWIW, the `$server` parameter was added as back then the `FindAndModify` operation used it in its `createCommandDocument` method to appropriately add the `bypassDocumentValidation` option. However, this was removed as part of #893 when we cleaned up pre-3.6 logic. There could be future cases in which the command document might depend on a selected server to add the correct options, in which case we'd be adding the parameter back to every `Explainable`.

GromNaN:
Interesting. To reduce coupling, it might be replaced by the server info array when the needs to check server version pops again.

jmikola:
> Code changes LGTM. I'm a little bit on the fence about whether we should make this change.

@alcaeus: Thanks for saving me the trouble of digging up the historical references.

> To reduce coupling, it might be replaced by the server info array when the needs to check server version pops again.

The dominate use case for Server arguments in `execute()` methods (similar to this) is calling `server_supports_feature()`, which does only need `Server::getInfo()`. I don't feel strongly about this, though, since it would just put the onus on each operation to extract the min/max wire versions.

I reckon the only real benefit is we might be able to write unit tests for `server_supports_feature()` (IIRC mocking the server isn't feasible).

PR #1103, opened 2023-06-12
PHPLIB-1120: Remove redundant options from with_transaction() docs examples

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1120


GromNaN (approved):
LGTM. Not need for this options when set to their default value.

PR #1104, opened 2023-06-14
PHPLIB-1142: Use `disableMD5` when provided to GridFS Bucket constructor

GromNaN:
Fix PHPLIB-1142

Bugfix, the option value set to the constructor was ignored. The only way to disable md5 was to  set the option when calling `Bucket::uploadFromStream`.

Added a functional test to ensure this option works. But it does not cover this specific bug.


tests/GridFS/BucketFunctionalTest.php
```diff
@@ -669,6 +670,19 @@ public function testUploadingAnEmptyFile(): void
         $this->assertSameDocument($expected, $fileDocument);
     }
 
+    public function testDisableMD5(): void
```
alcaeus, line 673: While this test checks that the `disableMD5` option is used when passed to `uploadFromStream`, the PR title suggests the issue was with the option not being respected when passed to the constructor. Can you add a test for that as well?

GromNaN: Indeed, test added.

alcaeus (approved):
LGTM!

PR #1105, opened 2023-06-14
PHPLIB-1144: Add unit tests on `Explainable::getCommandDocument()`

GromNaN:
Fix PHPLIB-1144

- Add `let` and `comment` options to explain `delete`.
- Remove `explain` option for Aggregate, it is incompatible with the `explain` command.

Adding a unit test will full options for each implementation of `Explainable::getCommandDocument`:
- `Count` ✅
- `Delete` ✅ 
- `DeleteMany` covered by `Delete`
- `DeleteOne` covered by `Delete`
- `Distinct` ✅ 
- `EstimatedDocumentCount` covered by `Count`
- `Find` ✅ 
- `FindAndModify` ✅ 
- `FindOne` covered by `Find`
- `FindOneAndDelete` ✅ + covered by `FindAndModify`
- `FindOneAndReplace` ✅ + covered by `FindAndModify`
- `FindOneAndUpdate` ✅ + covered by `FindAndModify`
- `Update` ✅ 
- `UpdateMany` covered by `Update`
- `UpdateOne` covered by `Update`

These tests allow us to review exactly which options are sent to the `explain` command.

The comparison of `ReadConcern` values will become stronger with PHPC-2234.


tests/Operation/FindAndModifyTest.php
```diff
@@ -83,4 +84,46 @@ public function testConstructorUpdateAndRemoveOptionsAreMutuallyExclusive(): voi
...
+        ];
+        $operation = new FindAndModify($this->getDatabaseName(), $this->getCollectionName(), $options);
+
+        $expected = [
+            'findAndModify' => $this->getCollectionName(),
+            'new' => true,
+            'upsert' => true,
+            'collation' => (object) ['locale' => 'fr'],
+            'fields' => (object) ['_id' => 0],
+            'let' => (object) ['a' => 3],
```
GromNaN, line 116: `let` option is provided here, but skipped for `delete` operation. I'm not sure if it's expected or a bug?

`FindAndModify::getCommandDocument()` returns unchanged command, it's modified for other operations.

alcaeus: `let` should not be skipped for delete operations. The best course of action would be to add specific tests for the various `FindOneAnd...` operation classes and assert that they behave correctly.

That said, I looked through both `FindAndModify` and `FindOneAndDelete` operation classes and the `let` option seems to be handled correctly. What do you mean with "it's modified for other operations"?

GromNaN: `FindAndModify::getCommandDocument` is a direct copy of `createCommandDocument`.

https://github.com/mongodb/mongo-php-library/blob/b3607d91b71b4efb2d92e52e68073308c9696b39/src/Operation/FindAndModify.php#L263-L266

In other Operation classes, the `getCommandDocument` is customized for explain needs.
https://github.com/mongodb/mongo-php-library/blob/b3607d91b71b4efb2d92e52e68073308c9696b39/src/Operation/Delete.php#L178-L181

https://github.com/mongodb/mongo-php-library/blob/b3607d91b71b4efb2d92e52e68073308c9696b39/src/Operation/Find.php#L331-L341

tests/Operation/DeleteTest.php
```diff
@@ -65,4 +65,30 @@ public function provideInvalidConstructorOptions()
...
+
+    public function testExplainableCommandDocument(): void
+    {
+        $options = [
+            'collation' => ['locale' => 'fr'],
+            'hint' => '_id_',
+            // Ignored options
+            'let' => ['a' => 1],
+            'ordered' => true,
+            'comment' => 'explain me',
```
jmikola, line 77: `ordered` shouldn't be included here, as it's not documented as an option for Delete.

As for `let` and `comment`, those should probably appear in the generated command document as top-level fields. Quoting [Explain: Command Fields](https://www.mongodb.com/docs/manual/reference/command/explain/#command-fields):

> If you specify explain without a comment, it inherits any comment in the command specified to explain.

So, the `comment` option is certainly relevant to include. I'm not sure about `let`, but I don't think there's any harm in including that as well. You can also hunt down someone from the server query team if you want a definitive answer on that.

Also, I recall that `writeConcern` was removed in https://github.com/mongodb/mongo-php-library/commit/118f08409fec7ce8be8664dc7bea0803842f8d2c for [PHPLIB-802](https://jira.mongodb.org/browse/PHPLIB-802). For operations where you explicitly excluding `writeConcern` from `Explainable::getCommandDocument()`, I think it'd make sense to include it in `$options` with a comment noting that it is intentionally ignored.

tests/Operation/DistinctTest.php
```diff
@@ -51,4 +53,28 @@ public function provideInvalidConstructorOptions()
...
         return $options;
     }
+
+    public function testExplainableCommandDocument(): void
+    {
+        $options = [
+            'collation' => ['locale' => 'fr'],
+            'maxTimeMS' => 100,
+            'readConcern' => new ReadConcern(ReadConcern::LOCAL),
+            'readPreference' => new ReadPreference(ReadPreference::SECONDARY_PREFERRED),
```
jmikola, line 63: Depending on the topology to which the driver is connected, this might result in a `$readPreference` option appearing in the outgoing command (see: [OP_MSG: Global Command Arguments](https://github.com/mongodb/specifications/blob/master/source/message/OP_MSG.rst#global-command-arguments)); however, I have no idea if that's relevant to `explain`. This may be another thing to ask the server query team about if you're so inclined.

Otherwise, lumping this towards the end of the `$options` array under `// Intentionally omitted options` (or something to that effect) seems good. Similar to what you'd do for Explainable operations that take a `writeConcern` option.

tests/Operation/FindAndModifyTest.php
jmikola: > FindAndModify::getCommandDocument is a direct copy of createCommandDocument.

This is generally the case for most operations that are executed using `executeCommand()` (or a related function). An exception is where some `execute` options need to be injected, as you recently did with adding `readConcern` and removing `writeConcern`.

Operations using `executeBulkWrite()` and `executeQuery()` are special because PHPLIB has no reason to construct the outgoing command apart from `Explainable::getCommandDocument()`.

In the case of Find, it's a bit odd that `createCommandDocument()` only exists for the purpose of `getCommandDocument()`. It looks like that dates back to [c7b2b03](https://github.com/mongodb/mongo-php-library/commit/c7b2b033b327b29a619ee0ed2bcd9d055b4ab9d1#diff-ff201f3e96f2f4300f6d25d87430d60784224c2b5509ed1f4d52d5ab03d52d92R304). You're welcome to factor out that private method if you wish (either in this PR or down the line).

tests/Operation/FindOneAndDeleteTest.php
```diff
@@ -31,4 +32,34 @@ public function provideInvalidConstructorOptions()
...
+    {
+        $options = [
+            'collation' => ['locale' => 'fr'],
+            'comment' => 'explain me',
+            'hint' => '_id_',
+            'maxTimeMS' => 100,
+            'projection' => ['_id' => 0],
+            'sort' => ['x' => 1],
+            'let' => ['a' => 3],
+            'typeMap' => ['root' => 'array'],
```
jmikola, line 46: I suggest adding a comment before any ignored options.

tests/Operation/FindTest.php
```diff
@@ -154,4 +156,61 @@ public function provideInvalidConstructorCursorTypeOptions()
...
+            'allowPartialResults' => true,
+            'batchSize' => 123,
+            'comment' => 'explain me',
+            'hint' => '_id_',
+            'limit' => 15,
+            'maxTimeMS' => 100,
+            'noCursorTimeout' => true,
+            'oplogReplay' => true,
+            'projection' => ['_id' => 0],
+            'readConcern' => new ReadConcern(ReadConcern::LOCAL),
```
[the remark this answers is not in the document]
jmikola: I suppose it's not worth adding a `@todo` comment here since the code won't need to be changed in PHPLIB. This is worth calling out in the commit message, though (with a reference to PHPC-2234).

src/Operation/Aggregate.php
```diff
@@ -300,6 +300,10 @@ public function getCommandDocument()
     {
         $cmd = $this->createCommandDocument();
 
+        // The 'explain' option is incompatible with the explain command
+        unset($cmd['explain']);
```
GromNaN, line 304: I get the following error if the option `explain` is set, even if it's `false`.

> MongoDB\Driver\Exception\CommandException: The 'explain' option is illegal when a explain verbosity is also provided

tests/Operation/FindTest.php
```diff
@@ -154,4 +156,62 @@ public function provideInvalidConstructorCursorTypeOptions()
...
+            'typeMap' => ['root' => 'array'],
+        ];
+        $operation = new Find($this->getDatabaseName(), $this->getCollectionName(), ['x' => 1], $options);
+
+        $expected = [
+            'find' => $this->getCollectionName(),
+            'filter' => (object) ['x' => 1],
+            'allowDiskUse' => true,
+            'allowPartialResults' => true,
+            'batchSize' => 123,
```
GromNaN, line 197: The `batchSize` seems to be useless for an explain. It is not supported by the aggregate command with `explain` option:
https://github.com/mongodb/mongo-php-library/blob/78cc92542e8710925e6a8f8cffee54051f4385b9/src/Operation/Aggregate.php#L225-L229

jmikola: > The batchSize seems to be useless for an explain

I imagine so, but I don't think there's any reason for PHPLIB to care about that. We can rely on the server to ignore it or raise an error. I wouldn't expect an error (not going to bother testing myself), but if so the user could always revise their operation options.

The case for unsetting `batchSize` when using the `explain` option with `aggregate` is different. That logic has nothing to do with it being relevant/supported when explaining a pipeline, but rather the fact that `aggregate` with `explain: true` returns a single document. The `batchSize` option is only relevant if `aggregate` yields a cursor.

src/Operation/Aggregate.php
jmikola: I'm OK relying on the server to raise an error here if you'd like to remove this.

I'd argue this is fundementally different from intentionally omitting `writeConcern` ([PHPLIB-802](https://jira.mongodb.org/browse/PHPLIB-802)) because that option is easily inherited from a parent context.

If a user encounters the error you highlighted above, it's because they've explicitly crafted an Aggregate operation with an `explain` option, and then decided to pass that to `Collection::explain()`.

src/Operation/Find.php
```diff
@@ -353,6 +338,11 @@ private function createCommandDocument(): array
             return $cmd;
         }
 
+        // Read concern can change the query plan
+        if (isset($this->options['readConcern'])) {
```
jmikola, line 342: I think this conditional can be removed.

`createQueryOptions()` actually includes the `readConcern` option, so it will already be added by the final `$cmd + $options` union in the return statement. This is because the `readConcern` option is passed to the Query constructor instead of `Server::executeQuery()` (as it might be for `executeReadCommand()` in other operations).

Hypothetically, if that wasn't the case, the position of this conditional would still be problematic due to the early return above if `createQueryOptions()` returns an empty array. The edge case would be constructing Find with only a `readConcern` option.

While writing this, I realized the docs regarding read concern inheritance for queries is not very clear. Opened [PHPC-2249](https://jira.mongodb.org/browse/PHPC-2249) to improve that.

jmikola (approved):
LGTM with at least the Psalm error resolved.

I'll defer to you on the other suggested changes.

GromNaN:
CS and Psalm checks OK. The failing tests are unrelated, due to https://github.com/mongodb/mongo-php-driver/commit/358fd2fc174faed00ce5c9f91f645a35da1fcbe9

PR #1107, opened 2023-06-14
PHPLIB-1153: Add `FunctionalTestCase::skipIfServerVersion($op, $version)`

GromNaN:
Fix PHPLIB-1153

### DIFF
--- a/tests/FunctionalTestCase.php
+++ b/tests/FunctionalTestCase.php
@@ -484,6 +484,13 @@ abstract class FunctionalTestCase extends TestCase
         return preg_match('@^.*/.*:\d+@', $document['host']);
     }
 
+    protected function skipIfServerVersion(string $operator, string $version, ?string $message = null): void
+    {
+        if (version_compare($this->getServerVersion(), $version, $operator)) {
+            $this->markTestSkipped($message ?? sprintf('Server version is %s %s', $operator, $version));
+        }
+    }
+
     protected function skipIfChangeStreamIsNotSupported(): void
     {
         switch ($this->getPrimaryServer()->getType()) {
--- a/tests/Operation/CreateIndexesFunctionalTest.php
+++ b/tests/Operation/CreateIndexesFunctionalTest.php
@@ -16,7 +16,6 @@ use MongoDB\Tests\CommandObserver;
 use function call_user_func;
 use function is_callable;
 use function sprintf;
-use function version_compare;
 
 class CreateIndexesFunctionalTest extends FunctionalTestCase
 {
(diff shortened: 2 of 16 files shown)


tests/FunctionalTestCase.php
```diff
@@ -484,6 +484,13 @@ protected function isShardedClusterUsingReplicasets()
         return preg_match('@^.*/.*:\d+@', $document['host']);
     }
 
+    protected function skipIfServerVersionNotSupported(string $operator, string $version, ?string $message = null): void
```
jmikola, line 487: How do you feel about naming this `skipIfServerVersion()? That would both make the LOC shorter and allow it to read like a natural sentence. For example:

```php
$this->skipIfServerVersion('<', '7.0.0', '...');
```

tests/Operation/CreateIndexesFunctionalTest.php
```diff
@@ -216,9 +213,7 @@ function (array $event): void {
 
     public function testCommitQuorumUnsupported(): void
     {
-        if (version_compare($this->getServerVersion(), '4.3.4', '>=')) {
-            $this->markTestSkipped('commitQuorum is supported');
-        }
+        $this->skipIfServerVersionNotSupported('>=', '4.3.4', 'commitQuorum is supported');
```
jmikola, line 216: Note: this is one example where "NotSupported" in the method name doesn't actually make sense. We're intentionally skipping a test where a feature _is_ supported in the server.

Likewise for some `hint` tests below.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -61,9 +60,7 @@ public function setUp(): void
      */
     public function testGetResumeToken(): void
     {
-        if ($this->isPostBatchResumeTokenSupported()) {
-            $this->markTestSkipped('postBatchResumeToken is supported');
-        }
+        $this->skipIfPostBatchResumeTokenNotSupported();
```
jmikola, line 63: This logic is incorrect. The test needs to be skipped if `postBatchResumeToken` _is_ supported (i.e. server version >= 4.0.7).

tests/Operation/WatchFunctionalTest.php
```diff
@@ -1596,9 +1567,9 @@ private function insertDocument($document): void
         $this->assertEquals(1, $writeResult->getInsertedCount());
     }
 
-    private function isPostBatchResumeTokenSupported()
+    private function skipIfPostBatchResumeTokenNotSupported(): void
     {
-        return version_compare($this->getServerVersion(), '4.0.7', '>=');
+        $this->skipIfServerVersionNotSupported('>=', '4.0.7', 'postBatchResumeToken is supported');
```
jmikola, line 1572: As noted above, this method doesn't cover all use cases for `isPostBatchResumeTokenSupported()`. Three tests are skipped when `postBatchResumeToken` _is_ supported, and two are skipped when it's _not_ supported.

If you'd like to refactor this, I'd propose just dropping this method entirely and replacing the calling code with the appropriate `$this->skipIfServerVersion()` call. It's unlikely that we'd add many more tests for the feature so I'm not concerned with duplicating "4.0.7" in a few places.

jmikola (changes requested):

tests/FunctionalTestCase.php
GromNaN: Less words is more correct 🥂

tests/Operation/WatchFunctionalTest.php
GromNaN: Good idea, I've totally removed `isPostBatchResumeTokenSupported`. This has more sense.

jmikola (approved):
One suggestion for the assertion message.

LGTM after that is fixed.

alcaeus (approved):

PR #1112, opened 2023-06-20
PHPLIB-1147, PHPLIB-1148: Int64 improvements

alcaeus:
PHPLIB-1147
PHPLIB-1148

This PR removes the previous workarounds for creating Int64 instances through `unserialise` and adds a comparator for Int64 values. Using this comparator renders other workarounds for comparing Int64 instances obsolete.

I've built the comparator to only become active if at least one value is an Int64 instance and the other value is an int or numeric string. On 64-bit platforms we're comparing int values directly, whereas we compare string representations on 32-bit platforms to ensure correct behaviour for 64-bit values.

A note to contributors: if you have your own `phpunit.xml` file, make sure to update the `bootstrap` configuration as seen in `phpunit.xml.dist`. This is necessary to ensure that the new comparator is loaded.

### DIFF
--- a/tests/Comparator/Int64Comparator.php
+++ b/tests/Comparator/Int64Comparator.php
@@ -1,0 +1,61 @@
+<?php
+
+namespace MongoDB\Tests\Comparator;
+
+use MongoDB\BSON\Int64;
+use SebastianBergmann\Comparator\Comparator;
+use SebastianBergmann\Comparator\ComparisonFailure;
+
+use function is_int;
+use function is_numeric;
+use function is_string;
+use function sprintf;
+
+use const PHP_INT_SIZE;
+
+class Int64Comparator extends Comparator
+{
+    public function accepts($expected, $actual)
+    {
+        // Only compare if either value is an Int64
+        return ($expected instanceof Int64 && $this->isComparable($actual))
+            || ($actual instanceof Int64 && $this->isComparable($expected));
+    }
+
+    public function assertEquals($expected, $actual, $delta = 0.0, $canonicalize = false, $ignoreCase = false): void
+    {
+        if (PHP_INT_SIZE == 8) {
+            // On 64-bit systems, compare integers directly
+            $expectedValue = (int) $expected;
+            $actualValue = (int) $actual;
+        } else {
+            // On 32-bit systems, compare integers as strings
+            $expectedValue = (string) $expected;
+            $actualValue = (string) $actual;
+        }
+
+        if ($expectedValue === $actualValue) {
+            return;
+        }
+
+        throw new ComparisonFailure(
+            $expected,
+            $actual,
+            '',
+            '',
+            false,
+            sprintf(
+                'Failed asserting that %s matches expected %s.',
+                $this->exporter->export($actual),
+                $this->exporter->export($expected)
+            )
+        );
+    }
+
+    private function isComparable($value): bool
+    {
+        return $value instanceof Int64 // Int64 instances
+            || is_int($value) // Integer values
+            || (is_string($value) && is_numeric($value)); // Numeric strings (is_numeric accepts floats)
+    }
+}
--- a/tests/Comparator/Int64Comparator.php
+++ b/tests/Comparator/Int64Comparator.php
@@ -6,35 +6,21 @@ use MongoDB\BSON\Int64;
 use SebastianBergmann\Comparator\Comparator;
 use SebastianBergmann\Comparator\ComparisonFailure;
 
-use function is_int;
 use function is_numeric;
-use function is_string;
 use function sprintf;
-
-use const PHP_INT_SIZE;
 
 class Int64Comparator extends Comparator
 {
     public function accepts($expected, $actual)
     {
-        // Only compare if either value is an Int64
+        // Only compare if either value is an Int64 and the other value is numeric
         return ($expected instanceof Int64 && $this->isComparable($actual))
             || ($actual instanceof Int64 && $this->isComparable($expected));
     }
 
     public function assertEquals($expected, $actual, $delta = 0.0, $canonicalize = false, $ignoreCase = false): void
     {
-        if (PHP_INT_SIZE == 8) {
-            // On 64-bit systems, compare integers directly
-            $expectedValue = (int) $expected;
-            $actualValue = (int) $actual;
-        } else {
-            // On 32-bit systems, compare integers as strings
-            $expectedValue = (string) $expected;
-            $actualValue = (string) $actual;
-        }
-
-        if ($expectedValue === $actualValue) {
+        if ($expected == $actual) {
             return;
         }
 
@@ -54,8 +40,6 @@ class Int64Comparator extends Comparator
 
     private function isComparable($value): bool
     {
-        return $value instanceof Int64 // Int64 instances
-            || is_int($value) // Integer values
-            || (is_string($value) && is_numeric($value)); // Numeric strings (is_numeric accepts floats)
+        return $value instanceof Int64 || is_numeric($value);
     }
 }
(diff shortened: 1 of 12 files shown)


tests/Comparator/Int64Comparator.php
```diff
@@ -0,0 +1,61 @@
...
+                $this->exporter->export($expected)
+            )
+        );
+    }
+
+    private function isComparable($value): bool
+    {
+        return $value instanceof Int64 // Int64 instances
+            || is_int($value) // Integer values
+            || (is_string($value) && is_numeric($value)); // Numeric strings (is_numeric accepts floats)
```
GromNaN, line 59: Do you use parentheses to explicit? They are not necessary since [`&&` precede `||`](https://www.php.net/manual/en/language.operators.precedence.php).
```suggestion
            || is_string($value) && is_numeric($value); // Numeric strings (is_numeric accepts floats)
```

tests/bootstrap.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+use MongoDB\Tests\Comparator\Int64Comparator;
+use SebastianBergmann\Comparator\Factory as ComparatorFactory;
+
+require __DIR__ . '/../vendor/autoload.php';
+
+// Register comparator factories
```
GromNaN, line 8: I don't get this comment. Copilot?

tests/Comparator/Int64Comparator.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+namespace MongoDB\Tests\Comparator;
```
GromNaN, line 3: This class could be in the `src` directory, with namespace `MongoDB\Test\Comparator`. The need to compare Int64 also exist for the developers using this lib.

GromNaN (commented):
Great idea to use this comparator. It's not even intrusive. I think it should be made available to the user of the library.

GromNaN (approved):
Can be merged as this, once updated, to unlock the CI of other PRs.

tests/Comparator/Int64Comparator.php
```diff
@@ -0,0 +1,61 @@
...
+            $actualValue = (int) $actual;
+        } else {
+            // On 32-bit systems, compare integers as strings
+            $expectedValue = (string) $expected;
+            $actualValue = (string) $actual;
+        }
+
+        if ($expectedValue === $actualValue) {
+            return;
+        }
```
jmikola, line 39: Is this necessary given how comparisons for Int64 work? Would it be preferable to use `==` here, or might that cause problems on 32-bit platforms?

[bson-int64-compare-001.phpt](https://github.com/mongodb/mongo-php-driver/blob/master/tests/bson/bson-int64-compare-001.phpt) doesn't test comparisons against other types, but I think this is handled via the cast handlers (also added in mongodb/mongo-php-driver/commit/b1ff48a00c831cd1f18d6ab9b94ec2c61002162d for [PHPC-2214](https://jira.mongodb.org/browse/PHPC-2214)). Consider:

```
> 5 == new MongoDB\BSON\Int64(5)
= true

> 5.2 == new MongoDB\BSON\Int64(5)
= false

> '5.0' == new MongoDB\BSON\Int64(5)
= true

> 5.0 == new MongoDB\BSON\Int64(5)
= true

> '5.2' == new MongoDB\BSON\Int64(5)
= false
```

My concern for 32-bit platforms is rooted in the string casting. Consider:

```
> (string) new MongoDB\BSON\Int64(5) === (string) '5.0'
= false

> (int) new MongoDB\BSON\Int64(5) === (int) '5.0'
= true
```

`assertEquals()` should behave consistently on any platform.

This should also provide you with a bunch of new regression tests with fractional floats and float-like string. And if there's some way to test the 32-bit branch on 64-bit platforms, that might also help. As-is, I assume we have no actual test coverage of that code path.

tests/bootstrap.php
alcaeus: Whoops. I've corrected the comment.

tests/Comparator/Int64Comparator.php
alcaeus: > Do you use parentheses to explicit?

What Jeremy said - it reduces the cognitive load on the reader (and writer in this case since I can't always remember operator precedence either)

Either way, this check has now been simplified.

alcaeus:
Turns out I've completely overthought this. I've added more comparison tests in PHPC (see https://github.com/mongodb/mongo-php-driver/pull/1439), and the only reason we're seeing these test failures is the order of comparators in PHPUnit:
* when comparing a string with an object that supports `__toString` (which the Int64 class does), `ScalarComparator` is used,
* when comparing an Int64 instance with any other value, a `TypeComparator` is used which obviously fails.

In this case, our comparator will become active as long as we're comparing an Int64 instance with another Int64 instance or numeric value. The rest is handled correctly by PHP itself using casts and type comparisons.

I've deliberately used `2**33` as test values here, as it's a 64-bit integer that also fits into a double without loss of precision, so we can test comparison with floats.

tests/Comparator/Int64Comparator.php
GromNaN: Understood, exposing such PHPUnit extension requires that we think about the supported PHPUnit versions BC promise. If user feels the need for this comparator, please raise an issue or add a comment to the new task PHPLIB-1165.

tests/Comparator/Int64ComparatorTest.php
```diff
@@ -0,0 +1,211 @@
...
+            'actualValue' => new Int64(123),
+        ];
+
+        yield 'Expects non-numeric string, Actual Int64' => [
+            'expectedResult' => false,
+            'expectedValue' => 'foo',
+            'actualValue' => new Int64(123),
+        ];
+
+        yield 'Expects float, Actual Float' => [
```
jmikola, line 86:
```suggestion
        yield 'Expects float, Actual float' => [
```

tests/Comparator/Int64ComparatorTest.php
```diff
@@ -0,0 +1,211 @@
...
+    public static function provideMatchingAssertions(): Generator
+    {
+        yield 'Expected Int64, Actual Int64' => [
+            'expected' => new Int64(8589934592),
+            'actual' => new Int64(8589934592),
+        ];
+
+        yield 'Expected Int64, Actual int' => [
+            'expected' => new Int64(8589934592),
+            'actual' => 8589934592,
```
jmikola, line 117: I realize we don't test PHPLIB on 32-bit platforms, but is this integer literal problematic or would PHP interpret it as a float?

jmikola (approved):
One suggestion and a question, but LGTM.

jmikola (commented):
Two more suggested changes:

`tests/SpecTests/ClientSideEncryptionSpecTest.php` defines a private [`createInt64()`](https://github.com/mongodb/mongo-php-library/blob/eb2a813abdc1e00a46c7cbf703ed9cbce490d084/tests/SpecTests/ClientSideEncryptionSpecTest.php#L1865) method, which is used in a few places.

`tests/SpecTests/ClientSideEncryption/FunctionalTestCase.php` also defines a protected [`createInt64()`](https://github.com/mongodb/mongo-php-library/blob/eb2a813abdc1e00a46c7cbf703ed9cbce490d084/tests/SpecTests/ClientSideEncryption/FunctionalTestCase.php#LL72C56-L72C56) method, which is used in [Prose22_RangeExplicitEncryptionTest](https://github.com/mongodb/mongo-php-library/blob/eb2a813abdc1e00a46c7cbf703ed9cbce490d084/tests/SpecTests/ClientSideEncryption/Prose22_RangeExplicitEncryptionTest.php#L171).

Both of those can be factored out in this PR.

PR #1115, opened 2023-06-22
Refactor psalm and phpcs workflows to get static names

alcaeus:
Using a matrix strategy results in changing names whenever we change the driver or PHP version, requiring an update to protected branch settings. Moving this configuration to env variables prevents this issue.


GromNaN (approved):
LGTM. Indeed it was a 1-1 matrix 🥲.

PR #1117, opened 2023-06-22
PHPLIB-1137: Reject PackedArrays when expecting documents

alcaeus:
PHPLIB-1137

This PR adds a new `is_document` function that is used to validate input options. While we accept `array|object` for documents, a `PackedArray` instance is one of those objects where we are absolutely sure that it is not a document.

This PR also includes a commit to refactor most data providers that provide options to operation classes. This was done to give the individual data sets a name, making it easier to spot failing tests. Due to the large amount of unrelated changes in that first commit, I suggest reviewing the commits separately.

### DIFF
--- a/src/Command/ListCollections.php
+++ b/src/Command/ListCollections.php
@@ -79,7 +79,7 @@ class ListCollections implements Executable
         }
 
         if (isset($options['filter']) && ! is_document($options['filter'])) {
-            throw InvalidArgumentException::invalidType('"filter" option', $options['filter'], 'document');
+            throw InvalidArgumentException::expectedDocumentType('"filter" option', $options['filter']);
         }
 
         if (isset($options['maxTimeMS']) && ! is_integer($options['maxTimeMS'])) {
--- a/src/Command/ListCollections.php
+++ b/src/Command/ListCollections.php
@@ -25,10 +25,9 @@ use MongoDB\Exception\InvalidArgumentException;
 use MongoDB\Model\CachingIterator;
 use MongoDB\Operation\Executable;
 
-use function is_array;
 use function is_bool;
 use function is_integer;
-use function is_object;
+use function MongoDB\is_document;
 
 /**
  * Wrapper for the listCollections command.
@@ -79,8 +78,8 @@ class ListCollections implements Executable
             throw InvalidArgumentException::invalidType('"authorizedCollections" option', $options['authorizedCollections'], 'boolean');
         }
 
-        if (isset($options['filter']) && ! is_array($options['filter']) && ! is_object($options['filter'])) {
-            throw InvalidArgumentException::invalidType('"filter" option', $options['filter'], 'array or object');
+        if (isset($options['filter']) && ! is_document($options['filter'])) {
+            throw InvalidArgumentException::invalidType('"filter" option', $options['filter'], 'document');
         }
 
         if (isset($options['maxTimeMS']) && ! is_integer($options['maxTimeMS'])) {
--- a/src/functions.php
+++ b/src/functions.php
@@ -87,8 +87,8 @@ function all_servers_support_write_stage_on_secondary(array $servers): bool
  */
 function apply_type_map_to_document($document, array $typeMap)
 {
-    if (! is_array($document) && ! is_object($document)) {
-        throw InvalidArgumentException::invalidType('$document', $document, 'array or object');
+    if (! is_document($document)) {
+        throw InvalidArgumentException::invalidType('$document', $document, 'document');
     }
 
     return toPHP(fromPHP($document), $typeMap);
@@ -133,7 +133,7 @@ function document_to_array($document): array
     }
 
     if (! is_array($document)) {
-        throw InvalidArgumentException::invalidType('$document', $document, 'array or object');
+        throw InvalidArgumentException::invalidType('$document', $document, 'document');
     }
 
     return $document;
@@ -187,6 +187,19 @@ function get_encrypted_fields_from_server(string $databaseName, string $collecti
 }
 
 /**
+ * Returns whether a given value is a valid document.
+ *
+ * This method returns true for any array or object, but specifically excludes
+ * BSON PackedArray instances
+ *
+ * @param mixed $document
+ */
+function is_document($document): bool
+{
+    return is_array($document) || (is_object($document) && ! $document instanceof PackedArray);
+}
+
+/**
  * Return whether the first key in the document starts with a "$" character.
  *
  * This is used for validating aggregation pipeline stages and differentiating
@@ -200,6 +213,10 @@ function get_encrypted_fields_from_server(string $databaseName, string $collecti
  */
 function is_first_key_operator($document): bool
 {
+    if ($document instanceof PackedArray) {
+        return false;
+    }
+
     $document = document_to_array($document);
 
     $firstKey = array_key_first($document);
@@ -266,7 +283,7 @@ function is_pipeline($pipeline, bool $allowEmpty = false): bool
     }
 
     foreach ($pipeline as $stage) {
-        if (! is_array($stage) && ! is_object($stage)) {
+        if (! is_document($stage)) {
             return false;
         }
 
--- a/src/functions.php
+++ b/src/functions.php
@@ -88,7 +88,7 @@ function all_servers_support_write_stage_on_secondary(array $servers): bool
 function apply_type_map_to_document($document, array $typeMap)
 {
     if (! is_document($document)) {
-        throw InvalidArgumentException::invalidType('$document', $document, 'document');
+        throw InvalidArgumentException::expectedDocumentType('$document', $document);
     }
 
     return toPHP(fromPHP($document), $typeMap);
@@ -133,7 +133,7 @@ function document_to_array($document): array
     }
 
     if (! is_array($document)) {
-        throw InvalidArgumentException::invalidType('$document', $document, 'document');
+        throw InvalidArgumentException::expectedDocumentType('$document', $document);
     }
 
     return $document;
--- a/tests/Collection/CollectionFunctionalTest.php
+++ b/tests/Collection/CollectionFunctionalTest.php
@@ -61,27 +61,14 @@ class CollectionFunctionalTest extends FunctionalTestCase
         new Collection($this->manager, $this->getDatabaseName(), $this->getCollectionName(), $options);
     }
 
-    public function provideInvalidConstructorOptions()
+    public function provideInvalidConstructorOptions(): void
     {
-        $options = [];
-
-        foreach ($this->getInvalidReadConcernValues() as $value) {
-            $options[][] = ['readConcern' => $value];
-        }
-
-        foreach ($this->getInvalidReadPreferenceValues() as $value) {
-            $options[][] = ['readPreference' => $value];
-        }
-
-        foreach ($this->getInvalidArrayValues() as $value) {
-            $options[][] = ['typeMap' => $value];
-        }
-
-        foreach ($this->getInvalidWriteConcernValues() as $value) {
-            $options[][] = ['writeConcern' => $value];
-        }
-
-        return $options;
+        $this->createOptionDataProvider([
+            'readConcern' => $this->getInvalidReadConcernValues(),
+            'readPreference' => $this->getInvalidReadPreferenceValues(),
+            'typeMap' => $this->getInvalidArrayValues(),
+            'writeConcern' => $this->getInvalidWriteConcernValues(),
+        ]);
     }
 
     public function testGetManager(): void
--- a/tests/Collection/CollectionFunctionalTest.php
+++ b/tests/Collection/CollectionFunctionalTest.php
@@ -61,9 +61,9 @@ class CollectionFunctionalTest extends FunctionalTestCase
         new Collection($this->manager, $this->getDatabaseName(), $this->getCollectionName(), $options);
     }
 
-    public function provideInvalidConstructorOptions(): void
+    public function provideInvalidConstructorOptions(): array
     {
-        $this->createOptionDataProvider([
+        return $this->createOptionDataProvider([
             'readConcern' => $this->getInvalidReadConcernValues(),
             'readPreference' => $this->getInvalidReadPreferenceValues(),
             'typeMap' => $this->getInvalidArrayValues(),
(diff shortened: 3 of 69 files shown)


src/functions.php
```diff
@@ -186,6 +186,17 @@ function get_encrypted_fields_from_server(string $databaseName, string $collecti
...
 
+/**
+ * Returns whether a given document is a valid document.
+ *
+ * This method accepts any array and object, but specifically excludes BSON
+ * PackedArray instances
+ */
+function is_document($document): bool
+{
+    return is_array($document) || (is_object($document) && ! $document instanceof PackedArray);
```
jmikola, line 197: Noting that we don't bother checking if array is a list, since APIs that take arrays as documents generally use an object cast to ensure proper BSON encoding as a document type.

tests/Operation/BulkWriteTest.php
```diff
@@ -57,7 +57,7 @@ public function testInsertOneDocumentArgumentMissing(): void
     public function testInsertOneDocumentArgumentTypeCheck($document): void
     {
         $this->expectException(InvalidArgumentException::class);
-        $this->expectExceptionMessageMatches('/Expected \$operations\[0\]\["insertOne"\]\[0\] to have type "array or object" but found "[\w ]+"/');
+        $this->expectExceptionMessageMatches('/Expected \$operations\[0\]\["insertOne"\]\[0\] to have type "document" but found ".+"/');
```
jmikola, line 60: I don't recall why we accepted a space in the original pattern.

tests/Operation/BulkWriteTest.php
```diff
@@ -192,7 +192,7 @@ public function testReplaceOneReplacementArgumentProhibitsUpdateDocument($replac
     public function testReplaceOneReplacementArgumentProhibitsUpdatePipeline($replacement): void
     {
         $this->expectException(InvalidArgumentException::class);
-        $this->expectExceptionMessage('$operations[0]["replaceOne"][1] is an update pipeline');
+        $this->expectExceptionMessageMatches('#(\$operations\[0\]\["replaceOne"\]\[1\] is an update pipeline)|(\$operations\[0\]\["replaceOne"\]\[1\] to have type "document")#');
```
jmikola, line 195: I noted the `replaceOne` validation has three possible exceptions in order:

 * not a document
 * first key is an operator
 * is a pipeline (with `allowEmpty=true`)

Is the "update pipeline" still reachable because `is_document()` simply returns `true` for PHP arrays? What's the data set that triggers "not a document" to be reached here?

tests/Operation/BulkWriteTest.php
```diff
@@ -267,7 +267,7 @@ public function testUpdateManyUpdateArgumentMissing(): void
     public function testUpdateManyUpdateArgumentTypeCheck($update): void
     {
         $this->expectException(InvalidArgumentException::class);
-        $this->expectExceptionMessageMatches('/Expected \$operations\[0\]\["updateMany"\]\[1\] to have type "array or object" but found "[\w ]+"/');
+        $this->expectExceptionMessageMatches('/(non-empty pipeline for \$operations\[0\]\["updateMany"\]\[1\])|(\$operations\[0\]\["updateMany"\]\[1\] to have type "array or object" but found ".+")/');
```
jmikola, line 270: I notice there's still an "array or object" check for `updateMany` and `updateOne`. Does it make sense to combine that with the `! is_first_key_operator && ! is_pipeline` check that appears later?

```
is_document && ! is_first_key_operator && ! is_pipeline
```

Or might that exclude PackedArray? Aggregation operations require an actual array for their pipelines, so I think it's just update and findAndModify ops that might subtlety allow a PackedArray because they currently accept `array|object`.

tests/Operation/BulkWriteTest.php
```diff
@@ -360,7 +360,7 @@ public function testUpdateOneUpdateArgumentMissing(): void
     public function testUpdateOneUpdateArgumentTypeCheck($update): void
     {
         $this->expectException(InvalidArgumentException::class);
-        $this->expectExceptionMessageMatches('/Expected \$operations\[0\]\["updateOne"\]\[1\] to have type "array or object" but found "[\w ]+"/');
+        $this->expectExceptionMessageMatches('/(non-empty pipeline for \$operations\[0\]\["updateOne"\]\[1\])|(\$operations\[0\]\["updateOne"\]\[1\] to have type "array or object" but found ".+")/');
```
jmikola, line 363: Relates to my earlier question above about `updateMany` and `updateOne`.

tests/Operation/WatchTest.php
```diff
@@ -50,6 +50,7 @@ public function provideInvalidConstructorOptions()
             'readPreference' => $this->getInvalidReadPreferenceValues(true),
             'resumeAfter' => $this->getInvalidDocumentValues(),
             'session' => $this->getInvalidSessionValues(),
+            'startAfter' => $this->getInvalidDocumentValues(),
```
jmikola, line 53: Good catch.

tests/TestCase.php
```diff
@@ -139,6 +140,11 @@ public function provideInvalidIntegerValues()
         return $this->wrapValuesForDataProvider($this->getInvalidIntegerValues());
     }
 
+    public function provideInvalidUpdateValues()
```
jmikola, line 143: Is this only used for Operation classes? If so, you can consider moving it to `MongoDB\Tests\Operation\TestCase`, which is where I put other providers.

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -61,27 +61,14 @@ public function testConstructorOptionTypeChecks(array $options): void
         new Collection($this->manager, $this->getDatabaseName(), $this->getCollectionName(), $options);
     }
 
-    public function provideInvalidConstructorOptions()
+    public function provideInvalidConstructorOptions(): void
```
jmikola, line 64: Why does this have a `void` return type? It looks like you're discarding the return value from `createOptionDataProvider()`.

tests/Command/ListCollectionsTest.php
```diff
@@ -15,26 +15,13 @@ public function testConstructorOptionTypeChecks(array $options): void
         new ListCollections($this->getDatabaseName(), $options);
     }
 
-    public function provideInvalidConstructorOptions()
+    public function provideInvalidConstructorOptions(): void
```
jmikola, line 18: Likewise here. Why `void`?

tests/TestCase.php
```diff
@@ -154,6 +156,25 @@ protected function assertDeprecated(callable $execution): void
...
+
+        // Loop through each option and get possible values
+        foreach ($options as $option => $values) {
+            // Create an option array with a named data provider for each value
+            foreach ($values as $key => $value) {
+                if (is_int($key)) {
+                    $key = get_debug_type($value);
+                }
+
+                $data[$option . '_' . $key] = [[$option => $value]];
```
jmikola, line 177: Does this potentially break if a data provider includes two values with the same type? I suppose PHPUnit would raise an error for that case, so we can leave as-is and address it if it ever becomes an issue down the line.

tests/TestCase.php
```diff
@@ -154,6 +156,25 @@ protected function assertDeprecated(callable $execution): void
         $this->assertCount(1, $errors);
     }
 
+    protected function createOptionDataProvider(array $options): array
+    {
+        $data = [];
+
+        // Loop through each option and get possible values
```
jmikola, line 169: This comment doesn't add much value. Is copilot the new copypasta?

alcaeus: The method was more complex originally (it involved a couple of closures), so I decided to add some comments. Then I removed the closure and made this easier but didn't remove the comment.

The only Copilot generated code in this PR is the repetitive `'<option>' => $this->getSomeValues(),` lines 😅

tests/Operation/BulkWriteTest.php
alcaeus: "Update pipeline" is indeed still reachable because of PHP arrays. Since some of the update pipelines are provided as `PackedArray` instances, these will trigger the "not a document" check before getting to the pipeline checks later on.

tests/Operation/BulkWriteTest.php
alcaeus: This indeed allows update pipelines to be passed as `PackedArray` instances. However, giving this some thought, the following options are valid:
- document with update operator as first key
- update pipeline (either as `PackedArray`, array, or `Serializable` that serialises to an array

With that in mind, we can refactor this to throw on this condition:
```
(! is_document || ! is_first_key_operator) && ! is_pipeline
```

I've made the change as the expected type (`array|object`) is less useful than the info that the operation requires an update document or update pipeline.

tests/Collection/CollectionFunctionalTest.php
alcaeus: Good catch - not sure how that slipped through the cracks. I've updated this and the other occurrence and double-checked that I didn't make the same mistake more often.

tests/TestCase.php
alcaeus: PHPUnit wouldn't raise for that case, but we'd silently ignore tests. Since that is extremely hard to detect and can cause significant problems, I've changed this to use `<option>_<key>_<type>` as data set name for numeric keys, and `<option>_<key>` for string keys. That should remove all possible duplicates.

tests/TestCase.php
alcaeus: Ah, good point. I've moved this and `getInvalidUpdateValues` for consistency.

This also provides a nice refactoring opportunity down the line once we make all our data providers static (which is required in later PHPUnit versions): we can offload these data providers to separate classes or traits and import them as necessary.

tests/Operation/BulkWriteTest.php
alcaeus: "I'm sure it seemed like a good idea at the time" 😅

tests/Model/ChangeStreamIteratorTest.php
```diff
@@ -67,7 +67,7 @@ public function testPostBatchResumeTokenArgumentTypeCheck($postBatchResumeToken)
 
     public function provideInvalidObjectValues()
     {
-        return $this->wrapValuesForDataProvider(array_merge($this->getInvalidDocumentValues(), [[]]));
```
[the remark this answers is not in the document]
alcaeus: Fixed psalm errors in the respective commits.

src/functions.php
```diff
@@ -186,6 +186,17 @@ function get_encrypted_fields_from_server(string $databaseName, string $collecti
     return null;
 }
 
+/**
+ * Returns whether a given document is a valid document.
+ *
+ * This method accepts any array and object, but specifically excludes BSON
+ * PackedArray instances
+ */
```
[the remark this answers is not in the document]
alcaeus: Missed that, along with the parameter type hint. Added both.

src/Command/ListCollections.php
```diff
@@ -79,8 +78,8 @@ public function __construct(string $databaseName, array $options = [])
             throw InvalidArgumentException::invalidType('"authorizedCollections" option', $options['authorizedCollections'], 'boolean');
         }
 
-        if (isset($options['filter']) && ! is_array($options['filter']) && ! is_object($options['filter'])) {
-            throw InvalidArgumentException::invalidType('"filter" option', $options['filter'], 'array or object');
+        if (isset($options['filter']) && ! is_document($options['filter'])) {
+            throw InvalidArgumentException::invalidType('"filter" option', $options['filter'], 'document');
```
GromNaN, line 82: "document" is not a native php type. The term should be defined somewhere. Maybe update the message in `MongoDB\Exception\InvalidArgumentException`.

```diff
- Expected %s to have type "document" but found "string"
+ Expected %s to have type "document" (array or object) but found "string"
```

src/functions.php
GromNaN: Ok, but can a list be a valid value for the commands? Can `0` be used as a field name?

If not, it may help developers if they're notified as soon as possible that the value they've passed is a list of documents, when a single document is expected.

GromNaN (commented):
It's great to have factored this recurring code pattern.

src/functions.php
alcaeus: > Ok, but can a list be a valid value for the commands?

Strictly speaking, a list is not a valid document, even though it has been treated as such in PHPLIB/PHPC.

> Can `0` be used as a field name?

Yes. Arrays are serialised like objects in BSON, but denoted through a special type. Likewise, inserting a structure like `[0 => 'foo', 2 => 'bar']` would always yield an object with `"0"` and `"2"` as keys (since the value isn't a list).

> If not, it may help developers if they're notified as soon as possible that the value they've passed is a list of documents, when a single document is expected.

I agree, which is why we're prohibiting `PackedArray`, as we're always sure that it's an array, not an object. For PHP arrays however this isn't as simple and sometimes performance sensitive, which is why it wasn't added before (I'll leave it to @jmikola to confirm this). There is one more difference: with a structure like `[0 => 'foo', 1 => 'bar']`, the trained eye sees an array, but it can easily be used as an object. A `PackedArray` is by definition always an array, which is why we're prohibiting it.

src/Command/ListCollections.php
alcaeus: Ah, yes. I was trying to avoid "array or object" since that would include all objects including `PackedArray`. I can add it if we want to glance over the packed array case. What do you think @jmikola?

jmikola: I thought I left a comment about my concern with using "document" as a type since we use actual PHP types everywhere else. I can't find that so perhaps I deleted the draft before submitting my last review.

In any event, I agree that something like `"document" (array or object)` would be preferable. We don't need to call out PackedArray specifically, as anyone using that should understand that it's not a document -- and if it leads to a bug report it'd be a nice opportunity to educate someone :)

jmikola: If you pass `document (array or object)` to the `invalidType()` factory method I suppose it's going to be wrapped in double quotes. Perhaps you just want to pass `document` and let the factory method deal with formatting -- but if that's a bit hairy (especially with the existing logic to handle arrays) then maybe just let it get quoted as-is.

GromNaN: > Perhaps you just want to pass document and let the factory method deal with formatting

jmikola (approved):
LGTM once you add "array or object" to the "document" type exception message.

tests/TestCase.php
alcaeus: Yes, static data providers have been allowed for quite some time but have only become required as of PHPUnit 10. We could make that change already.

src/Command/ListCollections.php
alcaeus: I've created a new `expectedDocumentType` factory for this case - it seemed a more sensible choice than messing around with the formatting.

PR #1122, opened 2023-07-03
PHPLIB-1176: Various improvements for In-Use Encryption tutorial

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1176

Note: this is rebased on https://github.com/mongodb/mongo-php-library/pull/1123, which fixes the docs build.

Also, I'll plan to create a separate PR (or DOCSP ticket) to setup the redirect in [mongodb/docs-php-library](https://github.com/mongodb/docs-php-library/blob/master/config/redirects).

### Summary of changes

Adds additional non-enterprise examples from the PyMongo tutorial: "Explicit Encryption with Automatic Decryption" and "Explicit Queryable Encryption".

Examples are now broken out into separate files, which are tested in ExamplesTest.

Renames "Client-Side Encryption" to "In-Use Encryption" (PHPLIB-997). This will warrant adding a redirect from "/tutorial/client-side-encryption/" to "/tutorial/encryption/" in the related docs-php-library project.

Adds docs for crypt_shared and mongocryptd (PHPLIB-985).


alcaeus (approved):
This tutorial is much better than the previous version. Not sure if `literalinclude` is supported on Snooty (whenever we migrate to it), but that's something to worry about once the time comes.

jmikola:
Added test coverage for example script output. Also fixed some issues in the original example scripts:

 * Ensure that we drop the key vault collection going into each script
 * The local schema was previously setting a server-side schema (copypasta). Fixed this to actually set the `schemaMap` option in the client auto encryption opts.

jmikola:
> Not sure if literalinclude is supported on Snooty

See: https://github.com/mongodb/snooty-parser/blob/master/snooty/rstspec.toml#L194

docs/examples/csfle-automatic_encryption-local_schema.php
```diff
@@ -0,0 +1,72 @@
...
+                        ],
+                    ],
+                ],
+            ],
+        ],
+    ],
+]);
+
+// Drop and create the collection.
+$encryptedClient->selectDatabase('test')->dropCollection('coll');
```
GromNaN, line 54: I guess you drop the collection in case the script runs several time on the same db, that should be explained.

docs/examples/csfle-automatic_encryption-local_schema.php
```diff
@@ -0,0 +1,72 @@
...
+$encryptedCollection->insertOne([
+    '_id' => 1,
+    'encryptedField' => 'mySecret',
+]);
+
+print_r($encryptedCollection->findOne(['_id' => 1]));
+
+/* Using the client configured without encryption, find the same document and
+ * observe that the field is not automatically decrypted. Additionally, the JSON
+ * schema will prohibit inserting a document with an unencrypted field value. */
```
GromNaN, line 69: This last sentence is not linked to a code example. If an error is throw, you may comment the failing code.

jmikola:
@alcaeus, @GromNaN: Feel free to take another pass on this (or opt out). I fixed up the test cases since Andreas' last review in order to get CI passing. Also rebased after merging https://github.com/mongodb/mongo-php-library/pull/1123.

docs/examples/csfle-explicit_encryption.php
```diff
@@ -0,0 +1,48 @@
...
+$collection = $client->selectCollection('test', 'coll');
+$collection->drop();
+
+// Insert a document with a manually encrypted field
+$encryptedValue = $clientEncryption->encrypt('mySecret', [
+    'algorithm' => ClientEncryption::AEAD_AES_256_CBC_HMAC_SHA_512_DETERMINISTIC,
+    'keyId' => $keyId,
+]);
+
+$collection->insertOne(['encryptedField' => $encryptedValue]);
```
GromNaN, line 40: Reading this example, I had to scroll up to find which client was used. You may add a comment that we can insert the value with the unencrypted client.

tests/ExamplesTest.php
```diff
@@ -195,6 +193,259 @@ public function testWithTransaction(): void
...
+
+        // Clean up metadata and key vault collections
+        $this->dropCollection('test', 'coll', ['encryptedFields' => []]);
+        $this->dropCollection('encryption', '__keyVault');
+    }
+
+    public static function provideEncryptionExamples(): Generator
+    {
+        $expectedOutput = <<<'OUTPUT'
+MongoDB\BSON\Binary Object
```
GromNaN, line 214: > **Note** this is a good use-case for [snapshot testing](https://github.com/spatie/phpunit-snapshot-assertions).

GromNaN (commented):
Thanks, CSFLE holds no secrets for me now 👍🏻

docs/examples/csfle-automatic_encryption-local_schema.php
```diff
@@ -0,0 +1,72 @@
...
+$clientEncryption = $client->createClientEncryption([
+    'keyVaultNamespace' => 'encryption.__keyVault',
+    'kmsProviders' => [
+        'local' => ['key' => $localKey],
+    ],
+]);
+
+/* Drop the key vault collection and create an encryption key. Alternatively,
+ * this key ID could be read from a configuration file. */
+$client->selectCollection('encryption', '__keyVault')->drop();
```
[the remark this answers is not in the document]
jmikola: I'm doing this for consistency in all of the examples, but it's technically only necessary in examples where we create a key with an alternate name.

In practice, a key vault should also have a unique index on the `keyAltNames` field (as discussed in [Key Vault Collections](https://www.mongodb.com/docs/manual/core/csfle/fundamentals/keys-key-vaults/#key-vault-collections)). I suppose I can add that to the `key_alt_name.php.php` file.

docs/examples/csfle-automatic_encryption-local_schema.php
jmikola: I can revise these comments to explain that I'm doing so to make the scripts idempotent (not sure if that's the correct term to use here, but certainly able to be run multiple times w/o error).

docs/examples/csfle-automatic_encryption-local_schema.php
jmikola: I'll add a code segment that triggers and catches the exception, and update the example test accordingly.

docs/examples/csfle-explicit_encryption.php
jmikola: I'll rename `$collection` in this example to `$unencryptedCollection`, as is done in `queryable_encryption-automatic.php`.

`csfle-explicit_encryption.php` and `csfle-explicit_encryption_automatic_decryption.php` only create a single client, so there is no ambiguity there.

tests/ExamplesTest.php
jmikola: I read through the README but didn't see anything that would apply to these tests since we rely on pattern matching (akin to phpt [EXPECTF](https://qa.php.net/expectf_details.php) codes).

`csfle-explicit_encryption_automatic_decryption.php` is the only one that has entirely static output ("mySecret").

docs/examples/csfle-automatic_encryption-local_schema.php
GromNaN: You may group all the cleanup operations so that the rest of the example is simpler and can be copy-pasted.

Or do the cleanup at the end of the example, so in reenforce the understanding on what have been created and clean the server when running tests.

docs/examples/csfle-explicit_encryption.php
GromNaN: > `csfle-explicit_encryption.php` and `csfle-explicit_encryption_automatic_decryption.php` only create a single client, so there is no ambiguity there.

In facts, there is 2 variables: `$client` and `$clientEncryption`. Both are used.

tests/ExamplesTest.php
GromNaN: Indeed, I hadn't noticed the use of placeholders.

docs/examples/csfle-explicit_encryption.php
jmikola: Those scripts have one Client and ClientEncryption objects, which differs from other scripts that create multiple clients. In the scripts with multiple Client objects, I've revised the examples to prefix the collection variable with "unencrypted" or "encrypted" accordingly.

docs/examples/csfle-automatic_encryption-local_schema.php
jmikola: Exceptions added to examples.

docs/examples/csfle-automatic_encryption-local_schema.php
jmikola: The setup operations depend on client objects that are created in the context of the script, so I don't think there's a good way to segregate that. And for some scripts, collection creation uses encryption options.

I purposely omitted all cleanup operations from the scripts, as I can handle that ExamplesTest.php. The drop commands are really setup operations to ensure we're starting with a fresh slate.

I'm amenable to revisiting this later if we end up adding more inline code examples (much of the PHPLIB docs omit examples for API reference), but I'll note that our docs example in the MongoDB manual (DocumentationExamplesTest.php) don't demonstrate setup/cleanup at all -- which makes them a bit less useful for users IMO.

PR #1124, opened 2023-07-04
PHPLIB-1098: Test on supported platforms

alcaeus:
PHPLIB-1098

This PR updates the build pipeline to test on supported platforms:
* Add Ubuntu 22.04 and 20.04
* Add RHEL 9.0, 8.2, 8.1, and 7.6

The main tests are currently disabled for RHEL 7.6 and Ubuntu 20.04. Ubuntu 20.04 has issues with the "Start KMS servers" task due to dependency conflicts that I'll have to investigate separately. RHEL 7.6 has issues installing the extension via PECL as it fails with an SSL error when attempting to connect to pecl.php.net. I suspect a certificate problem but need to investigate further.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -175,7 +175,7 @@ functions:
       params:
         script: |
           ${PREPARE_SHELL}
-          MONGODB_VERSION=${MONGODB_VERSION} ORCHESTRATION_FILE=${ORCHESTRATION_FILE} TOPOLOGY=${TOPOLOGY} AUTH=${AUTH} SSL=${SSL} STORAGE_ENGINE=${STORAGE_ENGINE} LOAD_BALANCER=${LOAD_BALANCER} REQUIRE_API_VERSION=${REQUIRE_API_VERSION} sh ${DRIVERS_TOOLS}/.evergreen/run-orchestration.sh
+          SKIP_LEGACY_SHELL=true MONGODB_VERSION=${MONGODB_VERSION} ORCHESTRATION_FILE=${ORCHESTRATION_FILE} TOPOLOGY=${TOPOLOGY} AUTH=${AUTH} SSL=${SSL} STORAGE_ENGINE=${STORAGE_ENGINE} LOAD_BALANCER=${LOAD_BALANCER} REQUIRE_API_VERSION=${REQUIRE_API_VERSION} sh ${DRIVERS_TOOLS}/.evergreen/run-orchestration.sh
     # run-orchestration generates expansion file with MONGODB_URI and CRYPT_SHARED_LIB_PATH
     - command: expansions.update
       params:
@@ -773,27 +773,35 @@ axes:
   - id: os
     display_name: OS
     values:
+      # Debian
       - id: debian11
         display_name: "Debian 11"
-        run_on: debian11
+        run_on: debian11-small
       - id: debian10
         display_name: "Debian 10"
-        run_on: debian10
+        run_on: debian10-small
       - id: debian92
         display_name: "Debian 9.2"
-        run_on: debian92
-      - id: rhel70
-        display_name: "RHEL 7.0"
-        run_on: rhel70
-      - id: rhel71-power8
-        display_name: "RHEL 7.1 Power 8"
-        run_on: rhel71-power8-build
-      - id: rhel72-zseries
-        display_name: "RHEL 7.2 zSeries"
-        run_on: rhel72-zseries-build
-      - id: ubuntu1804-arm64
-        display_name: "Ubuntu 18.04 ARM64"
-        run_on: ubuntu1804-arm64-test
+        run_on: debian92-small
+
+      # RHEL
+      - id: rhel90
+        display_name: "RHEL 9.0"
+        run_on: rhel90-small
+
+      # Ubuntu LTS
+      - id: ubuntu2204
+        display_name: "Ubuntu 22.04 x64"
+        run_on: ubuntu2204-small
+      - id: ubuntu2204-arm64
+        display_name: "Ubuntu 22.04 ARM64"
+        run_on: ubuntu2204-arm64-small
+      - id: ubuntu2004
+        display_name: "Ubuntu 20.04 x64"
+        run_on: ubuntu2004-small
+      - id: ubuntu2004-arm64
+        display_name: "Ubuntu 20.04 ARM64"
+        run_on: ubuntu2204-arm64-small
 
   - id: topology
     display_name: Topology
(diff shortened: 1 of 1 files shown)


.evergreen/config.yml
```diff
@@ -860,19 +878,38 @@ axes:
...
+      - debian10
+      - rhel90
+      # TODO: RHEL 8.x is not working yet
+      # - rhel82-arm64
+      # - rhel80
+      # TODO: RHEL 7.6 fails as it's not able to connect to pecl.php.net to install the library
+      # - rhel76
+      - ubuntu2204
+      - ubuntu2204-arm64
+      # TODO: KMS servers fail to start on Ubuntu 20.04
```
jmikola, line 895: Noted that we previously weren't testing on _any_ Ubuntu platforms. `ubuntu1804-arm64` was defined in the 'os` matrix but never used.

Do you plan on opening a "Build Failure" ticket for this? If so, let's add a ticket reference here.

I don't feel strongly about it as the OS matrix is more relevant for PHPC than PHPLIB.

.evergreen/config.yml
```diff
@@ -860,19 +878,38 @@ axes:
...
-  matrix_spec: { "os": "debian11", "mongodb-edge-versions": "latest-stable", "php-versions": "*", "driver-versions": "latest-stable" }
+  matrix_spec:
+    os:
+      - debian11
+      - debian10
+      - rhel90
+      # TODO: RHEL 8.x is not working yet
+      # - rhel82-arm64
+      # - rhel80
+      # TODO: RHEL 7.6 fails as it's not able to connect to pecl.php.net to install the library
```
jmikola, line 891: If we intend to resolve this, I'd suggest opening a "Build Failure" ticket and referencing it here.

.evergreen/config.yml
```diff
@@ -860,19 +878,38 @@ axes:
...
-# Test all PHP versions with latest-stable MongoDB and PHPC on Debian
+# Test all PHP versions with latest-stable MongoDB and PHPC, RHEL, and Ubuntu
 - matrix_name: "test-php-versions"
-  matrix_spec: { "os": "debian11", "mongodb-edge-versions": "latest-stable", "php-versions": "*", "driver-versions": "latest-stable" }
+  matrix_spec:
+    os:
+      - debian11
+      - debian10
+      - rhel90
+      # TODO: RHEL 8.x is not working yet
```
jmikola, line 888: What's the issue with RHEL 8.x? I noted that it was [omitted from the corresponding PHPC PR](https://github.com/mongodb/mongo-php-driver/pull/1442/files#diff-2bc841e86ce96b7b422ae203fd8315d0b2a461956cecbe0e096420656fc3fb12R1142). Do we need a ticket to track this?

Also, looking at that PR, I wonder if `rhel80` is even relevant if we're testing `rhel90` here. `rhel80` might be more relevant for PHPC.

`rhel82-arm64` may be relevant as an ARM64 platform (as an alternative to `ubuntu2204-arm64`), but perhaps not necessary for PHPLIB as much as it would be for PHPC.

.evergreen/config.yml
```diff
@@ -860,19 +878,38 @@ axes:
...
     - { "os": "debian11", "mongodb-edge-versions": "latest-stable", "php-versions": "8.2", "driver-versions": "latest-stable" }
+    # Exclude PHP versions older than 8.1 on RHEL 9 and Ubuntu 22.04 (OpenSSL 3 is only supported on PHP 8.1+)
+    - { "os": ["rhel90", "ubuntu2204-arm64", "ubuntu2204"], "php-versions": ["7.2", "7.3", "7.4", "8.0"], "mongodb-edge-versions": "*", "driver-versions": "*" }
   tasks:
     - name: "test-standalone"
     - name: "test-replica_set"
     - name: "test-sharded_cluster"
 
-# Test all topologies and MongoDB versions with latest-stable PHP and PHPC on Debian
+# Test all topologies and MongoDB versions with latest-stable PHP and PHPC on Debian 11
```
jmikola, line 912:
```suggestion
# Test all topologies and MongoDB versions with latest-stable PHP and PHPC on Debian 9.2 or 11 (per server version)
```

Alternatively:

> ...on Debian 9.2 (server <= 5.0) and 11 (6.0+)

.evergreen/config.yml
```diff
@@ -773,27 +773,45 @@ axes:
   - id: os
     display_name: OS
     values:
+      # Debian
```
jmikola, line 776: I like grouping these by OS name :+1:

.evergreen/config.yml
alcaeus: I'm happy to slim down the OS matrix as the platform is more relevant to PHPC than PHPLIB. For PHPLIB, I'd say the important systems are Ubuntu and Debian, as those are were most people run their applications.

An important difference are docker users, where Alpine is a popular choice, but that's not available. I'll note that the standard images (e.g. `php-cli`) are built on Debian, so we have those covered.

.evergreen/config.yml
alcaeus: While I'm not convinced we need to test on RHEL here, this matrix exposed problems installing extensions via PECL due to SSL issues. I'm not sure how relevant that is, but maybe @remicollet can shed some light on how popular installing via PECL is (and maybe even what the source of the certificate errors is).

.evergreen/config.yml
alcaeus: The toolchain was not installed on RHEL 8 hosts, which has now been done. PHPC will test on those, but I'm not convinced we need to test all PHP versions on those platforms.

.evergreen/config.yml
jmikola: FYI, [PHPC-1555](https://jira.mongodb.org/browse/PHPC-1555) is an old tracking ticket to add Alpine, which we had to resolve as "Won't Fix" due to missing support in Evergreen. If we can get Alpine working in GitHub actions then perhaps we can revisit this down the line.

.evergreen/config.yml
jmikola: Do you have details on the SSL error running `pecl install`? He's probably need to see that to provide any insight.

.evergreen/config.yml
jmikola: I concur that we don't need this for PHPLIB. Feel free to remove.

.evergreen/config.yml
alcaeus: Thanks @remicollet for confirming. I'll check with our build team to see what's going on with our build hosts.

.evergreen/config.yml
alcaeus: As per suggestions, I have slimmed down the OS matrix, omitting RHEL 7 and RHEL 8 completely as the platform doesn't have an impact on the library. We're testing on RHEL 9, which also doesn't have the certificate problem I saw on earlier RHEL versions.

jmikola: @alcaeus Will you be creating a ticket for "KMS servers fail to start on Ubuntu 20.04"?

jmikola (approved):
Suggested creating a ticket for the KMS build failure on Ubuntu 20.04, but actual changes in the PR LGTM.

.evergreen/config.yml
alcaeus: > @alcaeus Will you be creating a ticket for "KMS servers fail to start on Ubuntu 20.04"?

I wanted to grab a fresh copy of the error message, but the patch I created for that purpose succeeded: https://spruce.mongodb.com/version/64ad54d157e85a3f891c1afe/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

I'm now running a new patch on all Ubuntu 20.04 platforms - if that succeeds I'll add them in here; otherwise I'll create a ticket to have it looked at.

PR #1126, opened 2023-07-07
PHPLIB-1184: Improve CallbackIterator

alcaeus:
PHPLIB-1184

#1059 contained some improvements to the `CallbackIterator`. These have been extracted here:

* Pass element key to the callback
* Define template for types in CallbackIterator
* Declare the expected type for the callback
* Accept any callable as callback instead of only accepting a `Closure` instance

### DIFF
--- a/src/Model/CallbackIterator.php
+++ b/src/Model/CallbackIterator.php
@@ -17,11 +17,12 @@
 
 namespace MongoDB\Model;
 
-use Closure;
 use Iterator;
 use IteratorIterator;
 use ReturnTypeWillChange;
 use Traversable;
+
+use function call_user_func;
 
 /**
  * Iterator to apply a callback before returning an element
@@ -35,20 +36,20 @@ use Traversable;
  */
 class CallbackIterator implements Iterator
 {
-    /** @var Closure(TValue, TKey): TCallbackValue */
-    private $callback;
+    /** @var callable(TValue, TKey): TCallbackValue */
+    private $callable;
 
     /** @var Iterator<TKey, TValue> */
     private $iterator;
 
     /**
-     * @param Traversable<TKey, TValue>             $traversable
-     * @param Closure(TValue, TKey): TCallbackValue $callback
+     * @param Traversable<TKey, TValue>              $traversable
+     * @param callable(TValue, TKey): TCallbackValue $callable
      */
-    public function __construct(Traversable $traversable, Closure $callback)
+    public function __construct(Traversable $traversable, callable $callable)
     {
         $this->iterator = $traversable instanceof Iterator ? $traversable : new IteratorIterator($traversable);
-        $this->callback = $callback;
+        $this->callable = $callable;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Model/CallbackIterator.php
```diff
@@ -17,44 +17,54 @@
...
+ * @template TValue
+ * @template TCallbackValue
+ * @template-implements Iterator<TKey, TCallbackValue>
  */
 class CallbackIterator implements Iterator
 {
-    /** @var Closure */
-    private $callback;
+    /** @var callable(TValue, TKey): TCallbackValue */
+    private $callable;
```
GromNaN, line 40: The property name can stay `callback`: `callable` is a type, not a purpose.

tests/Model/CallbackIteratorTest.php
```diff
@@ -0,0 +1,70 @@
...
+
+        $this->assertSame(['a' => 2, 'b' => 4, 'c' => 6], iterator_to_array($callbackIterator));
+    }
+
+    public function testWithCallable(): void
+    {
+        $original = ['foo', 'bar', 'baz'];
+
+        $callbackIterator = new CallbackIterator(
+            new ArrayIterator($original),
```
GromNaN, line 59: To cover the `new IteratorIterator($traversable)` in the constructor, you can add 1 case that uses `IteratorAggregate`.

tests/Model/CallbackIteratorTest.php
```diff
@@ -0,0 +1,70 @@
...
+use MongoDB\Model\CallbackIterator;
+use MongoDB\Tests\TestCase;
+
+use function array_keys;
+use function iterator_to_array;
+use function strrev;
+
+class CallbackIteratorTest extends TestCase
+{
+    public function testArrayIteration(): void
```
jmikola, line 15: I'd propose rewriting all of these tests to use the following matrix:

* Callbacks in [various forms](https://www.php.net/manual/en/language.types.callable.php#language.types.callable.passing) that apply the same transformation (`*= 2` seems fine)
  * Closure
  * Object method (e.g. static method on this class)
* Iterables
  * ArrayIterator constructed from an indexed array
  * ArrayIterator constructed from an associative array (to show that keys are preserved)
  * IteratorAggregate for both of the above (per @GromNaN's [suggestion below](https://github.com/mongodb/mongo-php-library/pull/1126/files#r1255751584))

Using a data provider for iterables and expected arrays would probably be most efficient.

GromNaN: @alcaeus I can make the changes if it frees you up.

src/Model/CallbackIterator.php
alcaeus: Good point, updated!

tests/Model/CallbackIteratorTest.php
alcaeus: Ah, yes. I added a test using an `IteratorAggregate` instance and reworked the tests to use data providers for convenience.

PR #1127, opened 2023-07-12
PHPLIB-1008: Test crypt_shared with older server versions

alcaeus:
PHPLIB-1008

This changes CSFLE testing to not only test without crypt_shared, but to also test crypt_shared on all server versions, including those that do not ship with the shared library. For those versions, drivers-evergreen-tools downloads crypt_shared for MongoDB 6.0. This requires us to test on a platform that supports MongoDB 4.2 up to 6.0, hence the usage of RHEL 8.0 for that pipeline.

As an aside, I'm not sure if the previous tests properly tested without crypt_shared. After adding the initial commit to output some basic information about the test environment, I noticed that crypt_shared was still available. The `SKIP_CRYPT_SHARED` variable was incorrectly set in the call to "run tests", but needs to be set when starting mongo-orchestration as that function downloads crypt_shared unless asked not to do so.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -175,7 +175,7 @@ functions:
       params:
         script: |
           ${PREPARE_SHELL}
-          SKIP_LEGACY_SHELL=true MONGODB_VERSION=${MONGODB_VERSION} ORCHESTRATION_FILE=${ORCHESTRATION_FILE} TOPOLOGY=${TOPOLOGY} AUTH=${AUTH} SSL=${SSL} STORAGE_ENGINE=${STORAGE_ENGINE} LOAD_BALANCER=${LOAD_BALANCER} REQUIRE_API_VERSION=${REQUIRE_API_VERSION} sh ${DRIVERS_TOOLS}/.evergreen/run-orchestration.sh
+          SKIP_CRYPT_SHARED=${SKIP_CRYPT_SHARED} SKIP_LEGACY_SHELL=true MONGODB_VERSION=${MONGODB_VERSION} ORCHESTRATION_FILE=${ORCHESTRATION_FILE} TOPOLOGY=${TOPOLOGY} AUTH=${AUTH} SSL=${SSL} STORAGE_ENGINE=${STORAGE_ENGINE} LOAD_BALANCER=${LOAD_BALANCER} REQUIRE_API_VERSION=${REQUIRE_API_VERSION} sh ${DRIVERS_TOOLS}/.evergreen/run-orchestration.sh
     # run-orchestration generates expansion file with MONGODB_URI and CRYPT_SHARED_LIB_PATH
     - command: expansions.update
       params:
@@ -632,7 +632,7 @@ tasks:
             SSL: "yes"
         # Note: "stop load balancer" will be called from "post"
 
-    - name: "test-skip_crypt_shared"
+    - name: "test-csfle-crypt_shared"
       commands:
         - func: "bootstrap mongo-orchestration"
           vars:
@@ -641,7 +641,18 @@ tasks:
         - func: "set aws temp creds"
         - func: "run tests"
           vars:
+            TESTS: "csfle"
+
+    - name: "test-csfle-mongocryptd"
+      commands:
+        - func: "bootstrap mongo-orchestration"
+          vars:
+            TOPOLOGY: "replica_set"
             SKIP_CRYPT_SHARED: "yes"
+        - func: "start kms servers"
+        - func: "set aws temp creds"
+        - func: "run tests"
+          vars:
             TESTS: "csfle"
 
     - name: "test-without_aws_creds"
@@ -788,6 +799,9 @@ axes:
       - id: rhel90
         display_name: "RHEL 9.0"
         run_on: rhel90-small
+      - id: rhel80
+        display_name: "RHEL 8.0"
+        run_on: rhel80-small
 
       # Ubuntu LTS
       - id: ubuntu2204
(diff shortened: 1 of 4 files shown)


.evergreen/config.yml
```diff
@@ -175,7 +175,7 @@ functions:
       params:
         script: |
           ${PREPARE_SHELL}
-          SKIP_LEGACY_SHELL=true MONGODB_VERSION=${MONGODB_VERSION} ORCHESTRATION_FILE=${ORCHESTRATION_FILE} TOPOLOGY=${TOPOLOGY} AUTH=${AUTH} SSL=${SSL} STORAGE_ENGINE=${STORAGE_ENGINE} LOAD_BALANCER=${LOAD_BALANCER} REQUIRE_API_VERSION=${REQUIRE_API_VERSION} sh ${DRIVERS_TOOLS}/.evergreen/run-orchestration.sh
+          SKIP_CRYPT_SHARED=${SKIP_CRYPT_SHARED} SKIP_LEGACY_SHELL=true MONGODB_VERSION=${MONGODB_VERSION} ORCHESTRATION_FILE=${ORCHESTRATION_FILE} TOPOLOGY=${TOPOLOGY} AUTH=${AUTH} SSL=${SSL} STORAGE_ENGINE=${STORAGE_ENGINE} LOAD_BALANCER=${LOAD_BALANCER} REQUIRE_API_VERSION=${REQUIRE_API_VERSION} sh ${DRIVERS_TOOLS}/.evergreen/run-orchestration.sh
```
jmikola, line 178: I just noticed this env var wasn't documented for `run-orchestration.sh` and opened https://github.com/mongodb-labs/drivers-evergreen-tools/pull/328 to fix that.

Also, I noticed that `run-orchestration.sh` suggests `SKIP_CRYPT_SHARED=1` but we use strings. The actual implementation uses `-z` and only checks if it's an empty string, so we're OK to continue using `SKIP_CRYPT_SHARED=yes`. Likewise for other vars like `REQUIRE_API_VERSION`.

.evergreen/config.yml
```diff
@@ -947,14 +961,25 @@ buildvariants:
   tasks:
     - name: "test-loadBalanced"
 
-- matrix_name: "test-csfle-skip_crypt_shared"
-  matrix_spec: { "os": "debian11", "mongodb-versions": "*", "php-edge-versions": "latest-stable", "driver-versions": "latest-stable" }
-  display_name: "CSFLE skip_crypt_shared - ${mongodb-versions}"
+# CSFLE tests (crypt_shared and mongocryptd) are tested on RHEL 7 as it's the only version that supports
```
jmikola, line 964:
```suggestion
# CSFLE tests (crypt_shared and mongocryptd) are tested on RHEL 8 as it's the only version that supports
```

Was this a typo? The PR description said:

> This requires us to test on a platform that supports MongoDB 4.2 up to 6.0, hence the usage of RHEL 8.0 for that pipeline.

tests/bootstrap.php
```diff
@@ -1,9 +1,20 @@
...
+
+/* Ugly workaround for event system changes in PHPUnit
+ * PHPUnit 10 introduces a new event system, and at the same time removes the
+ * event system present in previous versions. This makes it near impossible to
+ * support PHPUnit 8.5 (required for PHP 7.2) and PHPUnit 9 (PHP < 8.1)
+ * alongside PHPUnit 10. For this reason, we use this bootstrap file to print
+ * test configuration summary.
+ * @todo PHP_VERSION_ID >= 80100: Remove this and use an extension
+ */
+TestCase::printConfiguration();
```
jmikola, line 20: This looks like it's only necessary for debugging. How is the "event system" (first I've heard of that term with respect to PHPUnit) relevant to us? AFAIK, we weren't printing any configuration options previously.

.evergreen/config.yml
```diff
@@ -641,7 +641,18 @@ tasks:
...
         - func: "run tests"
           vars:
+            TESTS: "csfle"
+
+    - name: "test-csfle-mongocryptd"
+      commands:
+        - func: "bootstrap mongo-orchestration"
+          vars:
+            TOPOLOGY: "replica_set"
             SKIP_CRYPT_SHARED: "yes"
```
jmikola, line 651: > I'm not sure if the previous tests properly tested without crypt_shared. After adding the initial commit to output some basic information about the test environment, I noticed that crypt_shared was still available. The SKIP_CRYPT_SHARED variable was incorrectly set in the call to "run tests", but needs to be set when starting mongo-orchestration as that function downloads crypt_shared unless asked not to do so.

`SKIP_CRYPT_SHARED` was introduced to `run-tests.sh` in [0fe1a0f](https://github.com/mongodb/mongo-php-library/commit/0fe1a0f0b9ee54ce270980b2651b94fcc58921ff#diff-3954529b61794d5a9a1f941f18a65a3233f21167980e973154ba1aa8324901a5R11) for [PHPLIB-911](https://jira.mongodb.org/browse/PHPLIB-911), but the environment variable was never actually used in that script. Instead `CRYPT_SHARED_LIB_PATH` is always exported.



I think my original intention for PHPLIB-911 was to only export `CRYPT_SHARED_LIB_PATH` to PHPUnit if `SKIP_CRYPT_SHARED=no` (default). But that would only have worked if the downloaded copy of crypt_shared never ends up in a default system path. `run-tests.sh` ultimately has no control over whether one of the download scripts places crypt_shared in a system path (in which case `CRYPT_SHARED_LIB_PATH` isn't even relevant).

This suggests that `SKIP_CRYPT_SHARED` isn't relevant to `run-tests.sh` at all and perhaps is only needed for `download-mongodb.sh` and `download-crypt.sh`, which are called via `run-orchestration.sh` and `serverless/create-instance.sh`, respectively. For context, see [this search for `CRYPT_SHARED_LIB_PATH` in drivers-evergreen-tools](https://github.com/search?q=repo%3Amongodb-labs%2Fdrivers-evergreen-tools%20CRYPT_SHARED_LIB_PATH&type=code).

I propose we remove `SKIP_CRYPT_SHARED` from `run-tests.sh`. `CRYPT_SHARED_LIB_PATH` should remain.

----

Side note: AFAIK, there's no mechanism to completely disable loading of crypt_shared (see: [CSFLE: extraOptions](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#extraoptions)). This may be related to some extra testing-specific options some drivers have (see: [this Slack thread in #drivers-fle](https://mongodb.slack.com/archives/CFPHLTRMK/p1689084650285559)), none of which would be available to PHP anyway as they're private libmongoc APIs.

.evergreen/config.yml
alcaeus: Yes, it was left over from a previous iteration of the changes. I've updated the comment.

tests/bootstrap.php
alcaeus: Unfortunately, there is a hard break with those systems: the existing listener system is removed in PHPUnit 10, but the new event system is only introduced in PHPUnit 10. In order to avoid supporting both versions I decided to opt for this simple variant. From what I've seen, symfony/phpunit-bridge does not have a compatibility layer in place for this change.

Once we use PHPUnit 10 exclusively (IIRC once dropping PHP < 8.1) we can use an extension to print this automatically and remove the workaround.

.evergreen/config.yml
alcaeus: FWIW, that's the reason for the configuration output I added to bootstrap.php. I wanted to see the configuration to make sure it does what we expect, that's how I noticed the above.

I've removed `SKIP_CRYPT_SHARED` from run-tests.sh and its invocations.

jmikola (approved):

PR #1128, opened 2023-07-13
PHPLIB-1118: Drop support for PHP 7.2 and 7.3

alcaeus:
PHPLIB-1118

This drops support for PHP 7.2 and 7.3 and updates code for PHP 7.4. Changes have been made automatically using rector and phpcbf. A few manual fixes were applied (a few property types were fixed to exclude `null`).

### DIFF
--- a/examples/bulk.php
+++ b/examples/bulk.php
@@ -68,7 +68,7 @@ $collection->bulkWrite(
                 ['x' => 10], // Document
             ],
         ],
-    ]
+    ],
 );
 
 $cursor = $collection->find([]);
--- a/examples/persistable.php
+++ b/examples/persistable.php
@@ -16,14 +16,12 @@ require __DIR__ . '/../vendor/autoload.php';
 
 class PersistableEntry implements Persistable
 {
-    /** @var ObjectId */
-    private $id;
+    private ObjectId $id;
 
-    /** @var string */
-    public $name;
+    public string $name;
 
     /** @var array<PersistableEmail> */
-    public $emails = [];
+    public array $emails = [];
 
     public function __construct(string $name)
     {
@@ -65,11 +63,9 @@ class PersistableEntry implements Persistable
 
 class PersistableEmail implements Persistable
 {
-    /** @var string */
-    public $type;
+    public string $type;
 
-    /** @var string */
-    public $address;
+    public string $address;
 
     public function __construct(string $type, string $address)
     {
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -83,17 +83,13 @@ class Collection
 
     private Manager $manager;
 
-    /** @var ReadConcern */
-    private $readConcern;
+    private ReadConcern $readConcern;
 
-    /** @var ReadPreference */
-    private $readPreference;
+    private ReadPreference $readPreference;
 
-    /** @var array */
-    private $typeMap;
+    private array $typeMap;
 
-    /** @var WriteConcern */
-    private $writeConcern;
+    private WriteConcern $writeConcern;
 
     /**
      * Constructs new Collection instance.
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -77,14 +77,11 @@ class Collection
 
     private const WIRE_VERSION_FOR_READ_CONCERN_WITH_WRITE_STAGE = 8;
 
-    /** @var string */
-    private $collectionName;
+    private string $collectionName;
 
-    /** @var string */
-    private $databaseName;
+    private string $databaseName;
 
-    /** @var Manager */
-    private $manager;
+    private Manager $manager;
 
     /** @var ReadConcern */
     private $readConcern;
--- a/src/Database.php
+++ b/src/Database.php
@@ -65,17 +65,13 @@ class Database
 
     private Manager $manager;
 
-    /** @var ReadConcern */
-    private $readConcern;
+    private ReadConcern $readConcern;
 
-    /** @var ReadPreference */
-    private $readPreference;
+    private ReadPreference $readPreference;
 
-    /** @var array */
-    private $typeMap;
+    private array $typeMap;
 
-    /** @var WriteConcern */
-    private $writeConcern;
+    private WriteConcern $writeConcern;
 
     /**
      * Constructs new Database instance.
--- a/src/Database.php
+++ b/src/Database.php
@@ -61,11 +61,9 @@ class Database
 
     private const WIRE_VERSION_FOR_READ_CONCERN_WITH_WRITE_STAGE = 8;
 
-    /** @var string */
-    private $databaseName;
+    private string $databaseName;
 
-    /** @var Manager */
-    private $manager;
+    private Manager $manager;
 
     /** @var ReadConcern */
     private $readConcern;
--- a/src/Model/BSONIterator.php
+++ b/src/Model/BSONIterator.php
@@ -36,23 +36,18 @@ class BSONIterator implements Iterator
 {
     private const BSON_SIZE = 4;
 
-    /** @var string */
-    private $buffer;
+    private string $buffer;
 
-    /** @var integer */
-    private $bufferLength;
+    private int $bufferLength;
 
-    /** @var mixed */
-    private $current;
+    /** @var array|object|null */
+    private $current = null;
 
-    /** @var integer */
-    private $key = 0;
+    private int $key = 0;
 
-    /** @var integer */
-    private $position = 0;
+    private int $position = 0;
 
-    /** @var array */
-    private $options;
+    private array $options;
 
     /**
      * Constructs a BSON Iterator.
--- a/src/Operation/Aggregate.php
+++ b/src/Operation/Aggregate.php
@@ -50,23 +50,17 @@ use function MongoDB\is_pipeline;
  */
 class Aggregate implements Executable, Explainable
 {
-    /** @var string */
-    private $databaseName;
+    private string $databaseName;
 
-    /** @var string|null */
-    private $collectionName;
+    private ?string $collectionName = null;
 
-    /** @var array */
-    private $pipeline;
+    private array $pipeline;
 
-    /** @var array */
-    private $options;
+    private array $options;
 
-    /** @var bool */
-    private $isExplain;
+    private bool $isExplain;
 
-    /** @var bool */
-    private $isWrite;
+    private bool $isWrite;
 
     /**
      * Constructs an aggregate command.
@@ -267,7 +261,7 @@ class Aggregate implements Executable, Explainable
 
         $command = new Command(
             $this->createCommandDocument(),
-            $this->createCommandOptions()
+            $this->createCommandOptions(),
         );
 
         $cursor = $this->executeCommand($server, $command);
--- a/src/Operation/CreateIndexes.php
+++ b/src/Operation/CreateIndexes.php
@@ -45,17 +45,13 @@ class CreateIndexes implements Executable
 {
     private const WIRE_VERSION_FOR_COMMIT_QUORUM = 9;
 
-    /** @var string */
-    private $databaseName;
+    private string $databaseName;
 
-    /** @var string */
-    private $collectionName;
+    private string $collectionName;
 
-    /** @var array */
-    private $indexes = [];
+    private array $indexes = [];
 
-    /** @var array */
-    private $options = [];
+    private array $options = [];
 
     /**
      * Constructs a createIndexes command.
@@ -143,9 +139,10 @@ class CreateIndexes implements Executable
 
         $this->executeCommand($server);
 
-        return array_map(function (IndexInput $index) {
-            return (string) $index;
-        }, $this->indexes);
+        return array_map(
+            fn (IndexInput $index): string => (string) $index,
+            $this->indexes,
+        );
     }
 
     /**
--- a/src/Operation/CreateIndexes.php
+++ b/src/Operation/CreateIndexes.php
@@ -49,6 +49,7 @@ class CreateIndexes implements Executable
 
     private string $collectionName;
 
+    /** @var list<IndexInput> */
     private array $indexes = [];
 
     private array $options = [];
@@ -140,7 +141,7 @@ class CreateIndexes implements Executable
         $this->executeCommand($server);
 
         return array_map(
-            fn (IndexInput $index): string => (string) $index,
+            'strval',
             $this->indexes,
         );
     }
--- a/src/Operation/InsertMany.php
+++ b/src/Operation/InsertMany.php
@@ -43,7 +43,7 @@ class InsertMany implements Executable
 
     private string $collectionName;
 
-    /** @var object[]|array[] */
+    /** @var list<object>|list<array> */
     private array $documents;
 
     private array $options;
--- a/src/Operation/InsertMany.php
+++ b/src/Operation/InsertMany.php
@@ -43,7 +43,7 @@ class InsertMany implements Executable
 
     private string $collectionName;
 
-    /** @var list<object>|list<array> */
+    /** @var list<object|array> */
     private array $documents;
 
     private array $options;
--- a/src/Operation/InsertMany.php
+++ b/src/Operation/InsertMany.php
@@ -39,17 +39,14 @@ use function sprintf;
  */
 class InsertMany implements Executable
 {
-    /** @var string */
-    private $databaseName;
+    private string $databaseName;
 
-    /** @var string */
-    private $collectionName;
+    private string $collectionName;
 
     /** @var object[]|array[] */
-    private $documents;
+    private array $documents;
 
-    /** @var array */
-    private $options;
+    private array $options;
 
     /**
      * Constructs an insert command.
--- a/src/Operation/ListCollectionNames.php
+++ b/src/Operation/ListCollectionNames.php
@@ -32,8 +32,7 @@ use MongoDB\Model\CallbackIterator;
  */
 class ListCollectionNames implements Executable
 {
-    /** @var ListCollectionsCommand */
-    private $listCollections;
+    private ListCollectionsCommand $listCollections;
 
     /**
      * Constructs a listCollections command.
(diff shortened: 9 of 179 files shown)


src/Model/BSONIterator.php
```diff
@@ -36,23 +36,18 @@ class BSONIterator implements Iterator
...
+    private string $buffer;
 
-    /** @var integer */
-    private $bufferLength;
+    private int $bufferLength;
 
-    /** @var mixed */
-    private $current;
+    /** @var array|object|null */
+    private $current = null;
```
alcaeus, line 44: Note that rector keeps changing this to `private ?object $current = null;`, which is incorrect as the return type of `toPHP` depends on the type map provided. @GromNaN is there a way to tell rector about this?

examples/bulk.php
```diff
@@ -68,7 +68,7 @@ function toJSON(object $document): string
                 ['x' => 10], // Document
             ],
         ],
-    ]
+    ],
```
alcaeus, line 71: This change is responsible for most of the diff in this PR: a trailing comma after the last argument in multi-line method calls is now required. This is consistent with the rules already in place for multi-line arrays.

examples/persistable.php
```diff
@@ -16,14 +16,12 @@
 
 class PersistableEntry implements Persistable
 {
-    /** @var ObjectId */
-    private $id;
+    private ObjectId $id;
```
alcaeus, line 19: I went through the various type changes and confirmed they were accurate, but a second look certainly won't hurt.

src/Model/BSONIterator.php
GromNaN: It looks like we are using rector with an outdated version of phpstan stubs. https://github.com/phpstan/phpstan-src/blob/1.10.21/resources/functionMap.php#L6735

GromNaN: To disable the problematic rule, add this to `rector.php`:
```php
    $rectorConfig->skip([
        // ...
        TypedPropertyFromAssignsRector::class => [
            __DIR__ . '/src/Model/BSONIterator.php',
        ],
```

src/Collection.php
```diff
@@ -77,14 +77,11 @@ class Collection
...
-    /** @var string */
-    private $databaseName;
+    private string $databaseName;
 
-    /** @var Manager */
-    private $manager;
+    private Manager $manager;
 
     /** @var ReadConcern */
     private $readConcern;
```
GromNaN, line 87: The type have been missed on this properties.

src/Database.php
```diff
@@ -61,11 +61,9 @@ class Database
...
-    /** @var string */
-    private $databaseName;
+    private string $databaseName;
 
-    /** @var Manager */
-    private $manager;
+    private Manager $manager;
 
     /** @var ReadConcern */
     private $readConcern;
```
GromNaN, line 69: Missing type

src/Operation/CreateIndexes.php
```diff
@@ -143,9 +139,10 @@ public function execute(Server $server)
 
         $this->executeCommand($server);
 
-        return array_map(function (IndexInput $index) {
-            return (string) $index;
-        }, $this->indexes);
+        return array_map(
+            fn (IndexInput $index): string => (string) $index,
```
GromNaN, line 143: This function is `strval`.

src/Operation/ListCollectionNames.php
```diff
@@ -76,9 +75,7 @@ public function execute(Server $server): Iterator
     {
         return new CallbackIterator(
             $this->listCollections->execute($server),
-            function (array $collectionInfo) {
-                return $collectionInfo['name'];
-            }
+            fn (array $collectionInfo): string => (string) $collectionInfo['name']
```
GromNaN, line 78:
```suggestion
            'strval'
```

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -182,8 +184,8 @@ public function testDistinctWithTypeMap(array $typeMap, array $expectedDocuments
                     return 1;
                 }
 
-                $a = json_encode($a);
-                $b = json_encode($b);
+                $a = json_encode($a, JSON_THROW_ON_ERROR);
+                $b = json_encode($b, JSON_THROW_ON_ERROR);
```
GromNaN, line 188: Good improvement. Even if that doesn't change anything for the test.

tests/UnifiedSpecTests/UnifiedTestRunner.php
```diff
@@ -60,26 +61,20 @@ final class UnifiedTestRunner
...
-    /** @var bool */
-    private $allowKillAllSessions = true;
+    private bool $allowKillAllSessions = true;
 
-    /** @var EntityMap */
-    private $entityMap;
+    private ?EntityMap $entityMap = null;
 
     /** @var callable(EntityMap):void */
     private $entityMapObserver;
```
GromNaN, line 73: This property is never set. https://github.com/mongodb/mongo-php-library/pull/1129

GromNaN (approved):
LGTM. With a few optional comments.

examples/persistable.php
GromNaN: Good to me. Some missing property types from `ext-mongodb`, maybe due to outdated stubs.

When types are added to properties, the biggest risk is to have an error when trying to access an uninitialized property.

tests/UnifiedSpecTests/UnifiedTestRunner.php
jmikola: Used by Atlas Testing. See https://github.com/mongodb/mongo-php-library/pull/1129#discussion_r1262872050.

src/Operation/ListCollectionNames.php
jmikola: No objection here unless we wanted to introduce a CollectionInfo type hint to the callable instead.

src/Operation/Watch.php
```diff
@@ -70,38 +70,29 @@ class Watch implements Executable, /* @internal */ CommandSubscriber
...
 
-    /** @var Aggregate */
-    private $aggregate;
+    private Aggregate $aggregate;
 
     /** @var array */
     private $aggregateOptions;
 
     /** @var array */
     private $changeStreamOptions;
```
jmikola, line 79: These are missing an `array` type.

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -703,19 +705,15 @@ public function collectionReadMethodClosures(): array
     {
         return array_filter(
             $this->collectionMethodClosures(),
-            function ($rw) {
-                return str_contains($rw[1], 'r');
-            }
+            fn ($rw) => str_contains($rw[1], 'r')
```
jmikola, line 708: Since this is part of a multi-line function call, shouldn't there be a trailing comma?

Likewise for the diff immediately below.

tests/Collection/CrudSpecFunctionalTest.php
```diff
@@ -87,7 +88,7 @@ public function provideSpecificationTests()
         $testArgs = [];
 
         foreach (glob(__DIR__ . '/spec-tests/*/*.json') as $filename) {
-            $json = json_decode(file_get_contents($filename), true);
+            $json = json_decode(file_get_contents($filename), true, 512, JSON_THROW_ON_ERROR);
```
jmikola, line 91: Noted that `512` is the default value for `$depth`.

tests/DocumentationExamplesTest.php
```diff
@@ -1713,7 +1711,7 @@ public function testVersionedApiMigration(): void
         try {
             $count = $db->sales->count();
         } catch (\MongoDB\Driver\Exception\CommandException $e) {
-            echo json_encode($e->getResultDocument());
+            echo json_encode($e->getResultDocument(), JSON_THROW_ON_ERROR);
```
jmikola, line 1714: Since this appears within a documentation example, I think we should use a leading backslash.

```suggestion
            echo json_encode($e->getResultDocument(), \JSON_THROW_ON_ERROR);
```

tests/Operation/CreateEncryptedCollectionFunctionalTest.php
```diff
@@ -5,7 +5,7 @@
 use MongoDB\BSON\Binary;
 use MongoDB\BSON\Document;
 use MongoDB\Client;
-use MongoDB\ClientEncryption;
+use MongoDB\Driver\ClientEncryption;
```
jmikola, line 8: Any reason this wasn't caught before?

tests/SpecTests/Operation.php
```diff
@@ -568,9 +559,7 @@ private function executeForSession(Session $session, FunctionalTestCase $test, C
 
             case 'withTransaction':
                 /** @var self[] $callbackOperations */
-                $callbackOperations = array_map(function ($operation) {
-                    return self::fromConvenientTransactions($operation);
-                }, $this->arguments['callback']->operations);
+                $callbackOperations = array_map(fn ($operation) => self::fromConvenientTransactions($operation), $this->arguments['callback']->operations);
```
jmikola, line 562: Do you want to split the `array_map()` parameters across multiple lines for readability? This extends past 120 characters.

tests/SpecTests/Operation.php
```diff
@@ -659,10 +648,8 @@ private function executeForTestRunner(FunctionalTestCase $test, Context $context
     private function getIndexNames(Context $context, string $databaseName, string $collectionName): array
     {
         return array_map(
-            function (IndexInfo $indexInfo) {
-                return $indexInfo->getName();
-            },
-            iterator_to_array($context->getInternalClient()->selectCollection($databaseName, $collectionName)->listIndexes())
+            fn (IndexInfo $indexInfo) => $indexInfo->getName(),
```
jmikola, line 651: This could also be `strval` if we don't care about the IndexInfo type-checking.

tests/SpecTests/RetryableReadsSpecTest.php
```diff
@@ -60,7 +58,7 @@ public function testRetryableReads(stdClass $test, ?array $runOn, $data, string
         }
 
         foreach (self::$skippedOperations as $operation => $skipReason) {
-            if (strpos($this->dataDescription(), $operation) === 0) {
+            if (strpos($this->dataDescription(), (string) $operation) === 0) {
```
jmikola, line 61: This is necessary because we don't specify an array shape for `$skippedOperations`, correct?

tests/SpecTests/RetryableWritesSpecTest.php
```diff
@@ -107,7 +107,7 @@ public function testNoWritesPerformedErrorReturnsOriginalError(): void
         ]);
 
         $subscriber = new class ($this) implements CommandSubscriber {
-            private $testCase;
+            private FunctionalTestCase $testCase;
```
jmikola, line 110: Since this is only used within this class, would it make sense to use `self`? This was conveniently [added in PHP 7.4](https://php.watch/versions/7.4/typed-properties).

FunctionalTestCase here refers to SpecTests/FunctionalTestCase, even though this actually only requires the base FunctionalTestCase class.

tests/UnifiedSpecTests/RunOnRequirement.php
```diff
@@ -31,38 +31,29 @@ class RunOnRequirement
...
-    private $maxServerVersion;
+    private ?string $maxServerVersion = null;
 
-    /** @var array */
-    private $topologies;
+    private ?array $topologies = null;
 
-    /** @var stdClass */
-    private $serverParameters;
+    private ?object $serverParameters = null;
```
jmikola, line 40: Should we have kept `stdClass` here?

tests/UnifiedSpecTests/UnifiedTestCase.php
```diff
@@ -25,20 +25,15 @@
  */
 final class UnifiedTestCase implements IteratorAggregate
 {
-    /** @var stdClass */
-    private $test;
+    private stdClass $test;
```
jmikola, line 28: Interesting that rector left `stdClass` here but changed it to `object` (I assume) in `RunOnRequirement.php`.

tests/Collection/CollectionFunctionalTest.php
alcaeus: I can only imagine that the sniff gets tripped up by the closure, but I'll have to investigate.

tests/Operation/CreateEncryptedCollectionFunctionalTest.php
alcaeus: I think psalm would've caught this, but we haven't enabled it for tests. I'll enable it in PHPLIB-1119 and add all current test errors to the baseline.

tests/Collection/CollectionFunctionalTest.php
alcaeus: Since we're in a new class definition, `self` won't do what we expect it to do, as it will refer to the anonymous class: https://3v4l.org/GavRg

tests/Collection/CrudSpecFunctionalTest.php
alcaeus: Yes, I believe that the `assertIsObject` check later in this file tripped rector up a bit.

src/Operation/Watch.php
alcaeus: Correct. I've added a type hint to avoid the cast.

tests/DocumentationExamplesTest.php
GromNaN: For the purpose of documentation, we should not add this option : it complicates the example without adding anything useful. The JSON returned by mongo is always valid. 
The rector rule can be disabled for this file.

alcaeus: I agree. I excluded the file from that rector.

tests/UnifiedSpecTests/RunOnRequirement.php
alcaeus: It got tripped up by a `assertIsObject` check later in the file.

src/Operation/ListCollectionNames.php
alcaeus: I undid that check as this is in the CallbackIterator: the callback is invoked with two arguments, but `strval` only accepts one.

tests/SpecTests/RetryableReadsSpecTest.php
alcaeus: Correct. I removed this in favour of a proper `array<string, string>` annotation on the property.

tests/SpecTests/RetryableWritesSpecTest.php
alcaeus: Since we're in a new class definition, `self` won't do what we expect it to do, as it will refer to the anonymous class: https://3v4l.org/GavRg

tests/DocumentationExamplesTest.php
jmikola: Ah, I didn't realize rector was adding this. I agree it's not needed for example code.

tests/UnifiedSpecTests/EntityMap.php
```diff
@@ -29,19 +29,17 @@
...
-    /** @var array */
-    private $map = [];
+    private array $map = [];
 
     /**
      * Track lsids so they can be accessed after Session::getLogicalSessionId()
      * has been called.
      *
-     * @var stdClass[]
+     * @var list<stdClass>
```
jmikola, line 38: This is incorrect. The `set()` method assigns the lsid object with a string key.

```suggestion
     * @var array<string, stdClass>
```

jmikola (approved):
LGTM after the necessary change in `EntityMap.php`.

Everything else is suggestions, for which I'll defer to you.

src/Operation/Aggregate.php
```diff
@@ -50,23 +50,17 @@
...
-    private $databaseName;
+    private string $databaseName;
 
-    /** @var string|null */
-    private $collectionName;
+    private ?string $collectionName = null;
 
-    /** @var array */
-    private $pipeline;
+    private array $pipeline;
```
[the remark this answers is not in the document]
alcaeus: I'd wait until we start using array shapes more, as that change would trigger a number of psalm issues down the line.

src/Operation/InsertMany.php
```diff
@@ -39,17 +39,14 @@
...
-    private $databaseName;
+    private string $databaseName;
 
-    /** @var string */
-    private $collectionName;
+    private string $collectionName;
 
-    /** @var object[]|array[] */
-    private $documents;
+    /** @var list<object>|list<array> */
```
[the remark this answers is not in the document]
alcaeus: I chose this as it's what the original type was about, but thinking about it made me realise that there's nothing that would prevent a mixed list of arrays and objects to be passed. I've updated it.

tests/SpecTests/ErrorExpectation.php
```diff
@@ -22,36 +22,28 @@
  */
 final class ErrorExpectation
 {
-    /**
-     * @see https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.err
-     *
-     * @var array
-     */
-    private static $codeNameMap = [
+    /** @see https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.err */
```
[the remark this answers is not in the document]
alcaeus: That one was added manually to avoid the cast. I've added a shape here as well.

tests/SpecTests/ErrorExpectation.php
```diff
@@ -22,36 +22,28 @@
...
+    private static array $codeNameMap = [
```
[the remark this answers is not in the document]
alcaeus: Not sure why those weren't addressed, I'll let @GromNaN follow up on this.

tests/UnifiedSpecTests/EntityMap.php
alcaeus: Oh, I missed that. Thanks for catching it!

tests/SpecTests/RetryableWritesSpecTest.php
alcaeus: That's a good point. I've updated it to `RetryableWritesSpecTest`.

examples/persistable.php
alcaeus: Note: the missing property types weren't caught by Rector for some reason, and I only later realised that we had an exclusion in the phpcs config to prevent native types from being added. This has since been removed and types for properties are now enforced.

PR #1130, opened 2023-07-14
PHPLIB-1192: Remove useCursor option in aggregate

alcaeus:
PHPLIB-1192

I came across this while working on codecs, and removing this legacy logic makes the logic significantly simpler.

### DIFF
--- a/docs/reference/method/MongoDBCollection-aggregate.txt
+++ b/docs/reference/method/MongoDBCollection-aggregate.txt
@@ -50,13 +50,8 @@ Errors/Exceptions
 Behavior
 --------
 
-:phpmethod:`MongoDB\\Collection::aggregate()`'s return value depends on the
-MongoDB server version and whether the ``useCursor`` option is specified. If
-``useCursor`` is ``true``, a :php:`MongoDB\\Driver\\Cursor
-<class.mongodb-driver-cursor>` object is returned. If ``useCursor`` is
-``false``, an :php:`ArrayIterator <arrayiterator>` is returned that wraps the
-``result`` array from the command response document. In both cases, the return
-value will be :php:`Traversable <traversable>`.
+:phpmethod:`MongoDB\\Collection::aggregate()`'s returns a
+:php:`MongoDB\\Driver\\Cursor <class.mongodb-driver-cursor>` object.
 
 Examples
 --------
--- a/src/Operation/Aggregate.php
+++ b/src/Operation/Aggregate.php
@@ -55,8 +55,6 @@ class Aggregate implements Executable, Explainable
 
     private array $options;
 
-    private bool $isExplain;
-
     private bool $isWrite;
 
     /**
@@ -195,8 +193,7 @@ class Aggregate implements Executable, Explainable
             unset($options['writeConcern']);
         }
 
-        $this->isExplain = ! empty($options['explain']);
-        $this->isWrite = is_last_pipeline_operator_write($pipeline) && ! $this->isExplain;
+        $this->isWrite = is_last_pipeline_operator_write($pipeline) && ! ($options['explain'] ?? false);
 
         if ($this->isWrite) {
             /* Ignore batchSize for writes, since no documents are returned and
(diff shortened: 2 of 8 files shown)


src/Operation/Aggregate.php
```diff
@@ -337,11 +298,9 @@ private function createCommandDocument(): array
...
         }
 
-        if ($this->options['useCursor']) {
-            $cmd['cursor'] = isset($this->options["batchSize"])
-                ? ['batchSize' => $this->options["batchSize"]]
-                : new stdClass();
-        }
+        $cmd['cursor'] = isset($this->options["batchSize"])
+            ? ['batchSize' => $this->options["batchSize"]]
+            : new stdClass();
```
jmikola, line 303: I assume always supplying a `cursor` option for the `explain` option is kosher?

Quoting [Aggregate: Example](https://www.mongodb.com/docs/manual/reference/command/aggregate/#example) from the server docs:

> MongoDB 3.6 removes the use of [aggregate](https://www.mongodb.com/docs/manual/reference/command/aggregate/#mongodb-dbcommand-dbcmd.aggregate)  command without the cursor option unless the command includes the explain option. Unless you include the explain option, you must specify the cursor option.

It doesn't say anything about `explain` _prohibiting_ a `cursor` option, but `cursor` does seem to be ignored. Do you think we should gate this with a `! $this->isExplain` condition?

src/Operation/Aggregate.php
```diff
@@ -272,25 +247,11 @@ public function execute(Server $server)
...
         }
 
-        $result = current($cursor->toArray());
-
-        if (! is_object($result) || ! isset($result->result) || ! is_array($result->result)) {
-            throw new UnexpectedValueException('aggregate command did not return a "result" array');
-        }
-
-        return new ArrayIterator($result->result);
+        return $cursor;
```
jmikola, line 254: I'll note that even if `explain: true` is used and `aggregate` _does not_ return a command cursor, we can rely on PHPC to consistently return a synthetic cursor with the inline command response as a single document.

src/Operation/Aggregate.php
```diff
@@ -223,12 +204,6 @@ public function __construct(string $databaseName, ?string $collectionName, array
         $this->isExplain = ! empty($options['explain']);
         $this->isWrite = is_last_pipeline_operator_write($pipeline) && ! $this->isExplain;
```
jmikola, line 205: After these changes, `$this->isExplain` is only used on this line.

If you don't decide to add the condition I suggested later in `createCommandDocument()`, then perhaps we can just remove the property.

docs/reference/method/MongoDBCollection-aggregate.txt
```diff
@@ -50,13 +50,8 @@ Errors/Exceptions
...
 
-:phpmethod:`MongoDB\\Collection::aggregate()`'s return value depends on the
-MongoDB server version and whether the ``useCursor`` option is specified. If
-``useCursor`` is ``true``, a :php:`MongoDB\\Driver\\Cursor
-<class.mongodb-driver-cursor>` object is returned. If ``useCursor`` is
-``false``, an :php:`ArrayIterator <arrayiterator>` is returned that wraps the
-``result`` array from the command response document. In both cases, the return
-value will be :php:`Traversable <traversable>`.
+:phpmethod:`MongoDB\\Collection::aggregate()`'s returns a
+:php:`MongoDB\\Driver\\Cursor <class.mongodb-driver-cursor>` object.
```
jmikola, line 54: I suppose this applies regardless of whether `explain: true` is used. SGTM.

src/Operation/Aggregate.php
alcaeus: I'd leave the option in unless `explain` specifically prohibits it - this is consistent with other commands where we also send `batchSize` as part of an `explain`.

src/Operation/Aggregate.php
alcaeus: Removed the property and inlined the usage - I also removed the `! empty` in favour of a null coalesce operator.

GromNaN (approved):

jmikola (approved):

PR #1131, opened 2023-07-14
PHPLIB-1119: Update to Psalm 5

alcaeus:
PHPLIB-1119

Along with the update, I decided to fix some low-hanging fruit from the baseline. I've also moved some other issues to the exclusion list in the config as they will appear more often until we have actual parameter types throughout the codebase.

### DIFF
--- a/docs/examples/csfle-explicit_encryption.php
+++ b/docs/examples/csfle-explicit_encryption.php
@@ -41,6 +41,8 @@ $collection->insertOne(['_id' => 1, 'encryptedField' => $encryptedValue]);
 
 /* Using the client configured without encryption, find the document and observe
  * that the field is not automatically decrypted. */
+
+/** @var object{encryptedField: Binary} $document */
 $document = $collection->findOne();
 
 print_r($document);
--- a/examples/changestream.php
+++ b/examples/changestream.php
@@ -53,7 +53,7 @@ while (true) {
     $changeStream->next();
 
     if (time() - $startTime > 3) {
-        printf("Aborting after 3 seconds...\n");
+        echo "Aborting after 3 seconds...\n";
         break;
     }
 }
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -59,37 +59,6 @@
       <code><![CDATA[$options['session']]]></code>
     </MixedAssignment>
   </file>
-  <file src="src/Exception/BadMethodCallException.php">
-    <UnsafeInstantiation>
-      <code><![CDATA[new static(sprintf('%s is immutable', $class))]]></code>
-      <code><![CDATA[new static(sprintf('%s should not be called for an unacknowledged write result', $method))]]></code>
-    </UnsafeInstantiation>
-  </file>
-  <file src="src/Exception/InvalidArgumentException.php">
-    <UnsafeInstantiation>
-      <code><![CDATA[new static(sprintf('Expected %s to have type "%s" but found "%s"', $name, $expectedType, get_debug_type($value)))]]></code>
-    </UnsafeInstantiation>
-  </file>
-  <file src="src/Exception/ResumeTokenException.php">
-    <UnsafeInstantiation>
-      <code><![CDATA[new static('Resume token not found in change document')]]></code>
-      <code><![CDATA[new static(sprintf('Expected resume token to have type "array or object" but found "%s"', get_debug_type($value)))]]></code>
-    </UnsafeInstantiation>
-  </file>
-  <file src="src/Exception/UnsupportedException.php">
-    <UnsafeInstantiation>
-      <code><![CDATA[new static('Array filters are not supported by the server executing this operation')]]></code>
-      <code><![CDATA[new static('Collations are not supported by the server executing this operation')]]></code>
-      <code><![CDATA[new static('Explain is not supported by the server executing this operation')]]></code>
-      <code><![CDATA[new static('Hint is not supported by the server executing this operation')]]></code>
-      <code><![CDATA[new static('Read concern is not supported by the server executing this command')]]></code>
-      <code><![CDATA[new static('The "allowDiskUse" option is not supported by the server executing this operation')]]></code>
-      <code><![CDATA[new static('The "commitQuorum" option is not supported by the server executing this operation')]]></code>
-      <code><![CDATA[new static('The "readConcern" option cannot be specified within a transaction. Instead, specify it when starting the transaction.')]]></code>
-      <code><![CDATA[new static('The "writeConcern" option cannot be specified within a transaction. Instead, specify it when starting the transaction.')]]></code>
-      <code><![CDATA[new static('Write concern is not supported by the server executing this command')]]></code>
-    </UnsafeInstantiation>
-  </file>
   <file src="src/GridFS/Bucket.php">
     <DocblockTypeContradiction>
       <code>! is_resource($destination)</code>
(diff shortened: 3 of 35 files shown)


docs/examples/csfle-explicit_encryption.php
```diff
@@ -41,6 +41,8 @@
 
 /* Using the client configured without encryption, find the document and observe
  * that the field is not automatically decrypted. */
+
+/** @var object{encryptedField: Binary} $document */
```
GromNaN, line 45: Looking at `psalm-baseline.xml`, this was not enough to enforce the type of `$document->encryptedField`?

psalm-baseline.xml
```diff
@@ -1,381 +1,191 @@
...
-      <code>new static(sprintf('Downloading file from "%s" to "%s" failed. GridFS identifier: "%s"', $sourceMetadata['uri'], $destinationMetadata['uri'], $idString))</code>
-      <code>new static(sprintf('Uploading file from "%s" to "%s" failed. GridFS filename: "%s"', $sourceMetadata['uri'], $destinationUri, $filename))</code>
-    </UnsafeInstantiation>
+    <MixedArgument>
+      <code><![CDATA[$options['revision']]]></code>
+      <code><![CDATA[$options['revision']]]></code>
+    </MixedArgument>
   </file>
   <file src="src/GridFS/ReadableStream.php">
-    <MixedArgument occurrences="2">
```
GromNaN, line 172: That's nice they [removed the `occurrences` number](https://github.com/vimeo/psalm/pull/8967). That will reduce merge conflicts.

psalm.xml.dist
```diff
@@ -15,9 +17,22 @@
...
+    <issueHandlers>
+        <!-- Class is meant to be Unserialized by the extension only -->
+        <PropertyNotSetInConstructor>
+            <errorLevel type="suppress">
+                <file name="examples/typemap.php" />
+            </errorLevel>
+        </PropertyNotSetInConstructor>
+
+        <!-- This is often the result of type checks due to missing native types -->
+        <DocblockTypeContradiction errorLevel="info" />
```
GromNaN, line 36: 👍🏻

src/Model/CachingIterator.php
```diff
@@ -83,12 +88,13 @@ public function current()
     {
         $currentItem = current($this->items);
 
-        return $currentItem !== false ? $currentItem[self::FIELD_VALUE] : false;
+        return $currentItem !== false ? $currentItem[self::FIELD_VALUE] : null;
```
GromNaN, line 91: This is something that surprised me when I worked on this class #1118.

The return value is not specified. `ArrayIterator` returns `null` while `EmptyIterator` [throws an exception](https://www.php.net/manual/en/emptyiterator.current.php).

The fix looks good.

GromNaN (approved):
LGTM

docs/examples/csfle-explicit_encryption.php
alcaeus: The error is regarding the return type of `decrypt`, which is mixed and can't be made any more specific:
> ERROR: MixedArgument - docs/examples/csfle-explicit_encryption.php:51:27 - Argument 2 of printf cannot be mixed, expecting float|int|string

The `@var` annotation here prevents a different Psalm error regarding accessing a potentially undefined property.

src/Model/CachingIterator.php
jmikola: The class originally used [`current()`](https://www.php.net/current), which returns `false`. That was preserved when it was later refactored in [3049b5d](https://github.com/mongodb/mongo-php-library/commit/3049b5dba3fd0ef5b95a5f2d12a140a8b6a2a180#diff-be61b04c89f7e9634e8b3d05aede2915bdbadefd72995eb1ecee98bcae3eb420L90-R94).

That said, I have no object to being consistent with ArrayIterator and returning `null`.

src/Model/ChangeStreamIterator.php
```diff
@@ -122,12 +125,13 @@ final public function commandSucceeded(CommandSucceededEvent $event): void
...
      * @see https://php.net/iteratoriterator.current
-     * @return mixed
+     * @return array|object|null
+     * @psalm-return TValue|null
      */
     #[ReturnTypeWillChange]
     public function current()
     {
-        return $this->isValid ? parent::current() : null;
+        return $this->valid() ? parent::current() : null;
```
jmikola, line 134: Was this necessary to use the `@psalm-assert-if-true` annotation on `valid()`? Otherwise, I see no reason to add a method call.

src/Model/CollectionInfoIterator.php
```diff
@@ -26,6 +26,7 @@
  * This iterator is used for enumerating collections in a database.
  *
  * @see \MongoDB\Database::listCollections()
+ * @template-extends Iterator<int, CollectionInfo>
```
jmikola, line 29: Related to my earlier question, I assume there's no notation for a list-like Traversable?

src/Model/IndexInfoIteratorIterator.php
```diff
@@ -34,11 +34,13 @@
  * @see https://github.com/mongodb/specifications/blob/master/source/enumerate-indexes.rst
  * @see https://mongodb.com/docs/manual/reference/command/listIndexes/
  * @see https://mongodb.com/docs/manual/reference/system-collections/
+ * @template-extends IteratorIterator<int, array, Traversable<int, array>>
```
jmikola, line 37: Why are there three types under `IteratorIterator<>`? Are the first two types always consistent with the two types for Traversable? Presumably "Traversable" here is just whatever is returned by `getIterator()`?

src/Operation/ListIndexes.php
```diff
@@ -144,6 +144,9 @@ private function executeCommand(Server $server): IndexInfoIteratorIterator
 
         $cursor->setTypeMap(['root' => 'array', 'document' => 'array']);
 
-        return new IndexInfoIteratorIterator(new CachingIterator($cursor), $this->databaseName . '.' . $this->collectionName);
+        /** @var CachingIterator<int, array> $iterator */
```
jmikola, line 147: Is it necessary to repeat the variable name in these annotations? I have some recollection of _not_ doing so previously.

examples/changestream.php
```diff
@@ -53,7 +53,7 @@ function toJSON(object $document): string
     $changeStream->next();
 
     if (time() - $startTime > 3) {
-        printf("Aborting after 3 seconds...\n");
+        echo "Aborting after 3 seconds...\n";
```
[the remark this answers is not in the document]
alcaeus: Psalm objected to this, complaining that `printf` needs two arguments.

src/Command/ListCollections.php
```diff
@@ -99,11 +100,13 @@ public function __construct(string $databaseName, array $options = [])
     /**
      * Execute the operation.
      *
+     * @return CachingIterator<int, array>
      * @see Executable::execute()
      * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
      */
     public function execute(Server $server): CachingIterator
     {
+        /** @var Cursor<array> $cursor */
```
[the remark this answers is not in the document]
alcaeus: I believe this was part of Psalm 4 already; the iterator stubs have templates to indicate the key and value types. Since I was working on these classes anyways, I added the annotations for consistency.

src/MapReduceResult.php
```diff
@@ -82,7 +88,7 @@ public function getExecutionTimeMS()
      * Return the mapReduce results as a Traversable.
      *
      * @see https://php.net/iteratoraggregate.getiterator
-     * @return Traversable
+     * @return Traversable<int, array|object>
```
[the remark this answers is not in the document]
alcaeus: Correct, I needed to import this elsewhere, hence the `@psalm-type`. I'm not quite sure how much the various PHPDoc tooling (e.g. PhpStorm) can do with the Psalm types, so I tried to not overdo it with Psalm types in regular PHPDoc annotations.

src/Model/CachingIterator.php
```diff
@@ -36,14 +36,19 @@
...
+ * @template TKey of array-key
+ * @template TValue
+ * @template-implements Iterator<TKey, TValue>
  */
 class CachingIterator implements Countable, Iterator
 {
     private const FIELD_KEY = 0;
     private const FIELD_VALUE = 1;
 
+    /** @var list<array{0: TKey, 1: TValue}> */
```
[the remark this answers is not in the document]
alcaeus: `list<TKey, TValue>` is invalid, as `list<TValue>` resolves to `array<int, TValue>`. In this case the array shape is correct, as we only expect two elements in the list: one for the key, one for the value.

src/Model/ChangeStreamIterator.php
alcaeus: That is correct. I'll do some checking to see if `@psalm-assert-if-true` also works on variables and if psalm picks up on that to avoid the method call.

src/Model/IndexInfoIteratorIterator.php
alcaeus: The template for `IteratorIterator` includes the inner iterator, unfortunately also locking it to the same key type: https://github.com/vimeo/psalm/blob/9c814c8a69b0fd80f7fb81de5740434ed1ce7022/stubs/CoreGenericIterators.phpstub#L105..L111. This has already proved problematic in the `AsListIterator` I was working on for our Codec implementation, so I'll have to revisit this to figure out what the purpose of the third template parameter is. This works for us, so we're good.

src/Model/ChangeStreamIterator.php
alcaeus: Did some checking on the Playground, and unfortunately it doesn't work with variables: https://psalm.dev/r/bf4920bc0b

src/Operation/ListIndexes.php
alcaeus: It's unnecessary when used on a property, but required in the code.

PR #1132, opened 2023-07-17
Skip queryable encryption tests on standalone servers

alcaeus:
Fixes two build failures on 7.0 and latest that started appearing recently.

### DIFF
--- a/tests/ExamplesTest.php
+++ b/tests/ExamplesTest.php
@@ -360,6 +360,10 @@ OUTPUT;
 
         $this->skipIfServerVersion('<', '7.0.0', 'Queryable encryption tests require MongoDB 7.0 or later');
 
+        if ($this->isStandalone()) {
+            $this->markTestSkipped('Encrypted collections are not supported on standalone');
+        }
+
         /* Ensure that the key vault, collection under test, and any metadata
          * collections are cleaned up before and after the example is run. */
         $this->dropCollection('test', 'coll', ['encryptedFields' => []]);
--- a/tests/ExamplesTest.php
+++ b/tests/ExamplesTest.php
@@ -361,7 +361,7 @@ OUTPUT;
         $this->skipIfServerVersion('<', '7.0.0', 'Queryable encryption tests require MongoDB 7.0 or later');
 
         if ($this->isStandalone()) {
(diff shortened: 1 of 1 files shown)


tests/ExamplesTest.php
```diff
@@ -360,6 +360,10 @@ public function testQueryableEncryptionExamples(string $file, string $expectedOu
 
         $this->skipIfServerVersion('<', '7.0.0', 'Queryable encryption tests require MongoDB 7.0 or later');
 
+        if (! $this->isReplicaSet()) {
```
jmikola, line 363:
```suggestion
        if ($this->isStandalone()) {
```

This should permit sharded clusters, as is done in: https://github.com/mongodb/mongo-php-library/blob/d4cdf057a67cb99a32db8984a16959bfa7ca7eb5/tests/DocumentationExamplesTest.php#L1808-L1810

While writing this, I also realized that `isShardedClusterUsingReplicasets()` is redundant with MongoDB 3.6+. Opened [PHPLIB-1202](https://jira.mongodb.org/browse/PHPLIB-1202) to track that.

jmikola (changes requested):

tests/ExamplesTest.php
alcaeus: Ah, yes. I used the negative checks since examples are never tested on sharded clusters, but nonetheless it makes more sense to explicitly skip on stand-alone in case we end up testing on sharded clusters after all.

tests/ExamplesTest.php
```diff
@@ -360,6 +360,10 @@ public function testQueryableEncryptionExamples(string $file, string $expectedOu
 
         $this->skipIfServerVersion('<', '7.0.0', 'Queryable encryption tests require MongoDB 7.0 or later');
 
+        if ($this->isStandalone()) {
+            $this->markTestSkipped('Encrypted collections are not supported on standalone');
```
jmikola, line 364:
```suggestion
            $this->markTestSkipped('Queryable encryption requires replica sets');
```

Optional suggestion if you want to use the same error message I did in https://github.com/mongodb/mongo-php-library/pull/1133 (in more places).

jmikola (approved):

Issue #1137, opened 2023-07-20
The StreamWrapper class is currently unusable without the bucket


fgm:
### Bug Report

Context: I am implementing GridFS support in Drupal 10.

For that purpose, I created a stream_wrapper of our own, and implemented a first rough version using the GridFS Bucket. However, I noticed that the Bucket actually relies on a fully functional existing stream wrapper for the `gridfs` scheme, meaning I was implementing a stream wrapper on top of the Bucket construction using an existing stream wrapper, which means lot of useless wrapping and unwrapping.

I then looked for a way to use the `gridfs` stream wrapper directly, but in the current implementation, it seems there is no way to use this stream wrapper directly unless passing an explicit `$context` parameter to `fopen()` calls, which basically removes most of the interest of this abstraction.

### Environment

Environment is not relevant here, as the issue is with the API shape, not its interaction with the extension and servers.

### Test Script

n.a.

### Expected and Actual Behavior

The expected behavior would be to have the StreamWrapper be aware of the relevant `Database`, obviating the need for wrapper users to pass a context, although that could still be done to override that default.

Given the lack of parameters on stream wrapper classes, this probably means setting the database as a static property on the StreamWrapper class. From there, operations could still query properties from the context, which the Bucket could still pass, and fallback to these defaults when not available.

It would allow clients  not to use the Bucket, and just register the StreamWrapper  and add their Database information as a static property on that class before the first use of the wrapper.

### Debug Log

n.a.

fgm:
Hello @—} . Thanks for answering so fast The API is "just" full support for any file operation, without the caller having to pass a stream context. That is the Drupal core expectation for any stream wrapper. So, for example, being able to just do a `file_get_contents("$scheme://$path")`.

In the meantime, I read more of the library implementation, and AFAICS it would not be directly applicable, because it only handles `fopen` modes `r` and `w`, while for example using `file_get_contents()` uses mode `rb`. And that is without even considering all the combined modes like "r+" we support in Drupal stream wrappers. The existing S3 wrapper, for example, normalizes modes by dropping `b` and `'t`, thus actually always using the implicit "b"; and they have 3 kinds of stream: read, append, and write; with all of them supporting reads. They support operations in the middle of files because S3 supports Range read/writes; for GridFS the equivalent is likely to selectively access chunks instead of rebuilding the whole file on every operation.

[...]

PR #1138, opened 2023-07-21
PHPLIB-1206 Allow global registration of GridFS buckets with `gridfs://` protocol

GromNaN:
Fix [PHPLIB-1206](https://jira.mongodb.org/browse/PHPLIB-1206)
Fix https://github.com/mongodb/mongo-php-library/issues/1137

Allows to register an alias for the stream wrapper. File urls looks like:
```
gridfs://<bucket-alias>/<filename>
```

Where `<bucket-alias>` is an alias for the bucket

```php
$client = new Client();
$bucket = $client->selectDatabase('test_db')->selectGridFSBucket();
$bucket->registerGlobalStreamWrapperAlias('mybucket');

file_put_contents("gridfs://mybucket/filename.txt", 'Hello GridFS!');

if (file_exists("gridfs://mybucket/filename.txt")) {
    echo file_get_contents("gridfs://mybucket/filename.txt") . PHP_EOL;
}
```

Todo:

- [x] Add reference doc for the method `Bucket::registerGlobalStreamWrapperAlias`

### DIFF
--- a/docs/reference/method/MongoDBGridFSBucket-registerGlobalStreamWrapperAlias.txt
+++ b/docs/reference/method/MongoDBGridFSBucket-registerGlobalStreamWrapperAlias.txt
@@ -1,0 +1,132 @@
+===========================================================
+MongoDB\\GridFS\\Bucket::registerGlobalStreamWrapperAlias()
+===========================================================
+
+.. versionadded:: 1.18
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Definition
+----------
+
+.. phpmethod:: MongoDB\\GridFS\\Bucket::registerGlobalStreamWrapperAlias()
+
+   Registers an alias for the bucket, which enables files within the bucket to
+   be accessed using a basic filename string (e.g.
+   `gridfs://<bucket-alias>/<filename>`).
+
+   .. code-block:: php
+
+      function registerGlobalStreamWrapperAlias(string $alias): void
+
+Parameters
+----------
+
+``$alias`` : array
+  A non-empty string used to identify the GridFS bucket when accessing files
+  using the ``gridfs://`` stream wrapper.
+
+Behavior
+--------
+
+After registering an alias for the bucket, the most recent revision of a file
+can be accessed using a filename string in the form ``gridfs://<bucket-alias>/<filename>``.
+
+Supported stream functions:
+
+- :php:`copy() <copy>`
+- :php:`file_exists() <file_exists>`
+- :php:`file_get_contents() <file_get_contents>`
+- :php:`file_put_contents() <file_put_contents>`
+- :php:`filemtime() <filemtime>`
+- :php:`filesize() <filesize>`
+- :php:`file() <file>`
+- :php:`fopen() <fopen>` (with "r", "rb", "w", and "wb" modes)
+
+In read mode, the stream context can contain the option ``gridfs['revision']``
+to specify the revision number of the file to read. If omitted, the most recent
+revision is read (revision ``-1``).
+
+In write mode, the stream context can contain the option ``gridfs['chunkSizeBytes']``.
+If omitted, the defaults are inherited from the ``Bucket`` instance option.
+
+Example
+-------
+
+Read and write to a GridFS bucket using the ``gridfs://`` stream wrapper
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following example demonstrates how to register an alias for a GridFS bucket
+and use the functions ``file_exists()``, ``file_get_contents()``, and
+``file_put_contents()`` to read and write to the bucket.
+
+Each call to these functions makes a request to the server.
+
+.. code-block:: php
+
+   <?php
+
+   $database = (new MongoDB\Client)->selectDatabase('test');
+   $bucket = $database->selectGridFSBucket();
+
+   $bucket->registerGlobalStreamWrapperAlias('mybucket');
+
+   var_dump(file_exists('gridfs://mybucket/hello.txt'));
+
+   file_put_contents('gridfs://mybucket/hello.txt', 'Hello, GridFS!');
+
+   var_dump(file_exists('gridfs://mybucket/hello.txt'));
+
+   echo file_get_contents('gridfs://mybucket/hello.txt');
+
+The output would then resemble:
+
+.. code-block:: none
+
+   bool(false)
+   bool(true)
+   Hello, GridFS!
+
+Read a specific revision of a file
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Using a stream context, you can specify the revision number of the file to
+read. If omitted, the most recent revision is read.
+
+.. code-block:: php
+
+   <?php
+
+   $database = (new MongoDB\Client)->selectDatabase('test');
+   $bucket = $database->selectGridFSBucket();
+
+   $bucket->registerGlobalStreamWrapperAlias('mybucket');
+
+   // Creating revision 0
+   $handle = fopen('gridfs://mybucket/hello.txt', 'w');
+   fwrite($handle, 'Hello, GridFS! (v0)');
+   fclose($handle);
+
+   // Creating revision 1
+   $handle = fopen('gridfs://mybucket/hello.txt', 'w');
+   fwrite($handle, 'Hello, GridFS! (v1)');
+   fclose($handle);
+
+   // Read revision 0
+   $context = stream_context_create([
+        'gridfs' => ['revision' => 0],
+   ]);
+   $handle = fopen('gridfs://mybucket/hello.txt', 'r', false, $context);
+   echo fread($handle, 1024);
+
+The output would then resemble:
+
+.. code-block:: none
+
+   Hello, GridFS! (v0)
--- a/src/GridFS/Bucket.php
+++ b/src/GridFS/Bucket.php
@@ -31,6 +31,7 @@ use MongoDB\Exception\InvalidArgumentException;
 use MongoDB\Exception\UnsupportedException;
 use MongoDB\GridFS\Exception\CorruptFileException;
 use MongoDB\GridFS\Exception\FileNotFoundException;
+use MongoDB\GridFS\Exception\LogicException;
 use MongoDB\GridFS\Exception\StreamException;
 use MongoDB\Model\BSONArray;
 use MongoDB\Model\BSONDocument;
@@ -39,6 +40,7 @@ use MongoDB\Operation\Find;
 use function array_intersect_key;
 use function array_key_exists;
 use function assert;
+use function explode;
 use function fopen;
 use function get_resource_type;
 use function in_array;
@@ -54,6 +56,7 @@ use function MongoDB\BSON\fromPHP;
 use function MongoDB\BSON\toJSON;
 use function property_exists;
 use function sprintf;
+use function str_contains;
 use function stream_context_create;
 use function stream_copy_to_stream;
 use function stream_get_meta_data;
@@ -588,6 +591,29 @@ class Bucket
     }
 
     /**
+     * Register an alias to enable basic filename access for this bucket.
+     *
+     * For applications that need to interact with GridFS using only a filename
+     * string, a bucket can be registered with an alias. Files can then be
+     * accessed using the following pattern:
+     *
+     *     gridfs://<bucket-alias>/<filename>
+     *
+     * Read operations will always target the most recent revision of a file.
+     *
+     * @param non-empty-string string $alias The alias to use for the bucket
+     */
+    public function registerGlobalStreamWrapperAlias(string $alias): void
+    {
+        if ($alias === '' || str_contains($alias, '/')) {
+            throw new InvalidArgumentException(sprintf('The bucket alias must be a non-empty string without any slash, "%s" given', $alias));
+        }
+
+        // Use a closure to expose the private method into another class
+        StreamWrapper::setContextResolver($alias, fn (string $path, string $mode, array $context) => $this->resolveStreamContext($path, $mode, $context));
+    }
+
+    /**
      * Renames the GridFS file with the specified ID.
      *
      * @param mixed  $id          File ID
@@ -755,5 +781,51 @@ class Bucket
         }
 
         StreamWrapper::register(self::STREAM_WRAPPER_PROTOCOL);
+    }
+
+    /**
+     * Create a stream context from the path and mode provided to fopen().
+     *
+     * @see StreamWrapper::setContextResolver()
+     *
+     * @param string                                                         $path    The full url provided to fopen(). It contains the filename.
+     *                                                                                gridfs://database_name/collection_name.files/file_name
+     * @param array{revision?: int, chunkSizeBytes?: int, disableMD5?: bool} $context The options provided to fopen()
+     *
+     * @return array{collectionWrapper: CollectionWrapper, file: object}|array{collectionWrapper: CollectionWrapper, filename: string, options: array}
+     *
+     * @throws FileNotFoundException
+     * @throws LogicException
+     */
+    private function resolveStreamContext(string $path, string $mode, array $context): array
+    {
+        // Fallback to an empty filename if the path does not contain one: "gridfs://alias"
+        $filename = explode('/', $path, 4)[3] ?? '';
+
+        if ($mode === 'r' || $mode === 'rb') {
+            $file = $this->collectionWrapper->findFileByFilenameAndRevision($filename, $context['revision'] ?? -1);
+
+            if (! is_object($file)) {
+                throw FileNotFoundException::byFilenameAndRevision($filename, $context['revision'] ?? -1, $path);
+            }
+
+            return [
+                'collectionWrapper' => $this->collectionWrapper,
+                'file' => $file,
+            ];
+        }
+
+        if ($mode === 'w' || $mode === 'wb') {
+            return [
+                'collectionWrapper' => $this->collectionWrapper,
+                'filename' => $filename,
+                'options' => $context + [
+                    'chunkSizeBytes' => $this->chunkSizeBytes,
+                    'disableMD5' => $this->disableMD5,
+                ],
+            ];
+        }
+
+        throw LogicException::openModeNotSupported($mode);
     }
 }
--- a/src/GridFS/StreamWrapper.php
+++ b/src/GridFS/StreamWrapper.php
@@ -17,11 +17,15 @@
 
 namespace MongoDB\GridFS;
 
+use Closure;
 use MongoDB\BSON\UTCDateTime;
+use MongoDB\GridFS\Exception\FileNotFoundException;
+use MongoDB\GridFS\Exception\LogicException;
 
 use function assert;
 use function explode;
 use function in_array;
+use function is_array;
 use function is_integer;
 use function is_resource;
 use function stream_context_get_options;
@@ -40,16 +44,18 @@ use const STREAM_IS_URL;
  * @internal
  * @see Bucket::openUploadStream()
  * @see Bucket::openDownloadStream()
+ * @psalm-type ContextOptions = array{collectionWrapper: CollectionWrapper, file: object}|array{collectionWrapper: CollectionWrapper, filename: string, options: array}
  */
 class StreamWrapper
 {
     /** @var resource|null Stream context (set by PHP) */
     public $context;
 
-    private ?string $protocol = null;
-
     /** @var ReadableStream|WritableStream|null */
     private $stream;
+
+    /** @var array<string, Closure(string, string, array): ContextOptions> */
+    private static array $contextResolvers = [];
 
     public function __destruct()
     {
@@ -82,6 +88,20 @@ class StreamWrapper
         }
 
         stream_wrapper_register($protocol, static::class, STREAM_IS_URL);
+    }
+
+    /**
+     * @see Bucket::resolveStreamContext()
+     *
+     * @param Closure(string, string, array):ContextOptions|null $resolver
+     */
+    public static function setContextResolver(string $name, ?Closure $resolver): void
+    {
+        if ($resolver === null) {
+            unset(self::$contextResolvers[$name]);
+        } else {
+            self::$contextResolvers[$name] = $resolver;
+        }
     }
 
     /**
@@ -123,17 +143,44 @@ class StreamWrapper
      */
     public function stream_open(string $path, string $mode, int $options, ?string &$openedPath): bool
     {
-        $this->initProtocol($path);
+        $context = [];
 
-        if ($mode === 'r') {
-            return $this->initReadableStream();
+        /**
+         * The Bucket methods { @see Bucket::openUploadStream() } and { @see Bucket::openDownloadStreamByFile() }
+         * always set an internal context. But the context can also be set by the user.
+         */
+        if (is_resource($this->context)) {
+            $context = stream_context_get_options($this->context)['gridfs'] ?? [];
+
+            if (! is_array($context)) {
+                throw LogicException::invalidContext($context);
+            }
         }
 
-        if ($mode === 'w') {
-            return $this->initWritableStream();
+        // When the stream is opened using fopen(), the context is not required, it can contain only options.
+        if (! isset($context['collectionWrapper'])) {
+            $bucketAlias = explode('/', $path, 4)[2] ?? '';
+
+            if (! isset(self::$contextResolvers[$bucketAlias])) {
+                throw LogicException::bucketAliasNotRegistered($bucketAlias);
+            }
+
+            $context = self::$contextResolvers[$bucketAlias]($path, $mode, $context);
         }
 
-        return false;
+        if (! $context['collectionWrapper'] instanceof CollectionWrapper) {
+            throw LogicException::invalidContextCollectionWrapper($context['collectionWrapper']);
+        }
+
+        if ($mode === 'r' || $mode === 'rb') {
+            return $this->initReadableStream($context);
+        }
+
+        if ($mode === 'w' || $mode === 'wb') {
+            return $this->initWritableStream($context);
+        }
+
+        throw LogicException::openModeNotSupported($mode);
     }
 
     /**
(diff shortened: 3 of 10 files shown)


src/GridFS/StreamWrapper.php
```diff
@@ -280,8 +292,24 @@ private function getStatTemplate(): array
...
+        $this->protocol = $parts['scheme'] ?: 'gridfs';
+
+        $context = stream_context_get_options($this->context);
+
+        if (isset($context[$this->protocol]['collectionWrapper'])) {
+            $this->contextOptions = $context[$this->protocol];
+        } elseif (! isset(self::$defaultBucket)) {
+            throw new RuntimeException('No default GridFS bucket available for stream wrapper.');
+        } else {
+            $file = explode('/', $parts['path'], 3)[2] ?? '';
```
jmikola, line 305: It looks like this `else` branch is setting the union of all possible context options for readable and writable streams.

Paths for a readable stream resemble `gridfs://{db}/{coll}.files/{id}`, where {id} is a lossy string representation of the `fs.files` document's `_id` field. For a writable stream, the path resembles `gridfs://{db}/{coll}.files`.

If we reach this point for a writable file, the `filename` context option will be an empty string and result in PHPLIB creating an incomplete `fs.files` document.

For a readable file, the behavior also seems incorrect because you're selecting a `fs.files` document where `_id` equals some lossy string representation. I think that's unlikely to match anything, and if it does it will likely be the wrong `fs.files` document.

src/GridFS/StreamWrapper.php
```diff
@@ -280,8 +292,24 @@ private function getStatTemplate(): array
      */
     private function initProtocol(string $path): void
     {
-        $parts = explode('://', $path, 2);
-        $this->protocol = $parts[0] ?: 'gridfs';
+        $parts = parse_url($path);
```
jmikola, line 295: I don't think this is a good use case for `parse_url()`. You only access `scheme` and `path` after this point, and `parse_url()` also captures the database name in a `host` component, which seems odd.

If we parse a typical readable path of `gridfs://{db}/{coll}.files/{id}`, the `path` component will be `/{coll}.files/{id}`. You'll later explode that to yield an array with three elements: `['', '{coll}.files', '{id}']`. The later logic to explode the `path` and extract the ID works fine, but it seems unnecessary.

The `path` component is only relevant if we enter the branch where we have no context options.

I'd suggest reverting this to explode once on `://` to extract the protocol. From there, you can then explode again if needed to extract the ID, and just account for the database component appearing in the path. I actually don't think the second `explode()` would change at all, since there'd be no leading slash in `$parts[1]` after the initial split.

src/GridFS/StreamWrapper.php
```diff
@@ -296,8 +324,8 @@ private function initReadableStream(): bool
 
         assert($this->protocol !== null);
         $this->stream = new ReadableStream(
-            $context[$this->protocol]['collectionWrapper'],
-            $context[$this->protocol]['file'],
+            $this->contextOptions['collectionWrapper'],
```
jmikola, line 327: The first two lines of this method with the `assert()` and `stream_context_get_options()` call no longer seem relevant. Likewise for `initWritableStream()` below.

src/GridFS/StreamWrapper.php
```diff
@@ -280,8 +292,24 @@ private function getStatTemplate(): array
      */
     private function initProtocol(string $path): void
```
jmikola, line 293: If the changes below remain, `initProtocol()` is no longer an accurate name for this method. Something like `initProtocolAndContextOptions()` would be more appropriate.

I'm not a fan of initializing the union of context options for readable and writable streams, but I think that can be revisited once the implementation is actually working (noting the issues with inferring a filename and querying for an `fs.files` document).

src/GridFS/StreamWrapper.php
GromNaN: No longer using `parse_url`. The filename is extracted after the default "base path" of the bucket.

src/GridFS/Bucket.php
```diff
@@ -728,4 +745,53 @@ private function registerStreamWrapper(): void
...
+        // The file can be read only if it belongs to this bucket
+        $basePath = $this->createPathForFile((object) ['_id' => '']);
+        if (! str_starts_with($path, $basePath)) {
+            return null;
+        }
+
+        $filename = urldecode(substr($path, strlen($basePath)));
+
+        if (str_contains($mode, 'r')) {
+            $file = $this->collectionWrapper->findFileByFilenameAndRevision($filename, -1);
```
GromNaN, line 771: In order to have the same path for read and write, we have to pass the filename in the path. So it's impossible to read a file from its `_id`, the file must have a filename and only the last revision is accessible.

src/GridFS/StreamWrapper.php
```diff
@@ -122,14 +143,44 @@ public function stream_eof(): bool
...
     {
-        $this->initProtocol($path);
+        $protocol = $this->initProtocol($path);
+
+        assert(is_resource($this->context));
+        $contextOptions = stream_context_get_options($this->context)[$protocol] ?? null;
+
+        if ($contextOptions === null) {
+            if (! isset(self::$contextResolver)) {
+                if ($options & STREAM_REPORT_ERRORS) {
```
jmikola, line 153: TIL this constant exists. Looks like it was introduced a [very long time ago](https://github.com/php/php-src/commit/32a4f35580436c0c7e74499b78f2390ab7fe367b#diff-aa9fffa7bf1ce5999dd7fb28a1dc1cb2d3841f97880dbde51d5b12a109669699R54).

This seems like a change we should track in a separate ticket for changelog visibility.

src/GridFS/Bucket.php
jmikola: I'm not sure how this is supposed to work. `filename` is not even a unique field in GridFS -- not on its own nor in conjunction with a an upload date (see: [GridFS: Indexes](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst#indexes)).

If we're going to use a URI to uniquely identify a GridFS file, the last component will need to be some encoded form of the `fs.files` document's `_id` field.

src/GridFS/StreamWrapper.php
```diff
@@ -278,26 +329,24 @@ private function getStatTemplate(): array
      *
      * @see StreamWrapper::stream_open()
      */
-    private function initProtocol(string $path): void
+    private function initProtocol(string $path): string
```
jmikola, line 332: If this is no longer setting a `$protocol` property, I think we should either rename this to `getProtocol()` or simply inline it within `stream_open()`. The latter probably makes sense if this is only called once.

This could just be a one-liner:

```
$protocol = explode('://', $path, 2)[0] ?: 'gridfs';
```

That said, while [testing this](https://3v4l.org/TIMYf#v7.4.33) it doesn't look like it does much. If a completely bogus path is provided without a scheme, the path itself is going to be picked up as the protocol.

It's been a while since I worked with the stream wrapper, so perhaps this isn't anything to worry about since we'd only ever receive a path _with_ a scheme. But in that case, I'm nto sure why we ever worry about using `?:` here. `explode()` only returns an empty array for empty input. Any string without a schema is going to return a single-element array with that string.

src/GridFS/Bucket.php
```diff
@@ -728,4 +745,53 @@ private function registerStreamWrapper(): void
...
     }
+
+    /**
+     * Create a stream context from
+     *
+     * @see StreamWrapper::setDefaultContextResolver()
+     * @see stream_context_create()
+     *
+     * @param string $path The full url provided to fopen(). It contains the filename.
+     *                     gridfs://database_name/collection_name.files/file_name
```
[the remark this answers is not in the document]
GromNaN: Let's remove the `.fs` suffix that is adding complexity for users, without provided any information.

src/GridFS/StreamWrapper.php
GromNaN: I'm not sure we can pass this option with the current code, because `fopen` cannot be used outside of the Bucket class. So I d'ont think this could impact current users. but I may be wrong as I didn't fully understood how this option is set.

Tracked in PHPLIB-1318.

alcaeus:
A couple of high level questions about the API. Let's take this code from the PR description:
```php
$bucket = $client->selectDatabase('test_db')->selectGridFSBucket();
$bucket->registerAsDefaultStreamWrapper();

file_put_contents("gridfs://test_db/fs.files/filename.txt", 'Hello GridFS!');
```

If we're already registering a bucket as the default stream wrapper, why does the URI still contain the database and bucket names? The bucket instance is tied to a client, database, and bucket, so I'd expect to use something like this:
```php
file_put_contents('gridfs://filename.txt', 'Hello GridFS!');
```

This might create conflicts with us using GridFS internally. However, we can work around that by registering a new stream wrapper for each bucket:
```php
$client->selectDatabase('test_db')->selectGridFSBucket()->registerStreamWrapper('foo');
$client->selectDatabase('test_db')->selectGridFSBucket(['name' => 'cache'])->registerStreamWrapper('cache');

file_put_contents('foo://filename.txt', 'Hello GridFS!'); // Writes to the fs.files collection
file_put_contents('cache://filename.txt', 'Hello GridFS!'); // Writes to the cache.files collection
```

src/GridFS/StreamWrapper.php
jmikola: This should not have any impact on users since streams are only created internally in PHPLIB. If you plan on changing that, then I expect you'd want to handle this option.

src/GridFS/StreamWrapper.php
GromNaN: Function removed.

src/GridFS/StreamWrapper.php
```diff
@@ -123,14 +148,71 @@ public function stream_eof(): bool
...
-        $this->initProtocol($path);
+        [$protocol] = explode('://', $path, 2);
+
+        $context = [];
+        if (is_resource($this->context)) {
+            $context = stream_context_get_options($this->context)[$protocol] ?? [];
+
+            if (! is_array($context)) {
+                if ($options & STREAM_REPORT_ERRORS) {
+                    trigger_error(sprintf('Invalid context for "%s" protocol.', $protocol), E_USER_ERROR);
```
GromNaN, line 159: Let's use exceptions instead. This flag is never sent. https://jira.mongodb.org/browse/PHPLIB-1318

src/GridFS/StreamWrapper.php
```diff
@@ -123,14 +148,71 @@ public function stream_eof(): bool
...
+                    trigger_error(sprintf('Unknown GridFS Bucket "%1$s". Call $bucket->asStreamWrap(\'%1$s\') on the Bucket you want to access.', $parts[2]), E_USER_ERROR);
+                }
+
+                return false;
+            }
+
+            $context = $resolver($path, $mode, $context);
+            if ($context === null) {
+                if ($options & STREAM_REPORT_ERRORS) {
+                    trigger_error(sprintf('File not found "%s".', $path), E_USER_ERROR);
```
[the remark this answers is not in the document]
GromNaN: I replaced all the `trigger_error` from an other age, to the more brutal exceptions of nowadays.

src/GridFS/StreamWrapper.php
```diff
@@ -123,14 +148,71 @@ public function stream_eof(): bool
...
+        }
+
+        if ($mode === 'w' || $mode === 'wb') {
+            assert(isset($context['filename']) && is_string($context['filename']));
+            assert(isset($context['options']) && is_array($context['options']));
+
+            return $this->initWritableStream($context);
         }
 
         return false;
```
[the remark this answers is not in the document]
GromNaN: Done.

src/GridFS/StreamWrapper.php
```diff
@@ -123,14 +148,71 @@ public function stream_eof(): bool
...
-            return $this->initReadableStream();
+                return false;
+            }
         }
 
-        if ($mode === 'w') {
-            return $this->initWritableStream();
+        // Opening stream from gridfs
+        if (! isset($context['collectionWrapper'])) {
+            $parts = explode('/', $path, 4);
```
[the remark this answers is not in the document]
GromNaN: Replaced with a more explicit `$bucketAlias = explode('/', $path, 4)[2] ?? '';`.

Also, `count($parts) < 4` is no longer checked and we fallback to an empty filename: https://github.com/mongodb/mongo-php-library/pull/1138/files#r1424759860

src/GridFS/Bucket.php
```diff
@@ -587,6 +589,24 @@ public function openUploadStream(string $filename, array $options = [])
         return fopen($path, 'w', false, $context);
     }
 
+    /**
+     * For legacy applications that needs to open stream from a string, the bucket can be registered in the stream
+     * wrapper to be used by default when no GridFS context is provided.
+     * Supported file name must follow this pattern:
+     *     gridfs://<bucket-alias>/<filename>
```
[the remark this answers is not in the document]
GromNaN: `unlink` is not supported ATM.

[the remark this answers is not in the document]
jmikola: > `unlink` is not supported ATM.

Oh, you're correct. We don't define an `unlink()` method on the stream wrapper, so that wouldn't be handled at all.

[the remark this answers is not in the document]
GromNaN: I created PHPLIB-1323

src/GridFS/StreamWrapper.php
```diff
@@ -123,14 +148,71 @@ public function stream_eof(): bool
...
+                    trigger_error(sprintf('Invalid GridFS file name: "%s"', $path), E_USER_ERROR);
+                }
+
+                return false;
+            }
+
+            $resolver = self::$contextResolvers[$parts[2]] ?? null;
+            if (null === $resolver) {
+                if ($options & STREAM_REPORT_ERRORS) {
+                    trigger_error(sprintf('Unknown GridFS Bucket "%1$s". Call $bucket->asStreamWrap(\'%1$s\') on the Bucket you want to access.', $parts[2]), E_USER_ERROR);
```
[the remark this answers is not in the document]
GromNaN: Replaced by an exception.

src/GridFS/StreamWrapper.php
```diff
@@ -70,6 +76,18 @@ public function getFile(): object
         return $this->stream->getFile();
     }
 
+    /** @return false|array */
+    public function url_stat(string $path, int $flags)
+    {
+        try {
+            $this->stream_open($path, 'r', 0, $openedPath);
+        } catch (FileNotFoundException $e) {
```
jmikola: Noted that this relies on `Bucket::resolveStreamContext()` throwing the exception.

Would `url_stat()` only be used for the basic file access with a registered alias? I'm curious what would happen if this was given an internal file path. Presumably `stream_open()` would fail because a context was not provided and the internal path string cannot be used to obtain a context.

It may be worth making a note of this here.

tests/GridFS/FunctionalTestCase.php
```diff
@@ -34,6 +35,13 @@ public function setUp(): void
         $this->filesCollection = $this->createCollection($this->getDatabaseName(), 'fs.files');
     }
 
+    public function tearDown(): void
+    {
+        StreamWrapper::setContextResolver('default', null);
```
jmikola, line 40: Where does "default" come from? In `StreamWrapperFunctionalTest::testStreamWithContextResolver()` you use "bucket".

Is there any reason to have this `tearDown()` logic in this file if aliases are never registered here? Might make more sense to clean up closer to the actual tests that create aliases.

I'd also be in favor of segregating all of the alias-related logic to its own test file, unless that'd make things harder to maintain.

docs/reference/method/MongoDBGridFSBucket-registerGlobalStreamWrapperAlias.txt
```diff
@@ -0,0 +1,80 @@
...
+Files can then be accessed using the following pattern: ``gridfs://<bucket-alias>/<filename>``
+
+Supported stream functions:
+
+- ``file_exists()``
+- ``file_get_contents()``
+- ``file_put_contents()``
+- ``fopen()`` (with "r", "rb", "w", and "wb" modes)
+- ``filesize()``
+- ``filemtime()``
```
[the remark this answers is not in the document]
GromNaN: I added `file()` and `copy()`.

`rename()` and `unlink()` are not supported.

docs/reference/method/MongoDBGridFSBucket-registerGlobalStreamWrapperAlias.txt
```diff
@@ -0,0 +1,80 @@
...
+``$alias`` : array
+  A non-empty string used to identify the GridFS bucket when accessing files
+  using the ``gridfs://`` stream wrapper.
+
+
+
+Behavior
+--------
+
+Files can then be accessed using the following pattern: ``gridfs://<bucket-alias>/<filename>``
```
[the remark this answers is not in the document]
GromNaN: Updated. The revision can be set in the stream context, but fallback to `-1` by default.

docs/reference/method/MongoDBGridFSBucket-registerGlobalStreamWrapperAlias.txt
```diff
@@ -0,0 +1,80 @@
...
+
+   $bucket->registerGlobalStreamWrapperAlias('mybucket');
+
+   var_dump(file_exists('gridfs://mybucket/hello.txt'));
+
+   file_put_contents('gridfs://mybucket/hello.txt', 'Hello, GridFS!');
+
+   var_dump(file_exists('gridfs://mybucket/hello.txt'));
+
+   echo file_get_contents('gridfs://mybucket/hello.txt');
```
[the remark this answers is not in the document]
GromNaN: My fear is that the examples will become very long if extracted into executable files.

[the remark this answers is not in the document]
jmikola: It's possible to pull in only specific lines from a file, but we don't need to worry about this now.

docs/reference/method/MongoDBGridFSBucket-registerGlobalStreamWrapperAlias.txt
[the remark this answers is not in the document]
jmikola: > The revision can be set in the stream context, but fallback to -1 by default.

Ah, I didn't consider that you can still pass in a context via one of the additional arguments on the file functions. Are any of the examples demonstrating that?

src/GridFS/StreamWrapper.php
```diff
@@ -70,6 +76,18 @@ public function getFile(): object
         return $this->stream->getFile();
     }
 
+    /** @return false|array */
+    public function url_stat(string $path, int $flags)
```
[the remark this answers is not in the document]
jmikola: :point_up:

src/GridFS/StreamWrapper.php
jmikola: :point_up:

docs/reference/method/MongoDBGridFSBucket-registerGlobalStreamWrapperAlias.txt
```diff
@@ -0,0 +1,133 @@
...
+- ``filesize()``
+- ``file()``
+- ``fopen()`` (with "r", "rb", "w", and "wb" modes)
+
+In read mode, the stream context can contain the option ``gridfs['revision']``
+to specify the revision number of the file to read. If omitted, the most recent
+revision is read (revision ``-1``).
+
+In write mode, the stream context can contain the options ``gridfs['chunkSizeBytes']``
+and ``gridfs['disableMD5']`` options. If omitted, the defaults are inherited from the
```
jmikola, line 57: The `md5` field is already deprecated ([DRIVERS-2656](https://jira.mongodb.org/browse/DRIVERS-2656)). Is it worth introducing another MD5-related option into a public API at this point?

jmikola (commented):
Two outstanding comments about `url_stat()`, which GitHub thinks are outdated.

docs/reference/method/MongoDBGridFSBucket-registerGlobalStreamWrapperAlias.txt
alcaeus: If we do end up adding an MD5 option, let's deprecate it straight away for removal in 2.0.

GromNaN: I'll remove this option as I don't expect anyone to use it. Should I set it to false by default?

src/GridFS/StreamWrapper.php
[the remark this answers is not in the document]
GromNaN: context options are accepted when using a `gridfs://alias/file` url, with `revision` or `chunkSizeBytes` options.

The internal mecanism is triggered only when `gridfs[collectionWrapper]` is a `CollectionWrapper` instance, which is an internal class. So nobody can trigger the "internal" behavior without using a internal class. I don't see where the breach could be.

src/GridFS/StreamWrapper.php
GromNaN: Yes, it's called only with `file_exists`, `is_file` and such functions.

src/GridFS/StreamWrapper.php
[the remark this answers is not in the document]
jmikola: Thanks for checking. I just realized that the internal `$context` property has no purpose outside of initialization (previously the `init` methods and now `stream_open()` directly).

src/GridFS/Bucket.php
```diff
@@ -821,7 +821,7 @@ private function resolveStreamContext(string $path, string $mode, array $context
                 'filename' => $filename,
                 'options' => $context + [
                     'chunkSizeBytes' => $this->chunkSizeBytes,
-                    'disableMD5' => $this->disableMD5,
+                    'disableMD5' => true,
```
jmikola, line 824: I don't think we need to go this far, and create a divergence between this stream alias API and the Bucket itself. Why not just create a ticket to change the default value of `Bucket::$disableMD5` and target the same release?

Or would that be considered a BC break even though the options API itself remains unchanged?

docs/reference/method/MongoDBGridFSBucket-registerGlobalStreamWrapperAlias.txt
```diff
@@ -0,0 +1,132 @@
...
+
+Behavior
+--------
+
+After registering an alias for the bucket, the most recent revision of a file
+can be accessed using a filename string in the form ``gridfs://<bucket-alias>/<filename>``.
+
+Supported stream functions:
+
+- ``copy()``
```
jmikola, line 43: Sorry to be a bother, but since these are PHP core functions, did you want to link to their docs pages? OK it not, just thought I'd ask in case you weren't aware that we had a shortcut for that.

For instance, this would be:

```
:php:`copy() <copy>`
```

The text in the angle brackets gets appended to `http://php.net/` (see: https://github.com/mongodb/docs-php-library/blob/master/conf.py#L60).

Edit: I just realized the "php" link doesn't use HTTPS. Opened https://github.com/mongodb/docs-php-library/pull/57 to correct that.

jmikola (approved):
I'll defer to you and Andreas to decide if `disableMD5` should be changed.

docs/reference/method/MongoDBGridFSBucket-registerGlobalStreamWrapperAlias.txt
jmikola: > Disabling MD5 by default.

Slight disagree if it's only done for this bucket alias API. See: https://github.com/mongodb/mongo-php-library/pull/1138#discussion_r1428158432

GromNaN: So I kept the option. Just don't document it (that's actually what is in the code).

docs/reference/method/MongoDBGridFSBucket-registerGlobalStreamWrapperAlias.txt
GromNaN: Since it's that simple, I've added the link

src/GridFS/Bucket.php
GromNaN: Reverted the default to the bucket config.

docs/reference/method/MongoDBGridFSBucket-registerGlobalStreamWrapperAlias.txt
```diff
@@ -0,0 +1,132 @@
...
+Supported stream functions:
+
+- :php:`copy() <copy>`
+- :php:`file_exists() <file_exists>`
+- :php:`file_get_contents() <file_get_contents>`
+- :php:`file_put_contents() <file_put_contents>`
+- :php:`filemtime() <filemtime>`
+- :php:`filesize() <filesize>`
+- :php:`file() <file>`
+- :php:`fopen() <fopen>`` (with "r", "rb", "w", and "wb" modes)
```
jmikola, line 50:
```suggestion
- :php:`fopen() <fopen>` (with "r", "rb", "w", and "wb" modes)
```

jmikola:
[This CI failure](https://github.com/mongodb/mongo-php-library/actions/runs/7225817914/job/19690035390?pr=1138#step:9:223) looks relevant:

```
1) MongoDB\Tests\GridFS\StreamWrapperFunctionalTest::testOpenSpecificRevision
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'version 1'
+'version 0'

/home/runner/work/mongo-php-library/mongo-php-library/tests/GridFS/StreamWrapperFunctionalTest.php:300
```

I noticed some [other CI failures](https://spruce.mongodb.com/task/mongo_php_library_test_debian11_php82_local_test_mongodb_latest_replicaset_noauth_nossl_patch_8f562a701c908b6b7e09e344062094d245050fd3_657c9d3232f4174ede56fea5_23_12_15_18_38_44?execution=0&sortBy=STATUS&sortDir=ASC) related to Atlas Search spec tests failing to match error messages, but that's probably related to [DRIVERS-2794](https://jira.mongodb.org/projects/DRIVERS/issues/DRIVERS-2794) as it only fails on mongodb-latest.

tests/GridFS/StreamWrapperFunctionalTest.php
```diff
@@ -204,4 +229,138 @@ public function testWritableStreamWrite(): void
...
+
+    public function testOpenSpecificRevision(): void
+    {
+        $this->bucket->registerGlobalStreamWrapperAlias('bucket');
+
+        $filename = 'gridfs://bucket/path/to/filename';
+
+        // Insert 3 revisions, wait 1ms between each to ensure they have different uploadDate
+        file_put_contents($filename, 'version 0');
+        usleep(1000);
```
GromNaN, line 295: @jmikola Added this 1ms delay to fix https://github.com/mongodb/mongo-php-library/pull/1138#issuecomment-1858364567

alcaeus (approved):

PR #1139, opened 2023-07-24
PHPLIB-1208: Split example namespaces

alcaeus:
PHPLIB-1208

The appstract/laravel-opcache package looks compiles all php files using `opcache_compile_file`. This creates some duplicate function declarations in our example files. Since we want to ship the examples for users' benefit, these files need to be excluded by another measure. ~Luckily, the package [excludes files containing a shebang](https://github.com/appstract/laravel-opcache/blob/d2ce88cddda6af54c14d1f9ceaaf94b54f38f9d3/src/OpcacheClass.php#L61), so adding that and making the examples executable not only fixes the opcache issues, but also makes them a little easier to run.~

The above solution does not work on all PHP versions, as our tests require the files and the addition of the shebang makes the `declare` and `namespace` declarations no longer the first statements in the file. To work around the conflicts observed when compiling for opcache, the examples now use separate namespaces.

### DIFF
--- a/examples/aggregate.php
+++ b/examples/aggregate.php
@@ -1,7 +1,7 @@
 <?php
 declare(strict_types=1);
 
-namespace MongoDB\Examples;
+namespace MongoDB\Examples\Aggregate;
 
 use MongoDB\Client;
 
--- a/examples/bulk.php
+++ b/examples/bulk.php
@@ -1,7 +1,7 @@
 <?php
 declare(strict_types=1);
 
-namespace MongoDB\Examples;
+namespace MongoDB\Examples\Bulk;
 
 use MongoDB\Client;
 use MongoDB\Driver\WriteConcern;
--- a/examples/changestream.php
+++ b/examples/changestream.php
@@ -1,7 +1,7 @@
 <?php
 declare(strict_types=1);
 
-namespace MongoDB\Examples;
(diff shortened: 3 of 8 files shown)


GromNaN (approved):
If nothing else, it's not bad.

IMO, the laravel package should have used the autoload paths from the composer config of each package.

I submitted https://github.com/appstract/laravel-opcache/pull/143 to exclude "examples" directory, that would benefit for all libraries that provide such examples.

examples/aggregate.php
```diff
@@ -1,7 +1,7 @@
 <?php
 declare(strict_types=1);
 
-namespace MongoDB\Examples;
+namespace MongoDB\Examples\Aggregate;
```
jmikola, line 4: Are we going to have the same problem with the [CSFLE examples](https://github.com/mongodb/mongo-php-library/tree/master/docs/examples)? Or perhaps not because those don't define any functions?

alcaeus: Potentially, yes. In this particular case we won't, however I consider this an upstream issue, as blindly compiling every PHP file regardless of whether it's even available to be autoloader is an error. The PR @GromNaN created in the library excludes all files in `examples` folders from compilation, which fixes the issue for good. This is just a temporary workaround to resolve the immediate issue until it can be fixed upstream.

PR #1140, opened 2023-07-24
PHPLIB-1182: Support codec option in operation classes

alcaeus:
PHPLIB-1182

This PR adds support for a `codec` option to the `MongoDB\Collection` class as well as the following operations:
* aggregate
* bulkWrite (for `insertOne` and `replaceOne` operations)
* find
* findOne
* findOneAndDelete (returned documents only)
* findOneAndUpdate (returned documents only)
* findOneAndReplace (replacement document and returned documents)
* insertMany
* insertOne
* replaceOne
* watch

When specifying a codec, the `typeMap` option is ignored entirely, and a `bson` type is applied for the `root` element. The collection class supports a class-level `codec` option, which will be applied when not explicitly specifying a `codec` option to any of the above operations. Specifying a `null` value for the `codec` option allows users to disable using the collection-level codec and instead rely on type maps to deserialise data.

### DIFF
--- a/src/ChangeStream.php
+++ b/src/ChangeStream.php
@@ -18,6 +18,8 @@
 namespace MongoDB;
 
 use Iterator;
+use MongoDB\BSON\Document;
+use MongoDB\Codec\DocumentCodec;
 use MongoDB\Driver\CursorId;
 use MongoDB\Driver\Exception\ConnectionException;
 use MongoDB\Driver\Exception\RuntimeException;
@@ -27,6 +29,7 @@ use MongoDB\Exception\ResumeTokenException;
 use MongoDB\Model\ChangeStreamIterator;
 use ReturnTypeWillChange;
 
+use function assert;
 use function call_user_func;
 use function in_array;
 
@@ -82,15 +85,22 @@ class ChangeStream implements Iterator
      */
     private bool $hasAdvanced = false;
 
+    private ?DocumentCodec $codec;
+
     /**
      * @internal
      *
      * @param ResumeCallable $resumeCallable
      */
-    public function __construct(ChangeStreamIterator $iterator, callable $resumeCallable)
+    public function __construct(ChangeStreamIterator $iterator, callable $resumeCallable, ?DocumentCodec $codec = null)
     {
         $this->iterator = $iterator;
         $this->resumeCallable = $resumeCallable;
+        $this->codec = $codec;
+
+        if ($codec) {
+            $this->iterator->getInnerIterator()->setTypeMap(['root' => 'bson']);
+        }
     }
 
     /**
@@ -100,7 +110,15 @@ class ChangeStream implements Iterator
     #[ReturnTypeWillChange]
     public function current()
     {
-        return $this->iterator->current();
+        $value = $this->iterator->current();
+
+        if (! $this->codec) {
+            return $value;
+        }
+
+        assert($value === null || $value instanceof Document);
+
+        return $this->codec->decodeIfSupported($value);
     }
 
     /** @return CursorId */
@@ -251,6 +269,10 @@ class ChangeStream implements Iterator
         $this->iterator = call_user_func($this->resumeCallable, $this->getResumeToken(), $this->hasAdvanced);
 
         $this->iterator->rewind();
+
+        if ($this->codec) {
+            $this->iterator->getInnerIterator()->setTypeMap(['root' => 'bson']);
+        }
 
         $this->onIteration($this->hasAdvanced);
     }
--- a/src/ChangeStream.php
+++ b/src/ChangeStream.php
@@ -18,7 +18,6 @@
 namespace MongoDB;
 
 use Iterator;
-use MongoDB\BSON\Document;
 use MongoDB\Codec\DocumentCodec;
 use MongoDB\Driver\CursorId;
 use MongoDB\Driver\Exception\ConnectionException;
@@ -29,7 +28,6 @@ use MongoDB\Exception\ResumeTokenException;
 use MongoDB\Model\ChangeStreamIterator;
 use ReturnTypeWillChange;
 
-use function assert;
 use function call_user_func;
 use function in_array;
 
@@ -115,8 +113,6 @@ class ChangeStream implements Iterator
         if (! $this->codec) {
             return $value;
         }
-
-        assert($value === null || $value instanceof Document);
 
         return $this->codec->decodeIfSupported($value);
     }
--- a/src/ChangeStream.php
+++ b/src/ChangeStream.php
@@ -18,6 +18,7 @@
 namespace MongoDB;
 
 use Iterator;
+use MongoDB\BSON\Document;
 use MongoDB\Codec\DocumentCodec;
 use MongoDB\Driver\CursorId;
 use MongoDB\Driver\Exception\ConnectionException;
@@ -28,6 +29,7 @@ use MongoDB\Exception\ResumeTokenException;
 use MongoDB\Model\ChangeStreamIterator;
 use ReturnTypeWillChange;
 
+use function assert;
 use function call_user_func;
 use function in_array;
 
@@ -114,7 +116,9 @@ class ChangeStream implements Iterator
             return $value;
         }
 
-        return $this->codec->decodeIfSupported($value);
+        assert($value instanceof Document);
+
+        return $this->codec->decode($value);
     }
 
     /** @return CursorId */
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -258,9 +258,8 @@ class Collection
      */
     public function bulkWrite(array $operations, array $options = [])
     {
-        $options = $this->inheritWriteOptions(
-            $this->inheritCodec($options),
-        );
+        $options = $this->inheritWriteOptions($options);
+        $options = $this->inheritCodec($options);
 
         $operation = new BulkWrite($this->databaseName, $this->collectionName, $operations, $options);
 
@@ -426,9 +425,8 @@ class Collection
      */
     public function distinct(string $fieldName, $filter = [], array $options = [])
     {
-        $options = $this->inheritReadOptions(
-            $this->inheritTypeMap($options),
-        );
+        $options = $this->inheritReadOptions($options);
+        $options = $this->inheritTypeMap($options);
 
         $operation = new Distinct($this->databaseName, $this->collectionName, $fieldName, $filter, $options);
 
@@ -449,9 +447,8 @@ class Collection
     {
         $server = select_server($this->manager, $options);
 
-        $options = $this->inheritWriteOptions(
-            $this->inheritTypeMap($options),
-        );
+        $options = $this->inheritWriteOptions($options);
+        $options = $this->inheritTypeMap($options);
 
         if (! isset($options['encryptedFields'])) {
             $options['encryptedFields'] = get_encrypted_fields_from_driver($this->databaseName, $this->collectionName, $this->manager)
@@ -484,9 +481,8 @@ class Collection
             throw new InvalidArgumentException('dropIndexes() must be used to drop multiple indexes');
         }
 
-        $options = $this->inheritWriteOptions(
-            $this->inheritTypeMap($options),
-        );
+        $options = $this->inheritWriteOptions($options);
+        $options = $this->inheritTypeMap($options);
 
         $operation = new DropIndexes($this->databaseName, $this->collectionName, $indexName, $options);
 
@@ -505,9 +501,8 @@ class Collection
      */
     public function dropIndexes(array $options = [])
     {
-        $options = $this->inheritWriteOptions(
-            $this->inheritTypeMap($options),
-        );
+        $options = $this->inheritWriteOptions($options);
+        $options = $this->inheritTypeMap($options);
 
         $operation = new DropIndexes($this->databaseName, $this->collectionName, '*', $options);
 
@@ -573,9 +568,8 @@ class Collection
      */
     public function find($filter = [], array $options = [])
     {
-        $options = $this->inheritReadOptions(
-            $this->inheritCodecOrTypeMap($options),
-        );
+        $options = $this->inheritReadOptions($options);
+        $options = $this->inheritCodecOrTypeMap($options);
 
         $operation = new Find($this->databaseName, $this->collectionName, $filter, $options);
 
@@ -596,9 +590,8 @@ class Collection
      */
     public function findOne($filter = [], array $options = [])
     {
-        $options = $this->inheritReadOptions(
-            $this->inheritCodecOrTypeMap($options),
-        );
+        $options = $this->inheritReadOptions($options);
+        $options = $this->inheritCodecOrTypeMap($options);
 
         $operation = new FindOne($this->databaseName, $this->collectionName, $filter, $options);
 
@@ -622,9 +615,8 @@ class Collection
      */
     public function findOneAndDelete($filter, array $options = [])
     {
-        $options = $this->inheritWriteOptions(
-            $this->inheritCodecOrTypeMap($options),
-        );
+        $options = $this->inheritWriteOptions($options);
+        $options = $this->inheritCodecOrTypeMap($options);
 
         $operation = new FindOneAndDelete($this->databaseName, $this->collectionName, $filter, $options);
 
@@ -653,9 +645,8 @@ class Collection
      */
     public function findOneAndReplace($filter, $replacement, array $options = [])
     {
-        $options = $this->inheritWriteOptions(
-            $this->inheritCodecOrTypeMap($options),
-        );
+        $options = $this->inheritWriteOptions($options);
+        $options = $this->inheritCodecOrTypeMap($options);
 
         $operation = new FindOneAndReplace($this->databaseName, $this->collectionName, $filter, $replacement, $options);
 
@@ -684,9 +675,8 @@ class Collection
      */
     public function findOneAndUpdate($filter, $update, array $options = [])
     {
-        $options = $this->inheritWriteOptions(
-            $this->inheritCodecOrTypeMap($options),
-        );
+        $options = $this->inheritWriteOptions($options);
+        $options = $this->inheritCodecOrTypeMap($options);
 
         $operation = new FindOneAndUpdate($this->databaseName, $this->collectionName, $filter, $update, $options);
 
@@ -789,9 +779,8 @@ class Collection
      */
     public function insertMany(array $documents, array $options = [])
     {
-        $options = $this->inheritWriteOptions(
-            $this->inheritCodec($options),
-        );
+        $options = $this->inheritWriteOptions($options);
+        $options = $this->inheritCodec($options);
 
         $operation = new InsertMany($this->databaseName, $this->collectionName, $documents, $options);
 
@@ -811,9 +800,8 @@ class Collection
      */
     public function insertOne($document, array $options = [])
     {
-        $options = $this->inheritWriteOptions(
-            $this->inheritCodec($options),
-        );
+        $options = $this->inheritWriteOptions($options);
+        $options = $this->inheritCodec($options);
 
         $operation = new InsertOne($this->databaseName, $this->collectionName, $document, $options);
 
@@ -872,9 +860,8 @@ class Collection
             $options['readConcern'] = $this->readConcern;
         }
 
-        $options = $this->inheritWriteOptions(
-            $this->inheritTypeMap($options),
-        );
+        $options = $this->inheritWriteOptions($options);
+        $options = $this->inheritTypeMap($options);
 
         $operation = new MapReduce($this->databaseName, $this->collectionName, $map, $reduce, $out, $options);
 
@@ -899,9 +886,8 @@ class Collection
             $toDatabaseName = $this->databaseName;
         }
 
-        $options = $this->inheritWriteOptions(
-            $this->inheritTypeMap($options),
-        );
+        $options = $this->inheritWriteOptions($options);
+        $options = $this->inheritTypeMap($options);
 
         $operation = new RenameCollection($this->databaseName, $this->collectionName, $toDatabaseName, $toCollectionName, $options);
 
@@ -923,9 +909,8 @@ class Collection
      */
     public function replaceOne($filter, $replacement, array $options = [])
     {
-        $options = $this->inheritWriteOptions(
-            $this->inheritCodec($options),
-        );
+        $options = $this->inheritWriteOptions($options);
+        $options = $this->inheritCodec($options);
 
         $operation = new ReplaceOne($this->databaseName, $this->collectionName, $filter, $replacement, $options);
 
@@ -987,9 +972,8 @@ class Collection
      */
     public function watch(array $pipeline = [], array $options = [])
     {
-        $options = $this->inheritReadOptions(
-            $this->inheritCodecOrTypeMap($options),
-        );
+        $options = $this->inheritReadOptions($options);
+        $options = $this->inheritCodecOrTypeMap($options);
 
         $operation = new Watch($this->manager, $this->databaseName, $this->collectionName, $pipeline, $options);
 
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -19,6 +19,7 @@ namespace MongoDB;
 
 use Iterator;
 use MongoDB\BSON\JavascriptInterface;
+use MongoDB\Codec\DocumentCodec;
 use MongoDB\Driver\CursorInterface;
 use MongoDB\Driver\Exception\RuntimeException as DriverRuntimeException;
 use MongoDB\Driver\Manager;
@@ -63,6 +64,7 @@ use MongoDB\Operation\Watch;
 
 use function array_diff_key;
 use function array_intersect_key;
+use function array_key_exists;
 use function current;
 use function is_array;
 use function strlen;
@@ -76,6 +78,8 @@ class Collection
     ];
 
     private const WIRE_VERSION_FOR_READ_CONCERN_WITH_WRITE_STAGE = 8;
+
+    private ?DocumentCodec $codec = null;
 
     private string $collectionName;
 
@@ -98,6 +102,9 @@ class Collection
      * CRUD (i.e. create, read, update, and delete) and index management.
      *
      * Supported options:
+     *
+     *  * codec (MongoDB\Codec\DocumentCodec): Codec used to decode documents
+     *    from BSON to PHP objects.
      *
      *  * readConcern (MongoDB\Driver\ReadConcern): The default read concern to
      *    use for collection operations. Defaults to the Manager's read concern.
@@ -128,6 +135,10 @@ class Collection
             throw new InvalidArgumentException('$collectionName is invalid: ' . $collectionName);
         }
 
+        if (isset($options['codec']) && ! $options['codec'] instanceof DocumentCodec) {
+            throw InvalidArgumentException::invalidType('"codec" option', $options['codec'], DocumentCodec::class);
+        }
+
         if (isset($options['readConcern']) && ! $options['readConcern'] instanceof ReadConcern) {
             throw InvalidArgumentException::invalidType('"readConcern" option', $options['readConcern'], ReadConcern::class);
         }
@@ -147,6 +158,8 @@ class Collection
         $this->manager = $manager;
         $this->databaseName = $databaseName;
         $this->collectionName = $collectionName;
+
+        $this->codec = $options['codec'] ?? null;
         $this->readConcern = $options['readConcern'] ?? $this->manager->getReadConcern();
         $this->readPreference = $options['readPreference'] ?? $this->manager->getReadPreference();
         $this->typeMap = $options['typeMap'] ?? self::DEFAULT_TYPE_MAP;
@@ -162,6 +175,7 @@ class Collection
     public function __debugInfo()
     {
         return [
+            'codec' => $this->codec,
             'collectionName' => $this->collectionName,
             'databaseName' => $this->databaseName,
             'manager' => $this->manager,
@@ -220,6 +234,10 @@ class Collection
             $options['readConcern'] = $this->readConcern;
         }
 
+        if (! array_key_exists('codec', $options)) {
+            $options['codec'] = $this->codec;
+        }
+
         if (! isset($options['typeMap'])) {
             $options['typeMap'] = $this->typeMap;
         }
@@ -246,6 +264,10 @@ class Collection
      */
     public function bulkWrite(array $operations, array $options = [])
     {
+        if (! array_key_exists('codec', $options)) {
+            $options['codec'] = $this->codec;
+        }
+
         if (! isset($options['writeConcern']) && ! is_in_transaction($options)) {
             $options['writeConcern'] = $this->writeConcern;
         }
@@ -626,6 +648,10 @@ class Collection
      */
     public function find($filter = [], array $options = [])
     {
+        if (! array_key_exists('codec', $options)) {
+            $options['codec'] = $this->codec;
+        }
+
         if (! isset($options['readPreference']) && ! is_in_transaction($options)) {
             $options['readPreference'] = $this->readPreference;
         }
@@ -659,6 +685,10 @@ class Collection
      */
     public function findOne($filter = [], array $options = [])
     {
+        if (! array_key_exists('codec', $options)) {
+            $options['codec'] = $this->codec;
+        }
+
         if (! isset($options['readPreference']) && ! is_in_transaction($options)) {
             $options['readPreference'] = $this->readPreference;
         }
@@ -695,6 +725,10 @@ class Collection
      */
     public function findOneAndDelete($filter, array $options = [])
     {
+        if (! array_key_exists('codec', $options)) {
+            $options['codec'] = $this->codec;
+        }
+
         $server = select_server($this->manager, $options);
 
         if (! isset($options['writeConcern']) && ! is_in_transaction($options)) {
@@ -732,6 +766,10 @@ class Collection
      */
     public function findOneAndReplace($filter, $replacement, array $options = [])
     {
+        if (! array_key_exists('codec', $options)) {
+            $options['codec'] = $this->codec;
+        }
+
         $server = select_server($this->manager, $options);
 
         if (! isset($options['writeConcern']) && ! is_in_transaction($options)) {
@@ -769,6 +807,10 @@ class Collection
      */
     public function findOneAndUpdate($filter, $update, array $options = [])
     {
+        if (! array_key_exists('codec', $options)) {
+            $options['codec'] = $this->codec;
+        }
+
         $server = select_server($this->manager, $options);
 
         if (! isset($options['writeConcern']) && ! is_in_transaction($options)) {
@@ -880,6 +922,10 @@ class Collection
      */
     public function insertMany(array $documents, array $options = [])
     {
+        if (! array_key_exists('codec', $options)) {
+            $options['codec'] = $this->codec;
+        }
+
         if (! isset($options['writeConcern']) && ! is_in_transaction($options)) {
             $options['writeConcern'] = $this->writeConcern;
         }
@@ -903,6 +949,10 @@ class Collection
      */
     public function insertOne($document, array $options = [])
     {
+        if (! array_key_exists('codec', $options)) {
+            $options['codec'] = $this->codec;
+        }
+
         if (! isset($options['writeConcern']) && ! is_in_transaction($options)) {
             $options['writeConcern'] = $this->writeConcern;
         }
@@ -1029,6 +1079,10 @@ class Collection
      */
     public function replaceOne($filter, $replacement, array $options = [])
     {
+        if (! array_key_exists('codec', $options)) {
+            $options['codec'] = $this->codec;
+        }
+
         if (! isset($options['writeConcern']) && ! is_in_transaction($options)) {
             $options['writeConcern'] = $this->writeConcern;
         }
@@ -1100,6 +1154,10 @@ class Collection
      */
     public function watch(array $pipeline = [], array $options = [])
     {
+        if (! array_key_exists('codec', $options)) {
+            $options['codec'] = $this->codec;
+        }
+
         if (! isset($options['readPreference']) && ! is_in_transaction($options)) {
             $options['readPreference'] = $this->readPreference;
         }
@@ -1137,6 +1195,7 @@ class Collection
     public function withOptions(array $options = [])
     {
         $options += [
+            'codec' => $this->codec,
             'readConcern' => $this->readConcern,
             'readPreference' => $this->readPreference,
             'typeMap' => $this->typeMap,
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -234,16 +234,10 @@ class Collection
             $options['readConcern'] = $this->readConcern;
         }
 
-        if (! array_key_exists('codec', $options) && ! isset($options['typeMap'])) {
-            $options['codec'] = $this->codec;
-        }
+        $options = $this->inheritCodecOrTypeMap($options);
 
-        if (! isset($options['typeMap'])) {
-            $options['typeMap'] = $this->typeMap;
-        }
-
-        if ($hasWriteStage && ! isset($options['writeConcern']) && ! is_in_transaction($options)) {
-            $options['writeConcern'] = $this->writeConcern;
+        if ($hasWriteStage) {
+            $options = $this->inheritWriteOptions($options);
         }
 
         $operation = new Aggregate($this->databaseName, $this->collectionName, $pipeline, $options);
@@ -264,13 +258,9 @@ class Collection
      */
     public function bulkWrite(array $operations, array $options = [])
     {
-        if (! array_key_exists('codec', $options) && ! isset($options['typeMap'])) {
-            $options['codec'] = $this->codec;
-        }
-
-        if (! isset($options['writeConcern']) && ! is_in_transaction($options)) {
-            $options['writeConcern'] = $this->writeConcern;
-        }
+        $options = $this->inheritWriteOptions(
+            $this->inheritCodec($options),
+        );
 
         $operation = new BulkWrite($this->databaseName, $this->collectionName, $operations, $options);
         $server = select_server($this->manager, $options);
@@ -294,15 +284,9 @@ class Collection
      */
     public function count($filter = [], array $options = [])
     {
-        if (! isset($options['readPreference']) && ! is_in_transaction($options)) {
-            $options['readPreference'] = $this->readPreference;
-        }
+        $options = $this->inheritReadOptions($options);
 
         $server = select_server($this->manager, $options);
-
-        if (! isset($options['readConcern']) && ! is_in_transaction($options)) {
-            $options['readConcern'] = $this->readConcern;
-        }
 
         $operation = new Count($this->databaseName, $this->collectionName, $filter, $options);
 
@@ -323,15 +307,9 @@ class Collection
      */
     public function countDocuments($filter = [], array $options = [])
     {
-        if (! isset($options['readPreference']) && ! is_in_transaction($options)) {
-            $options['readPreference'] = $this->readPreference;
-        }
+        $options = $this->inheritReadOptions($options);
 
         $server = select_server($this->manager, $options);
-
-        if (! isset($options['readConcern']) && ! is_in_transaction($options)) {
-            $options['readConcern'] = $this->readConcern;
-        }
 
         $operation = new CountDocuments($this->databaseName, $this->collectionName, $filter, $options);
 
@@ -391,9 +369,7 @@ class Collection
     {
         $server = select_server($this->manager, $options);
 
-        if (! isset($options['writeConcern']) && ! is_in_transaction($options)) {
-            $options['writeConcern'] = $this->writeConcern;
-        }
+        $options = $this->inheritWriteOptions($options);
 
         $operation = new CreateIndexes($this->databaseName, $this->collectionName, $indexes, $options);
 
@@ -414,9 +390,7 @@ class Collection
      */
     public function deleteMany($filter, array $options = [])
     {
-        if (! isset($options['writeConcern']) && ! is_in_transaction($options)) {
-            $options['writeConcern'] = $this->writeConcern;
-        }
+        $options = $this->inheritWriteOptions($options);
 
         $operation = new DeleteMany($this->databaseName, $this->collectionName, $filter, $options);
         $server = select_server($this->manager, $options);
@@ -438,9 +412,7 @@ class Collection
      */
     public function deleteOne($filter, array $options = [])
     {
-        if (! isset($options['writeConcern']) && ! is_in_transaction($options)) {
-            $options['writeConcern'] = $this->writeConcern;
-        }
+        $options = $this->inheritWriteOptions($options);
 
         $operation = new DeleteOne($this->databaseName, $this->collectionName, $filter, $options);
         $server = select_server($this->manager, $options);
@@ -463,19 +435,11 @@ class Collection
      */
     public function distinct(string $fieldName, $filter = [], array $options = [])
     {
-        if (! isset($options['readPreference']) && ! is_in_transaction($options)) {
-            $options['readPreference'] = $this->readPreference;
-        }
-
-        if (! isset($options['typeMap'])) {
-            $options['typeMap'] = $this->typeMap;
-        }
+        $options = $this->inheritReadOptions(
+            $this->inheritTypeMap($options),
+        );
 
         $server = select_server($this->manager, $options);
-
-        if (! isset($options['readConcern']) && ! is_in_transaction($options)) {
-            $options['readConcern'] = $this->readConcern;
-        }
 
         $operation = new Distinct($this->databaseName, $this->collectionName, $fieldName, $filter, $options);
 
@@ -494,15 +458,11 @@ class Collection
      */
     public function drop(array $options = [])
     {
-        if (! isset($options['typeMap'])) {
-            $options['typeMap'] = $this->typeMap;
-        }
-
         $server = select_server($this->manager, $options);
 
-        if (! isset($options['writeConcern']) && ! is_in_transaction($options)) {
-            $options['writeConcern'] = $this->writeConcern;
-        }
+        $options = $this->inheritWriteOptions(
+            $this->inheritTypeMap($options),
+        );
 
         if (! isset($options['encryptedFields'])) {
             $options['encryptedFields'] = get_encrypted_fields_from_driver($this->databaseName, $this->collectionName, $this->manager)
@@ -535,15 +495,11 @@ class Collection
             throw new InvalidArgumentException('dropIndexes() must be used to drop multiple indexes');
         }
 
-        if (! isset($options['typeMap'])) {
-            $options['typeMap'] = $this->typeMap;
-        }
-
         $server = select_server($this->manager, $options);
 
-        if (! isset($options['writeConcern']) && ! is_in_transaction($options)) {
-            $options['writeConcern'] = $this->writeConcern;
-        }
+        $options = $this->inheritWriteOptions(
+            $this->inheritTypeMap($options),
+        );
 
         $operation = new DropIndexes($this->databaseName, $this->collectionName, $indexName, $options);
 
@@ -562,15 +518,11 @@ class Collection
      */
     public function dropIndexes(array $options = [])
     {
-        if (! isset($options['typeMap'])) {
-            $options['typeMap'] = $this->typeMap;
-        }
-
         $server = select_server($this->manager, $options);
 
-        if (! isset($options['writeConcern']) && ! is_in_transaction($options)) {
-            $options['writeConcern'] = $this->writeConcern;
-        }
+        $options = $this->inheritWriteOptions(
+            $this->inheritTypeMap($options),
+        );
 
         $operation = new DropIndexes($this->databaseName, $this->collectionName, '*', $options);
 
@@ -590,15 +542,9 @@ class Collection
      */
     public function estimatedDocumentCount(array $options = [])
     {
-        if (! isset($options['readPreference']) && ! is_in_transaction($options)) {
-            $options['readPreference'] = $this->readPreference;
-        }
+        $options = $this->inheritReadOptions($options);
 
         $server = select_server($this->manager, $options);
-
-        if (! isset($options['readConcern']) && ! is_in_transaction($options)) {
-            $options['readConcern'] = $this->readConcern;
-        }
 
         $operation = new EstimatedDocumentCount($this->databaseName, $this->collectionName, $options);
 
@@ -623,9 +569,7 @@ class Collection
             $options['readPreference'] = $this->readPreference;
         }
 
-        if (! isset($options['typeMap'])) {
-            $options['typeMap'] = $this->typeMap;
-        }
+        $options = $this->inheritTypeMap($options);
 
         $server = select_server($this->manager, $options);
 
@@ -648,23 +592,11 @@ class Collection
      */
     public function find($filter = [], array $options = [])
     {
-        if (! array_key_exists('codec', $options) && ! isset($options['typeMap'])) {
-            $options['codec'] = $this->codec;
-        }
-
-        if (! isset($options['readPreference']) && ! is_in_transaction($options)) {
-            $options['readPreference'] = $this->readPreference;
-        }
+        $options = $this->inheritReadOptions(
+            $this->inheritCodecOrTypeMap($options),
+        );
 
         $server = select_server($this->manager, $options);
-
-        if (! isset($options['readConcern']) && ! is_in_transaction($options)) {
-            $options['readConcern'] = $this->readConcern;
-        }
-
-        if (! isset($options['typeMap'])) {
-            $options['typeMap'] = $this->typeMap;
-        }
 
         $operation = new Find($this->databaseName, $this->collectionName, $filter, $options);
 
@@ -685,23 +617,11 @@ class Collection
      */
     public function findOne($filter = [], array $options = [])
     {
-        if (! array_key_exists('codec', $options) && ! isset($options['typeMap'])) {
-            $options['codec'] = $this->codec;
-        }
-
-        if (! isset($options['readPreference']) && ! is_in_transaction($options)) {
-            $options['readPreference'] = $this->readPreference;
-        }
+        $options = $this->inheritReadOptions(
+            $this->inheritCodecOrTypeMap($options),
+        );
 
         $server = select_server($this->manager, $options);
-
-        if (! isset($options['readConcern']) && ! is_in_transaction($options)) {
-            $options['readConcern'] = $this->readConcern;
-        }
-
-        if (! isset($options['typeMap'])) {
-            $options['typeMap'] = $this->typeMap;
-        }
 
         $operation = new FindOne($this->databaseName, $this->collectionName, $filter, $options);
 
@@ -725,19 +645,11 @@ class Collection
      */
     public function findOneAndDelete($filter, array $options = [])
     {
-        if (! array_key_exists('codec', $options) && ! isset($options['typeMap'])) {
-            $options['codec'] = $this->codec;
-        }
-
         $server = select_server($this->manager, $options);
 
-        if (! isset($options['writeConcern']) && ! is_in_transaction($options)) {
-            $options['writeConcern'] = $this->writeConcern;
-        }
-
-        if (! isset($options['typeMap'])) {
-            $options['typeMap'] = $this->typeMap;
-        }
+        $options = $this->inheritWriteOptions(
+            $this->inheritCodecOrTypeMap($options),
+        );
 
         $operation = new FindOneAndDelete($this->databaseName, $this->collectionName, $filter, $options);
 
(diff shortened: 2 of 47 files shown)


src/Collection.php
```diff
@@ -219,6 +234,10 @@ public function aggregate(array $pipeline, array $options = [])
             $options['readConcern'] = $this->readConcern;
         }
 
+        if (! array_key_exists('codec', $options)) {
```
GromNaN, line 237: I understand that `array_key_exists` is used to be able to remove the default codec option.

src/Collection.php
```diff
@@ -188,7 +203,7 @@ public function __toString()
      * @see Aggregate::__construct() for supported options
      * @param array $pipeline Aggregation pipeline
      * @param array $options  Command options
-     * @return Cursor
+     * @return CursorInterface&Iterator
```
GromNaN, line 206: This return type change is a breaking change when using static analysis tools. The user code can have type hint on the returned value.
But the returned value type will not change until the `codec` option is set. I don't think is can break any code.

Example of code that could break if a codec is provided:
```php
// Type hint
function process(Cursor $cursor) {
    // ...
}

/** @var Cursor is currently deducted from Collection::aggregate return type */
$result = $collection->aggregate([]);

process($result);
```

src/GridFS/ReadableStream.php
```diff
@@ -47,7 +48,8 @@ class ReadableStream
 
     private int $chunkOffset = 0;
 
-    private ?Cursor $chunksIterator = null;
+    /** @var (CursorInterface&Iterator)|null */
+    private $chunksIterator = null;
```
GromNaN, line 52: You can set one of the union type or does that make a psalm error?
```suggestion
    private Iterator $chunksIterator = null;
```

src/Model/ChangeStreamIterator.php
```diff
@@ -244,9 +246,19 @@ private function extractResumeToken($document)
             return $this->extractResumeToken($document->bsonSerialize());
         }
 
-        $resumeToken = is_array($document)
-            ? ($document['_id'] ?? null)
-            : ($document->_id ?? null);
+        if ($document instanceof Document) {
+            $resumeToken = $document->has('_id')
```
GromNaN, line 250: I don't find the documentation on the behavior of `Document::get('_id')` when the key is not set? Reading your code I think it throw an error and doesn't return `null` like a a access to an undefined property of an object.

src/Model/ChangeStreamIterator.php
```diff
@@ -244,9 +246,19 @@ private function extractResumeToken($document)
...
+                ? $document->get('_id')
+                : null;
+
+            if ($resumeToken instanceof Document) {
+                $resumeToken = $resumeToken->toPHP();
```
GromNaN, line 255: It seems you've considered several situations when writing this code, leading to multiple code execution paths. Are they all tested?

src/Model/ChangeStreamIterator.php
```diff
@@ -244,9 +246,19 @@ private function extractResumeToken($document)
             return $this->extractResumeToken($document->bsonSerialize());
         }
 
-        $resumeToken = is_array($document)
-            ? ($document['_id'] ?? null)
-            : ($document->_id ?? null);
+        if ($document instanceof Document) {
```
alcaeus, line 249: This is to support `watch` operations with a codec supplied, so we can avoid messing with the type map on every iteration.

src/Model/ChangeStreamIterator.php
```diff
@@ -244,9 +246,19 @@ private function extractResumeToken($document)
...
+            $resumeToken = $document->has('_id')
+                ? $document->get('_id')
+                : null;
+
+            if ($resumeToken instanceof Document) {
+                $resumeToken = $resumeToken->toPHP();
+            }
```
alcaeus, line 256: Passing a `Document` instance as resume token would be valid, but not converting it to PHP causes test failures. In our tests, we extract the resume token from the command response, which is always in PHP format. To avoid issues when testing watch operations with a codec, this is always converted to a PHP value here as well.

src/Model/CodecCursor.php
```diff
@@ -0,0 +1,126 @@
...
+
+    public function rewind(): void
+    {
+        $this->current = null;
+        $this->cursor->rewind();
+    }
+
+    public function setTypeMap(array $typemap): void
+    {
+        // Not supported
```
alcaeus, line 106: I decided to make this a no-op, as throwing an exception would most violate the Liskov substitution principle.

src/Model/ChangeStreamIterator.php
alcaeus: `get` throws an exception if the key wasn't found. I'm not sure if we can make the assumption that an `_id` is always present, as users can pass their own pipeline to `watch` which may remove the `_id` field from results.

src/Collection.php
alcaeus: Yes, that's a downside of the current change. Unfortunately, there is no way to change this without either removing `final` from our cursor implementation or making this small change. Since the behaviour is opt-in, I figured the change was benign enough to make in a minor release.

src/Operation/BulkWrite.php
```diff
@@ -330,13 +341,31 @@ public function execute(Server $server)
                     break;
 
                 case self::INSERT_ONE:
-                    $insertedIds[$i] = $bulk->insert($args[0]);
+                    $insertedDocument = isset($this->options['codec'])
+                        ? $this->options['codec']->encodeIfSupported($args[0])
+                        : $args[0];
+                    // Psalm's assert-if-true annotation does not work with unions, so
+                    // assert the type manually instead of using is_document
```
jmikola, line 348: Is there an upstream bug report we want to reference here?

src/Operation/Find.php
```diff
@@ -313,6 +323,10 @@ public function execute(Server $server)
 
         $cursor = $server->executeQuery($this->databaseName . '.' . $this->collectionName, new Query($this->filter, $this->createQueryOptions()), $this->createExecuteOptions());
 
+        if (isset($this->options['codec'])) {
+            return CodecCursor::fromCursor($cursor, $this->options['codec']);
+        }
+
         if (isset($this->options['typeMap'])) {
```
jmikola, line 330: The PR description says:

> When specifying a codec, the typeMap option is ignored entirely, and a bson type is applied for the root element.

Does it make sense to throw if both options are specified? I wouldn't consider that a BC break since we're introducing a new option anyway. Seems like it would be a safe way to avoid users doing something incorrect (as we do for some other operations by warning about conflicting options).

Also, I noticed there are no docs in this PR. Did you intend to lump those changes together or handle it later? If later, please add `documentation-needed` to PHPLIB-1182.

src/Operation/FindAndModify.php
```diff
@@ -243,6 +253,17 @@ public function execute(Server $server)
 
         $cursor = $server->executeWriteCommand($this->databaseName, new Command($this->createCommandDocument()), $this->createOptions());
 
+        if (isset($this->options['codec'])) {
+            $cursor->setTypeMap(['root' => 'bson']);
+            $result = current($cursor->toArray());
+            assert($result instanceof Document);
+
+            $decoded = $this->options['codec']->decodeIfSupported($result->get('value') ?? null);
```
jmikola, line 261: Note that a bit later in this method we do `$result->value ?? null` to gracefully handle a case where `value` does not exist. I'm not sure what the point of a [null-coalescing operator](https://www.php.net/manual/en/language.operators.comparison.php#language.operators.comparison.coalesce) is for `Document::get()`, since the method throws if the field doesn't exist.

[findAndModify: value](https://www.mongodb.com/docs/manual/reference/command/findAndModify/#std-label-findandmodify-value-field) suggests that the field will always be either a BSON object or null, so perhaps this was only needed for older server versions. If we're not able to confirm that, I might suggest adding some common code to gracefully extract `value` here.

I don't think we can do it for a common case earlier in this method because the `typeMap` option handling requires application to the cursor before we access the value.

src/Operation/FindOneAndReplace.php
```diff
@@ -147,10 +158,17 @@ public function __construct(string $databaseName, string $collectionName, $filte
 
         unset($options['projection'], $options['returnDocument']);
 
+        $replacementDocument = isset($options['codec'])
+            ? $options['codec']->encodeIfSupported($replacement)
+            : $replacement;
+        // Psalm's assert-if-true annotation does not work with unions, so
+        // assert the type manually instead of using is_document
+        assert(is_array($replacementDocument) || is_object($replacementDocument));
```
jmikola, line 166: Consider the following:

```php
if (isset($options['codec'])) {
    $replacement = $options['codec']->encodeIfSupported($replacement);
}

// Psalm's assert-if-true annotation does not work with unions, so
// assert the type manually instead of using is_document
assert(is_array($replacement) || is_object($replacement));
```

You could also conditionally move the `assert()` into the condition if that's all that's needed to satisfy Psalm. I would imagine so, as we already check `$replacement` earlier in the method.

----

I'm suggesting this because we don't really use `$replacement` for anything else, so I didn't see the harm in filtering/transforming the original value. This would apply in other contexts where you're modifying things, as well.

Perhaps not BulkWrite, as an explicitly named variable makes more sense there than `$args[i]`.

src/Operation/InsertMany.php
```diff
@@ -142,7 +153,14 @@ public function execute(Server $server)
...
         foreach ($this->documents as $i => $document) {
-            $insertedIds[$i] = $bulk->insert($document);
+            $insertedDocument = isset($this->options['codec'])
+                ? $this->options['codec']->encodeIfSupported($document)
+                : $document;
+            // Psalm's assert-if-true annotation does not work with unions, so
+            // assert the type manually instead of using is_document
+            assert(is_array($insertedDocument) || is_object($insertedDocument));
+
+            $insertedIds[$i] = $bulk->insert($insertedDocument);
```
jmikola, line 163: Similar to my suggestion in FindOneAndReplace:

```suggestion
            if (isset($this->options['codec'])) {
                $document = $this->options['codec']->encodeIfSupported($document);
                // Psalm's assert-if-true annotation does not work with unions, so
                // assert the type manually instead of using is_document
                assert(is_array($document) || is_object($document));
            }
            
            $insertedIds[$i] = $bulk->insert($document);
```

Likewise in InsertOne below.

src/Operation/Watch.php
```diff
@@ -314,6 +326,7 @@ public function execute(Server $server)
         return new ChangeStream(
             $this->createChangeStreamIterator($server),
             fn ($resumeToken, $hasAdvanced): ChangeStreamIterator => $this->resume($resumeToken, $hasAdvanced),
+            $this->codec,
```
jmikola, line 329: `createChangeStreamIterator($server)` invokes `executeAggregate()`, which executes the Aggregate object created from the `$aggregateOptions` property. Notably, `$aggregateOptions` is never given the codec.

What's the reasoning for only passing it to the ChangeStream object? I noted that the ChangeStream applies the codec to its inner iterator (i.e. Aggregate cursor). Is it worth having the Aggregate operation apply the option itself since it already needs to support it? Or might that pose problems for accessing the resume token?

src/Model/ChangeStreamIterator.php
```diff
@@ -69,9 +71,9 @@ class ChangeStreamIterator extends IteratorIterator implements CommandSubscriber
     /**
      * @internal
      * @param array|object|null $initialResumeToken
-     * @psalm-param Cursor<TValue> $cursor
+     * @psalm-param CursorInterface<int, TValue>&Iterator<int, TValue> $cursor
```
jmikola, line 74: Would you want to `assert()` on both types in the constructor? That would be temporary until we could utilize [intersection types](https://php.watch/versions/8.1/intersection-types) in 8.1+.

src/Collection.php
```diff
@@ -658,6 +685,10 @@ public function find($filter = [], array $options = [])
      */
     public function findOne($filter = [], array $options = [])
     {
+        if (! array_key_exists('codec', $options)) {
+            $options['codec'] = $this->codec;
+        }
```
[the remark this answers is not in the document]
alcaeus: I'll take a look at how we can support this. For operations that return documents (either in a Cursor or a single document) this may be feasible, but for other operations like `replace` and `insert` that encode incoming values it's a little more complicated. Incoming values are passed in the constructor to the parent operation, so the middleware can't just wrap an existing operation and modify its return value. Furthermore, the type map option needs to be updated accordingly, which means that the middleware would have to be the one creating the original operation in order to encode incoming values and update the type map accordingly.

src/Operation/Find.php
alcaeus: I wasn't sure whether to lump the docs in here or not. I've created PHPLIB-1209 to track this explicitly.

src/Operation/InsertMany.php
alcaeus: Updated.

src/Operation/Watch.php
alcaeus: If the `codec` option is passed to the `Aggregate` operation, the change stream iterator will receive the result of the codec operation. Once that happens, we can no longer make any assumptions about fields (which is why I purposefully used `id` as property for the `_id` field in the tests). The change stream iterator however needs to be able to extract the `_id` field to store it as a resume token. To allow for this, we don't pass the `codec` to the `Aggregate` operation, but set a `bson` type map on the resulting cursor if a codec is present. This allows us to extract the `_id` from the resulting document and then passing it on to the codec before returning it to the user.

I'll add that this might already be a problem if we pass a type map that results in an object that no longer allows accessing an `_id` property, but we haven't considered this so far.

---

While working on the `watch` operation, I was considering whether the codec given to the operation should be applied on the root level or whether we should wrap it in a special `ChangeStreamCodec` to ensure that we're handling the default document structure gracefully. I decided against this as users are able to provide their own pipeline, so we again can't make any assumptions about the documents returned in a change stream.

So, instead of not supporting codecs in `Watch` at all, I decided to implement them this way to allow users the flexibility of using them if they would like to. Given the complexity around change streams (especially with multiple events being returned that all have a different structure), I don't think users will be doing so very often, so I'm open to reconsider our approach regarding `Watch`.

src/Model/ChangeStreamIterator.php
alcaeus: Good point. I've added a check that throws an `InvalidArgumentException`.

src/Operation/FindAndModify.php
alcaeus: That was a brainfart on my part. I've removed the null-coalesce from this.

src/Collection.php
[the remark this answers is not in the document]
jmikola: Oh, I wasn't actually suggesting middleware to handle documents on the way in/out. It was more about wrapping operations with common behaviors, such as inheriting an option or doing transaction checks. So much of the code in the helpers is repeated option handling and it's mostly pre-execution since we often return the result of `execute()` directly.

src/Operation/Find.php
jmikola: What about prohibiting `codec` and `typeMap`?

src/Operation/Watch.php
jmikola: > I'll add that this might already be a problem if we pass a type map that results in an object that no longer allows accessing an _id property, but we haven't considered this so far.

Good point. Happy to wait until that bug report comes in, if it ever does.

----

Quoting the [Change Stream spec](https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst#helper-method):

> The server will produce an error if `_id` is projected out, but a user should otherwise be able to modify the shape of the change stream event as desired.

I don't think we need to worry about the `_id` not existing; however, it becoming inaccessible due to a `typeMap` is a legitimate (albeit unlikely) concern.

src/Operation/BulkWrite.php
alcaeus: Added a reference to https://github.com/vimeo/psalm/issues/6831

src/GridFS/ReadableStream.php
alcaeus: Using `?Iterator` works as expected.

src/Model/ChangeStreamIterator.php
alcaeus: Ok, correction based on Jeremy's comment below: the `_id` field will always be present, so I've removed the check for the field.

I'll note that `get` throws an exception instead of returning `null` to allow differentiating between a missing field and a field with an explicit `null` value.

src/Operation/Watch.php
alcaeus: Agree about not worrying about this until it comes in. Since Codecs might potentially make type maps obsolete, there's a very good chance that this won't happen at all.

src/Collection.php
```diff
@@ -127,6 +135,10 @@ public function __construct(Manager $manager, string $databaseName, string $coll
             throw new InvalidArgumentException('$collectionName is invalid: ' . $collectionName);
         }
 
+        if (isset($options['codec']) && ! $options['codec'] instanceof DocumentCodec) {
```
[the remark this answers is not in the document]
alcaeus: Since a codec library might return something other than an `object` instance, this requires a `DocumentCodec`. While there technically are ways to support multiple document codecs in a library, this would require a specific library that only accepts `DocumentCodec` instances. For this first iteration, I decided to only allow a `DocumentCodec` instance, as a potential future `DocumentCodecLibrary` would always implement this interface, so this is forward compatible with such a change.

src/Collection.php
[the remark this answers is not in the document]
alcaeus: Ah, I see, so I misunderstood. I believe the key here would be to have distinct classes for options as other drivers already have them. This would allow us to use types to cut down on such checks and also add custom logic to each options class while keeping common logic reusable using traits. I can draft such a change so we know the impact and can decide when and how to make it.

src/Operation/Find.php
alcaeus: I added logic to prohibit providing both a type map and a non-null codec. I also used the oppoertunity to de-duplicate some of the inheritance logic in the Collection class.

src/ChangeStream.php
```diff
@@ -100,7 +110,15 @@ public function __construct(ChangeStreamIterator $iterator, callable $resumeCall
...
     public function current()
     {
-        return $this->iterator->current();
+        $value = $this->iterator->current();
+
+        if (! $this->codec) {
+            return $value;
+        }
+
+        assert($value === null || $value instanceof Document);
```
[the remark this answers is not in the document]
alcaeus: That is correct - the need for this assertion went away so I've removed it.

src/Collection.php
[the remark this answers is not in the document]
jmikola: SGTM. I've long wondered if there was a use case for using something like Symfony's [OptionsResolver component](https://symfony.com/doc/current/components/options_resolver.html) but was worried the overhead might not be worth it.

I definitely think there's some room for improvement vs. what we have now, though. Would be helpful to revisit after get benchmarks in place and can do some profiling.

src/Model/ChangeStreamIterator.php
```diff
@@ -140,10 +150,10 @@ public function current()
      * but it's very much an invalid use-case. This method can be dropped in 2.0
      * once the class is final.
      */
-    final public function getInnerIterator(): Cursor
+    final public function getInnerIterator(): CursorInterface
```
jmikola, line 153: Since the constructor was changed to require a `CursorInterface<int, TValue>&Iterator<int, TValue>` should we use Psalm to document that return type as well? I understand we can't use actual PHP syntax to do so yet.

Likewise, does it make sense to assert that the inner iterator is _both_ CursorInterface and Iterator below?

src/Model/ChangeStreamIterator.php
jmikola: @alcaeus: Did you address this comment?

src/Model/CodecCursor.php
jmikola: Would `E_USER_WARNING` be appropriate? If this is a new class that only becomes relevant when using codecs, I'm curious how users would hit this unless they had already modified their application to use codecs. In that sense, this doesn't seem much different than changing the return type elsewhere from `Cursor` to `CursorInterface&Iterator` and not worrying about the theoretical (but impossible in practice?) BC break.

src/Operation/FindOneAndReplace.php
```diff
@@ -147,6 +158,15 @@ public function __construct(string $databaseName, string $collectionName, $filte
...
         unset($options['projection'], $options['returnDocument']);
 
+        if (isset($options['codec'])) {
+            $replacement = $options['codec']->encodeIfSupported($replacement);
+        }
+
+        // Psalm's assert-if-true annotation does not work with unions, so
+        // assert the type manually instead of using is_document
+        // See https://github.com/vimeo/psalm/issues/6831
+        assert(is_array($replacement) || is_object($replacement));
```
jmikola, line 168: We already do `if (! is_document($replacement))` earlier in this method, so I think this assertion can be moved within `if (isset($options['codec']))`. I assume it's only needed to assert the result of `encodeIfSupported()`, correct?

I think that was in my original suggestion but overlooked.

src/Operation/BulkWrite.php
```diff
@@ -330,13 +341,33 @@ public function execute(Server $server)
...
+                    // See https://github.com/vimeo/psalm/issues/6831
+                    assert(is_array($insertedDocument) || is_object($insertedDocument));
```
jmikola, line 350: Similar to my suggestion in FindOneAndReplace, I think this assertion is only relevant for `if (isset($this->options['codec']))`. The constructor already checks `is_document($args[0])` for all code paths, and `is_document($args[1])` for replaceOne.

In this case, I don't object to something like:

```php
if (isset($this->options['codec'])) {
    $args[0] = $this->options['codec']->encodeIfSupported($args[0]);
    // Psalm comment...
    assert(is_array($args[0]) || is_object($args[0]));
}
```

And likewise for replaceOne below. Doing so would also allow us to keep the fall through to a common `$bulk->update($args[0], $args[1], $args[2]);`. I'd suggest adding a `// Fall through` comment after the `REPLACE_ONE` case like we do in the test suite in various classes that deal with spec test operations.

src/Operation/InsertMany.php
```diff
@@ -142,6 +153,15 @@ public function execute(Server $server)
...
 
         foreach ($this->documents as $i => $document) {
+            if (isset($this->options['codec'])) {
+                $document = $this->options['codec']->encodeIfSupported($document);
+            }
+
+            // Psalm's assert-if-true annotation does not work with unions, so
+            // assert the type manually instead of using is_document
+            // See https://github.com/vimeo/psalm/issues/6831
+            assert(is_array($document) || is_object($document));
```
jmikola, line 163: I think this assertion can be moved within `if (isset($this->options['codec']))` above.

src/Operation/InsertOne.php
```diff
@@ -94,6 +105,15 @@ public function __construct(string $databaseName, string $collectionName, $docum
             unset($options['writeConcern']);
         }
 
+        if (isset($options['codec'])) {
+            $document = $options['codec']->encodeIfSupported($document);
+
+            // Psalm's assert-if-true annotation does not work with unions, so
+            // assert the type manually instead of using is_document
+            // See https://github.com/vimeo/psalm/issues/6831
+            assert(is_array($document) || is_object($document));
```
jmikola, line 114: Interesting that you remembered to nest the assertion within the `if` condition here.

src/Operation/ReplaceOne.php
```diff
@@ -96,6 +103,23 @@ public function __construct(string $databaseName, string $collectionName, $filte
...
+        if (isset($options['codec']) && ! $options['codec'] instanceof DocumentCodec) {
+            throw InvalidArgumentException::invalidType('"codec" option', $options['codec'], DocumentCodec::class);
+        }
+
+        if (isset($options['codec'])) {
+            if (isset($options['typeMap'])) {
+                throw InvalidArgumentException::cannotCombineCodecAndTypeMap();
+            }
+
+            $replacement = $options['codec']->encodeIfSupported($replacement);
```
jmikola, line 115: I'll note that this differs from BulkWrite, which calls `encodeIfSupported()` at execution time. I understand that's necessary because we're just delegating to a common Update operation here, but it may be worth a comment.

Also, I think the assertion below can be moved within this `if (isset($options['codec']))` condition.

tests/Fixtures/Document/TestObject.php
```diff
@@ -0,0 +1,39 @@
...
+
+final class TestObject
+{
+    public int $id;
+
+    public TestNestedObject $x;
+
+    public bool $decoded;
+
+    public static function createForFixture(int $i, bool $decoded = false): self
```
jmikola, line 28: Would it make more sense to rename `$i` to `$id` here?

tests/Fixtures/Document/TestObject.php
```diff
@@ -0,0 +1,39 @@
...
+    {
+        $instance = new self();
+        $instance->id = $i;
+        $instance->decoded = $decoded;
+
+        $instance->x = new TestNestedObject();
+        $instance->x->foo = 'bar';
```
jmikola, line 35: I assume the `x` property and its value is not meant to be configurable, or you could always create a new factory method if needed?

tests/Fixtures/Document/TestObject.php
```diff
@@ -0,0 +1,39 @@
...
+
+    public TestNestedObject $x;
+
+    public bool $decoded;
+
+    public static function createForFixture(int $i, bool $decoded = false): self
+    {
+        $instance = new self();
+        $instance->id = $i;
+        $instance->decoded = $decoded;
```
jmikola, line 32: What is the point of this property? Looking at TestDocumentCodec, it looks like the `decode()` method always sets this to `true`. I assume `false` is a default here because you're constructing via a factory method and the resulting object _is not_ the result of `decode()`.

I noted that the use case for calling `createForFixture()` with `$decoded=true` is when asserting decoded documents for an operation. Rather than overload `createForFixture()`, I'd suggest creating a separate factory method that invokes `createForFixture()` and _then_ assigns `$decoded` accordingly. Consider:

```
public static function createDecodedForFixture($int $id): self
{
    $instance = self::createForFixture($id);
    $instance->decoded = true;
    return $instance;
}
```

Then the invocations would be self-documenting.

Along with this, I'd then suggest defaulting the `$decoded` property to `false` in its declaration.

----

I also see that the codec's `encode()` method sets an `$encoded` property on the returned Document instance. Does that pose no problem for PHP 8.2 prohibiting [dynamic property assignment](https://php.watch/versions/8.2/dynamic-properties-deprecated) because we never end up with a TestObject containing an `encoded` property?

tests/Operation/FindOneAndReplaceFunctionalTest.php
```diff
@@ -0,0 +1,66 @@
...
+     */
+    private function createFixtures(int $n): void
+    {
+        $bulkWrite = new BulkWrite(['ordered' => true]);
+
+        for ($i = 1; $i <= $n; $i++) {
+            $bulkWrite->insert([
+                '_id' => $i,
+                'x' => (object) ['foo' => 'bar'],
+            ]);
```
jmikola, line 59: This seems closely coupled with TestObject. Did you consider inserting `TestObject::createForFixture($i)` here instead? That may require a Serializable implementation to ensure `$id` gets mapped to an `_id` field.

I'm not sure if there are other places where you've coupled the structure of TestObject in a fixture method, but the same suggestion would apply there as well if so.

tests/Operation/InsertManyFunctionalTest.php
```diff
@@ -200,4 +202,48 @@ public function testUnacknowledgedWriteConcernAccessesInsertedId(InsertManyResul
...
+    {
+        $documents = [
+            TestObject::createForFixture(1),
+            TestObject::createForFixture(2),
+            TestObject::createForFixture(3),
+        ];
+
+        $expectedDocuments = [
+            (object) [
+                '_id' => 1,
```
jmikola, line 216: I've noted that TestDocumentCodec is ensuring the `$id` property gets mapped to `_id` here.

tests/Collection/CodecCollectionFunctionalTest.php
```diff
@@ -0,0 +1,528 @@
...
+    }
+
+    private static function createFixtureResult(int $id): BSONDocument
+    {
+        return new BSONDocument(['_id' => $id, 'x' => new BSONDocument(['foo' => 'bar'])]);
+    }
+
+    private static function createObjectFixtureResult(int $id, bool $isInserted = false): BSONDocument
+    {
+        return new BSONDocument(['_id' => $isInserted ? null : $id, 'id' => $id, 'x' => new BSONDocument(['foo' => 'bar']), 'decoded' => false]);
```
jmikola, line 526: Consider splitting this out on multiple lines for readability.

tests/Collection/CodecCollectionFunctionalTest.php
```diff
@@ -0,0 +1,528 @@
...
+        ];
+
+        $this->expectExceptionObject(InvalidArgumentException::cannotCombineCodecAndTypeMap());
+        $this->collection->replaceOne(['_id' => 1], ['foo' => 'bar'], $options);
+    }
+
+    /**
+     * Create data fixtures.
+     */
+    private function createFixtures(int $n, array $executeBulkWriteOptions = []): void
```
jmikola, line 503: Noted that this is another place where you coupled the structure of TestObject into a fixture creation method. I'd suggest making TestObject directly Serializable to work around this, and you can leave a comment on the `bsonSerialize()` method explaining why you do so.

----

One thing I didn't consider is whether you intentionally _didn't_ implement Serializable to ensure that the document is only encoded correctly when a codec is provided to the operation.

If that's the case then I think we can work around this by introducing a new fixture creation method in TestObject that returns a Document instance (using `Document::fromPHP()`) that we can pass directly to `BulkWrite::insert()`. That would still allow us to consolidate the TestObject schema to its own class without scattering it throughout a few functional tests' `createFixtures()` methods.

src/Collection.php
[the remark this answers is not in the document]
alcaeus: Extracting options handling to separate classes is being tracked in PHPLIB-1210.

src/Model/ChangeStreamIterator.php
alcaeus: I've updated the return type once again to be `Iterator` (as that's what the parent class requires) and added a `@return` annotation to annotate the full type. I've also added assertions for both `CursorInterface` and `Iterator` in the method body.

src/Operation/FindOneAndReplace.php
alcaeus: Good point. Changed 👍

src/Operation/BulkWrite.php
alcaeus: Updated along with a comment documenting that the missing `break` is intentional.

src/Operation/InsertOne.php
alcaeus: At least I'm consistently inconsistent at some point. Or is it called "eventually consistent"? 🤔

src/Operation/ReplaceOne.php
alcaeus: That is correct. When we create the nested `UpdateOne` operation, we'd have to inspect the update document to figure out whether we're doing an atomic update or a replace, so I opted for handling encoding of data in `ReplaceOne`, even though this causes an exception at a different time than in other cases.

I've moved the assertion here as well.

tests/Fixtures/Document/TestObject.php
alcaeus: I could, but only some tests touch it at all. Since the properties are public, I decided to not add yet another factory or argument to change the value of `foo`.

tests/Fixtures/Document/TestObject.php
alcaeus: The `decoded` property allows us to ensure that the object was created through the Codec, and not through other means like a type map (if it ever implements Unserializable down the road).

As for the `encoded` property that's added in `encode()`, that property is never read in `decode`, so it will never get used. That is also only used to check that the object went through the Codec instead of being converted to an object by other means.

tests/Operation/FindOneAndReplaceFunctionalTest.php
alcaeus: Updated to create a document through a factory instead. I've also updated other instances of that same fixture method in these tests.

tests/Collection/CodecCollectionFunctionalTest.php
alcaeus: Not making it `Serializable` was on purpose, as the latter would make errors less explicit when inserting or updating such documents without a codec. I've added a factory to create a `Document` instance for use in the `createFixtures` methods.

src/Model/ChangeStreamIterator.php
alcaeus: Sorry, that comment seems to have gotten lost.

Since our tests extract the `_id` field through command subscribers which always convert BSON to PHP values, I've decided to convert the resume token to PHP values in the iterator as well. This makes it easier for us to assert against the extracted resume token in tests.

src/Operation/BulkWrite.php
```diff
@@ -341,28 +341,30 @@ public function execute(Server $server)
...
-                    assert(is_array($insertedDocument) || is_object($insertedDocument));
-
-                    $insertedIds[$i] = $bulk->insert($insertedDocument);
+                    if (isset($this->options['codec'])) {
+                        $args[0] = $this->options['codec']->encodeIfSupported($args[0]);
+
+                        // Psalm's assert-if-true annotation does not work with unions, so
+                        // assert the type manually instead of using is_document
+                        // See https://github.com/vimeo/psalm/issues/6831
+                        assert(is_array($args[0]) || is_object($args[0]));
```
jmikola, line 350: Should we worry about PackedArray here? `is_document()` does:

```
is_array($document) || (is_object($document) && ! $document instanceof PackedArray)
```

My gut says, "Probably?", since the intention here was to inline `is_document()` for Psalm's benefit.

Before addressing this, consider my comment below for `REPLACE_ONE`, though.

src/Operation/BulkWrite.php
```diff
@@ -341,28 +341,30 @@ public function execute(Server $server)
...
-
-                    $bulk->update($args[0], $replacementDocument, $args[2]);
-                    break;
+                    if (isset($this->options['codec'])) {
+                        $args[1] = $this->options['codec']->encodeIfSupported($args[1]);
+
+                        // Psalm's assert-if-true annotation does not work with unions, so
+                        // assert the type manually instead of using is_document
+                        // See https://github.com/vimeo/psalm/issues/6831
+                        assert(is_array($args[1]) || is_object($args[1]));
```
jmikola, line 363: Similar to `INSERT_ONE` above, I wonder if we should duplicate the replacement document validation here.

The constructor performs additional checks on `$args[1]` that we're missing here:

 * Treat empty arrays as replacement documents for BC
 * Prohibit `is_first_key_operator($args[1])`
 * Prohibit `is_pipeline($args[1], true /* allowEmpty */)`

----

Having said this, I'm now wondering why you decided to encode at execution time instead of assembly time in the constructor:

```php
foreach ($operations as $i => $operation) {
    ...

    switch ($type) {
        case self::INSERT_ONE:
            // here

        ...

        case self::REPLACE_ONE:
            // and here
```

I realize `encodeIfSupported()` doesn't throw (at least our implementations shouldn't), but this would consolidate all of the validation errors to construction time. If we're going to be utilizing a codec, I think it makes sense to _only_ validate the result of the encoding instead of both sides as we seem to be doing here (and noting that the post-encoding validation is incomplete).

src/Operation/FindOneAndReplace.php
```diff
@@ -160,12 +160,12 @@ public function __construct(string $databaseName, string $collectionName, $filte
...
 
-        // Psalm's assert-if-true annotation does not work with unions, so
-        // assert the type manually instead of using is_document
-        // See https://github.com/vimeo/psalm/issues/6831
-        assert(is_array($replacement) || is_object($replacement));
+            // Psalm's assert-if-true annotation does not work with unions, so
+            // assert the type manually instead of using is_document
+            // See https://github.com/vimeo/psalm/issues/6831
+            assert(is_array($replacement) || is_object($replacement));
+        }
```
jmikola, line 168: Similar to my comments in BulkWrite, I think it makes sense to address this earlier in the constructor _before_ the existing `$replacement` validation. I'm assuming that only the _result_ of the encoding is relevant. Hypothetically, if the application defined an encoder that somehow stripped out update operators from a document and yielded a valid replacement document, that should be valid.

I'll note that this class is already ensuring that encoding happens in the constructor, so we don't have the same issue as BulkWrite with that being deferred to execution time.

src/Operation/InsertMany.php
```diff
@@ -155,12 +155,12 @@ public function execute(Server $server)
...
-            }
 
-            // Psalm's assert-if-true annotation does not work with unions, so
-            // assert the type manually instead of using is_document
-            // See https://github.com/vimeo/psalm/issues/6831
-            assert(is_array($document) || is_object($document));
+                // Psalm's assert-if-true annotation does not work with unions, so
+                // assert the type manually instead of using is_document
+                // See https://github.com/vimeo/psalm/issues/6831
+                assert(is_array($document) || is_object($document));
```
jmikola, line 162: This has the same issue I mentioned in BulkWrite. I think it makes sense to address this in the constructor and do encoding _before_ we attempt any validation of `$document`. Interestingly, InsertOne _does not_ have the same problem and applies the codec in its constructor.

I'll confess that this is a departure from our existing convention of dealing with named arguments prior to anything in `$options`, but I think there's a good reason for doing so in this case.

src/Operation/InsertOne.php
```diff
@@ -94,6 +105,15 @@ public function __construct(string $databaseName, string $collectionName, $docum
             unset($options['writeConcern']);
         }
 
+        if (isset($options['codec'])) {
```
jmikola, line 108: This is processing the document in the constructor (unlike BulkWrite and InsertMany), but I think we should do this before any `$document` validation above.

src/Operation/ReplaceOne.php
```diff
@@ -113,12 +113,12 @@ public function __construct(string $databaseName, string $collectionName, $filte
...
-        }
 
-        // Psalm's assert-if-true annotation does not work with unions, so
-        // assert the type manually instead of using is_document
-        // See https://github.com/vimeo/psalm/issues/6831
-        assert(is_array($replacement) || is_object($replacement));
+            // Psalm's assert-if-true annotation does not work with unions, so
+            // assert the type manually instead of using is_document
+            // See https://github.com/vimeo/psalm/issues/6831
+            assert(is_array($replacement) || is_object($replacement));
```
jmikola, line 120: Similar to my comments on InsertOne, I think we should apply the conditional encoding _before_ we validate `$replacement`. `encodeIfSupported()` is meant to accept anything, so I presume we can just pass whatever the user sends us.

Thinking about this more abstractly, this is somewhat related to my previous thread about middleware or decorating operations (https://github.com/mongodb/mongo-php-library/pull/1140#discussion_r1272520344), which lead to the creation of [PHPLIB-1210](https://jira.mongodb.org/browse/PHPLIB-1210). Applying a codec is something I envision as happening before the operation itself, although I understand we probably want to embed it in the operation class to avoid code duplication. And in some cases where we work with cursors (e.g. change streams), we have no choice but to handle it within.

tests/Model/CodecCursorFunctionalTest.php
```diff
@@ -0,0 +1,66 @@
...
+
+        $this->expectWarning();
+        $this->expectWarningMessage('Discarding type map for MongoDB\Model\CodecCursor::setTypeMap');
+
+        $codecCursor->setTypeMap(['root' => 'array']);
+    }
+
+    private function createCodec(): DocumentCodec
+    {
+        return new class implements DocumentCodec {
```
jmikola, line 34: This doesn't look like a meaningful implementation. Since `testSetTypeMap()` only calls `setTypeMap()`, perhaps you can just pass a mock DocumentCodec to `CodecCursor::fromCursor()`? I noted that the `$codec` property is only used in `current()`, which won't be called by the test.

alcaeus: I keep forgetting about mocks, but you're correct. Changed accordingly.

alcaeus:
@jmikola I've changed the handling in operations that write (BulkWrite, FindOneAndReplace, InsertOne, InsertMany, and ReplaceOne):
* Encoding is now always done through `encode` instead of `encodeIfSupported` - when a codec is involved we're expecting that codec to be able to handle incoming values.
* Encoding is now done before all other checks, despite having knowledge that `encode` will always return a `Document` (so checks like `is_update_pipeline` will return `false`). For replace operations the check for atomic update operators is still relevant, so I decided it was easiest to encode first, check later.

The first change leaves open how we want to proceed with decoding values: we may want to use `decode` instead of `decodeIfSupported` - if nothing else the user then has a guarantee that the return values are of the type they expect, instead of having to account for the possibility of the codec not being able to decode a given value and returning it as-is. If you agree I can make that change as well.

tests/Operation/ReplaceOneTest.php
```diff
@@ -48,4 +50,25 @@ public function testConstructorReplacementArgumentProhibitsUpdatePipeline($repla
...
+        return $this->createOptionDataProvider([
+            'codec' => $this->getInvalidDocumentCodecValues(),
+        ]);
+    }
+
+    public function testCodecRejectsInvalidDocuments(): void
+    {
+        $this->expectExceptionObject(UnsupportedValueException::invalidEncodableValue([]));
+
+        new ReplaceOne($this->getDatabaseName(), $this->getCollectionName(), ['x' => 1], ['y' => 1], ['codec' => new TestDocumentCodec()]);
```
jmikola, line 72: Noted this fails because TestDocumentCodec only supports TestObject and an array is being passed.

tests/Operation/InsertOneTest.php
```diff
@@ -25,8 +27,16 @@ public function provideInvalidConstructorOptions()
...
             'session' => $this->getInvalidSessionValues(),
             'writeConcern' => $this->getInvalidWriteConcernValues(),
         ]);
     }
+
+    public function testCodecRejectsInvalidDocuments(): void
+    {
+        $this->expectExceptionObject(UnsupportedValueException::invalidEncodableValue([]));
+
+        new InsertOne($this->getDatabaseName(), $this->getCollectionName(), [], ['codec' => new TestDocumentCodec()]);
```
jmikola, line 40: I realize `[]` is perfectly valid (would result in a document with a generated `_id` field only), I think this makes the test more readable.

```suggestion
        new InsertOne($this->getDatabaseName(), $this->getCollectionName(), ['x' => 1], ['codec' => new TestDocumentCodec()]);
```

Edit: looks like you were already using `['x' => 1]` in BulkWriteTest.

tests/Operation/InsertManyTest.php
```diff
@@ -40,9 +42,17 @@ public function provideInvalidConstructorOptions()
...
             'session' => $this->getInvalidSessionValues(),
             'writeConcern' => $this->getInvalidWriteConcernValues(),
         ]);
     }
+
+    public function testCodecRejectsInvalidDocuments(): void
+    {
+        $this->expectExceptionObject(UnsupportedValueException::invalidEncodableValue([]));
+
+        new InsertMany($this->getDatabaseName(), $this->getCollectionName(), [[]], ['codec' => new TestDocumentCodec()]);
```
jmikola, line 56: Same reasoning as InsertOneTest.

```suggestion
        new InsertMany($this->getDatabaseName(), $this->getCollectionName(), [['x' => 1]], ['codec' => new TestDocumentCodec()]);
```

src/Operation/ReplaceOne.php
```diff
@@ -79,28 +83,21 @@ class ReplaceOne implements Executable
...
-        if (is_first_key_operator($replacement)) {
-            throw new InvalidArgumentException('First key in $replacement is an update operator');
-        }
-
-        if (is_pipeline($replacement, true /* allowEmpty */)) {
-            throw new InvalidArgumentException('$replacement is an update pipeline');
+        if (isset($options['codec'])) {
+            if (isset($options['typeMap'])) {
+                throw InvalidArgumentException::cannotCombineCodecAndTypeMap();
+            }
```
jmikola, line 93:
```suggestion
        if (isset($options['codec'], $options['typeMap'])) {
            throw InvalidArgumentException::cannotCombineCodecAndTypeMap();
```

Alternatively: `isset() && isset()` if you think that's more readable.

src/Operation/BulkWrite.php
```diff
@@ -156,6 +294,10 @@ public function __construct(string $databaseName, string $collectionName, array
 
             switch ($type) {
                 case self::INSERT_ONE:
+                    if ($codec) {
+                        $operations[$i][$type][0] = $codec->encode($args[0]);
+                    }
```
jmikola, line 299: `is_document()` is used to validate `$args[0]` _before_ the `switch` statement, which means it never checks the result of `DocumentCodec::encode()`. That seems OK because a Document object is certainly going to be considered a document, but I'd suggest making a note of this in a short comment to copy/paste above these calls to `$codec->encode($args[0]);` (or this conditional).

src/Operation/FindOneAndReplace.php
```diff
@@ -177,4 +170,34 @@ public function getCommandDocument()
...
+            throw InvalidArgumentException::expectedDocumentType('$replacement', $replacement);
+        }
+
+        // Treat empty arrays as replacement documents for BC
+        if ($replacement === []) {
+            $replacement = (object) $replacement;
+        }
+
+        if (is_first_key_operator($replacement)) {
+            throw new InvalidArgumentException('First key in $replacement is an update operator');
```
jmikola, line 194: Related to my other question about creating common utility functions to validate the replacement and insert documents, I probably overlooked that we want to provide contextual exception messages here. So in that case I won't object to keeping some of this duplicated logic as-is.

src/Collection.php
```diff
@@ -245,14 +258,13 @@ public function aggregate(array $pipeline, array $options = [])
      */
     public function bulkWrite(array $operations, array $options = [])
     {
-        if (! isset($options['writeConcern']) && ! is_in_transaction($options)) {
-            $options['writeConcern'] = $this->writeConcern;
-        }
+        $options = $this->inheritWriteOptions(
+            $this->inheritCodec($options),
+        );
```
jmikola, line 263: I realize the signatures allow these to be chained, but I think it'd be more readable on separate lines. Would make it easier to grep for things like `inheritWriteOptions($options)` and would also minimize future diffs if we modify inheritance logic.

src/Collection.php
```diff
@@ -271,19 +283,11 @@ public function bulkWrite(array $operations, array $options = [])
...
-
-        if (! isset($options['readConcern']) && ! is_in_transaction($options)) {
-            $options['readConcern'] = $this->readConcern;
-        }
+        $options = $this->inheritReadOptions($options);
 
         $operation = new Count($this->databaseName, $this->collectionName, $filter, $options);
 
-        return $operation->execute($server);
+        return $operation->execute(select_server($this->manager, $options));
```
jmikola, line 290: This is a subtle behavioral change because we're now deferring server selection until _after_ constructing the Count operation, but I'll note that other helper methods vary in this regard.

For examples, helpers like `aggregate()` must select servers in advance to handle option inheritance.

I think it does make sense to defer server selection as long as possible, since it makes it easier to catch other client-side validation errors sooner.

src/Collection.php
```diff
@@ -471,15 +447,11 @@ public function distinct(string $fieldName, $filter = [], array $options = [])
...
-        }
-
         $server = select_server($this->manager, $options);
 
-        if (! isset($options['writeConcern']) && ! is_in_transaction($options)) {
-            $options['writeConcern'] = $this->writeConcern;
-        }
+        $options = $this->inheritWriteOptions(
+            $this->inheritTypeMap($options),
+        );
```
jmikola, line 454: In the interest of deferring `select_server()` as late as possible, this can technically be moved up to the top of the `drop()` method. Doesn't make much functional difference, though, since these wouldn't throw, but it'd be consistent with the order you use in most other methods.

src/Collection.php
```diff
@@ -967,17 +872,13 @@ public function mapReduce(JavascriptInterface $map, JavascriptInterface $reduce,
...
 
-        if (! isset($options['typeMap'])) {
-            $options['typeMap'] = $this->typeMap;
-        }
-
-        if (! isset($options['writeConcern']) && ! is_in_transaction($options)) {
-            $options['writeConcern'] = $this->writeConcern;
-        }
+        $options = $this->inheritWriteOptions(
+            $this->inheritTypeMap($options),
```
jmikola, line 876: Thinking about how you had to retain `if (! isset($options['readPreference']) && ! is_in_transaction($options))` in this method, did you consider splitting the ReadPreference and ReadConcern inheritance into their own methods (like TypeMap and Codec) and then having a common `inheritReadOptions()` call out to the individual methods?

The special handling for ReadConcern inheritance only comes up a small number of times so I'm fine either way. Just curious if this was intentional.

src/Collection.php
```diff
@@ -1099,30 +987,13 @@ public function updateOne($filter, $update, array $options = [])
...
-
-        $server = select_server($this->manager, $options);
-
-        /* Although change streams require a newer version of the server than
-         * read concerns, perform the usual wire version check before inheriting
-         * the collection's read concern. In the event that the server is too
-         * old, this makes it more likely that users will encounter an error
-         * related to change streams being unsupported instead of an
-         * UnsupportedException regarding use of the "readConcern" option from
-         * the Aggregate operation class. */
```
jmikola, line 1114: I assume this is entirely irrelevant since PHPLIB requires MongoDB 3.6+ and change streams are always supported (assuming we're not a standalone).

jmikola (commented):
> The first change leaves open how we want to proceed with decoding values: we may want to use decode instead of decodeIfSupported - if nothing else the user then has a guarantee that the return values are of the type they expect, instead of having to account for the possibility of the codec not being able to decode a given value and returning it as-is. If you agree I can make that change as well.

My initial reaction is that strict enforcement with `decode()` is preferable, but that may run afoul of the [Robustness Principle](https://en.wikipedia.org/wiki/Robustness_principle). Relying on `encode()` ensures that we emit well-formed values instead of leaving garbage as-is (although it'd probably fail later validation if important).

For decoding, the risk seems to be that the server returns something the codec didn't expect. Today, users would just get those values as-is and a `typeMap` with field paths that don't exist might just NOP. If that sounds like behavior we should preserve then I'd be on board with `decodeIfSupported()`.

But to argue once more in the other direction: if the codec we expect to be used is something quite generic like DocumentCodec, then I don't see much harm in using `decode()`.

Sorry for bouncing back and forth here, but I'm struggling to conceive of all of the ways things could possibly go wrong with either approach. Maybe we should punt this to a video chat as well.

src/Operation/BulkWrite.php
alcaeus: Added an explanatory comment.

src/Collection.php
alcaeus: I removed chained invocations in favour of re-assigning `$options`.

src/Collection.php
alcaeus: I went back and added `inheritReadConcern` and `inheritReadPreference`, simplifying the conditions above as well.

src/Operation/FindAndModify.php
```diff
@@ -243,6 +257,16 @@ public function execute(Server $server)
 
         $cursor = $server->executeWriteCommand($this->databaseName, new Command($this->createCommandDocument()), $this->createOptions());
 
+        if (isset($this->options['codec'])) {
+            $cursor->setTypeMap(['root' => 'bson']);
+            $result = current($cursor->toArray());
+            assert($result instanceof Document);
+
+            $decoded = $this->options['codec']->decode($result->get('value'));
```
jmikola, line 265: Note that below we gracefully handle `value` being unset, which is possible depending on the `new` option and whether something was matched. May be worth adding a functional regression test for this since if it wasn't caught already by a test failure.

tests/Collection/CodecCollectionFunctionalTest.php
```diff
@@ -0,0 +1,530 @@
...
+            $expected[3] = Document::fromPHP($inserted);
+        }
+
+        $this->assertEquals(
+            $expected,
+            $this->collection->find([], $options)->toArray(),
+        );
+    }
+
+    public function provideFindOneAndModifyOptions(): Generator
```
jmikola, line 154: Consider adding a regression test for the issue I mentioned in https://github.com/mongodb/mongo-php-library/pull/1140/files#r1284677901. You'd need some `findAndModify` options to result in `value` being unset on the response.

See [findAndModify: value](https://www.mongodb.com/docs/manual/reference/command/findAndModify/#value) for two possible conditions to trigger that response.

tests/Collection/CodecCollectionFunctionalTest.php
```diff
@@ -0,0 +1,530 @@
...
+    public function testInsertOne($expected, $options): void
+    {
+        $result = $this->collection->insertOne(TestObject::createForFixture(1), $options);
+        $this->assertSame(1, $result->getInsertedCount());
+
+        if ($expected instanceof BSONDocument && $expected->_id === null) {
+            $expected->_id = $result->getInsertedId();
+        }
+
+        if ($expected instanceof Document && $expected->get('_id') === null) {
```
jmikola, line 441: Are these code paths relevant? Looking at the data provider the `_id` is always set. Either way, this appears to assume that `_id` at least always _exists_ on the document, since `get()` would throw otherwise.

tests/Collection/CodecCollectionFunctionalTest.php
```diff
@@ -0,0 +1,530 @@
...
+
+        $result = $this->collection->insertMany($documents, $options);
+        $this->assertSame(3, $result->getInsertedCount());
+
+        foreach ($expected as $index => &$expectedDocument) {
+            if ($expectedDocument instanceof BSONDocument && $expectedDocument->_id === null) {
+                $expectedDocument->_id = $result->getInsertedIds()[$index];
+            }
+
+            if ($expectedDocument instanceof Document && $expectedDocument->get('_id') === null) {
```
jmikola, line 403: Same question as below about whether these code paths are entered. Do any of the expected documents have a `null` ID?

jmikola (commented):
Note the two phpcs errors.

tests/Collection/CodecCollectionFunctionalTest.php
alcaeus: Those are relevant for the "No codec" data set, as the resulting BSON document will not have an "_id" field. I've added a clarifying comment here and in the other test.

src/Operation/FindAndModify.php
jmikola: @alcaeus: I think this is still outstanding. The method is still doing:

```php
return $this->options['codec']->decode($result->get('value'));
```

tests/Collection/CodecCollectionFunctionalTest.php
jmikola: The comment still suggests that the `_id` field won't exist ("will not have an _id field"). I think it'd make sense be clear and say that the data provider specifies a `null` value to indicate that the `_id` is unset and should be generated.

tests/Collection/CodecCollectionFunctionalTest.php
jmikola: I'll resolve this as you already answered it in https://github.com/mongodb/mongo-php-library/pull/1140#discussion_r1295736142, but do not my follow-up comment there.

src/GridFS/Bucket.php
```diff
@@ -301,7 +302,7 @@ public function drop()
      * @see Find::__construct() for supported options
      * @param array|object $filter  Query by which to filter documents
      * @param array        $options Additional options
-     * @return Cursor
+     * @return CursorInterface&Iterator
```
GromNaN, line 305: The `Collection` instance is created internally, so I don't see how someone could pass on a codec to get a different result. The return type will always be `Cursor`.

src/Model/ChangeStreamIterator.php
```diff
@@ -47,7 +49,7 @@
  *
  * @internal
  * @template TValue of array|object
- * @template-extends IteratorIterator<int, TValue, Cursor<TValue>>
+ * @template-extends IteratorIterator<int, TValue, Iterator<int, TValue>>
```
GromNaN, line 52: After checking, psalm seems to accept union types here.
```suggestion
 * @template-extends IteratorIterator<int, TValue, CursorInterface&Iterator<int, TValue>>
```

GromNaN (approved):

src/Operation/FindAndModify.php
alcaeus: My apologies, this got lost as I changed everything to `decode` instead of `decodeIfSupported` and my brain got confused. I changed the logic to directly return `null` values, but to use `decode` for all other values. This ensures that the result will always be of the type indicated by the codec, or `null`. Using `decodeIfSupported` could mean that we're returning a `Document` instance which may be unexpected.

tests/Collection/CodecCollectionFunctionalTest.php
alcaeus: I updated the comment - I hope this explains it. FWIW, this logic is found in a few other tests where we rely on automatically generated identifiers.

src/GridFS/Bucket.php
alcaeus: The `$options` passed to `Bucket::find` are passed down to `Collection::find` as-is, so a user could introduce a codec that way, which would result in a return value of a type other than `Cursor`.

That said, this makes me wonder whether we should directly support a `codec` option in the `Bucket` class as well (similar to how `typeMap` is already supported). I haven't considered that until now, so if we decide to go that route I'd do so in a separate ticket and pull request.

jmikola: > The $options passed to Bucket::find are passed down to Collection::find as-is, so a user could introduce a codec that way, which would result in a return value of a type other than Cursor.

Noted that this explains the `CursorInterface&Iterator` return type being added here.

> That said, this makes me wonder whether we should directly support a codec option in the Bucket class as well (similar to how typeMap is already supported).

The `typeMap` option is only used for `getFileDocumentForStream()` and `getFileIdForStream()`, and `uploadFromStream()` indirectly since that calls `getFileIdForStream()`.

I think it does make sense to support `codec` there as we do `typeMap` today. Please create a separate ticket for that.

src/Operation/FindAndModify.php
jmikola: The most recent version in this PR is doing the following for the `codec` code path:

```php
$value = $result->get('value');

return $value === null ? $value : $this->options['codec']->decode($value);
```

This doesn't address my original comment (https://github.com/mongodb/mongo-php-library/pull/1140#discussion_r1272534094) about `value` being unset; however, I actually think that's no longer a concern with MongoDB 3.6+. The original code path where we handle a missing `value` property may have been related to some Very Old™ code ([PHPLIB-95](https://jira.mongodb.org/browse/PHPLIB-95) and [SPEC-201](https://jira.mongodb.org/browse/SPEC-201).

The [MongoDB 3.6 documentation for `findAndModify`](https://www.mongodb.com/docs/v3.6/reference/command/findAndModify/#findandmodify-value-field) suggests that `value` is always present but may be `null`.

Feel free to resolve this after reading.

src/ChangeStream.php
```diff
@@ -82,15 +85,22 @@ class ChangeStream implements Iterator
...
 
+    private ?DocumentCodec $codec;
+
     /**
      * @internal
      *
      * @param ResumeCallable $resumeCallable
      */
-    public function __construct(ChangeStreamIterator $iterator, callable $resumeCallable)
+    public function __construct(ChangeStreamIterator $iterator, callable $resumeCallable, ?DocumentCodec $codec = null)
```
jmikola, line 95: Watch constructs the ChangeStreamIterator and has access to its internal cursor, even though it currently sends it as-is directly from the private `executeAggregate()` method. Watch also constructs the `$resumeCallable` here.

I was going to ask why ChangeStream bothers with calling `setTypeMap(['root' => 'bson'])` when that could otherwise be done in Watch; however, even if we _did_ move that logic to Watch, ChangeStream would still need to know about the codec in order to process return values for `current()`.

And you previously explained that Watch cannot apply the codec itself since doing so might interfere with resume token collection.

I think this is fine as-is but just want to confirm with you before signing off. Feel free to resolve after reading if my understanding is correct.

Another reason for asking was that ChangeStream is in the public API (although undocumented), so I'm paying extra attention the signature change above.

jmikola (commented):
Suggested exception and new test for Watch.

src/GridFS/Bucket.php
alcaeus: Created PHPLIB-1220.

src/Operation/Watch.php
```diff
@@ -199,6 +206,10 @@ public function __construct(Manager $manager, ?string $databaseName, ?string $co
             'readPreference' => new ReadPreference(ReadPreference::PRIMARY),
         ];
 
+        if (isset($options['codec']) && ! $options['codec'] instanceof DocumentCodec) {
```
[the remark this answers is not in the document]
alcaeus: Good point. I tested to confirm it works as expected with codec AND typeMap specified. I agree that specifying both options should be prohibited as we do in other operations. Added the exception and a test as suggested.

src/ChangeStream.php
alcaeus: Since `ChangeStream` needs to be in control of the codec used in the cursor (as it needs to be sure about the return types), I figured it would make more sense for `ChangeStream` to apply its codec to the underlying cursor instead of relying on that being set correctly. As is, `Watch` is only accepting the `codec` option on behalf of `ChangeStream`, and passing it on without needing to know about internal workings of `ChangeStream`.

The main problem with applying the codec in the original `Aggregate` operation is that while the aggregation pipeline results will always have an `_id` field to be used as resume token, we cannot make such guarantees after the codec has been applied. After that, the `_id` field may have been discarded or assigned to an object property we don't know about.

Also note that while `ChangeStream` and `ChangeStreamIterator` are part of the public API, both their constructors are marked as internal; indicating that while we expect users to interact with these classes, we don't expect them to instantiate them on their own. The signature change here is backward compatible, as the `$codec` parameter defaults to `null`, so any previously working instantiations of that class will continue to function as expected.

jmikola (approved):

PR #1141, opened 2023-08-04
PHPLIB-1183: Initial performance tests

alcaeus:
PHPLIB-1183

This PR adds initial performance tests. For the time being, this tests the performance of `MongoDB\BSON\Document` and `MongoDB\BSON\PackedArray`, and runs comparative benchmarks as defined in the BSON benchmark spec.

The performance tests for the BSON classes could theoretically be moved to the PHPC repo, but would require setting up an entirely new composer project. For the time being the tests can live here, but they should be moved to PHPC when we start regression testing so that any performance regressions can be caught directly while doing the work.


benchmark/BSON/DocumentBench.php
```diff
@@ -0,0 +1,100 @@
...
+
+#[BeforeMethods('prepareData')]
+final class DocumentBench extends BaseBench
+{
+    private static string $rawBSON;
+    private static Document $document;
+
+    public function prepareData(): void
+    {
+        self::$rawBSON = fromJSON(file_get_contents(self::LARGE_FILE_PATH));
```
jmikola, line 36: Was this split out from `getBSONDocument()` to distinguish _reading_ the JSON file from constructing the Document instance?

If neither method is included in the benchmark, I'd have expected you to use `Document::fromJSON()`.

----

Edit: I just noticed the `BeforeMethods('prepareData')` annotation. I guess that explains this.

If you run this method before the benchmark, why do you want `getBSONDocument()` included? I recall you saying that resulted in a slower first run, which balances out in subsequent tests. But in that case do you even want Document constructed counted at all?

benchmark/BSON/DocumentBench.php
```diff
@@ -0,0 +1,100 @@
...
+    public function benchToPHPArray(): void
+    {
+        self::getBSONDocument()->toPHP(['root' => 'array']);
+    }
+
+    public function benchIteration(): void
+    {
+        // phpcs:ignore Generic.CodeAnalysis.EmptyStatement.DetectedForeach
+        foreach (self::getBSONDocument() as $key => $value) {
+        }
```
jmikola, line 84:
```suggestion
        foreach (self::getBSONDocument() as $key => $value);
```

If the iteration has no body, I think you can get by with semicolons.

benchmark/BaseBench.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+namespace MongoDB\Benchmark;
```
jmikola, line 3: Is this file missing a copyright block?

Do we actually need those for benchmarks? We don't use it in our test files.

jmikola (approved):

benchmark/data/tweet.json
```diff
@@ -0,0 +1,66 @@
+{
+  "text": "@wildfits you're not getting one.....",
```
GromNaN, line 2: Now I'm wondering how you ended with [this tweet](https://twitter.com/byjessiekay/status/22824602300).

benchmark/ReadMultipleDocumentsBench.php
```diff
@@ -0,0 +1,168 @@
...
+                $document['entities']['user_mentions'][0]['screen_name'];
+                break;
+
+            case 'bson':
+                assert($document instanceof Document);
+                $document->get('entities')->get('user_mentions')->get(0)->get('screen_name');
+                break;
+
+            case 'object':
+                assert(is_object($document));
```
GromNaN, line 160: `Document` is also an object.

```suggestion
                assert($document instanceof stdClass);
```

benchmark/BaseBench.php
alcaeus: We don't use them for tests, so I omitted them here as well.

phpbench.ci.json
```diff
@@ -0,0 +1,8 @@
+{
+    "$schema": "./vendor/phpbench/phpbench/phpbench.schema.json",
+    "runner.bootstrap": "vendor/autoload.php",
+    "runner.path": "benchmark",
+    "runner.iterations": 3,
+    "runner.revs": 10,
```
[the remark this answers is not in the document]
alcaeus: Removed the extra files and included this in the dist configuration. I also added the dist file to .gitattributes.

benchmark/BSON/DocumentBench.php
alcaeus: This was based on a different benchmark which needed extra handling. I simplified the two benchmarks to create the BSON structures directly.

benchmark/ReadMultipleDocumentsBench.php
alcaeus: That is not true for the "Library default typemap" set, as the document will be returned as `BSONDocument` instance.

benchmark/data/tweet.json
alcaeus: The file is from the benchmark specification data, where it was added in https://github.com/mongodb/specifications/commit/19ec68039b12f431d30848203c7ccd55815b9319. It probably ended up in a test dataset at some point in time and just kept on being copied around since then.

GromNaN (approved):
LGTM.
I see the report is in the CI artifacts.
But there is no automatic comparison with previous build ATM?

alcaeus:
> But there is no automatic comparison with previous build ATM?

Not yet - there are a number of variables to control before making that viable, most notably ensuring that unrelated changes (e.g. newer server versions, operating systems, etc.) don't affect results. We'll also have to define metrics and add assertions to then make the build fail.

PR #1142, opened 2023-08-07
PHPLIB-1177: SDAM logger example

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1177

### DIFF
--- a/examples/sdam_logger.php
+++ b/examples/sdam_logger.php
@@ -1,0 +1,145 @@
+<?php
+declare(strict_types=1);
+
+namespace MongoDB\Examples;
+
+use MongoDB\Client;
+use MongoDB\Driver\Monitoring\SDAMSubscriber;
+use MongoDB\Driver\Monitoring\ServerChangedEvent;
+use MongoDB\Driver\Monitoring\ServerClosedEvent;
+use MongoDB\Driver\Monitoring\ServerHeartbeatFailedEvent;
+use MongoDB\Driver\Monitoring\ServerHeartbeatStartedEvent;
+use MongoDB\Driver\Monitoring\ServerHeartbeatSucceededEvent;
+use MongoDB\Driver\Monitoring\ServerOpeningEvent;
+use MongoDB\Driver\Monitoring\TopologyChangedEvent;
+use MongoDB\Driver\Monitoring\TopologyClosedEvent;
+use MongoDB\Driver\Monitoring\TopologyOpeningEvent;
+
+use function get_class;
+use function getenv;
+use function MongoDB\BSON\fromPHP;
+use function MongoDB\BSON\toRelaxedExtendedJSON;
+use function printf;
+
+require __DIR__ . '/../vendor/autoload.php';
+
+/** @param array|object $document */
+function toJSON($document): string
+{
+    return toRelaxedExtendedJSON(fromPHP($document));
+}
+
+class SDAMLogger implements SDAMSubscriber
+{
+    public function serverChanged(ServerChangedEvent $event): void
+    {
+        printf(
+            "serverChanged: %s:%d changed from %s to %s\n",
+            $event->getHost(),
+            $event->getPort(),
+            $event->getPreviousDescription()->getType(),
+            $event->getNewDescription()->getType(),
+        );
+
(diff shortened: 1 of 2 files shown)


tests/ExamplesTest.php
```diff
@@ -128,6 +128,23 @@ public static function provideExamples(): Generator
             'expectedOutput' => $expectedOutput,
         ];
 
+        /* Note: Do not assert output beyond the initial topology events, as it
+         * may vary based on the test environment. PHPUnit's matching behavior
+         * for "%A" also seems to differ slightly from run-tests.php, otherwise
+         * we could assert the final topologyClosed event. */
```
jmikola, line 134: I looked through [StringMatchesFormatDescription](https://github.com/sebastianbergmann/phpunit/blob/9.5.28/src/Framework/Constraint/String/StringMatchesFormatDescription.php) but couldn't figure out exactly why it differs from `run-tests.php`. In my local testing, PHPUnit had trouble matching expected output _after_ `%A`, in contrast to the attached [sdam_logger.phpt](https://github.com/mongodb/mongo-php-library/files/12283265/sdam_logger.txt) file (ignore the `txt` extension, which was needed to appease GitHub's validation).

Relevant format in the `phpt` test is:

```
--EXPECTF--
topologyOpening: %x was opened

topologyChanged: %x changed from Unknown to %s

%A

topologyChanged: %x changed from %s to %s

topologyClosed: %x was closed

===DONE===
```

GromNaN: The PHPUnit `StringMatchesFormatDescription` splits line-by-line. So you cannot match several lines with a single `%A`. This works on my machine:

```php
        $expectedOutput = <<<'OUTPUT'
topologyOpening: %x was opened

topologyChanged: %x changed from Unknown to %s

%A
%A
%A
%A
%A
%A
%A
%A
%A
%A

topologyChanged: %x changed from %s to %s

topologyClosed: %x was closed


OUTPUT;
```

jmikola: > The PHPUnit StringMatchesFormatDescription splits line-by-line

Can you point me to where that happens? I know that [StringMatchesFormatDescription](https://github.com/sebastianbergmann/phpunit/blob/main/src/Framework/Constraint/String/StringMatchesFormatDescription.php) has some logic to convert `\r\n` (e.g. https://github.com/sebastianbergmann/phpunit/commit/05db035d21de4f2499ea20a349eb2be082543c2d), but it looks like it defers to `preg_match` to handle multi-line strings.

A series of `%A` patterns on their own line should be redundant, as that could otherwise be written with `%S` (see: [EXPECTF](https://qa.php.net/phpt_details.php#expectf_section]). And the regex conversion seemingly _does_ attempt to handle newlines for `%a` and `%A` based on the conversion logic in `regularExpressionForFormatDescription()`:

```
'%s' => '[^\r\n]+',
'%S' => '[^\r\n]*',
'%a' => '.+',
'%A' => '.*',
```

Asserting the number of newlines in the output is not portable, as that will vary based on the topology used to run the test. A 3-member replica set is going to log more lines than a 2-mongos sharded cluster or standalone server.

examples/sdam_logger.php
```diff
@@ -0,0 +1,145 @@
...
+            "serverChanged: %s:%d changed from %s to %s\n",
+            $event->getHost(),
+            $event->getPort(),
+            $event->getPreviousDescription()->getType(),
+            $event->getNewDescription()->getType(),
+        );
+
+        printf("previous hello response: %s\n", toJson($event->getPreviousDescription()->getHelloResponse()));
+        printf("new hello response: %s\n", toJson($event->getNewDescription()->getHelloResponse()));
+        echo "\n";
```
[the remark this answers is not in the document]
jmikola: This is indeed more verbose, but I'd argue it makes the code easier for copy-pasting and modifying if the user decides to trim a preceding `printf()` statement. In my own testing, I was frequently commenting out the `hello` response logging just to focus on the single line (e.g. `"serverChanged: %s:%d changed from %s to %s\n"`).

`command_logger.php` has trailing `echo` statements in each subscriber method as well.

alcaeus (approved):
LGTM. No objection to partial matches, as we're only testing the examples to make sure they run.

PR #1147, opened 2023-08-31
Add Atlas Search example

GromNaN:
Follow-up #1097

The example requires an Atlas M10+ cluster with Sample data loaded. It creates an index and performs a search query.

[asciicast](https://asciinema.org/a/HI0JnAOjE813nqV54xMhIJbO2)

### DIFF
--- a/examples/atlas-search.php
+++ b/examples/atlas-search.php
@@ -1,0 +1,129 @@
+<?php
+
+/**
+ * This example demonstrates how to create an Atlas Search index and perform a search query.
+ * It requires a MongoDB Atlas M10+ cluster with Sample Dataset loaded.
+ *
+ * Use the MONGODB_URI environment variable to specify the connection string from the Atlas UI.
+ */
+
+declare(strict_types=1);
+
+namespace MongoDB\Examples;
+
+use Closure;
+use MongoDB\Client;
+use RuntimeException;
+
+use function define;
+use function getenv;
+use function hrtime;
+use function iterator_to_array;
+use function sleep;
+use function str_contains;
+
+require __DIR__ . '/../vendor/autoload.php';
+
+$uri = getenv('MONGODB_URI');
+if (! $uri || ! str_contains($uri, '.mongodb.net')) {
+    echo 'This example requires a MongoDB Atlas cluster.', "\n";
+    echo 'Make sure you set the MONGODB_URI environment variable.', "\n";
+    exit(1);
+}
+
+// Atlas Search index management operations are asynchronous.
+// They usually take less than 5 minutes to complete.
+define('WAIT_TIMEOUT_SEC', 300);
+
+$client = new Client($uri);
+$collection = $client->selectCollection('sample_airbnb', 'listingsAndReviews');
+
+$count = $collection->estimatedDocumentCount();
+if ($count === 0) {
+    echo 'This example requires the sample_airbnb database with the listingsAndReviews collection.', "\n";
+    echo 'Load the sample dataset in your MongoDB Atlas cluster before running this example:', "\n";
+    echo '    https://www.mongodb.com/docs/atlas/sample-data/', "\n";
+    exit(1);
+}
+
+// Delete the index if it already exists
+$indexes = iterator_to_array($collection->listSearchIndexes());
+foreach ($indexes as $index) {
+    if ($index->name === 'default') {
+        echo "The index already exists. Dropping it.\n";
+        $collection->dropSearchIndex($index->name);
+
+        // Wait for the index to be deleted.
+        wait(function () use ($collection) {
+            echo '.';
+            foreach ($collection->listSearchIndexes() as $index) {
+                if ($index->name === 'default') {
+                    return false;
+                }
+            }
+
+            return true;
+        });
+    }
+}
+
+// Create the search index
+echo "\nCreating the index.\n";
+$collection->createSearchIndex(
+    /* The index definition requires a mapping
+     * See: https://www.mongodb.com/docs/atlas/atlas-search/define-field-mappings/ */
+    ['mappings' => ['dynamic' => true]],
+    // "default" is the default index name, this config can be omitted.
+    ['name' => 'default'],
+);
+
+// Wait for the index to be ready.
+wait(function () use ($collection) {
+    echo '.';
+    foreach ($collection->listSearchIndexes() as $index) {
+        if ($index->name === 'default') {
+            return $index->queryable;
+        }
+    }
+
+    return false;
+});
+
+// Perform a text search
+echo "\n", 'Performing a text search...', "\n";
+$results = $collection->aggregate([
+    [
+        '$search' => [
+            'index' => 'default',
+            'text' => [
+                'query' => 'view beach ocean',
+                'path' => ['name'],
+            ],
+        ],
+    ],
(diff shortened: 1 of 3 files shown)


examples/atlas-search.php
```diff
@@ -0,0 +1,130 @@
...
+use function getenv;
+use function hrtime;
+use function iterator_to_array;
+use function preg_match;
+use function sleep;
+
+require __DIR__ . '/../vendor/autoload.php';
+
+$uri = getenv('MONGODB_URI');
+if (! $uri || ! preg_match('/\.(mongodb\.net|mongodb-dev\.net)/', $uri)) {
```
jmikola, line 26: Users would not be running on `mongodb-dev.net`. Should we remove this and stick with `strpos()` on "mongodb.net" (like the original unified test runner)?

jmikola (commented):
Should there be test for expected output, or is that intentionally excluded since it's redundant with our spec tests and not portable (although I suppose it could do the same check for an Atlas URI and skip accordingly)?

examples/atlas-search.php
GromNaN: That's the only way I can run this test. I don't have the privilege of running an M10 cluster on Atlas, but I can run one on cloud-dev.

examples/atlas-search.php
```diff
@@ -0,0 +1,130 @@
...
+            return $index->queryable;
+        }
+    }
+
+    return false;
+});
+
+// Perform a text search.
+echo "\nPerforming a text search...\n";
+$results = $collection->aggregate([
```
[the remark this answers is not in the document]
GromNaN: I didn't see anything about a search helper. That was out of scope of the search index management specs.

examples/atlas-search.php
GromNaN: Ok, I can do this: `MONGODB_URI="mongodb+srv://user:*****@cluster0.****.mongodb-dev.net/?hehe=.mongodb.net`

examples/atlas-search.php
```diff
@@ -0,0 +1,130 @@
...
+
+$uri = getenv('MONGODB_URI');
+if (! $uri || ! preg_match('/\.(mongodb\.net|mongodb-dev\.net)/', $uri)) {
+    echo "This example requires a MongoDB Atlas cluster.\n";
+    echo "Make sure you set the MONGODB_URI environment variable.\n";
+    exit(1);
+}
+
+define('WAIT_TIMEOUT_SEC', (int) getenv('MONGODB_WAIT_TIMEOUT_SEC') ?: 300);
+define('INDEX_NAME', 'default');
```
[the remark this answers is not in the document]
GromNaN: Removed.

GromNaN:
> Should there be test for expected output, or is that intentionally excluded since it's redundant with our spec tests and not portable (although I suppose it could do the same check for an Atlas URI and skip accordingly)?

It would be nice, but that would be very slow to run because it needs to load the sample dataset and the output depends of the state of the cluster: if the index already exists or not.

jmikola:
> It would be nice, but that would be very slow to run because it needs to load the sample dataset and the output depends of the state of the cluster: if the index already exists or not.

Makes sense. I suggest making a note of this in the final commit so there's some record that it was intentionally omitted.

examples/atlas-search.php
```diff
@@ -0,0 +1,135 @@
...
+use function define;
+use function getenv;
+use function hrtime;
+use function iterator_to_array;
+use function sleep;
+use function str_contains;
+
+require __DIR__ . '/../vendor/autoload.php';
+
+$uri = getenv('MONGODB_URI');
```
jmikola, line 27:
```suggestion
$uri = getenv('MONGODB_URI') ?? null;
```

This may address the PossiblyFalseArgument error below. Alternatively, default to an empty string and then you can change the following line to `if (! str_contains(...))` and remove the `! $uri` condition.

GromNaN (commented):
I added a test with a lot of skip conditions.

GromNaN:
Test can run in CI with https://github.com/mongodb/mongo-php-library/pull/1150 config.

tests/ExamplesTest.php
```diff
@@ -179,6 +182,37 @@ public static function provideExamples(): Generator
         ];
     }
 
+    /**
+     * MongoDB Atlas Search example requires a MongoDB Atlas M10+ cluster with MongoDB 7.0+ and sample data loaded.
+     * Tips for insiders: if using a cloud-dev server, append ".mongodb.net" to the MONGODB_URI.
+     */
```
alcaeus, line 188:
```suggestion
     * @group atlas
     */
```

examples/atlas-search.php
```diff
@@ -0,0 +1,135 @@
...
+    ['name' => 'default'],
+);
+
+// Wait for the index to be ready.
+wait(function () use ($collection) {
+    foreach ($collection->listSearchIndexes() as $index) {
+        if ($index->name === 'default') {
+            echo '.';
+
+            return $index->queryable;
```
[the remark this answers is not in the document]
jmikola: I see there's no Psalm error here. Is it worth casting this to a boolean and using `bool` return type hint for the callback? OK to leave this as-is if it doesn't matter.

tests/ExamplesTest.php
```diff
@@ -179,6 +182,39 @@ public static function provideExamples(): Generator
...
+            $this->markTestSkipped('Atlas Search examples are only supported on MongoDB Atlas');
+        }
+
+        $this->skipIfServerVersion('<', '7.0', 'Atlas Search examples require MongoDB 7.0 or later');
+
+        $client = new Client($uri);
+        $collection = $client->selectCollection('sample_airbnb', 'listingsAndReviews');
+        $count = $collection->estimatedDocumentCount();
+        if ($count === 0) {
+            $this->markTestSkipped('Atlas Search examples require the sample_airbnb database with the listingsAndReviews collection');
```
jmikola, line 204: Do I assume correctly that this is for the benefit of running the test locally, and that it will most likely be skipped in CI because we don't load sample data?

I'm OK either way, but just wanted to ask.

jmikola (approved):

tests/ExamplesTest.php
GromNaN: We added the `@group atlas`, but I think I can insert some data to unit the collection in tests if it does not exist. We would benefit from having the test running in CI.

jmikola (approved):
Good idea loading some sample data for the benefit of CI.

examples/atlas-search.php
```diff
@@ -0,0 +1,133 @@
...
+    echo '    https://www.mongodb.com/docs/atlas/sample-data/', "\n";
+    exit(1);
+}
+
+// Delete the index if it already exists
+$indexes = iterator_to_array($collection->listSearchIndexes());
+foreach ($indexes as $index) {
+    if ($index->name === 'default') {
+        echo "The index already exists. Dropping it.\n";
+        $collection->dropSearchIndex($index->name);
```
GromNaN, line 58: Since this is idempotent, I can remove surrounding loop.

GromNaN: In fact no, the error for `NamespaceNotFound` is silenced, not the error for `IndexNotFound`. 
https://github.com/mongodb/specifications/blob/master/source/index-management/index-management.rst#namespacenotfound-errors

PR #1150, opened 2023-09-06
PHPLIB-1168: Run atlas tests on an Atlas cluster

alcaeus:
PHPLIB-1168

This PR adds a new evergreen build configuration to run all tests in the `atlas` group on an Atlas cluster. At the moment, this is limited to search index spec tests, but may encompass more tests later.

To ensure the atlas cluster is shut down, this leverages a task group. No other tasks or build variants have changed, as that's handled in the separate CI cleanup.

I've included a run for the build variant in the patch build for this PR, but these tests will usually not be run as part of a pull request to save build resources.

### DIFF
--- a/.evergreen/run-tests.sh
+++ b/.evergreen/run-tests.sh
@@ -82,6 +82,10 @@ export MONGODB_MULTI_MONGOS_LB_URI="${MONGODB_MULTI_MONGOS_LB_URI}"
 
 # Run the tests, and store the results in a junit result file
 case "$TESTS" in
+   atlas*)
+      php vendor/bin/simple-phpunit $PHPUNIT_OPTS --group atlas
+      ;;
+
    atlas-data-lake*)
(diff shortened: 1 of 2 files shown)


GromNaN (approved):
LGTM

.evergreen/run-tests.sh
```diff
@@ -82,6 +82,10 @@ export MONGODB_MULTI_MONGOS_LB_URI="${MONGODB_MULTI_MONGOS_LB_URI}"
 
 # Run the tests, and store the results in a junit result file
 case "$TESTS" in
+   atlas*)
```
jmikola, line 85: Isn't `atlas*` going to conflict with `atlas-data-lake*` below? Should we drop the asterisks? According to https://stackoverflow.com/a/4555979/162228, they do function as a wildcard for the pattern.

Just realized this while reviewing https://github.com/mongodb/mongo-php-library/pull/1151.

PR #1153, opened 2023-09-07
Remove wildcards from cases in run-tests

alcaeus:
As mentioned by @jmikola in https://github.com/mongodb/mongo-php-library/pull/1150#discussion_r1317639037, the wildcard character breaks running the data-lake test suite.


GromNaN (approved):
LGTM.

PR #1154, opened 2023-09-07
PHPLIB-1209: Add tutorial to show working with Codecs

alcaeus:
PHPLIB-1209

The tutorial uses the same classes and functionality we show in other BSON tutorials, namely the `Person` class to show serialisation (this is used to show the behaviour of `Persistable` and `Serializable` interfaces), and our example of storing a date as `UTCDateTime` with timezone information.


docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,302 @@
...
+               throw UnsupportedValueException::invalidDecodableValue($utc);
+           }
+
+           $dateTime = $utc->toDateTime();
+
+           if ($value instanceof MongoDB\BSON\Document && $value->has('tz')) {
+               $dateTime->setTimeZone(new DateTimeZone($value->get('tz')));
+           }
+
+           return DateTimeImmutable::fromMutable($dateTime);
```
GromNaN, line 165: That make me think that `UTCDateTime::toDateTime` should return a `DateTimeImmutable`. Something for 2.0?

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,302 @@
...
+           ]);
+       }
+   }
+
+Codec Libraries
+---------------
+
+If you have a number of codecs that you want to use in multiple places, you can create a codec library. A codec library
+contains a list of codecs and checks each codec if it supports a value. If it does, it will use that codec to encode or
+decode the given value. The following code snippet changes the PersonCodec to use a codec library instead of a
```
GromNaN, line 231:
```suggestion
decode the given value. The following code snippet changes the ``PersonCodec`` to use a codec library instead of a
```

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,302 @@
...
+       // Other code omitted for brevity
+       public function decode($value): Person
+       {
+           if (! $this->canDecode($value)) {
+               throw UnsupportedValueException::invalidDecodableValue($value);
+           }
+
+           $person = new Person($value->get('name'));
+           $person->id = $value->get('_id');
+           $person->createdAt = $this->codecLibrary->decode($value->get('createdAt'));
```
GromNaN, line 255: I think it's not possible, but the codecLibrary could be used as a service locator because you know which codec you need to use here. And from my previous comment, if you have a `City` property, you must use the `CityCodec`.

I'm thinking of the performance impact of having to loop through all the codecs for each field.

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,302 @@
...
+
+   Be careful when adding ``DocumentCodec`` instances to a codec library. The codec library will only use the first
+   codec that supports a given value. While this will most likely work fine when encoding to BSON, it can lead to issues
+   when decoding from BSON as you'll have to inspect the BSON document to detect whether the codec supports it. For this
+   reason, the ``codec`` option in a collection does not accept a codec library.
+
+Library-aware Codecs
+~~~~~~~~~~~~~~~~~~~~
+
+If you have a codec that needs to be aware of the codec library, you can implement the ``MongoDB\Codec\LibraryAware``
```
GromNaN, line 296:
```suggestion
If you have a codec that needs to be aware of the codec library, you can implement the ``MongoDB\Codec\KnowsCodecLibrary``
```

GromNaN (commented):
I think we need an example with a nested document `person.city`

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,302 @@
...
+
+Codecs are used to decode BSON documents into PHP objects and vice versa. In contrast to other methods, they allow for
+more flexibility and customization of the process and how different data types are handled. They allow separating the
+logic for BSON encoding and decoding from the domain classes, allowing to decode BSON into plain old PHP objects (POPOs).
+
+Document Codec Usage
+--------------------
+
+The main logic is contained in a document codec. This class implements the ``MongoDB\Codec\DocumentCodec`` interface and
+defines what data types can be encoded/decoded and how. The following example defines a ``Person`` class and a codec to
```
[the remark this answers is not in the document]
alcaeus: I'll add a note on the top to explain the terminology.

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,302 @@
...
+
+   final class PersonCodec implements MongoDB\Codec\DocumentCodec
+   {
+       // These traits define commonly used functionality to avoid duplication
+       use MongoDB\Codec\DecodeIfSupported;
+       use MongoDB\Codec\EncodeIfSupported;
+
+       public function canDecode($value): bool
+       {
+           return $value instanceof MongoDB\BSON\Document && $value->has('name');
```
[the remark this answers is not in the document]
alcaeus: Yes, the `PersonCodec` would only be used when interacting with `Person` instances, i.e. on the `Person` collection or if any other object embedded a `Person`. The check here was to ensure we don't try to decode any document without a `name` - this shouldn't mean that all documents with a `name` property will be successfully decoded.

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,302 @@
...
+
+   $person = $collection->findOne();
+
+The example above selects a collection and instructs it to use the ``PersonCodec`` for encoding and decoding documents.
+When inserting data, the ``PersonCodec`` is used to encode the document. When retrieving data, the ``PersonCodec`` is
+used to decode BSON data into a ``Person`` instance.
+
+.. note::
+
+   When working on a collection with a codec, all incoming and outgoing data will be handled with the given codec. This
```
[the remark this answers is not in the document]
alcaeus: I'll split this up to explain what data is encoded. `find` and `aggregate` do not encode incoming data.

docs/tutorial/codecs.txt
alcaeus: I'm not sure if we've previously discussed this, but I do think it's worth a ticket for 2.0 - either for changing the return type of `toDateTime` or to add a separate method.

GromNaN: Created [PHPC-2286](https://jira.mongodb.org/browse/PHPC-2286)

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,421 @@
...
+       }
+   }
+
+To then use this codec with a collection, specify the ``codec`` option when selecting the collection:
+
+.. code-block:: php
+
+   <?php
+
+   $collection = (new MongoDB\Client())->selectCollection('test', 'person', ['codec' => new PersonCodec()]);
```
GromNaN, line 97: It's easier to read if things are separate. And you can put the option on its own line to make it stand out.
```suggestion
   $client = new MongoDB\Client();
   $collection = $client->selectCollection('test', 'person', [
       'codec' => new PersonCodec(),
   ]);
```

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,421 @@
...
+to be used in a collection, or when decoding embedded documents.
+
+When working on a collection with a codec, the codec will only accept and return data of that type for certain
+operations. The ``bulkWrite``, ```findOneAndReplace``, ``insertMany``, ``insertOne``, and ``replaceOne`` operations
+will attempt to encode the given data using the provided codec. Trying to insert or replace a document that cannot be
+encoded will result in an exception. The ``aggregate``, ``find``, ``findOne``, ``findOneAndDelete``,
+``findOneAndReplace``, and ``findOneAndUpdate`` operations will attempt to decode returned documents using the provided
+codec. If the codec does not support the data returned, an exception will be thrown. You can disable codec usage for a
+specific operation or use a different codec (e.g. to decode the result of an aggregation pipeline) by specifying the
+``codec`` option for the operation. This will override the collection-level codec.
```
GromNaN, line 117: This paragraph is much more explicit than the previous version.

You can indicate how to disable the codec by specifying that the option is nullable.

```suggestion
nullable ``codec`` option for the operation. This will override the collection-level codec.
```

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,421 @@
...
+   <?php
+
+   final class PersonCodec implements MongoDB\Codec\DocumentCodec
+   {
+       private DateTimeCodec $dateTimeCodec;
+
+       public function __construct()
+       {
+           $this->dateTimeCodec = new DateTimeCodec();
+       }
```
GromNaN, line 201: For the doc, can we use PHP 8.1 features: promoted property and [new in initializer](https://wiki.php.net/rfc/new_in_initializers).
```suggestion
       public function __construct(
           private DateTimeCodec $dateTimeCodec = new DateTimeCodec(),
       ) {}
```

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,421 @@
...
+           if (! $this->canEncode($value)) {
+               throw UnsupportedValueException::invalidEncodableValue($value);
+           }
+
+           return MongoDB\BSON\Document::fromPHP([
+               'street' => $value->street,
+               'postCode' => $value->postCode,
+               'city' => $value->city,
+               'country' => $value->country,
+           ]);
```
GromNaN, line 286: Would performance be better without an intermediate array?
```suggestion
           $doc = new MongoDB\BSON\Document();
           $doc->set('street', $value->street);
           $doc->set('postCode, $value->postCode);
           $doc->set('city', $value->city);
           $doc->set('country', $value->country);
           
           return $doc;
```

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,421 @@
...
+   <?php
+
+   final class PersonCodec implements MongoDB\Codec\DocumentCodec
+   {
+       private AddressCodec $addressCodec;
+
+       public function __construct()
+       {
+           $this->addressCodec = new AddressCodec();
+       }
```
GromNaN, line 307:
```suggestion
       public function __construct(
           private AddressCodec $addressCodec = new AddressCodec(),
       ) {}
```

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,421 @@
...
+
+   <?php
+
+   final class PersonCodec implements MongoDB\Codec\DocumentCodec
+   {
+       private MongoDB\Codec\CodecLibrary $library;
+
+       public function __construct(private MongoDB\Codec\CodecLibrary $codecLibrary)
+       {
+       }
```
GromNaN, line 363: Duplicate declaration.

```suggestion
       public function __construct(
           private MongoDB\Codec\CodecLibrary $codecLibrary,
       ) {}
```

docs/tutorial/codecs.txt
alcaeus: The BSON document class is currently read-only, so the only way to create one is through `fromPHP` or the other factory methods. The reason for this is that we only keep a `bson_t*` internally, which is essentially a byte stream of the BSON. In-place replacement only works for types with the same size, e.g. overwriting an int with another int. As soon as values grow or shrink in size, the BSON needs to be rewritten (i.e. create a new `bson_t*`, copy over everything except the property being changed, then write the changed property). Since the main motivation for adding the `Document` and `PackedArray` classes was to provide a way to access BSON data without decoding it to PHP objects, we refrained from adding a BSON writing mechanism. That said, there is PHPC-2172 which tracks creating an implementation of a BSON writer.

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,416 @@
...
+to be used in a collection, or when decoding embedded documents.
+
+When working on a collection with a codec, the codec will only accept and return data of that type for certain
+operations. The ``bulkWrite``, ```findOneAndReplace``, ``insertMany``, ``insertOne``, and ``replaceOne`` operations
+will attempt to encode the given data using the provided codec. Trying to insert or replace a document that cannot be
+encoded will result in an exception. The ``aggregate``, ``find``, ``findOne``, ``findOneAndDelete``,
+``findOneAndReplace``, and ``findOneAndUpdate`` operations will attempt to decode returned documents using the provided
+codec. If the codec does not support the data returned, an exception will be thrown. You can disable codec usage for a
+specific operation or use a different codec (e.g. to decode the result of an aggregation pipeline) by specifying the
+nullable ``codec`` option for the operation. This will override the collection-level codec.
```
GromNaN, line 120: You can add an agg example with disabled codec. (Always good to have agg examples).

GromNaN (approved):
LGTM.

GromNaN (approved):

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+The example above selects a collection and instructs it to use the ``PersonCodec`` for encoding and decoding documents.
+When inserting data, the ``PersonCodec`` is used to encode the document. When retrieving data, the ``PersonCodec`` is
+used to decode BSON data into a ``Person`` instance. Note that while the ``PersonCodec`` could technically decode any
+BSON document that contains a name field, we wouldn't want to use it for any other documents. Document codecs are meant
+to be used in a collection, or when decoding embedded documents.
+
+When working on a collection with a codec, the codec will only accept and return data of that type for certain
+operations. The ``bulkWrite``, ```findOneAndReplace``, ``insertMany``, ``insertOne``, and ``replaceOne`` operations
+will attempt to encode the given data using the provided codec. Trying to insert or replace a document that cannot be
+encoded will result in an exception. The ``aggregate``, ``find``, ``findOne``, ``findOneAndDelete``,
```
jmikola, line 116: Didn't we discuss _not_ having ``aggregate`` apply the collection's codec under the assumption that a document may be modified by the pipeline? I just checked the source and `Collection::aggregate()` definitely inherits the codec with `inheritCodecOrTypeMap()`.

Perhaps that discussion was only about change streams, `distinct()`, etc.

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+``findOneAndReplace``, and ``findOneAndUpdate`` operations will attempt to decode returned documents using the provided
+codec. If the codec does not support the data returned, an exception will be thrown. You can disable codec usage for a
```
jmikola, line 118: I think it'd be helpful to start a new paragraph here so readers are less likely to miss this:

```suggestion
codec. If the codec does not support the data returned, an exception will be thrown.

You can disable codec usage for a
```

This relates to my earlier point about this potentially being quiet common with `aggregate()`, since it's very easy to change the document shape.

I was going to ask if you wanted to call out ``aggregate`` again here, but I see you do so in the example code below. That seems reasonable.

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+specific operation or use a different codec (e.g. to decode the result of an aggregation pipeline) by specifying the
+nullable ``codec`` option for the operation. This will override the collection-level codec:
```
jmikola, line 120: > specifying the nullable ``codec`` option

The option itself is nullable because it accepts a null value. But does "specifying the nullable option" imply that the user is setting a `null` value? Would it be clearer to say:

> specifying ``null`` for the ``codec`` option

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+       'codec' => null,
+   ]);
+
+   // Disables codec usage as the aggregate result will have a different format
+   $collection->aggregate([['$group' => [
+       '_id' => '$accountType',
+       'count' => ['$sum' => 1],
+   ]]], [
+       'codec' => null,
+   ]);
```
jmikola, line 137: This formatting looks unlike anything else I've found in our docs or project. A subsequent line with the same indentation as the first would usually be the final line in the statement.

Should this instead be:

```
$collection->aggregate(
    [['$group' => [
        ...
    ]]],
    ['codec' => null]
);
```

That honestly looks odd as well. I don't think there's any good way to do this -- so it's more about just picking something consistent. AFAIK this isn't addressed by our CS configuration.

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+
+Generic Codecs
```
jmikola, line 139: Consider "Handling Custom Types" or "Handling Fields and Data Types".

Related to my earlier comment, this is probably what you'd want to link to from the existing [Custom Data-Types](https://www.mongodb.com/docs/php-library/current/tutorial/custom-types/) tutorial.

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+       use MongoDB\Codec\DecodeIfSupported;
+       use MongoDB\Codec\EncodeIfSupported;
+
+       public function canDecode($value): bool
+       {
+           // For maximum compatibility, this codec supports decoding both UTCDateTime instances and documents with a
+           // UTCDateTime instance and optional timezone
+           return
+               $value instanceof MongoDB\BSON\UTCDateTime ||
+               $value instanceof MongoDB\BSON\Document && $value->has('utc');
```
jmikola, line 162: > and documents with a UTCDateTime instance and optional timezone

I don't recall this coming up in any previous test suite examples. Isn't this likely to lead to problems if users want to use this codec alongside an existing DocumentCodec in CodecLibrary?

----

Maybe I'm misunderstanding this and the `MongoDB\BSON\Document` check here is actually necessary because you're going to encode a UTCDateTime into an embedded document. If so, I think the check here should consider _both_ fields in the embedded document: `utc` and `tz`.

And per my comment below, you might also do well to check their types here.

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+       {
+           if (! $this->canDecode($value)) {
+               throw UnsupportedValueException::invalidDecodableValue($value);
+           }
+
+           $utc = $value instanceof MongoDB\BSON\UTCDateTime
+               ? $value
+               : $value->get('utc');
+
+           if (! $utc instanceof UTCDateTime) {
```
jmikola, line 180: This question assumes we keep the design as-is (handling both a field and document).

`canDecode()` currently checks that a `utc` field exists but doesn't check its type at all. Why would you wait to do so until the `decode()` method? This opens the possibility that `decodeIfSupported()` could still throw, no?

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+           }
+
+           return MongoDB\BSON\Document::fromPHP([
+               'utc' => new UTCDateTime($value),
+               'tz' => $value->getTimezone()->getName(),
+           ]);
+       }
+   }
+
+This codec can now be leveraged by other codecs encode and decode dates. Let's return to the previous example and add a
```
jmikola, line 206:
```suggestion
This codec can now be leveraged by other codecs to handle date fields. Let's return to the previous example and add a
```

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+
+   <?php
+
+   final class PersonCodec implements MongoDB\Codec\DocumentCodec
+   {
+       public function __construct(
+           private readonly DateTimeCodec $dateTimeCodec = new DateTimeCodec(),
+       ) {}
+
+       // Other code omitted for brevity
```
jmikola, line 220: What is "other code" here? Other methods or code within this `decode()` method?

If it's other methods, I wouldn't place this comment directly alongside the `decode()` method declaration.

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+       'count' => ['$sum' => 1],
+   ]]], [
+       'codec' => null,
+   ]);
+
+Generic Codecs
+--------------
+
+The previous example showed how to define a codec for a specific class. However, sometimes you want to define codecs to
+handle a given type in all documents. For such codecs, you can implement the ``MongoDB\Codec\Codec`` interface. The
```
jmikola, line 143: Suggestion:

> The previous example showed how to define a codec for a specific class. However, you may want to create a codec that handles a particular data type in any document. This can be achieved by implementing the ``MongoDB\Codec\Codec`` interface.
>
> The following example defines...

This splits up the intro paragraph with a sentence introducing the example code.

My other motivation was the clarify that we're designing a codec for a particular type in _any_ document vs. "all documents", since you might still want to limit such a codec to certain use cases.

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+
+           return MongoDB\BSON\Document::fromPHP([
+               '_id' => $value->id,
+               'name' => $value->name,
+               'createdAt' => $this->dateTimeCodec->encode($value->createdAt),
+           ]);
+       }
+   }
+
+Handling embedded documents
```
jmikola, line 248:
```suggestion
Handling Embedded Documents
```

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+
+Overview
+--------
+
+Codecs are used to decode BSON documents into PHP objects, and encode PHP objects into BSON documents. In contrast to
+other methods, they allow for more flexibility and customization of the process and how different data types are
+handled. They allow separating the logic for BSON encoding and decoding from the domain classes, allowing to decode BSON
+into plain old PHP objects (POPOs).
+
+Document Codec Usage
```
jmikola, line 23: Consider renaming to "Handling Documents" for consistency with other headings.

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+               'name' => $value->name,
+               'createdAt' => $this->dateTimeCodec->encode($value->createdAt),
+           ]);
+       }
+   }
+
+Handling embedded documents
+---------------------------
+
+The previous example showed how to handle a single document. However, sometimes you want to handle fields that contain
```
jmikola, line 251: This reads as if it appears directly after the "Handling Documents" header.

```suggestion
A previous example showed how to handle a single document. However, sometimes you want to handle fields that contain
```

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+embedded documents. To show this, let's create an address document that we'll be embedding into our ``Person`` document.
+To ensure consistency, we're going to make this a readonly class:
```
jmikola, line 253:
```suggestion
To ensure consistency, we're going to make this a read-only class:
```

I realize the PHP keyword is `readonly`, but I think this reads better in English.

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+               'createdAt' => $this->dateTimeCodec->encode($value->createdAt),
+           ]);
+       }
+   }
+
+Handling embedded documents
+---------------------------
+
+The previous example showed how to handle a single document. However, sometimes you want to handle fields that contain
+embedded documents. To show this, let's create an address document that we'll be embedding into our ``Person`` document.
```
jmikola, line 252:
```suggestion
embedded documents. We will demonstrate this using an ``Address`` document, which we will embed within a ``Person`` document.
```

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+   }
+
+We can now create a document codec for this class:
+
+.. code-block:: php
+
+   <?php
+
+   final class AddressCodec implements MongoDB\Codec\DocumentCodec
+   {
```
jmikola, line 276: Does this codec intentionally not use the DecodeIfSupported and EncodeIfSupported traits from the previous examples?

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+               'street' => $value->street,
+               'postCode' => $value->postCode,
+               'city' => $value->city,
+               'country' => $value->country,
+           ]);
+       }
+   }
+
+This codec is quite similar to the ``PersonCodec`` we had before, except that we're not adding an identifier to it. The
+creation of the object also differs, as we have to pass all the fields to the constructor.
```
jmikola, line 308:
```suggestion
creation of the object also differs, as we have to pass all fields to the constructor.
```

Either "all fields" or "all of the fields", but not "all the fields".

Is this only necessary because the class is read-only? Did you want to elaborate on what you meant by "ensuring consistency" with this design?

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+
+In the ``PersonCodec``, we can now extend the ``decode`` and ``encode`` methods to handle the address field. Note that
+the example below excludes some code we've already shown in previous examples.
```
jmikola, line 311: If we end up embedding example code, I think we should just lean into repeating ourselves here. Users might benefit from being able to copy-paste working code.

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+               'city' => $value->city,
+               'country' => $value->country,
+           ]);
+       }
+   }
+
+This codec is quite similar to the ``PersonCodec`` we had before, except that we're not adding an identifier to it. The
+creation of the object also differs, as we have to pass all the fields to the constructor.
+
+In the ``PersonCodec``, we can now extend the ``decode`` and ``encode`` methods to handle the address field. Note that
```
jmikola, line 310: > we can now extend the ``decode`` and ``encode`` methods

This won't be misunderstood to suggest class inheritance, will it? If we need alternative language, you could say "enhance", "augment", or "modify".

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+       }
+   }
+
+Codec Libraries
+---------------
+
+If you have a number of codecs that you want to use in multiple places, you can create a codec library. A codec library
+contains a list of codecs and checks each codec if it supports a value. If it does, it will use that codec to encode or
+decode the given value. The following code snippet changes the ``PersonCodec`` to use a codec library instead of a
+hard-coded ``DateTimeCodec``:
```
jmikola, line 365: Should you clarify that a CodecLibrary is, itself, a codec, and accepted everywhere a singular codec can be provided?

> A codec library contains a list of codecs and presents itself as a codec. When encoding or decoding a value, it will consult each registered codec in sequence and delegate to the first supporting codec to encode or decode that value.
>
> The following code snippet changes the ``PersonCodec`` to use a codec library instead of a...

Intentionally broke up the paragraph for readability. In general, I think if you're explaining some complex topic (or introducing something) in a paragraph and then leading into an example, the example-preceding blurb is best written on its own.

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+
+    $codecLibrary = new MongoDB\Codec\CodecLibrary([
+         new DateTimeCodec(),
+    ]);
+
+    $personCodec = new PersonCodec($codecLibrary);
+
+.. note::
+
+   Be careful when adding ``DocumentCodec`` instances to a codec library. The codec library will only use the first
```
jmikola, line 419: I called this out in the intro paragraph but agree with calling more attention to it here.

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+   codec that supports a given value. While this will most likely work fine when encoding to BSON, it can lead to issues
+   when decoding from BSON as you'll have to inspect the BSON document to detect whether the codec supports it. For this
```
jmikola, line 421: > lead to issues when decoding from BSON as you'll have to inspect the BSON document to detect whether the codec supports it.

Can you elaborate on this? The `encode()` method turns a PHP value into BSON, so are you saying that it's unlikely for multiple encoders to support the same value because they'd typically reject BSON? If so, how is that different for `decode()` methods if they typically receive BSON and spit out a PHP object?

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+   reason, the ``codec`` option in a collection does not accept a codec library.
```
jmikola, line 422: This is enforced by Collection requiring a DocumentCodec, correct?

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+
+Library-aware Codecs
+~~~~~~~~~~~~~~~~~~~~
+
+If you have a codec that needs to be aware of the codec library, you can implement the
+``MongoDB\Codec\KnowsCodecLibrary`` interface. When the codec is added to a library, this library is then injected into
```
jmikola, line 428: Did we discuss calling this CodecLibraryAware in some previous PR? I don't feel strongly either way but seeing the section title here made me wonder why it's named KnowsCodecLibrary.

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+the codec. This allows you to create a common library for all value codecs and re-use it in multiple document codecs.
+
+.. note::
+
+   The ``attachCodecLibrary`` method will not be called if a library-aware codec is never added to a library.
```
jmikola, line 433: Is there an edge case where a codec is added to multiple libraries? In that case, `attachCodecLibrary()` may get called multiple times, no?

I realize KnowsCodecLibrary is only an interface and we don't ship a trait, so this is up to the user to implement. I'm wondering if they might shoot themselves in the foot with this, though.

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+       }
+
+       public function decode($value): Person
+       {
+           if (! $this->canDecode($value)) {
+               throw UnsupportedValueException::invalidDecodableValue($value);
+           }
+
+           $person = new Person($value->get('name'));
+           $person->id = $value->get('_id');
```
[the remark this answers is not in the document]
alcaeus: Changed the property to use constructor promotion and `readonly`:
```php
final class Person
{
    public function __construct(
        public string $name,
        public readonly MongoDB\BSON\ObjectId $id = new MongoDB\BSON\ObjectId(),
    ) {}
}
```

Private properties aren't that important, as people will have to use accessors for everything anyways, so their usage is simpler. I wanted to show public properties to show that with modern tooling you don't really need private properties and accessors anymore.

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+   $person = $collection->findOne();
+
+The example above selects a collection and instructs it to use the ``PersonCodec`` for encoding and decoding documents.
+When inserting data, the ``PersonCodec`` is used to encode the document. When retrieving data, the ``PersonCodec`` is
+used to decode BSON data into a ``Person`` instance. Note that while the ``PersonCodec`` could technically decode any
+BSON document that contains a name field, we wouldn't want to use it for any other documents. Document codecs are meant
+to be used in a collection, or when decoding embedded documents.
+
+When working on a collection with a codec, the codec will only accept and return data of that type for certain
+operations. The ``bulkWrite``, ```findOneAndReplace``, ``insertMany``, ``insertOne``, and ``replaceOne`` operations
```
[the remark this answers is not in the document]
alcaeus: Changed and applied the same change to read operations.

docs/tutorial/codecs.txt
alcaeus: I remember discussing `aggregate` in that context, but don't remember a decision on it. I'm on the fence about it and can see arguments made in both directions, so I'm not too sure what the best course of action is.

docs/tutorial/codecs.txt
alcaeus: I've started formatting pipelines a bit different, even though phpcs will throw a fit about it:

```php
$collection->aggregate([
    ['$group' => [
        ...,
    ]],
]);
```

It does become a bit awkward with options, at which point I usually extract pipeline and options to their own variables to have cleaner looking invocations. That said, I think this would be a good compromise:
```php
$collection->aggregate(
    $pipeline,
    ['codec' => null],
);
```

We don't necessarily need to introduce a `$pipeline` variable in the same snippet as it's rather self-explanatory, and it helps keeping focus on the `codec` option.

docs/tutorial/codecs.txt
alcaeus: The idea behind this is that it would handle any `UTCDateTime` value and any document with a `utc` property. Yes, that could clash with other codecs that use a similar structure, so we may want to add an extra field to be able to determine what this is.

On top of that, this presents an opportunity to once again mention that any `DocumentCodec` instances most likely shouldn't be in the default `CodecLibrary`, as they may pick up documents in error, and instead only using them directly in the `Codec` option. It also presents an opportunity to show how the functionality of an `odm` field used by the `Persistable` interface can be replicated.

docs/tutorial/codecs.txt
alcaeus: Reworded to "Other methods [...]" and separated the comment from the method by a newline.

docs/tutorial/codecs.txt
alcaeus: Added the `use` statements to all codecs. Also separate the comment about omitting other methods.

docs/tutorial/codecs.txt
alcaeus: Removed the entire paragraph: the `PersonCodec` was changed to also pass all fields in the constructor, and the first sentence feels lost by itself.

docs/tutorial/codecs.txt
alcaeus: `modify` sounds good. Changed.

docs/tutorial/codecs.txt
GromNaN: That's easier to remove the codec by setting `'codec' => null` than injecting the Codec instance to where the line `Collection::aggregate is called`. There is plenty of use-cases for using aggregate to return documents as in the collection (`$search`, `$geoNear`).

docs/tutorial/codecs.txt
alcaeus: Looking through #1125 I couldn't find anything related to it. That said, I do remember the discussion and I vaguely remember us agreeing that `CodecLibraryAware` was better, but I may have forgotten to change it. @GromNaN I believe it was you that suggested the alternative, do I remember correctly?

docs/tutorial/codecs.txt
alcaeus: That is correct, adding the same codec instance to multiple libraries may be an issue, but that shouldn't happen too often. Let me know if you think this is worth calling out in the docs.

docs/tutorial/codecs.txt
alcaeus: `encode` won't have an issue as it works on a specific class instance. `decode` on the other hand works on a `Document` and will have to inspect all fields to decide whether it can decode the given data. I agree that the wording is a bit awkward.

How about this:
> [...] it can lead to issues when decoding from BSON as the codec will have to inspect all fields in the BSON document to detect whether the codec actually supports it.

This would also provide an opportunity to talk about discriminator fields (e.g. the `odm` field used by `Persistable` or the `discriminatorField` functionality in Doctrine ODM).

docs/tutorial.txt
```diff
@@ -8,6 +8,7 @@ Tutorials
    /tutorial/connecting
    /tutorial/server-selection
    /tutorial/crud
+   /tutorial/codecs
```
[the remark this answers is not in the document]
alcaeus: Added a note. I didn't find an example how to link to a specific heading, maybe you could point me in the right direction?

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,433 @@
...
+into plain old PHP objects (POPOs).
+
+Document Codec Usage
+--------------------
+
+The main logic is contained in a document codec. This class implements the ``MongoDB\Codec\DocumentCodec`` interface and
+defines what data types can be encoded/decoded and how. The following example defines a ``Person`` class and a codec to
+encode/decode it:
+
+.. code-block:: php
```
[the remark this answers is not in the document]
alcaeus: Good point, I'll update those examples.

docs/tutorial/codecs.txt
alcaeus: Updated to inspect the document as a whole and added a comment to explain that this can be slow for large documents.

docs/tutorial/codecs.txt
alcaeus: Codec libraries were removed from the code and in this tutorial.

docs/tutorial/codecs.txt
jmikola: That language SGTM.

docs/tutorial.txt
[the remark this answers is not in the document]
jmikola: Some of the ChangeStream methods link to the `.. _php-tailable-cursor:` anchor in `tailable-cursor.txt` from their "See Also" sections:

```
- :ref:`Tailable Cursor Iteration <php-tailable-cursor>`
```

That might apply here. Looks to be similar to how `crud.rst` links to `php-query-projection` within itself, although it's cross-page so even more relevant to what you're trying to do here.

docs/tutorial/codecs.txt
jmikola: I believe we also talked this through over video. SGTM. I'll resolve.

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,108 @@
...
+
+When working on a collection with a codec, the codec will only accept and return data of that type for certain
+operations. Insert and replace operations (e.g. ``insertOne``, ```findOneAndReplace``, and some ``bulkWrite``
+operations) will attempt to encode the given data using the provided codec. Trying to insert or replace a document that
+cannot be encoded will result in an exception. Read operations (e.g. ``aggregate``, ``find``, and ``findOneAndUpdate``)
+operations will attempt to decode returned documents using the provided codec. If the codec does not support the data
+returned, an exception will be thrown.
+
+You can disable codec usage for a specific operation or use a different codec (e.g. to decode the result of an
+aggregation pipeline) by specifying the ``null`` for the ``codec`` option for any operation. This will override the
```
GromNaN, line 55:
```suggestion
aggregation pipeline) by specifying ``null`` for the ``codec`` option for any operation. This will override the
```

docs/examples/codecs/handling-data-types/Person.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+use MongoDB\BSON\ObjectId;
+
+final class Person
```
jmikola, line 5: Noted that this differs from `handling-documents/Person.php` due to having an additional `createdAt` property.

I just want to confirm that none of the example files are exact duplicates. Feel free to resolve if that's correct.

docs/examples/codecs/handling-documents/using-codec.php
```diff
@@ -0,0 +1,13 @@
...
+
+$client = new Client();
+$collection = $client->selectCollection('test', 'person', [
+    'codec' => new PersonCodec(),
+]);
+
+$person = new Person('Jane Doe');
+$collection->insertOne($person);
+
+$person = $collection->findOne();
```
jmikola, line 13: Would using a different variable name make more sense here? I think I've tried to do that in other example code, but it's by no means a rule.

I'll defer you to ignore/resolve.

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,108 @@
...
+To then use this codec with a collection, specify the ``codec`` option when selecting the collection:
+
+.. literalinclude:: /examples/codecs/handling-documents/using-codec.php
+   :language: php
+
+The example above selects a collection and instructs it to use the ``PersonCodec`` for encoding and decoding documents.
+When inserting data, the ``PersonCodec`` is used to encode the document. When retrieving data, the same ``PersonCodec``
+is used to decode BSON data into a ``Person`` instance. Note that while the ``PersonCodec`` could technically decode any
+BSON document that contains a name field, we wouldn't want to use it for any other documents. Document codecs are meant
+to be used in a collection, or when decoding embedded documents.
```
jmikola, line 45:
```suggestion
to be used with a :phpclass:`MongoDB\\Collection`, or when decoding embedded documents.
```

I think referring to the class name at least once works well here to communicate that we're talking about the library object and not just a database collection.

Note: I'm also intentionally about referring to the class name here instead of earlier in the paragraph, because we're specifically talking about using two PHP objects together.

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,108 @@
...
+
+When working on a collection with a codec, the codec will only accept and return data of that type for certain
```
jmikola, line 47:
```suggestion
When using a collection with a codec, the codec will only accept and return data of that type for certain
```

"Working on" reads a bit odd, and "working with" is too much word repetition.

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,108 @@
...
+operations. Insert and replace operations (e.g. ``insertOne``, ```findOneAndReplace``, and some ``bulkWrite``
+operations) will attempt to encode the given data using the provided codec. Trying to insert or replace a document that
+cannot be encoded will result in an exception. Read operations (e.g. ``aggregate``, ``find``, and ``findOneAndUpdate``)
+operations will attempt to decode returned documents using the provided codec. If the codec does not support the data
```
jmikola, line 51:
```suggestion
will attempt to decode returned documents using the provided codec. If the codec does not support the data
```

Duplicate word.

docs/tutorial/codecs.txt
jmikola, line 55:
```suggestion
aggregation pipeline) by specifying ``null`` for the ``codec`` option for any operation. This will override the
```

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,108 @@
...
+
+.. versionadded:: 1.17
+
+Overview
+--------
+
+Codecs are used to decode BSON documents into PHP objects, and encode PHP objects into BSON documents. In contrast to
+other methods (e.g. type maps), codecs allow for greater customization and handling of different data types. They allow
+separating the logic for BSON encoding and decoding from the domain classes, allowing to decode BSON into plain old PHP
+objects (POPOs).
```
jmikola, line 21: Enjoy this gem I came across while googling for "POPOs": http://www.javaleaks.org/open-source/php/plain-old-php-object.html

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,108 @@
...
+   :class: singlecol
+
+.. versionadded:: 1.17
+
+Overview
+--------
+
+Codecs are used to decode BSON documents into PHP objects, and encode PHP objects into BSON documents. In contrast to
+other methods (e.g. type maps), codecs allow for greater customization and handling of different data types. They allow
+separating the logic for BSON encoding and decoding from the domain classes, allowing to decode BSON into plain old PHP
```
jmikola, line 20: "allow" is used three times in two sentences. Consider rewriting as:

> In contrast to other methods (e.g. type maps), codecs allow for greater customization and handling of different data types. They separate logic for BSON encoding and decoding from the domain classes, which also enables BSON to be decoded into plain old PHP objects (POPOs).

You can also consider deleting "(POPOs)" here since we don't refer back to this acronym later in the document -- and I'm not sure how widely known it is by our audience.

docs/tutorial/codecs.txt
```diff
@@ -0,0 +1,108 @@
...
+   :language: php
+
+Handling Fields and Data Types
+------------------------------
+
+The previous example showed how to define a codec for a specific class. However, you may want to create a codec that
+handles a particular data type in any document. This can be achieved by implementing the ``MongoDB\Codec\Codec``
+interface.
+
+The following example defines a codec to store ``DateTimeInterface`` instances as BSON dates including the timezone:
```
jmikola, line 68:
```suggestion
The following example defines a codec that stores ``DateTimeInterface`` instances as an embedded document containing a BSON date and accompanying timezone string. Those same embedded documents can then be translated back into a ``DateTimeInterface`` during BSON decoding.
```

This drops the colon but should explain the entire example you're about to present. I think it's worth describing the encode _and_ decode steps in prose here, as the remainder of the section is pretty light on words.

docs/tutorial/custom-types.txt
```diff
@@ -4,6 +4,12 @@ Custom Data-Types
 
 .. default-domain:: mongodb
 
+.. note::
+
+   This tutorial explains implementing custom data types using the :php:`MongoDB\\BSON\\Persistable <class.mongodb-bson-persistable>`
+   interface found in the MongoDB extension. Consider using a codec instead to decouple the MongoDB persistence logic
+   from your business logic. See the :doc:`Codec tutorial </tutorial/codecs>` for an example.
```
jmikola, line 11:
```suggestion
   from your business logic. See the :doc:`Codecs tutorial </tutorial/codecs>` for an example.
```

Matches the actual page title. Alternatively, you can use "codec" here but keep it lowercase so it's clear you're not referring to the page title.

docs/tutorial/codecs.txt
GromNaN: You'll have an army of lawyers suing you for making POPOs.

docs/tutorial/codecs.txt
alcaeus: Forgot this comment - the section was removed. I'll keep this language in mind if we add something like this again.

docs/examples/codecs/handling-data-types/PersonCodec.php
```diff
@@ -0,0 +1,55 @@
...
+
+    public function decode(mixed $value): Person
+    {
+        if (! $this->canDecode($value)) {
+            throw UnsupportedValueException::invalidDecodableValue($value);
+        }
+
+        return new Person(
+            $value->get('name'),
+            $this->dateTimeCodec->decode($value->get('createdAt')),
```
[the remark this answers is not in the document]
alcaeus: It duplicates the data in memory. Each `Document` and `PackedArray` instance is tied to its own `bson_t*` so we don't have to start refcounting the libmongoc BSON structures.

docs/examples/codecs/handling-documents/disabling-codec.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+// Returns a document using the default type map
```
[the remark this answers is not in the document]
alcaeus: Ah, good point. Specifying an operation-level `typeMap` option also overrides the collection-level codec. I've added a sentence explaining this and updated the example to show both ways.

docs/examples/codecs/handling-data-types/DateTimeCodec.php
```diff
@@ -0,0 +1,61 @@
...
+    use EncodeIfSupported;
+
+    public function canDecode(mixed $value): bool
+    {
+        /*
+         * This codec inspects the BSON document to ensure it has the fields it expects, and that those fields are of
+         * the correct type. This is a robust approach to avoid decoding document that are not supported and would cause
+         * exceptions.
+         *
+         * For large documents, this can be inefficient as we're inspecting the entire document 4 times (once for each
```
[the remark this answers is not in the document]
alcaeus: That is correct. Inspecting large documents will be slow, but since we've removed libraries users are unlikely to run into such scenarios as they'd be using specific codecs for specific fields.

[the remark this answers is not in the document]
alcaeus: Ah, I was typically using 3 as the cut-off. Good to know 👍

docs/examples/codecs/handling-data-types/DateTimeCodec.php
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    public function decode(mixed $value): DateTimeImmutable
+    {
+        if (! $this->canDecode($value)) {
+            throw UnsupportedValueException::invalidDecodableValue($value);
+        }
+
+        $timeZone = new DateTimeZone($value->get('tz'));
+        $dateTime = $value->get('utc')
```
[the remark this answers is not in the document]
alcaeus: Yes, which is why I was removing them for simplicity. People familiar with Psalm will figure it out without a number of `@psalm-` annotations, whereas they would confuse people that aren't as familiar with the concept.

docs/examples/codecs/handling-data-types/Person.php
alcaeus: That is correct, they all have minor changes and thus can't be reused.

docs/tutorial/codecs.txt
alcaeus: Haha. Leaving this unresolved so that someone can eventually stumble upon this while cave diving into old pull requests.

docs/examples/codecs/handling-data-types/DateTimeCodec.php
```diff
@@ -0,0 +1,61 @@
...
+
+    public function canEncode(mixed $value): bool
+    {
+        return $value instanceof DateTimeInterface;
+    }
+
+    public function decode(mixed $value): DateTimeImmutable
+    {
+        if (! $this->canDecode($value)) {
+            throw UnsupportedValueException::invalidDecodableValue($value);
```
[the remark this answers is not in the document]
alcaeus: Yes. Will add in a separate PR.

jmikola (approved):

PR #1156, opened 2023-09-08
Fix signatures of bsonSerialize() methods

alcaeus:
https://github.com/mongodb/mongo-php-driver/pull/1449 changed the tentative return type of `MongoDB\BSON\Serializable` to no longer allow any `object` (which was incorrect as we never supported returning any object other than `stdClass` instances), instead restricting it to `stdClass` instances. This PR deals with the resulting deprecation notices emitted on PHP 8.1 and newer.

### DIFF
--- a/src/Model/SearchIndexInput.php
+++ b/src/Model/SearchIndexInput.php
@@ -19,6 +19,7 @@ namespace MongoDB\Model;
 
 use MongoDB\BSON\Serializable;
 use MongoDB\Exception\InvalidArgumentException;
+use stdClass;
 
 use function is_string;
 use function MongoDB\is_document;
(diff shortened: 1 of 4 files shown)


src/Model/SearchIndexInput.php
```diff
@@ -66,7 +67,7 @@ public function __construct(array $index)
      * @see \MongoDB\Collection::createSearchIndexes()
      * @see https://php.net/mongodb-bson-serializable.bsonserialize
      */
-    public function bsonSerialize(): object
+    public function bsonSerialize(): stdClass
     {
         return (object) $this->index;
```
GromNaN, line 72: The psalm inferred type is incorrect. `$this->index` is an array, so `(object) $this->index` cannot be anything different that `stdClass`.
Let's add this warnings to the baseline.

> src/Model/SearchIndexInput.php:70:38: MoreSpecificReturnType: The declared return type 'stdClass' for MongoDB\Model\SearchIndexInput::bsonSerialize is more specific than the inferred return type 'object' (see https://psalm.dev/070)

GromNaN: Reported issue: https://github.com/vimeo/psalm/issues/10019

GromNaN (approved):

PR #1158, opened 2023-09-12
PHPLIB-1243: Adapter to pipe ext-mongodb logs to a PSR-3 logger

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1243


src/PsrLogAdapter.php
```diff
@@ -0,0 +1,72 @@
...
+        $this->psrLogger->log(
+            $this->driverLevelToPsrLevel($level),
+            $domain . ': ' . $message,
+        );
+    }
+
+    // TODO: Refactor this method as a private const array map on PsrLogAdapter
+    private function driverLevelToPsrLevel(int $level): string
+    {
+        switch ($level) {
```
alcaeus, line 44: Noted that `match` would be nice to have here. Not long hopefully 🤞

src/PsrLogAdapter.php
```diff
@@ -0,0 +1,72 @@
...
+            /* PSR does not define a "trace" level, so route both levels to
+             * "debug". That said, trace logging is very verbose and should not
+             * be encouraged when using a PSR logger. */
+            case DriverLoggerInterface::LEVEL_DEBUG:
+            case DriverLoggerInterface::LEVEL_TRACE:
+                return PsrLogLevel::DEBUG;
+
+            /* A default case is intentionally not handled since libmongoc and
+             * PHPC should avoid emitting bogus levels; however, we can consider
+             * throwing an UnexpectedValueException here. */
```
alcaeus, line 69: Throwing may be undesirable, but we may want to emit a warning or error about an undefined level and logging the message with the same level for good measure

src/Client.php
```diff
@@ -329,6 +331,19 @@ public function selectDatabase(string $databaseName, array $options = [])
...
+    public function setLogger(LoggerInterface $logger): void
+    {
+        /* TODO: lazily initialize a PsrLogAdapter for this client and register
+         * it with PHPC. Consider what happens if multiple clients instances in
+         * an app register the same PSR logger. That logger will likely receive
+         * duplicate messages, since PHPC's own checks would only prevent the
+         * same PsrLogAdapter instance from being registered multiple times.
+         *
+         * Perhaps we can work around this by making PsrLogAdapter a singleton
+         * or utilize a static map of PSR loggers.
```
alcaeus, line 343: A static map of PSR loggers may be sensible - I don't think a singleton is the right choice here. Technically, we'd want to have one instance of `PsrLogAdapter` per `LoggerInterface` at most, so not quite a singleton.

src/PsrLogAdapter.php
GromNaN: This is the easiest place to log a message. I agree with @alcaeus that you should log a warning that the level is incorrect.

src/Client.php
GromNaN: The current implementation is one `PsrLogAdapter` per `LoggerInterface`. The proposition would be to have a single `PsrLogAdapter` (singleton) sending to all `LoggerInterface` that it contains. The adapter would have methods to add and remove `LoggerInterface` objects.

src/Client.php
```diff
@@ -329,6 +331,19 @@ public function selectDatabase(string $databaseName, array $options = [])
         return new Database($this->manager, $databaseName, $options);
     }
 
+    public function setLogger(LoggerInterface $logger): void
```
GromNaN, line 334: How would you remove the logger? It should nullable.

At the very least, it must be unregistered from the driver when the client is destroyed.

src/PsrLogAdapter.php
jmikola: > emit a warning or error about an undefined level

Can you clarify what you mean by "emit a warning or error"? Are you suggesting we log an additional message through the PSR logger since we already have access to it?

> logging the message with the same level for good measure

And here, are you suggesting we log the original message (received from PHPC) at a warning/error level?

What are you thoughts on logging a single message that reports the invalid level, and appends the original log message? Then we only end up with a single log line with all of the original information _and_ our error.

src/Client.php
jmikola: Sorry, I wasn't actually suggesting that we make PsrLogAdapter a singleton. A static property is what I had in mind, as that would allow all PsrLogAdapter instances (at most one per Client) to track which loggers have been globally registered.

```mermaid
graph TD
    A[Client1] --> B[PsrLogAdapter1]
    B --> C[PsrLogger]
    D[Client2] --> E[PsrLogAdapter2]
    E --> C
```

In this example, the static map shared by all PsrLogAdapter instances would ensure that PHPLIB never registers the PsrLogAdapter a second time.

Trying to think of how this could be problematic:

 1. Construct Client1 and assign PsrLogger1 to it. PsrLogAdapter1 is created and registered with PHPC.
 2. Construct Client2 and assign PsrLogger1 to it. PsrLogAdapter2 is created but not registered, since it knows another PsrLogAdapter instance has already been registered for PsrLogger1.
 3. Destroy Client1 and allow PsrAdapter1 to be unregistered.

In that case, Client2's logging no longer functions because PsrLogAdapter2 was never registered. I'll give this some more thought.

tests/PsrLogAdapterTest.php
```diff
@@ -0,0 +1,69 @@
...
+use Psr\Log\AbstractLogger;
+use Psr\Log\LoggerInterface;
+use Psr\Log\LogLevel;
+
+use function func_get_args;
+use function MongoDB\Driver\Logging\log;
+
+class PsrLogAdapterTest extends BaseTestCase
+{
+    private PsrLogAdapter $adapter;
```
GromNaN, line 17: Unused. And that's good.

src/PsrLogAdapter.php
```diff
@@ -0,0 +1,96 @@
...
+    public static function addLogger(PsrLogger $psrLogger): void
+    {
+        $instance = self::getInstance();
+
+        $instance->psrLoggers->attach($psrLogger);
+
+        addLogger($instance);
+    }
+
+    public static function getInstance(): self
```
GromNaN, line 59: By making the method private, you would be sure its `log` method is never be called user-side. But maybe you want to use it to log from the PHPLIB client? It should be `@internal` then.

src/PsrLogAdapter.php
```diff
@@ -0,0 +1,96 @@
...
+        DriverLogger::LEVEL_TRACE => PsrLogLevel::DEBUG,
+    ];
+
+    public static function addLogger(PsrLogger $psrLogger): void
+    {
+        $instance = self::getInstance();
+
+        $instance->psrLoggers->attach($psrLogger);
+
+        addLogger($instance);
```
GromNaN, line 56: You may skip this call if the logger is already registered.
```
        if ($instance->psrLoggers->count() === 0) {
            addLogger($instance);
        }
```

composer.json
```diff
@@ -22,6 +22,7 @@
     "require-dev": {
         "doctrine/coding-standard": "^11.1",
         "phpbench/phpbench": "^1.2",
+        "psr/log": "^1.1.4",
```
GromNaN, line 25: The versions 2 and 3 added respectively parameters and return types.
```suggestion
        "psr/log": "^1.1.4|^2|^3",
```

GromNaN (commented):
Much simpler, and it's easier to understand that the logger is global and not client-specific.

src/PsrLogAdapter.php
alcaeus: PHPC handles this already. Since this is not in the hot path I think we can do without duplicating such logic here.

composer.json
GromNaN: We may mark the dependency as "compatible" in composer, one day. https://github.com/composer/composer/issues/11635

src/PsrLogAdapter.php
```diff
@@ -0,0 +1,118 @@
...
+        $this->psrLoggers = new SplObjectStorage();
+    }
+
+    private static function getInstance(): self
+    {
+        if (self::$instance === null) {
+            self::$instance = new self();
+        }
+
+        return self::$instance;
```
GromNaN, line 116: There is an operator for that.
```suggestion
        return self::$instance ??= new self();
```

src/PsrLogAdapter.php
```diff
@@ -0,0 +1,118 @@
...
+        $instance = self::getInstance();
+        $psrLevel = self::PSR_LEVELS[$level];
+        $context = ['domain' => $domain];
+
+        foreach ($instance->psrLoggers as $psrLogger) {
+            $psrLogger->log($psrLevel, $message, $context);
+        }
+    }
+
+    public static function removeLogger(PsrLogger $psrLogger): void
```
GromNaN, line 85: removeLogger should be just after addLogger because they are related. And internal methods should be last (as privates).

src/PsrLogAdapter.php
jmikola: Thanks. I forgot that was introduced in PHP 7.4. :v:

composer.json
jmikola: Resolving this, as `|^2|^3` has been added.

src/PsrLogAdapter.php
jmikola: Aye, I've changed `getInstance()` to private.

tests/PsrLogAdapterTest.php
jmikola: Thanks for reminding me. This is now removed.

src/Client.php
jmikola: Resolving, as Client no longer implements LoggerAwareInterface.

src/PsrLogAdapter.php
jmikola: We now throw a UnexpectedValueException that includes the original level, domain, and message.

src/Client.php
jmikola: Closing the loop here. PsrLogAdapter is now a singleton and collects one or more PsrLoggers. The PsrLogAdapter registered itself when the first PsrLogger is added and unregistered itself when the last logger is removed.

src/PsrLogAdapter.php
jmikola: Spoke to @alcaeus about this. I'll add a skip for PsrLogAdapter in PedantryTest for now and re-order these methods. We can revisit removing PedantryTest altogether in a subsequent PR.

jmikola: I found it was easier to just add `@internal` to the sort logic. Opened [PHPLIB-1238](https://jira.mongodb.org/browse/PHPLIB-1238) to track this (already implemented locally) and will address in a subsequent PR.

jmikola: #1162 is the PR for method sorting changes. If that gets merged first I'll update this PR to be consistent before merging.

src/PsrLogAdapter.php
```diff
@@ -0,0 +1,118 @@
...
+            removeSubscriber($instance);
+        }
+    }
+
+    /**
+     * Writes a log message to all registered PSR loggers.
+     *
+     * @internal
+     */
+    public static function writeLog(int $level, string $domain, string $message): void
```
jmikola, line 100: Per https://github.com/mongodb/mongo-php-driver/pull/1395#discussion_r1326793685, PHPC defines a `mongoc_log()` function for internal tests. I'd ultimately like to have it defined only for test builds ([PHPC-2293](https://jira.mongodb.org/browse/PHPC-2293)), so I think it still makes sense to use our own `writeLog()` function here.

We also discussed only caring about logging through PsrLogAdapter for spec compliance, so it shouldn't be necessary to have PHPLIB log directly to a PHPC mechanism (be it the `mongodb.debug` stream or registered LogSubscriber object(s)).

jmikola: Since we're only dealing with PSR loggers here, does it make any sense to use the LogSubscriber levels here? The [PSR levels](https://www.php-fig.org/psr/psr-3/#5-psrlogloglevel) are consistent with our [logging spec](https://github.com/mongodb/specifications/blob/master/source/logging/logging.rst#log-severity-levels), although the spec defines an additional trace level (how ironic). That said, it's not used so I don't think there's any urgent need to support it.

If we use PSR levels, we'd need a non-optional dependency on `psr/log`. It may make more sense to define our own levels (exactly consistent with the spec) and then maintain a mapping. This seems to be what Monolog does internally (for both PSR and RFC5424).

jmikola:
Note: [PHPLIB-998](https://jira.mongodb.org/browse/PHPLIB-998) covers full compliance with the [Logging spec](https://github.com/mongodb/specifications/blob/master/source/logging/logging.rst), so I think there may be more to do here. For instance, logging is supposed to be controlled via environment variables.

I intend to through the spec again and build a list of outstanding features to implement.

composer.json
```diff
@@ -22,6 +22,7 @@
     "require-dev": {
         "doctrine/coding-standard": "^11.1",
         "phpbench/phpbench": "^1.2",
+        "psr/log": "^1.1.4|^2|^3",
```
alcaeus, line 25: We can add this to the `suggests` section so people know they can use a PSR logger.

src/PsrLogAdapter.php
alcaeus: I think people will get confused either way. Since PSR log levels are strings and libmongoc log levels are integers, how about accepting both?

composer.json
jmikola: Can do. In case you missed @GromNaN's [previous comment](https://github.com/mongodb/mongo-php-library/pull/1158#discussion_r1324489025):

> We may mark the dependency as "compatible" in composer, one day. https://github.com/composer/composer/issues/11635

jmikola: I think this is a moot point. If it was just type hints for the add/removeSubscriber methods, we could stick with `suggestions`, but PsrLogAdapter references the PSR LogLevel constants directly in a private constant, so we actually depend on `psr/log` being available.

src/PsrLogAdapter.php
jmikola: @GromNaN is correct that no users should be calling `writeLog()` directly. It is only intended for internal use within PHPLIB.

That said, after reading through the [Logging spec](https://github.com/mongodb/specifications/blob/master/source/logging/logging.rst#log-severity-levels) earlier today, I realized we should probably use those levels directly. My most recent commit introduces constants for the spec levels and changes `writeLog()` to take those and translate them internally to a PSR level (the only non-standard mapping is spec level `trace` to PSR level `debug`).

On a related note, I also changed `writeLog()` to call the PSR loggers directly instead of forwarding to `LogSubscriber::log()`, since forwarding would limit what levels PHPLIB could utilize (probably not relevant, but worth noting).

src/PsrLogAdapter.php
```diff
@@ -0,0 +1,159 @@
...
+use Psr\Log\LogLevel;
+use SplObjectStorage;
+
+use function MongoDB\Driver\Monitoring\addSubscriber;
+use function MongoDB\Driver\Monitoring\removeSubscriber;
+use function sprintf;
+
+final class PsrLogAdapter implements LogSubscriber
+{
+    /** @internal */
```
jmikola, line 32: This is meant to apply to all of the constants below, but I opted not to repeat this doc block for each declaration.

src/PsrLogAdapter.php
```diff
@@ -0,0 +1,159 @@
...
+use MongoDB\Exception\UnexpectedValueException;
+use Psr\Log\LoggerInterface;
+use Psr\Log\LogLevel;
+use SplObjectStorage;
+
+use function MongoDB\Driver\Monitoring\addSubscriber;
+use function MongoDB\Driver\Monitoring\removeSubscriber;
+use function sprintf;
+
+final class PsrLogAdapter implements LogSubscriber
```
jmikola, line 30: Ultimately, I think this entire class can be `@internal`, but I'm not going to do so now. The only user-facing API here is add/removeLogger, and I think those might be better presented as [functions](https://www.mongodb.com/docs/php-library/current/reference/functions/) like `with_transaction()`. Doing so would make it easier to keep this class out of view.

Also, the [Logging spec](https://www.mongodb.com/docs/php-library/current/reference/functions/) requires an API for configuring logging at a component level, which we presently can't implement because libmongoc itself doesn't. In the meantime I've asked the C team about the timeline for [CDRIVER-3775](https://jira.mongodb.org/browse/CDRIVER-3775) and likelihood that it would actually comply with the spec (since `mongoc-log.c` has several design conflicts).

Let's focus on getting https://github.com/mongodb/mongo-php-driver/pull/1395 merged first and I'll see what assurances I can extract from the C team to help inform the design we'll need in PHPLIB to comply with the spec.

Also, while we _could_ implement command and SDAM logging on our own using PHPC subscribers, that would set us up for duplicate logging once libmongoc gets around to implementing the spec and starts emitting its own messages. And since we can't prevent folks from using newer PHPC versions with older PHPLIB versions, there wouldn't be an easy way to work around the duplication.

tests/PsrLogAdapterTest.php
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    public function testLog(): void
+    {
+        /* This uses PHPC's internal mongoc_log() function to write messages
+         * directly to libmongoc. Those messages are then relayed to
+         * PsrLogAdapter and forwarded to each registered PSR logger.
+         *
+         * Note: it's not possible to test PsrLogAdapter::log() with an invalid
+         * level since mongoc_log() already validates its level parameter. */
```
jmikola, line 40: Although I ended up removing the previous test that called `log()` with invalid levels, it was adapted for `testWriteLogWithInvalidLevel()` below.

.evergreen/config.yml
```diff
@@ -766,17 +766,20 @@ axes:
     display_name: Driver Version
     values:
       - id: "oldest-supported"
-        display_name: "PHPC 1.16.0"
+        display_name: "PHPC 1.17.0"
         variables:
-          EXTENSION_VERSION: "1.16.0"
+          EXTENSION_REPO: "https://github.com/jmikola/mongo-php-driver"
+          EXTENSION_BRANCH: "phpc-2180"
```
alcaeus, line 772: Reminder to change this back to the main repo/branch.

.github/workflows/tests.yml
```diff
@@ -34,34 +34,34 @@ jobs:
         mongodb-version:
           - "4.4"
         driver-version:
-          - "stable"
+          - "jmikola/mongo-php-driver@phpc-2180"
```
alcaeus, line 37: Reminder to change this back to the main repo/branch.

src/functions.php
```diff
@@ -46,6 +47,26 @@
 use function MongoDB\BSON\toPHP;
 use function substr;
 
+/**
+ * Registers a PSR-3 logger to receive log messages from the driver/library.
+ *
+ * Adding a registered logger is a NOP.
+ */
+function add_logger(LoggerInterface $logger): void
```
alcaeus, line 55: Seeing this made me realise that all functions use snake_case naming instead of camelCase. I'm not sure why we opted to do that, but I'd be in favour of changing that. Most functions in here are internal, with `with_transaction` and `is_document` being the lone exceptions.

I've created PHPLIB-1244 to track this, so feel free to leave the names as-is for consistency, but if we agree on this I'd like to take care of this before releasing 1.17.

alcaeus (commented):
LGTM with the repo/branch changes for builds.

src/functions.php
jmikola: `with_transaction()` came much later, and probably should have been `withTransaction()` since we already had camelCase in PHPC at the time.

There's no good reason that other, internal functions were snake_case other than conforming to other PHP internal functions (`is_pipeline` vs. `is_int`).

composer.json
GromNaN: This is still an optional dependency, because logging is an optional feature. You could throw an exception before the PsrLogAdapter class declaration, telling a psr-3 compatible logger needs to be installed.

jmikola: @GromNaN: I expect at some point PHPLIB will need to write its own log messages for specs, at which point we'd need to introduce the dependency then.

If you feel strongly about making it optional now and deferring the hard requirement until then, I can implement an exception if the PSR classes don't exist.

alcaeus (approved):

src/functions.php
```diff
@@ -46,6 +47,26 @@
 use function MongoDB\BSON\toPHP;
 use function substr;
 
+/**
+ * Registers a PSR-3 logger to receive log messages from the driver/library.
+ *
+ * Adding a registered logger is a NOP.
```
GromNaN, line 53: "NOP" is a jargon I don't use.

```suggestion
 * Calling this method again with a logger that has already been added will have no effect.
```

src/functions.php
```diff
@@ -46,6 +47,26 @@
...
+ */
+function addLogger(LoggerInterface $logger): void
+{
+    PsrLogAdapter::addLogger($logger);
+}
+
+/**
+ * Unregisters a PSR-3 logger.
+ *
+ * Removing an unregistered logger is a NOP.
```
GromNaN, line 63:
```suggestion
 * Calling this method with a logger that has not been added will have no effect.
```

composer.json
GromNaN: Ok, let's stick with what's most standard with a required dependency.

GromNaN (approved):

src/PsrLogAdapter.php
```diff
@@ -0,0 +1,161 @@
...
+{
+    public const EMERGENCY = 0;
+    public const ALERT = 1;
+    public const CRITICAL = 2;
+    public const ERROR = 3;
+    public const WARN = 4;
+    public const NOTICE = 5;
+    public const INFO = 6;
+    public const DEBUG = 7;
+    public const TRACE = 8;
```
jmikola, line 48: Note: I removed the `LEVEL_` prefix on these constants, as it conflicted with the definitions on the LogSubscriber interface. PHP didn't seem to care about that, but Psalm did (noticed after correcting the extension version in the Psalm workflow).

These are for internal use only, so I don't particularly care about dropping the prefix. I also considered doing without these and relying on the PSR constants; however, we technically wouldn't have `TRACE` available (despite it being unlikely a spec would require its use).

src/PsrLogAdapter.php
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    private function __construct()
+    {
+        $this->loggers = new SplObjectStorage();
+    }
+
+    private static function getInstance(): self
+    {
+        return self::$instance ??= new self();
```
jmikola, line 159: @GromNaN: Also refactored this per your previous suggestion in https://github.com/mongodb/mongo-php-library/pull/1158#discussion_r1324991480.

src/PsrLogAdapter.php
```diff
@@ -0,0 +1,161 @@
...
+            removeSubscriber($instance);
+        }
+    }
+
+    /**
+     * Writes a log message to all registered PSR loggers.
+     *
+     * This function is intended for internal use within the library.
+     */
+    public static function writeLog(int $specLevel, string $domain, string $message): void
```
GromNaN, line 130: @alcaeus Why this method isn't `@internal`?

alcaeus: I don't think that's necessary since the entire class is internal.

PR #1160, opened 2023-09-13
Remove codec library classes

alcaeus:
The `CodecLibrary` class was introduced as it was necessary when working with lazy BSON codecs. Since those have been removed from the 1.17 release, it does not make sense to introduce the library either, as there's no direct need for it. In fact, in most frameworks (e.g. Laravel, Symfony), each codec would rely on a service locator to receive codecs it needs for data types or embedded documents as a constructor dependency. Until we have a clear need for such functionality, it's best to not introduce code so we are not bound by any BC promises.


GromNaN:
The architecture document needs to be updated. 

https://github.com/mongodb/mongo-php-library/blob/30be2cd1a75bc7d32a436f5a9b30dcf5479ce9e7/src/Codec/architecture.md?plain=1#L33-L39

GromNaN (approved):

PR #1161, opened 2023-09-14
PHPLIB-1235 Implements Single-Doc Benchmarks

GromNaN:
Fix [PHPLIB-1235](https://jira.mongodb.org/browse/PHPLIB-1235)

https://github.com/mongodb/specifications/blob/master/source/benchmarking/benchmarking.rst#single-doc-benchmarks

Result:
```
\MongoDB\Benchmark\DriverBench\SingleDocBench

    benchFindOneByID........................I9 - Mo1.146s (±4.74%)
    benchInsertOne # Small doc..............I0 - Mo3.050m (±0.00%)
    benchInsertOne # Large doc..............I0 - Mo640.560ms (±0.00%)
```


phpbench.json.dist
```diff
@@ -1,6 +1,7 @@
 {
     "$schema": "./vendor/phpbench/phpbench/phpbench.schema.json",
     "runner.bootstrap": "vendor/autoload.php",
+    "runner.file_pattern": ".*Bench.php$",
```
GromNaN, line 4: Asked by PHPBench itself.
```
[WARNING] File "PassThruCodec.php" has been identified as a benchmark file but it
 does not end with `Bench.php`. This behavior is incorrect and will be fixed in a
future version. Set `runner.file_pattern` to `*Bench.php` in `phpbench.json` to
avoid this.

[WARNING] File "ToObjectCodec.php" has been identified as a benchmark file but it
does not end with `Bench.php`. This behavior is incorrect and will be fixed in a
future version. Set `runner.file_pattern` to `*Bench.php` in `phpbench.json` to
avoid this.
```

benchmark/BaseBench.php
```diff
@@ -40,4 +46,9 @@ protected static function getDatabase(): string
     {
         return getenv('MONGODB_DATABASE') ?: 'phplib_test';
     }
+
+    protected static function readJsonFile(string $path): stdClass
+    {
+        return json_decode(file_get_contents($path), false, 512, JSON_THROW_ON_ERROR);
```
GromNaN, line 52: Factorized this code used in several tests.

benchmark/DriverBench/SingleDocBench.php
```diff
@@ -0,0 +1,65 @@
...
+    #[Bench\Revs(1)]
+    #[Bench\Iterations(10)]
+    public function benchFindOneByID(): void
+    {
+        $collection = self::getCollection();
+
+        for ($id = 1; $id <= 10_000; $id++) {
+            $collection->findOne(['_id' => $id]);
+        }
+    }
```
GromNaN, line 38: Spec: https://github.com/mongodb/specifications/blob/master/source/benchmarking/benchmarking.rst#find-one-by-id

I could use the `Bench\Rev` annotation to run the test several times instead of having the loop in the test.

benchmark/DriverBench/SingleDocBench.php
```diff
@@ -0,0 +1,65 @@
...
+    public function prepareDatabase(): void
+    {
+        self::getCollection()->drop();
+    }
+
+    public function setupFindOneByID(): void
+    {
+        $tweet = (array) self::readJsonFile(self::TWEET_FILE_PATH);
+        $docs = array_map(fn ($id) => $tweet + ['_id' => $id], range(1, 10_000));
+        self::getCollection()->insertMany($docs);
```
GromNaN, line 25: > For each of the 10,000 sequential `_id` numbers, issue a find command for that _id on the 'corpus' collection and retrieve the single-document result.

As we can see in the "small document insert one" benchmark, inserting one-by-one is very slow. That's why I used `insertMany`. That should not matter as this is a setup operation out of the performance measurements.

benchmark/DriverBench/SingleDocBench.php
```diff
@@ -0,0 +1,65 @@
...
+use function array_map;
+use function range;
+
+/** @see https://github.com/mongodb/specifications/blob/ddfc8b583d49aaf8c4c19fa01255afb66b36b92e/source/benchmarking/benchmarking.rst#single-doc-benchmarks */
+#[Bench\BeforeMethods('prepareDatabase')]
+final class SingleDocBench extends BaseBench
+{
+    public function prepareDatabase(): void
+    {
+        self::getCollection()->drop();
```
GromNaN, line 18: > Drop the 'perftest' database.

I don't think the database name matter. It's `phplib_test`.

https://github.com/mongodb/mongo-php-library/blob/758984085c6f516df119b98eb2613e92516e1a18/benchmark/BaseBench.php#L39-L42

And the collection is `perftest`, not `corpus`.

https://github.com/mongodb/mongo-php-library/blob/758984085c6f516df119b98eb2613e92516e1a18/benchmark/BaseBench.php#L27-L32

GromNaN (commented):
In the spec, [Measurement](https://github.com/mongodb/specifications/blob/master/source/benchmarking/benchmarking.rst#measurement) is supposed to output MB/s, I'm not sure this is something we can do with PHPBench, but I can explore reporting features.

benchmark/DriverBench/SingleDocBench.php
```diff
@@ -0,0 +1,65 @@
...
+        yield 'Large doc' => [
+            'document' => self::readJsonFile(self::LARGE_FILE_PATH),
+            'repeat' => 10,
+        ];
+    }
+
+    /** @param array{document: object, repeat: int} $params */
+    #[Bench\ParamProviders('provideInsertOneParams')]
+    #[Bench\Revs(1)]
+    #[Bench\Iterations(1)]
```
alcaeus, line 56: Same here, fall back to the default iterations:

```suggestion
```

alcaeus (commented):
A short note on the iterations: multiple iterations can be run to ensure that the benchmarks are consistent. For example, we could let phpbench fail the benchmark suite if any benchmark has a deviation of more than x%. We currently don't do this, but will eventually add such assertions in CI. Therefor, I would recommend not hardcoding any iteration number in the benchmarks, instead relying on what's configured in `phpbench.json.dist` and `phpbench.json`.

benchmark/DriverBench/SingleDocBench.php
```diff
@@ -0,0 +1,65 @@
...
+
+    #[Bench\BeforeMethods('setupFindOneByID')]
+    #[Bench\Revs(1)]
+    #[Bench\Iterations(10)]
+    public function benchFindOneByID(): void
+    {
+        $collection = self::getCollection();
+
+        for ($id = 1; $id <= 10_000; $id++) {
+            $collection->findOne(['_id' => $id]);
```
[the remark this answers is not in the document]
GromNaN: It's completely random on my laptop.

1st run
```
\MongoDB\Benchmark\DriverBench\SingleDocBench

    benchFindOneById # default..............I2 - Mo1.125s (±1.32%)
    benchFindOneById # bson typemap.........I2 - Mo1.003s (±2.78%)
```

2nd run
```
\MongoDB\Benchmark\DriverBench\SingleDocBench

    benchFindOneById # default..............I2 - Mo1.099s (±1.78%)
    benchFindOneById # bson typemap.........I2 - Mo1.167s (±8.18%)
```

benchmark/DriverBench/SingleDocBench.php
```diff
@@ -0,0 +1,65 @@
...
+    public function setupFindOneByID(): void
+    {
+        $tweet = (array) self::readJsonFile(self::TWEET_FILE_PATH);
+        $docs = array_map(fn ($id) => $tweet + ['_id' => $id], range(1, 10_000));
+        self::getCollection()->insertMany($docs);
+    }
+
+    #[Bench\BeforeMethods('setupFindOneByID')]
+    #[Bench\Revs(1)]
+    #[Bench\Iterations(10)]
```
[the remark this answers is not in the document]
GromNaN: We'll have to loop less than 10,000. With 10 iterations of 2m, it takes 20 minutes to run a single case!

benchmark/DriverBench/SingleDocBench.php
alcaeus: Yes. When you say it like that, we really shouldn't do this 😅

GromNaN:
Running this new benchmark takes 35 minutes on my laptop.

```console
❯ time ./vendor/bin/phpbench run --filter SingleDocBench
PHPBench (1.2.14) running benchmarks... #standwithukraine
with configuration file: /Users/jerome/Develop/mongo-php-library/phpbench.json.dist
with PHP version 8.2.9, xdebug ✔, opcache ❌

\MongoDB\Benchmark\DriverBench\SingleDocBench

    benchRunCommand.........................I2 - Mo693.331ms (±1.44%)
    benchFindOneById # default..............I2 - Mo1.097s (±6.82%)
    benchFindOneById # bson typemap.........I2 - Mo968.745ms (±0.45%)
    benchInsertOne # Small doc..............I2 - Mo2.812m (±1.93%)
    benchInsertOne # Small BSON doc.........I2 - Mo2.928m (±0.58%)
    benchInsertOne # Large doc..............I2 - Mo632.107ms (±4.04%)
    benchInsertOne # Large BSON doc.........I2 - Mo709.506ms (±4.88%)

Subjects: 3, Assertions: 0, Failures: 0, Errors: 0
./vendor/bin/phpbench run --filter SingleDocBench  19.94s user 7.76s system 1% cpu 35:20.41 total
```

benchmark/DriverBench/SingleDocBench.php
```diff
@@ -0,0 +1,65 @@
...
+namespace MongoDB\Benchmark\DriverBench;
+
+use Generator;
+use MongoDB\Benchmark\BaseBench;
+use PhpBench\Attributes as Bench;
+
+use function array_map;
+use function range;
+
+/** @see https://github.com/mongodb/specifications/blob/ddfc8b583d49aaf8c4c19fa01255afb66b36b92e/source/benchmarking/benchmarking.rst#single-doc-benchmarks */
```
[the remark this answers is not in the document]
GromNaN: Added. I didn't realize it was an actual test.

benchmark/DriverBench/SingleDocBench.php
```diff
@@ -0,0 +1,65 @@
...
+    {
+        yield 'Small doc' => [
+            'document' => self::readJsonFile(self::SMALL_FILE_PATH),
+            'repeat' => 10_000,
+        ];
+
+        yield 'Large doc' => [
+            'document' => self::readJsonFile(self::LARGE_FILE_PATH),
+            'repeat' => 10,
+        ];
```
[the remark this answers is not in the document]
GromNaN: Added the BSON Document version.

alcaeus:
> Running this new benchmark takes 35 minutes on my laptop.

The main cause for this seems to be the "Small doc" benchmark:
 
> ```
>     benchInsertOne # Small doc..............I2 - Mo2.812m (±1.93%)
>     benchInsertOne # Small BSON doc.........I2 - Mo2.928m (±0.58%)
> ```

I would suggest dropping the second data set, as decoding to raw BSON doesn't yield a significantly different result from the original, as network overhead is likely a big factor in the timing here. At the same time, limiting this test to a single iteration would make sense given the runtime. That would reduce the time to ~3 minutes, which is much more tolerable.

However, given these results I'd still like to check with other drivers to see how long this process takes in their drivers and whether we should expect this kind of time.

benchmark/DriverBench/SingleDocBench.php
```diff
@@ -0,0 +1,121 @@
...
+        self::getCollection()->drop();
+    }
+
+    /** @see https://github.com/mongodb/specifications/blob/ddfc8b583d49aaf8c4c19fa01255afb66b36b92e/source/benchmarking/benchmarking.rst#run-command */
+    public function benchRunCommand(): void
+    {
+        $manager = self::createClient()->getManager();
+
+        for ($i = 0; $i < 10_000; $i++) {
+            $manager->executeCommand(self::getDatabase(), new Command(['hello' => true]));
```
alcaeus, line 33: You could extract the method call for benchmarking purposes. 10k method calls will add a certain overhead which we don't want to include in our testing. Command creation can be in the loop as that would usually be part of this kind of workflow.

```suggestion
        $manager = self::createClient()->getManager();
        $database = self::getDatabase();

        for ($i = 0; $i < 10_000; $i++) {
            $manager->executeCommand($database, new Command(['hello' => true]));
```

alcaeus (approved):

jmikola (commented):
@GromNaN: Apologies. I realized I had some pending comments that were never published. Nothing major, so feel free to ignore or address in a separate commit as you like.

benchmark/DriverBench/SingleDocBench.php
```diff
@@ -0,0 +1,113 @@
...
+     * @param array{document: object|array, repeat: int, options?: array} $params
+     */
+    #[ParamProviders('provideInsertOneParams')]
+    #[Revs(1)]
+    public function benchInsertOne(array $params): void
+    {
+        $collection = self::getCollection();
+
+        for ($i = $params['repeat']; $i > 0; $i--) {
+            $collection->insertOne($params['document'], $params['options'] ?? []);
```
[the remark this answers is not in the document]
GromNaN: I removed it before the merge. Good catch.

benchmark/DriverBench/SingleDocBench.php
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /** @see https://github.com/mongodb/specifications/blob/ddfc8b583d49aaf8c4c19fa01255afb66b36b92e/source/benchmarking/benchmarking.rst#run-command */
+    public function benchRunCommand(): void
+    {
+        $manager = self::createClient()->getManager();
+        $database = self::getDatabase();
+
+        for ($i = 0; $i < 10_000; $i++) {
+            $manager->executeCommand($database, new Command(['hello' => true]));
```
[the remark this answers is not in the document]
GromNaN: The spec mention to use true explicitly.

PR #1163, opened 2023-09-15
Add server description to the benchmark report

GromNaN:
- Add custom environment description with MongoDB URI, topology and version
- Remove BaseBench so I don't call its static method from outside, use dedicated classes.
- Get Warmup change from https://github.com/mongodb/mongo-php-library/pull/1161#discussion_r1326505284 (I'll open an issue on phpbench)

This is to highlight that server topology have a huge impact on performance. https://github.com/mongodb/mongo-php-library/pull/1161#discussion_r1326508198

### Example
image


benchmark/Extension/EnvironmentProvider.php
```diff
@@ -0,0 +1,74 @@
...
+        return ['uri' => $uri];
+    }
+
+    private function getServerInfo(Manager $manager): array
+    {
+        $topology = match ($manager->selectServer()->getType()) {
+            Server::TYPE_STANDALONE => 'standalone',
+            Server::TYPE_LOAD_BALANCER => 'load-balanced',
+            Server::TYPE_RS_PRIMARY => 'replica-set',
+            Server::TYPE_MONGOS => 'shared',
```
alcaeus, line 54:
```suggestion
            Server::TYPE_MONGOS => 'sharded',
```

alcaeus (commented):
Found a tiny typo, LGTM otherwise. Good find about the environment provider, I think that's good information to print about the runs 👍

alcaeus (approved):

benchmark/Utils.php
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    private static function createClient(): Client
+    {
+        return new Client(self::getUri());
+    }
+
+    private static function createCollection(): Collection
+    {
+        return self::getClient()->selectCollection(self::getDatabase(), 'perftest');
```
jmikola, line 42: Is there any need to parameterize the collection name? I assume not now, and if this comes up down the line we can always switch over to a param with a default value.

GromNaN: Seen you comment after merging. We will figure out if we need more that one collection.
Same for client options were not used and that I removed in the PR.

PR #1166, opened 2023-09-19
PHPLIB-1237 Implement Parallel Benchmarks

GromNaN:
Fix [PHPLIB-1237](https://jira.mongodb.org/browse/PHPLIB-1237)

Parallel Benchmarks specs: [LDJSON multi-file import](https://github.com/mongodb/specifications/blob/e09b41df206f9efaa36ba4c332c47d04ddb7d6d1/source/benchmarking/benchmarking.rst#ldjson-multi-file-import)

Implementations:
- Using Driver's BulkWrite in a single thread
- Using library's Collection::insertMany in a single thread
- Using multiple forked threads
- Using amphp/parallel-functions with worker pool

To get the fastest result:
- Reading files is done using `stream_get_line`
- Document insertion is done using Driver's BulkInsert


benchmark/composer.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "mongodb/mongodb-benchmark",
```
GromNaN, line 2: This is a proposal to define different dependencies for the benchmark: create a dedicated composer project. There is the PHP version and now the amphp dependency (which requires PHP 8).

The library `mongodb/mongodb` is installed from the parent path.

benchmark/DriverBench/ParallelBench.php
```diff
@@ -0,0 +1,199 @@
...
+     * Using amphp/parallel-functions with worker pool
+     *
+     * @param array{processes:int, files:string[], batchSize:int} $params
+     */
+    #[BeforeMethods('beforeMultiFileImport')]
+    #[ParamProviders(['provideProcessesParameter'])]
+    #[Revs(1)]
+    public function benchMultiFileImportAmp(array $params): void
+    {
+        wait(parallelMap(
```
GromNaN, line 141: This implementation uses a [worker process](https://github.com/amphp/parallel/blob/1.x/lib/Context/Process.php) that communicate with the parent process to receive serialized functions to call and send serialized results.

benchmark/DriverBench/ParallelBench.php
```diff
@@ -0,0 +1,199 @@
...
+                $pid = pcntl_waitpid(-1, $status);
+                unset($pids[$pid]);
+            }
+
+            $pid = pcntl_fork();
+            if ($pid === 0) {
+                // If we reset, we can garantee that we get a new manager in the child process
+                // If we don't reset, we will get the same manager client_zval in the child process
+                // and share the libmongoc client.
+                Utils::reset();
```
GromNaN, line 108: Comment dictated by @jmikola.
Unsetting the Client/Database/Collection objects is required.

benchmark/DriverBench/ParallelBench.php
```diff
@@ -0,0 +1,199 @@
...
+     */
+    #[BeforeMethods('beforeMultiFileImport')]
+    #[Revs(1)]
+    public function benchMultiFileImport(): void
+    {
+        $collection = Utils::getCollection();
+        foreach (self::getFileNames() as $file) {
+            self::importFile($file, $collection);
+        }
+    }
```
GromNaN, line 82: Added this bench for baseline performance.

GromNaN (commented):
Only Multi-File import is implemented for now, with 3 stratedies:
- Single process, no parallel
- Forked processes using `pcntl` extension
- Worker processes using AMP async framework

Results:
```
\MongoDB\Benchmark\DriverBench\ParallelBench

    benchMultiFileImport....................I2 - Mo10.596s (±1.69%)
    benchMultiFileImportFork # 1 proc.......I2 - Mo12.458s (±0.29%)
    benchMultiFileImportFork # 2 proc.......I2 - Mo6.255s (±0.77%)
    benchMultiFileImportFork # 3 proc.......I2 - Mo3.915s (±4.47%)
    benchMultiFileImportFork # 4 proc.......I2 - Mo2.938s (±0.85%)
    benchMultiFileImportFork # 5 proc.......I2 - Mo2.368s (±1.87%)
    benchMultiFileImportFork # 7 proc.......I2 - Mo1.909s (±2.03%)
    benchMultiFileImportFork # 9 proc.......I2 - Mo1.683s (±7.78%)
    benchMultiFileImportFork # 12 proc......I2 - Mo1.763s (±6.02%)
    benchMultiFileImportFork # 15 proc......I2 - Mo1.806s (±6.66%)
    benchMultiFileImportFork # 19 proc......I2 - Mo1.667s (±9.02%)
    benchMultiFileImportFork # 24 proc......I2 - Mo1.827s (±4.88%)
    benchMultiFileImportFork # 30 proc......I2 - Mo1.474s (±3.26%)
    benchMultiFileImportAmp # 1 proc........I2 - Mo10.494s (±0.62%)
    benchMultiFileImportAmp # 2 proc........I2 - Mo4.849s (±4.03%)
    benchMultiFileImportAmp # 3 proc........I2 - Mo3.314s (±1.43%)
    benchMultiFileImportAmp # 4 proc........I2 - Mo2.504s (±1.19%)
    benchMultiFileImportAmp # 5 proc........I2 - Mo2.066s (±1.31%)
    benchMultiFileImportAmp # 7 proc........I2 - Mo1.667s (±1.84%)
    benchMultiFileImportAmp # 9 proc........I2 - Mo1.505s (±2.48%)
    benchMultiFileImportAmp # 12 proc.......I2 - Mo1.672s (±5.32%)
    benchMultiFileImportAmp # 15 proc.......I2 - Mo1.699s (±5.74%)
    benchMultiFileImportAmp # 19 proc.......I2 - Mo2.015s (±7.34%)
    benchMultiFileImportAmp # 24 proc.......I2 - Mo1.932s (±3.17%)
    benchMultiFileImportAmp # 30 proc.......I2 - Mo2.243s (±16.66%)

```

benchmark/src/DriverBench/ParallelBench.php
```diff
@@ -0,0 +1,199 @@
...
+     * Using amphp/parallel-functions with worker pool
+     *
+     * @param array{processes:int, files:string[], batchSize:int} $params
+     */
+    #[BeforeMethods('beforeMultiFileImport')]
+    #[ParamProviders(['provideProcessesParameter'])]
+    #[Revs(1)]
+    public function benchMultiFileImportAmp(array $params): void
+    {
+        wait(parallelMap(
```
GromNaN, line 141: This implementation uses a [worker process](https://github.com/amphp/parallel/blob/1.x/lib/Context/Process.php) that communicate with the parent process to receive serialized functions to call and send serialized results.

benchmark/src/DriverBench/ParallelBench.php
```diff
@@ -0,0 +1,199 @@
...
+                $pid = pcntl_waitpid(-1, $status);
+                unset($pids[$pid]);
+            }
+
+            $pid = pcntl_fork();
+            if ($pid === 0) {
+                // If we reset, we can garantee that we get a new manager in the child process
+                // If we don't reset, we will get the same manager client_zval in the child process
+                // and share the libmongoc client.
+                Utils::reset();
```
GromNaN, line 108: Comment dictated by @jmikola.
Unsetting the Client/Database/Collection objects is required.

benchmark/src/DriverBench/ParallelBench.php
```diff
@@ -0,0 +1,199 @@
...
+        self::$files = [];
+    }
+
+    /**
+     * Parallel: LDJSON multi-file import
+     * Using single thread
+     */
+    #[BeforeMethods('beforeMultiFileImport')]
+    #[Revs(1)]
+    public function benchMultiFileImport(): void
```
GromNaN, line 76: Added this bench for baseline performance.

benchmark/src/DriverBench/ParallelBench.php
jmikola: I suggest:

> Reset to ensure that the existing libmongoc client (via the Manager) is not re-used by the child process. When the child process constructs a new Manager, the differing PID will result in creation of a new libmongoc client.

Two additional thoughts:

You can consider handling this in `Utils::getClient()` directly by having it track the PID used to create the `$client` singleton. Then, if the PID ever changes, `getClient()` can reset on its own. This might require eliminating the `$database` and `$collection` singletons, though (or applying the PID logic to those getters).

Additionally, I think there's probably no harm in using `disableClientPersistence: true` when creating clients if you don't actually expect to create multiple instances. This wouldn't impact forking at all, though, so I wouldn't say it's necessary.

benchmark/src/DriverBench/ParallelBench.php
```diff
@@ -0,0 +1,199 @@
...
+    /** @var string[] */
+    private static array $files = [];
+
+    public static function beforeClass(): void
+    {
+        // Generate files
+        $fileContents = str_repeat(file_get_contents(Data::LDJSON_FILE_PATH), 5_000);
+        foreach (self::getFileNames() as $file) {
+            file_put_contents($file, $fileContents);
+        }
```
alcaeus, line 58: I like how you prepare the files for each run instead of relying on the archive that contains a few hundred MB of the same data 👍

benchmark/src/DriverBench/ParallelBench.php
```diff
@@ -0,0 +1,199 @@
...
+use const FILE_SKIP_EMPTY_LINES;
+
+/**
+ * For accurate results, run benchmarks on a standalone server.
+ *
+ * @see https://github.com/mongodb/specifications/blob/ddfc8b583d49aaf8c4c19fa01255afb66b36b92e/source/benchmarking/benchmarking.rst#parallel
+ */
+#[BeforeClassMethods('beforeClass')]
+#[AfterClassMethods('afterClass')]
+final class ParallelBench
```
alcaeus, line 47: Considering the runtime of these benchmarks, I think reducing the number of iterations to 1 makes sense here.

composer.json
```diff
@@ -33,13 +32,11 @@
...
     "autoload-dev": {
         "psr-4": {
-            "MongoDB\\Tests\\": "tests/",
-            "MongoDB\\Benchmark\\": "benchmark/"
+            "MongoDB\\Tests\\": "tests/"
         },
         "files": [ "tests/PHPUnit/Functions.php" ]
     },
     "scripts": {
-        "benchmark": "phpbench run --report=aggregate",
```
alcaeus, line 42: We could make this run a `composer update` in the benchmark directory and then invoke the benchmarking script from there. Just an idea to make this a little more accessible.

benchmark/src/DriverBench/ParallelBench.php
```diff
@@ -0,0 +1,199 @@
...
+            [self::class, 'importFile'],
+            // The pool size is the number of processes
+            new DefaultPool($params['processes']),
+        ));
+    }
+
+    public static function provideProcessesParameter(): Generator
+    {
+        // Max number of forked processes
+        for ($i = 1; $i <= 30; $i = (int) ceil($i * 1.25)) {
```
alcaeus, line 153: What do you think of trying exponential increase?
```suggestion
        for ($i = 2; $i <= 32; $i *= 2) {
```

That would run 5 benchmarks (with 2, 4, 8, 16, and 32 processes) instead of the current 12 benchmarks and thus save a bunch of time in addition to the savings we'd get from only running a single iteration.

benchmark/src/DriverBench/ParallelBench.php
GromNaN: I think it's better staying explicit in this benchmark: it teaches the best practice to potential readers and it doesn't impact other benchmarks that don't uses forks.

benchmark/src/DriverBench/ParallelBench.php
GromNaN: Done.

benchmark/src/DriverBench/ParallelBench.php
GromNaN: Comment updated.

composer.json
GromNaN: Added

benchmark/src/DriverBench/ParallelBench.php
GromNaN: I preferred to be verbose so I can set a number of processes that matches with the number of round trip that are necessary to handle 100 files.

GromNaN (commented):
Review applied. Here is the new results on my machine.

I switched to using a driver bulk write to get best performances and remove intermediate PHP objects for command and other things. It's 8.5% faster.

    benchMultiFileImportBulkWrite...........I0 - Mo9.959s (±0.00%)
    benchMultiFileImportInsertMany..........I0 - Mo10.801s (±0.00%)
    benchMultiFileImportFork # 1 proc.......I0 - Mo12.016s (±0.00%)
    benchMultiFileImportFork # 2 proc.......I0 - Mo6.000s (±0.00%)
    benchMultiFileImportFork # 4 proc.......I0 - Mo3.002s (±0.00%)
    benchMultiFileImportFork # 8 proc.......I0 - Mo1.548s (±0.00%)
    benchMultiFileImportFork # 13 proc......I0 - Mo1.504s (±0.00%)
    benchMultiFileImportFork # 20 proc......I0 - Mo1.524s (±0.00%)
    benchMultiFileImportFork # 34 proc......I0 - Mo1.469s (±0.00%)
    benchMultiFileImportAmp # 1 proc........I0 - Mo10.034s (±0.00%)
    benchMultiFileImportAmp # 2 proc........I0 - Mo5.013s (±0.00%)
    benchMultiFileImportAmp # 4 proc........I0 - Mo2.637s (±0.00%)
    benchMultiFileImportAmp # 8 proc........I0 - Mo1.705s (±0.00%)
    benchMultiFileImportAmp # 13 proc.......I0 - Mo1.644s (±0.00%)
    benchMultiFileImportAmp # 20 proc.......I0 - Mo1.783s (±0.00%)
    benchMultiFileImportAmp # 34 proc.......I0 - Mo2.011s (±0.00%)


In CI

benchmark | subject | set | mem_peak | mode | rstdev
-- | -- | -- | -- | -- | --
ParallelBench | benchMultiFileImportBulkWrite |   | 13.361mb | 17.157s | ± 0.00%
ParallelBench | benchMultiFileImportInsertMany |   | 14.100mb | 16.229s | ± 0.00%
ParallelBench | benchMultiFileImportFork | 1 proc | 1.220mb | 18.366s | ± 0.00%
ParallelBench | benchMultiFileImportFork | 2 proc | 1.220mb | 9.754s | ± 0.00%
ParallelBench | benchMultiFileImportFork | 4 proc | 1.220mb | 9.783s | ± 0.00%
ParallelBench | benchMultiFileImportFork | 8 proc | 1.220mb | 9.927s | ± 0.00%
ParallelBench | benchMultiFileImportFork | 13 proc | 1.220mb | 10.010s | ± 0.00%
ParallelBench | benchMultiFileImportFork | 20 proc | 1.220mb | 9.969s | ± 0.00%
ParallelBench | benchMultiFileImportFork | 34 proc | 1.220mb | 10.232s | ± 0.00%
ParallelBench | benchMultiFileImportAmp | 1 proc | 2.833mb | 16.461s | ± 0.00%
ParallelBench | benchMultiFileImportAmp | 2 proc | 2.871mb | 9.082s | ± 0.00%
ParallelBench | benchMultiFileImportAmp | 4 proc | 2.946mb | 8.973s | ± 0.00%
ParallelBench | benchMultiFileImportAmp | 8 proc | 3.099mb | 9.245s | ± 0.00%
ParallelBench | benchMultiFileImportAmp | 13 proc | 3.303mb | 9.487s | ± 0.00%
ParallelBench | benchMultiFileImportAmp | 20 proc | 3.601mb | 9.796s | ± 0.00%
ParallelBench | benchMultiFileImportAmp | 34 proc | 4.158mb | 10.494s | ± 0.00%

benchmark/src/DriverBench/ParallelBench.php
jmikola: > and it doesn't impact other benchmarks that don't uses forks.

Ah, if this is the only file that will utilize forking, then I agree with keeping the logic here.

benchmark/src/DriverBench/ParallelBench.php
```diff
@@ -0,0 +1,210 @@
...
+        $database->drop();
+        $database->createCollection(Utils::getCollectionName());
+    }
+
+    public static function importFile(string $file): void
+    {
+        $namespace = sprintf('%s.%s', Utils::getDatabaseName(), Utils::getCollectionName());
+
+        $bulkWrite = new BulkWrite();
+        foreach (file($file, FILE_IGNORE_NEW_LINES | FILE_NO_DEFAULT_CONTEXT) as $line) {
```
jmikola, line 191: I see `file()` used here and above in `benchMultiFileImportInsertMany()`. Is there any memory concern with loading the entire file into an array?

Asking as I'm not sure how large these are, and if we're also parsing them into BSON then we're likely storing multiple copies of the data in memory. If so, I suppose we could optimize the file reading to use something like [SplFileObject](https://www.php.net/SplFileObject) instead to stream the data.

I see this function is also used _within_ some benchmarks (e.g. `benchMultiFileImportAmp`). Is this something we should be doing in a setup step instead, or does the benchmark actually want to time reading the file from disk?

If the file reading is intentionally part of the benchmark, then maybe you can compare `file()` and `SplFileObject` and use whichever is more performant.

Feel free to resolve if this isn't an issue.

benchmark/src/Utils.php
```diff
@@ -16,7 +16,7 @@ final class Utils
 
     public static function getClient(): Client
     {
-        return self::$client ??= new Client(self::getUri());
+        return self::$client ??= new Client(self::getUri(), ['disableClientPersistence' => true]);
```
jmikola, line 19:
```suggestion
        return self::$client ??= new Client(self::getUri(), [], ['disableClientPersistence' => true]);
```

`disableClientPersistence` is a driver option (third param).

benchmark/src/DriverBench/ParallelBench.php
```diff
@@ -0,0 +1,210 @@
...
+    #[Revs(1)]
+    #[Iterations(1)]
+    public function benchMultiFileImportInsertMany(): void
+    {
+        $collection = Utils::getCollection();
+        foreach (self::getFileNames() as $file) {
+            // Read file contents into BSON documents
+            $docs = array_map(
+                static fn (string $line) => Document::fromJSON($line),
+                file($file, FILE_IGNORE_NEW_LINES | FILE_SKIP_EMPTY_LINES | FILE_NO_DEFAULT_CONTEXT),
```
jmikola, line 93: Same question as my comment in `importFile()` re: use of `file()`.

jmikola (approved):
One actual suggestion to fix a bug, but LGTM after that.

I'll defer to you on whether `file()` is preferable to reading in a stream.

GromNaN:
> I'll defer to you on whether `file()` is preferable to reading in a stream.

`file()` is actually slower than `fgets()` and `SplFileObject` when xdebug is disabled.
I'm switching the implementation to use `fgets()`.

with PHP version 8.2.9, xdebug ✅, opcache ❌

    benchReadUsingFile......................I9 - Mo1.899ms (±4.30%)
    benchReadUsingSplFileObject.............I9 - Mo3.409ms (±3.21%)
    benchReadUsingFgets.....................I9 - Mo1.619ms (±5.43%)


with PHP version 8.2.9, xdebug ❌, opcache ✅

    benchReadUsingFile......................I9 - Mo1.740ms (±3.10%)
    benchReadUsingSplFileObject.............I9 - Mo1.262ms (±4.38%)
    benchReadUsingFgets.....................I9 - Mo1.156ms (±2.80%)


```php
    #[Revs(10)]
    #[Warmup(1)]
    #[Iterations(10)]
    public function benchReadUsingFile(): void
    {
        $file = self::getFileNames()[0];
        foreach (file($file, FILE_IGNORE_NEW_LINES | FILE_NO_DEFAULT_CONTEXT) as $line) {
            // process $line
        }
    }

    #[Revs(10)]
    #[Warmup(1)]
    #[Iterations(10)]
    public function benchReadUsingSplFileObject(): void
    {
        $file = self::getFileNames()[0];
        $file = new \SplFileObject($file);
        foreach ($file as $line) {
            if ($line === '') {
                // process $line
            }
        }
    }

    #[Revs(10)]
    #[Warmup(1)]
    #[Iterations(10)]
    public function benchReadUsingFgets(): void
    {
        $file = self::getFileNames()[0];
        $fh = fopen($file, 'r');
        while (($line = fgets($fh)) !== false) {
            if ($line === '') {
                // process $line
            }
        }
        fclose($fh);
    }
```

.github/workflows/benchmark.yml
```diff
@@ -64,14 +64,15 @@ jobs:
...
       - name: "Run phpbench"
+        working-directory: "./benchmark"
         run: "vendor/bin/phpbench run --report=aggregate --report=bar_chart_time --report=env --output html"
 
       - name: Upload HTML report
         uses: actions/upload-artifact@v3
         with:
           name: phpbench-${{ github.sha }}.html
-          path: .phpbench/html/index.html
+          path: ./benchmark/.phpbench/html/index.html
```
jmikola, line 77: Should we document instructions for referencing benchmarks somewhere? This may warrant a new heading in CONTRIBUTING.md.

GromNaN: I think the goal is to run benchmarks on evergreen and publish reports there. [PHPLIB-1187](https://jira.mongodb.org/browse/PHPLIB-1187)

alcaeus (approved):

PR #1167, opened 2023-09-20
PHP 7.3 polyfill is not necessary because PHP 7.4 is required

GromNaN:
I think it has been forgotten in https://github.com/mongodb/mongo-php-library/pull/1128

- PHP 7.4 don't need PHP.3 polyfill
- the function `get_debug_type` from PHP 8.0 polyfill is used
- the function `array_is_list` from PHP 8.1 polyfill is used

The benchmark sub-project requires PHP 8.1, I disable the installation of this polyfills.

### DIFF
--- a/benchmark/composer.json
+++ b/benchmark/composer.json
@@ -1,26 +1,30 @@
 {
-  "name": "mongodb/mongodb-benchmark",
-  "type": "project",
-  "repositories": [
-    {
-      "type": "path",
-      "url": "../",
-      "symlink": true
+    "name": "mongodb/mongodb-benchmark",
+    "type": "project",
+    "repositories": [
+        {
+            "type": "path",
+            "url": "../",
+            "symlink": true
+        }
+    ],
+    "replace": {
+        "symfony/polyfill-php80": "*",
+        "symfony/polyfill-php81": "*"
+    },
+    "require": {
+        "php": ">=8.1",
+        "ext-pcntl": "*",
+        "amphp/parallel-functions": "^1.1",
+        "mongodb/mongodb": "@dev",
+        "phpbench/phpbench": "^1.2"
+    },
+    "autoload": {
+        "psr-4": {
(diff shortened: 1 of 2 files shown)


benchmark/composer.json
```diff
@@ -1,26 +1,30 @@
 {
-  "name": "mongodb/mongodb-benchmark",
-  "type": "project",
-  "repositories": [
-    {
-      "type": "path",
-      "url": "../",
-      "symlink": true
+    "name": "mongodb/mongodb-benchmark",
```
GromNaN, line 2: Fix indentation.

benchmark/composer.json
```diff
@@ -1,26 +1,30 @@
...
+        {
+            "type": "path",
+            "url": "../",
+            "symlink": true
+        }
+    ],
+    "replace": {
+        "symfony/polyfill-php80": "*",
+        "symfony/polyfill-php81": "*"
+    },
```
GromNaN, line 14: The benchmark sub-project requires PHP 8.1, I disable the installation of this polyfills.

jmikola: Is this the preferred method to avoid installing polyfills that are indirect dependencies? I suppose in an ideal world the polyfill packages would have some way to note that they're redundant for newer PHP versions (an inverse of `requires`), but there's no `composer.json` syntax for that.

jmikola (approved):

benchmark/composer.json
alcaeus: Strictly speaking, it's not necessary to prevent installing the polyfills. The polyfill itself [checks if the function exists](https://github.com/symfony/polyfill-php80/blob/6caa57379c4aec19c0a12a38b59b26487dcfe4b5/bootstrap.php#L37..L39) and doesn't register itself if it does.

That said, I'm fine including these directives here until we can use PHP 8.1.

alcaeus (approved):

PR #1174, opened 2023-09-25
PHPLIB-1249 Init code generator project for aggregation builder

GromNaN:
Fix [PHPLIB-1249](https://jira.mongodb.org/browse/PHPLIB-1249)

Reimplement [aggregation builder prototype](https://github.com/mongodb/mongo-php-library/compare/master...alcaeus:mongo-php-library:skunkworks/agg-builder-prototype).

The branch [feature/agg-builder](https://github.com/mongodb/mongo-php-library/tree/feature/agg-builder) will be the target of all the PRs of the aggregation builder feature.

Use [nette/php-generator](https://github.com/nette/php-generator) to generate operator code. 
Inspired by [AsyncAWS Code Generator](https://github.com/async-aws/aws/tree/master/src/CodeGenerator).


generator/config/config.php
```diff
@@ -0,0 +1,107 @@
...
+use MongoDB\CodeGenerator\ValueClassGenerator;
+
+$src = __DIR__ . '/../../src';
+
+return [
+    // Stages
+    [
+        'configFile' => __DIR__ . '/stages.yaml',
+        'generatorClass' => ValueClassGenerator::class,
+        'namespace' => MongoDB\Aggregation\Stage::class,
```
GromNaN, line 16: About Aggregation Builder namespacing. I'd like to put all the classes in the same `MongoDB\Operator` namespace. They are grouped under this name this in the documentation: https://www.mongodb.com/docs/manual/reference/operator/

```
MongoDB\Operator\Query
MongoDB\Operator\Project
MongoDB\Operator\Update
MongoDB\Operator\Stage
MongoDB\Operator\Pipeline
```
(update operator example is for future, not implemented in this project).

Each of this would be classes with static factory methods for value-object classes in their sub-namespace. So we will have:

```
MongoDB\Operator\Query\AndQuery
MongoDB\Operator\Query\EqQuery
MongoDB\Operator\Query\ElemMatchQuery
...
MongoDB\Operator\Project\ElemMatchProjection
...
MongoDB\Operator\Update\SetUpdate
MongoDB\Operator\Update\PopUpdate
...
MongoDB\Operator\Stage\BucketStage
MongoDB\Operator\Stage\ProjectStage
MongoDB\Operator\Stage\SetStage
...
MongoDB\Operator\Aggregation\AbsAggregation
MongoDB\Operator\Aggregation\BinarySizeAggregation
MongoDB\Operator\Aggregation\EqAggregation
MongoDB\Operator\Aggregation\BinarySizeAggregation
MongoDB\Operator\Aggregation\FirstNAggregation
MongoDB\Operator\Aggregation\MaxNAggregation
```

@jmikola what do you think?

generator/config/config.php
```diff
@@ -0,0 +1,107 @@
...
+        'filePath' => $src . '/Aggregation/Stage/',
+        'interfaces' => [Stage::class],
+        'classNameSuffix' => 'Stage',
+    ],
+    /*
+    // Stage codec
+    [
+        // Stage converters
+        'configFile' => __DIR__ . '/stages.yaml',
+        'generatorClass' => ConverterClassGenerator::class,
```
GromNaN, line 26: I'm going to leave the codec generator for now and start writing a generic codec that can encode any operator: less codec objects to instantiate and maybe easier to write.

generator/config/query-operators.yaml
```diff
@@ -0,0 +1,9 @@
+- name: and
+  args:
+    - name: query
+      type: resolvesToQueryOperator
+      isVariadic: true
```
GromNaN, line 5: We should throw an error if the number of query objects is lower that 1. In Doctrine, they use a required argument before the variadic one.

https://github.com/doctrine/mongodb-odm/blob/9b7ed61f4e12ac0f4f8a18f65dfcfd8a05066dab/lib/Doctrine/ODM/MongoDB/Aggregation/Expr.php#L162

generator/config/config.php
GromNaN: Note that we already have the `MongoDB\Operation` namespace and there could be confusion between `MongoDB\Operation\Aggregate` and `MongoDB\Operator\Aggregation`.

We can choose `MongoDB\Builder` as namespace. Like "query builder", "aggregation builder", "update builder" ...

generator/src/AbstractGenerator.php
```diff
@@ -0,0 +1,127 @@
...
+/** @internal */
+abstract class AbstractGenerator
+{
+    /** @var array<string, list<string|class-string>> */
+    protected array $typeAliases = [
+        'resolvesToExpression' => [Expression\ResolvesToExpression::class, 'array', 'object', 'string', 'int', 'float', 'bool', 'null'],
+        'resolvesToArrayExpression' => [Expression\ResolvesToArrayExpression::class, 'array', 'object', 'string'],
+        'resolvesToBoolExpression' => [Expression\ResolvesToBoolExpression::class, 'array', 'object', 'string', 'bool'],
+        'resolvesToMatchExpression' => ['array', 'object', Expression\ResolvesToMatchExpression::class],
+        'resolvesToNumberExpression' => [Expression\ResolvesToBoolExpression::class, 'array', 'object', 'string', 'int', 'float'],
```
alcaeus, line 34: Just a quick heads up that the type substitution rules have changed since I built the original prototype. Mainly, you can no longer use arrays, objects, or strings to pass in your own expressions or field paths: https://docs.google.com/document/d/1vFWXB_6nnYH6p136alKDnIHekW-JeNA7nH3vWSriDtY/edit#bookmark=id.3rypi8grpbbo

generator/src/AbstractGenerator.php
```diff
@@ -0,0 +1,127 @@
...
+        sort($docTypes);
+
+        if ($arg->isOptional) {
+            $nativeTypes[] = 'null';
+            $docTypes[] = 'null';
+        }
+
+        return [
+            'native' => implode('|', $nativeTypes),
+            'doc' => implode('|', $docTypes),
```
alcaeus, line 95: Note: you may want to use `array_unique` here in case we ever start out with a union in the config.

generator/config/config.php
jmikola: What is the intention of repeating the namespace component as a class name suffix? Is that to avoid requiring `use as` if multiple operators are included in the same file?

GromNaN: Avoid name restrictions (`And` and `Match` are not allowed as class names) and avoid `use as` when needing to use operators with the same name in the same file (`Query\Eq` and `Aggregation\Eq`).

But this classes will not be used outside of this library or ORMs.

generator/src/AbstractGenerator.php
GromNaN: Added.

generator/src/AbstractGenerator.php
GromNaN: I updated the expression type system, adding a config file to generate expression classes and interfaces.

tests/Builder/BuiderCodecTest.php
```diff
@@ -0,0 +1,120 @@
...
+use MongoDB\Tests\TestCase;
+
+use function array_merge;
+
+class BuiderCodecTest extends TestCase
+{
+    public function testPipeline(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::match(Aggregation::eq('$foo', 1)),
```
alcaeus, line 21: Note that the `$match` stage shouldn't accept the aggregation pipeline operators, except within `$expr`.

tests/Builder/BuiderCodecTest.php
```diff
@@ -0,0 +1,120 @@
...
+
+    /** @see https://www.mongodb.com/docs/manual/reference/operator/aggregation/match/#perform-a-count */
+    public function testPerformCount(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::match(Query::or(
+                ['score' => [Query::gt(70), Query::lt(90)]],
+                ['views' => Query::gte(1000)],
+            )),
+            Stage::group(null, ['count' => Aggregation::sum(1)]),
```
alcaeus, line 43: Consider the named argument form:

```suggestion
            Stage::group(
                _id: null, 
                count: Aggregation::sum(1)
            ),
```

tests/Builder/BuiderCodecTest.php
```diff
@@ -0,0 +1,120 @@
...
+    {
+        $pipeline = new Pipeline(
+            Stage::project([
+                'items' => Aggregation::filter(
+                    input: Expression::fieldPath('items'),
+                    cond: Aggregation::gte(Expression::variable('item.price'), 100),
+                    as: 'item',
+                    limit: $limit,
+                ),
+            ]),
```
alcaeus, line 80: Consider the named argument form:

```suggestion
            Stage::project(
                items: Aggregation::filter(
                    input: Expression::fieldPath('items'),
                    cond: Aggregation::gte(Expression::variable('item.price'), 100),
                    as: 'item',
                    limit: $limit,
                ),
            ),
```

src/Builder/BuilderCodec.php
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    public function encode($value): array|stdClass|string|int|float|bool|null
+    {
+        if (! $this->canEncode($value)) {
+            throw UnsupportedValueException::invalidEncodableValue($value);
+        }
+
+        // A pipeline is encoded as a list of stages
+        if ($value instanceof Pipeline) {
```
alcaeus, line 60: This check and all subsequent checks should be handled by their own encoder; we can then leverage an encoder map to encode values.

src/Builder/BuilderCodec.php
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    private function encodeAsObject($value): stdClass
+    {
+        $result = new stdClass();
+        foreach ($value as $key => $val) {
+            $val = $this->encodeIfSupported($val);
+            if ($val !== null) {
+                $result->{$key} = $val;
+            }
```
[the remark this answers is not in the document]
GromNaN: We will need a special "not specified" object to mark unspecified values...

tests/Builder/BuiderCodecTest.php
jmikola: There's also a typo in this class name (or maybe just the file name). Should be: BuilderCodecTest.

GromNaN: The implementation you proposed for `Stage::group` factory.
```php
    public static function group(mixed $_id = null, array|null|object ...$fields): GroupStage
    {
        if (is_array($fields) && count($fields) === 1 && array_key_first($fields) === 0) {
            $fields = $fields[0];
        }

        return new GroupStage($_id, $fields);
    }
```

This will mess with types because we will have to accept `array|null|object` or a single `array<array|null|object>`.

We can say that `array` are not accepted and replace named arguments by `...[]` syntax when dotted field names are necessary.

```php
Stage::group(
    _id: null, 
    ...['count' => Aggregation::sum(1)]
),
```

src/Builder/Pipeline.php
```diff
@@ -0,0 +1,23 @@
...
+use MongoDB\Exception\InvalidArgumentException;
+
+use function array_is_list;
+
+class Pipeline
+{
+    public array $stages;
+
+    public function __construct(
+        Stage ...$stages
```
GromNaN, line 15: `Pipelines` must be accepted and merged with stages. [PHPLIB-1264](https://jira.mongodb.org/browse/PHPLIB-1264)

src/Builder/BuilderCodec.php
GromNaN: This is a draft implementation to get an overview of how operators and stages are dumped. I'll work on specificities on this tickets:

[PHPLIB-1256](https://jira.mongodb.org/browse/PHPLIB-1256)
[PHPLIB-1257](https://jira.mongodb.org/browse/PHPLIB-1257)
[PHPLIB-1258](https://jira.mongodb.org/browse/PHPLIB-1258)

psalm.xml.dist
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0"?>
 <psalm
+    phpVersion="8.0"
```
GromNaN, line 3: This needs to be configured per directory.

generator/config/config.php
GromNaN: If you have alternative suggestions for the main namespace `MongoDB\Builder\` and `MongoDB\Builder\Aggregation` for aggregation pipeline operators, I'm all ears.

tests/Builder/BuiderCodecTest.php
GromNaN: Playground: https://3v4l.org/QBBie

GromNaN: Tracked in [PHPLIB-1267](https://jira.mongodb.org/browse/PHPLIB-1267)

tests/Builder/BuiderCodecTest.php
GromNaN: Tracked in [PHPLIB-1267](https://jira.mongodb.org/browse/PHPLIB-1267)

psalm-baseline.xml
```diff
@@ -59,6 +82,11 @@
       <code><![CDATA[$mergedDriver['platform']]]></code>
     </MixedAssignment>
   </file>
+  <file src="src/Codec/EncodeIfSupported.php">
+    <InvalidReturnType>
+      <code>($value is NativeType ? BSONType : $value)</code>
```
GromNaN, line 87: @alcaeus error details:

```
ERROR: InvalidReturnType
at src/Codec/EncodeIfSupported.php:46:22
The declared return type '(BSONType:MongoDB\Codec\EncodeIfSupported as mixed)|(TGeneratedFromParam0:fn-mongodb\codec\encodeifsupported::encodeifsupported as mixed)' for MongoDB\Codec\EncodeIfSupported::encodeIfSupported is incorrect, got '(TGeneratedFromParam0:fn-mongodb\codec\encodeifsupported::encodeifsupported as mixed)|array<array-key, mixed>|stdClass|string' (see https://psalm.dev/011)
     * @psalm-return ($value is NativeType ? BSONType : $value)
```

I'm having trouble using the template annotations you've defined.

tests/Builder/BuiderCodecTest.php
jmikola: I'm trying to wrap my head around the conflict between named args and unpacking:

```
/* PHP 8.1+ allows this and merges the named argument into unpacked args.
 * PHP 8.0 yields a fatal error ("Cannot combine named arguments and argument unpacking") */
acceptsVariadicNamed(...['foo.bar' => 'bar', 'bar' => 'baz'], baz: 'bla');

/* PHP 8.0+ yields a fatal error ("Cannot use argument unpacking after named arguments") */
acceptsVariadicNamed(baz: 'bla', ...['foo.bar' => 'bar', 'bar' => 'baz']);
```

I recall you proposed requiring PHP 8.1 for the query builder (might have been via Slack), so let's assume the first call style above is used.

Is that still problematic for `$group` because it would require expressing all fields _before_ the required `$_id` parameter?

> This will mess with types because we will have to accept `array|null|object` or a single `array<array|null|object>`.

If each optional field must be an object of the form `{ <accumulator1> : <expression1> }`, I have two questions:

 * Why would `null` be accepted?
 * If an `array` is accepted, wouldn't the actual type be an array with a single string key corresponding to an [accumulator operator](https://www.mongodb.com/docs/manual/reference/operator/aggregation/group/#std-label-accumulators-group)? In that case, wouldn't that be suitable to distinguish the type from a list of `array|object` values (at least in Psalm)? 

https://www.mongodb.com/docs/manual/reference/operator/aggregation/group/#accumulator-operator

src/Builder/Pipeline.php
jmikola: Was that an original requirement of the [technical design](https://docs.google.com/document/d/1vFWXB_6nnYH6p136alKDnIHekW-JeNA7nH3vWSriDtY/edit)? If so, I'm having trouble finding it in the document.

Is this syntactic sugar preferable to adding a method on a Pipeline to return its Stages as a list, and then merging those with a splat operator (e.g. `...$pipeline->getStages(),`)?

phpcs.xml.dist
```diff
@@ -13,10 +13,17 @@
...
+    <file>generator/src</file>
+    <file>generator/config</file>
+    <file>generator/bin</file>
     <file>tests</file>
     <file>tools</file>
     <file>rector.php</file>
 
+    <!-- Ignore auto-generated code -->
+    <exclude-pattern>src/Builder/(Aggregation|Expression|Query|Stage)\.php</exclude-pattern>
+    <exclude-pattern>src/Builder/(Aggregation|Expression|Query|Stage)/*\.php</exclude-pattern>
```
jmikola, line 25: I assume this is in sync with the generated paths in `.gitattributes`, yes?

phpcs.xml.dist
```diff
@@ -13,10 +13,17 @@
...
+
     <!-- Target minimum supported PHP version -->
     <config name="php_version" value="70400"/>
```
jmikola, line 28: Should this also get a TODO item referencing [PHPLIB-1268](https://jira.mongodb.org/browse/PHPLIB-1268)?

src/Builder/BuilderEncoder.php
```diff
@@ -0,0 +1,181 @@
...
+    use EncodeIfSupported;
+
+    /** The first property is the operator value */
+    public const ENCODE_AS_SINGLE = 'single';
+
+    /** Arguments as encoded as a map of properties to their values */
+    public const ENCODE_AS_OBJECT = 'object';
+
+    /** Properties are encoded as a list of values, names are ignored */
+    public const ENCODE_AS_ARRAY = 'array';
```
jmikola, line 36: Would an enum make more sense here? If the feature is going to require PHP 8.1+, I don't think there'd be any syntax limitation.

src/Builder/BuilderEncoder.php
```diff
@@ -0,0 +1,181 @@
...
+    {
+        if (! $this->canEncode($value)) {
+            throw UnsupportedValueException::invalidEncodableValue($value);
+        }
+
+        // A pipeline is encoded as a list of stages
+        if ($value instanceof Pipeline) {
+            $encoded = [];
+            foreach ($value->getIterator() as $stage) {
+                $encoded[] = $this->encodeIfSupported($stage);
```
jmikola, line 59: Is `encodeIfSupported()` necessary here? I would think if the builder is well-typed, we could be more strict and assume that a pipeline would only contain Stage objects. If so, `encode()` could be used instead to effectively assert that we have well-formed input.

src/Builder/BuilderEncoder.php
```diff
@@ -0,0 +1,181 @@
...
+            $result = new stdClass();
+            // Specific: fields are encoded as a map of properties to their values at the top level as _id
+            foreach ($value->specifications as $key => $val) {
+                $result->{$key} = $this->encodeIfSupported($val);
+            }
+
+            return $this->wrap($value, $result);
+        }
+
+        if ($value instanceof OrQuery) {
```
jmikola, line 95: Is this file ultimately going to grow to contain encoding logic for each Expression and Operator type? If so, would it make sense to start structuring it differently to avoid a very lengthy method body? Perhaps a map of class names to callbacks that delegates based on a sequence of `instanceof` checks.

Alternatively, would a CodecLibrary make sense here?

I'm not sure if there's a concern of overhead, but as-is I could see this becoming quite unwieldy and ending up like a generated hydrator class in Doctrine (probably undesirably for a hand-written file).

src/Builder/BuilderEncoder.php
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    private function encodeAsObject(ExpressionInterface|StageInterface $value): stdClass
+    {
+        $result = new stdClass();
+        /** @var mixed $val */
+        foreach (get_object_vars($value) as $key => $val) {
+            /** @var mixed $val */
+            $val = $this->encodeIfSupported($val);
+            // @todo check for undefined value vs null
```
jmikola, line 153: Would "undefined" be considered a sentinel value? If so, I was recently talking about this in the phpc (not ours) Discord with @crell and he suggested using a single-value enum.

image

This lead me to an interesting wiki article: [Elephant in Cairo](https://en.wikipedia.org/wiki/Elephant_in_Cairo)

src/Builder/Expression.php
```diff
@@ -0,0 +1,96 @@
+<?php
+
+/**
+ * THIS FILE IS AUTO-GENERATED. ANY CHANGES WILL BE LOST!
```
jmikola, line 4: I'll note that GitHub didn't auto-collapse this file. I think it's missing from `.gitattributes`; however, it is excluded in `phpcs.xml.dist`, which tells me those aren't entirely in sync (per my earlier comment).

tests/Builder/BuilderCodecTest.php
```diff
@@ -0,0 +1,146 @@
...
+ * @requires PHP 8.0
+ */
+class BuilderCodecTest extends TestCase
+{
+    /** @see https://www.mongodb.com/docs/manual/reference/operator/aggregation/match/#equality-match */
+    public function testPipeline(): void
+    {
+        $pipeline = new Pipeline(
+            // @todo array is accepted by the stage class, but we expect an object. The driver accepts both.
+            Stage::match((object) ['author' => 'dave']),
```
jmikola, line 30: Is the object-cast necessary here? The `match()` factory method and Match constructor only say they accept `mixed $query`, so I don't see anything that enforces the argument type.

Does that mean the implementation for Match is incomplete, or is that intentional? I'm curious what would happen if you passed a scalar value here. Would it just encode to an invalid stage like `{ "$match": 1 }`?

generator/src/OperatorClassGenerator.php
```diff
@@ -0,0 +1,119 @@
...
+     * Operator classes interfaces are defined by their return type as a MongoDB expression.
+     */
+    private function getInterfaces(OperatorDefinition $definition): array
+    {
+        if ($definition->type === null) {
+            return [];
+        }
+
+        if ($definition->type === 'stage') {
+            return ['\\' . StageInterface::class];
```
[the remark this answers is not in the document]
GromNaN: When `\` is missing, the generated code doesn't contain the `use` statement, so the classes are loaded from current namespace.

---

The `expression.php` config is specific to PHP as it reference native types. It was easier to use PHP to get class names.

---

For the `architecture.md`, that's an idea. Things are far from definitive at the moment.

phpcs.xml.dist
GromNaN: It is.

src/Builder/Expression.php
GromNaN: Added

tests/Builder/BuilderCodecTest.php
GromNaN: It's a perverse effect to convert all associative arrays into object with `objectify` to improve readability of expected result. The value could be an associative array here.

tests/Builder/BuiderCodecTest.php
GromNaN: I made the change in this branch.

generator/src/Definition/ArgumentDefinition.php
```diff
@@ -3,26 +3,23 @@
...
         public bool $isOptional = false,
         public bool $isVariadic = false,
         public int $variadicMin = 1,
     ) {
         if (is_array($type)) {
             foreach ($type as $t) {
-                if (! is_string($t)) {
-                    throw new InvalidArgumentException('Argument type must be a string or list of strings');
-                }
+                assert(is_string($t));
```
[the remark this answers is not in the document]
GromNaN: `@psalm-assert` tells that we assert the type definition in the function body. So I added an `assert(array_is_list())`

generator/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Contributing
+
+Updating the generated code can be done only by modifying the code generator, or its configuration.
+
+To run the generator, you need to have PHP 8.2+ installed and Composer.
+
+1. Move to the `generator` directory: `cd generator`
+1. Install dependencies: `composer install`
+1. Run the generator: `bin/console generate`
```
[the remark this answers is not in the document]
GromNaN: I removed the `bin/` directory and renamed to `generate`

src/Builder/BuilderEncoder.php
GromNaN: I added the `Optional` enum, with `Undefined` case. I like that "optional" is self-documenting that a parameter is optional.

generator/config/config.php
jmikola: I think the current namespaces are fine, especially if users are unlikely to interface with these directly (assuming they use factory methods). We can always revisit later if needed.

generator/README.md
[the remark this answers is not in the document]
jmikola: OK, I'm fine with either approach.

generator/src/AbstractGenerator.php
```diff
@@ -0,0 +1,92 @@
+<?php
```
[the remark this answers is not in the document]
jmikola: Fine to keep as-is, just curious. Ultimately, this is all internal for us for I have no strong feelings on namespace grouping.

tests/Builder/BuilderCodecTest.php
jmikola: Are you saying that the `(object)` cast was just done to make the test assertion more readable (presumably in the event it fails)? I wasn't thinking of `objectify()` when I wrote the above comment and was more just generally curious about how `Stage::match()` reacts to an invalid value being passed in since it only uses `mixed`.

tests/Builder/BuiderCodecTest.php
jmikola: Discussed over video. So there's no need to demonstrate the following in docs:

```
acceptsVariadicNamed(...['foo.bar' => 'bar', 'bar' => 'baz'], baz: 'bla');
```

Users can utilize named arguments or utilize array unpacking as they wish, and we don't have to encourage them to mix both forms.

jmikola: Also confirmed that the argument is now a map of expressions so my earlier question about accepting `null` no longer applies.

src/Builder/Pipeline.php
jmikola: Will track in [PHPLIB-1264](https://jira.mongodb.org/browse/PHPLIB-1264).

src/Builder/BuilderEncoder.php
jmikola: We can come back to this in a future PR.

tests/Builder/BuilderCodecTest.php
jmikola: Confirmed this was a temporary issue, and the type system will be improved down the line.

jmikola (approved):

generator/config/expressions.php
```diff
@@ -0,0 +1,117 @@
...
+        'implements' => [$resolvesTo],
+        'types' => ['string'],
+    ];
+}
+
+return [
+    'null' => ['scalar' => true, 'types' => ['null']],
+    'int' => ['scalar' => true, 'types' => ['int', BSON\Int64::class]],
+    'number' => ['scalar' => true, 'types' => ['int', BSON\Int64::class]],
+    'decimal' => ['scalar' => true, 'types' => ['float', BSON\Decimal128::class]],
```
alcaeus, line 26: A couple of things here:
* `double` is missing from the list; it would correspond to PHP's `float` type
* `number` can be any of the numeric types (i.e. `int`, `Int64`, `float`, `Decimal128`)
* An `int` is a valid value to be given whenever a `decimal` type is accepted.

Given that, I think this should look as follows:
```suggestion
    'int' => ['scalar' => true, 'types' => ['int', BSON\Int64::class]],
    'double' => ['scalar' => true, 'types' => ['int', BSON\Int64::class, 'float']],
    'decimal' => ['scalar' => true, 'types' => ['int', BSON\Int64::class, 'float', BSON\Decimal128::class]],
    'number' => ['scalar' => true, 'types' => ['int', BSON\Int64::class, 'float', BSON\Decimal128::class]],
```

I'll note that a 64-bit integer value may be truncated when converted to a `double`, but I believe that shouldn't be an issue in this case.

generator/config/pipeline-operators.yaml
```diff
@@ -0,0 +1,92 @@
+- name: and
+  type: resolvesToBool
+  args:
+    - name: expressions
+      type: expression
+      isVariadic: true
+      variadicMin: 1
+- name: eq
```
alcaeus, line 8: Not an issue here, but we should consider splitting this file into multiple smaller files in the future.

generator/config/stages.yaml
```diff
@@ -0,0 +1,33 @@
...
+  type: stage
+  args:
+    - name: query
+      type: expression
+- name: sort
+  type: stage
+  args:
+    # @todo should generate be a map, not a list
+    - name: sortSpecification
+      type: number # should be an enum PHPLIB-1269
```
alcaeus, line 11: `number` is incorrect: the sort specification also supports [Text Score Metadata Sort](https://www.mongodb.com/docs/manual/reference/operator/aggregation/sort/#std-label-sort-pipeline-metadata)

GromNaN: That's handled by [PHPLIB-1269](https://jira.mongodb.org/browse/PHPLIB-1269).

generator/config/expressions.php
GromNaN: Fixed.

generator/config/stages.yaml
GromNaN: See [5bf4753](https://github.com/mongodb/mongo-php-library/pull/1174/commits/5bf47537028684b7fb5697de97ad32b82ca33229)

alcaeus (approved):

GromNaN:
Tests failing due to PHP version incompatibility. This will be fixed by [PHPLIB-1268](https://jira.mongodb.org/browse/PHPLIB-1268)

PR #1178, opened 2023-09-26
PHPLIB-476: Consider transaction readPreference in select_server

jmikola:
https://jira.mongodb.org/browse/PHPLIB-476

This also refactors the conditionals in extract_session_from_options and extract_read_preference_from_options to improve readability.

select_server() previously did not consider the read preference of an active transaction. This isn't very significant, as transactions require a primary read preference, but it is correct to do so.

### DIFF
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -260,7 +260,7 @@ class Collection
 
         $operation = new BulkWrite($this->databaseName, $this->collectionName, $operations, $options);
 
-        return $operation->execute(select_server($this->manager, $options));
+        return $operation->execute(select_server_for_write($this->manager, $options));
     }
 
     /**
@@ -362,7 +362,7 @@ class Collection
 
         $operation = new CreateIndexes($this->databaseName, $this->collectionName, $indexes, $options);
 
-        return $operation->execute(select_server($this->manager, $options));
+        return $operation->execute(select_server_for_write($this->manager, $options));
     }
 
     /**
@@ -418,7 +418,7 @@ class Collection
     public function createSearchIndexes(array $indexes, array $options = []): array
     {
         $operation = new CreateSearchIndexes($this->databaseName, $this->collectionName, $indexes, $options);
-        $server = select_server($this->manager, $options);
+        $server = select_server_for_write($this->manager, $options);
 
         return $operation->execute($server);
     }
@@ -441,7 +441,7 @@ class Collection
 
         $operation = new DeleteMany($this->databaseName, $this->collectionName, $filter, $options);
 
-        return $operation->execute(select_server($this->manager, $options));
+        return $operation->execute(select_server_for_write($this->manager, $options));
     }
 
     /**
@@ -462,7 +462,7 @@ class Collection
 
         $operation = new DeleteOne($this->databaseName, $this->collectionName, $filter, $options);
 
-        return $operation->execute(select_server($this->manager, $options));
+        return $operation->execute(select_server_for_write($this->manager, $options));
     }
 
     /**
@@ -503,7 +503,7 @@ class Collection
         $options = $this->inheritWriteOptions($options);
         $options = $this->inheritTypeMap($options);
 
-        $server = select_server($this->manager, $options);
+        $server = select_server_for_write($this->manager, $options);
 
         if (! isset($options['encryptedFields'])) {
             $options['encryptedFields'] = get_encrypted_fields_from_driver($this->databaseName, $this->collectionName, $this->manager)
@@ -541,7 +541,7 @@ class Collection
 
         $operation = new DropIndexes($this->databaseName, $this->collectionName, $indexName, $options);
 
-        return $operation->execute(select_server($this->manager, $options));
+        return $operation->execute(select_server_for_write($this->manager, $options));
     }
 
     /**
@@ -561,7 +561,7 @@ class Collection
 
         $operation = new DropIndexes($this->databaseName, $this->collectionName, '*', $options);
 
-        return $operation->execute(select_server($this->manager, $options));
+        return $operation->execute(select_server_for_write($this->manager, $options));
     }
 
     /**
@@ -577,7 +577,7 @@ class Collection
     public function dropSearchIndex(string $name, array $options = []): void
     {
         $operation = new DropSearchIndex($this->databaseName, $this->collectionName, $name);
-        $server = select_server($this->manager, $options);
+        $server = select_server_for_write($this->manager, $options);
 
         $operation->execute($server);
     }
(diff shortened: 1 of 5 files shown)


src/functions.php
```diff
@@ -558,33 +558,42 @@ function extract_session_from_options(array $options): ?Session
...
-        $readPreference = new ReadPreference(ReadPreference::PRIMARY);
+
+    // Read preference for an active transaction takes priority
+    if ($readPreference === null && $session instanceof Session && $session->isInTransaction()) {
+        /* Session::getTransactionOptions() should always return an array if the
+         * session is in a transaction, but we can be defensive. */
+        $readPreference = extract_read_preference_from_options($session->getTransactionOptions() ?? []);
     }
 
+    // Manager::selectServer() defaults to a primary read preference
```
jmikola, line 596: Since 1.11 ([PHPC-1937](https://jira.mongodb.org/browse/PHPC-1937)).

tests/Functions/SelectServerFunctionalTest.php
```diff
@@ -0,0 +1,51 @@
...
+
+        $session = $client->startSession();
+        $session->startTransaction();
+
+        $collection = $client->selectCollection($this->getDatabaseName(), $this->getCollectionName());
+        $collection->find([], ['session' => $session]);
+
+        $this->assertTrue($session->isInTransaction());
+        $this->assertInstanceOf(Server::class, $session->getServer(), 'Session is pinned');
+        $this->assertEquals($session->getServer(), select_server($client->getManager(), ['session' => $session]));
```
jmikola, line 41: Note: this test is still a bit dubious, as we don't have any insight into what `Manager::selectServer()` is _actually_ doing. That may change with server selection logging ([PHPLIB-1000](https://jira.mongodb.org/browse/PHPLIB-1000)).

Currently, there's a chance that server selection just happens to select the same server that was pinned (see: [Topology type: Sharded](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#topology-type-sharded) in the Server Selection spec). Supplying an alternative read preference in the operation-level options doesn't really affect this, since it wouldn't influence mongos selection. Again, logging would help verify that.

Having said all that, I think adding a functional test is a good first step since we previously had no direct test coverage for this function. This file at least provides a starting point to add more coverage down the line. I've left a comment in [PHPLIB-1000](https://jira.mongodb.org/browse/PHPLIB-1000) to that effect.

src/functions.php
```diff
@@ -558,33 +558,42 @@ function extract_session_from_options(array $options): ?Session
...
 function select_server(Manager $manager, array $options): Server
 {
     $session = extract_session_from_options($options);
-    $server = $session instanceof Session ? $session->getServer() : null;
-    if ($server !== null) {
-        return $server;
+
+    // Pinned server for an active transaction takes priority
+    if ($session instanceof Session && $session->getServer() instanceof Server) {
+        return $session->getServer();
```
alcaeus, line 583: Note that the previous logic of storing the result of `$session->getServer()` in its own variable was to work around the Psalm error. `$session->getServer()` may return a different result on subsequent invocations (as unlikely it is in a single-threaded context). Storing the result in a variable and checking that variable ensures that we're always returning a `Server` instance.

alcaeus (approved):
LGTM, suggested a change to work around the Psalm error.

src/functions.php
jmikola: Thanks. I obviously forgot to run `psalm` locally before publishing the PR :smile:

src/functions.php
```diff
@@ -635,3 +649,18 @@ function select_server_for_aggregate_write_stage(Manager $manager, array &$optio
...
+ * Performs server selection for a write operation.
+ *
+ * The pinned server for an active transaction takes priority, followed by an
+ * operation-level read preference, followed by a primary read preference. This
+ * is similar to select_server() except that it ignores a read preference from
+ * an active transaction's options.
+ *
+ * @internal
+ */
+function select_server_for_write(Manager $manager, array $options): Server
```
jmikola, line 663: Quoting Transactions: readPreference](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#readpreference):

> The read preference to use for all read operations in this transaction.

The `select_server_for_write()` method here allows us to avoid considering a transaction's `readPreference` option for a known write command. I intentionally chose _not_ to use this method for read methods that are typically run on primaries (e.g. collection and database enumeration). And it's not used for `Database::command()` sinec that's considered a read operation (see: [runCommand spec](https://github.com/mongodb/specifications/blob/master/source/run-command/run-command.rst#readpreference)).

It's clear we don't have comprehensive test coverage for this across _all_ operations (for example, there's no test that tries to list databases within a transaction), but I'm glad that I was alerted to this oversight in some failing [legacy transaction spec tests](https://spruce.mongodb.com/version/6515c8f8d1fe07b7c48b1f74/tasks).

GromNaN (approved):

jmikola:
Test failures seem to be due to an unrelated change in drivers-evergreen-tools, which I'm tracking in https://github.com/mongodb-labs/drivers-evergreen-tools/pull/352/files#r1342969157. I'll wait for that to get resolved before moving forward with this PR.

src/Collection.php
```diff
@@ -418,7 +418,7 @@ public function createSearchIndex($definition, array $options = []): string
     public function createSearchIndexes(array $indexes, array $options = []): array
     {
         $operation = new CreateSearchIndexes($this->databaseName, $this->collectionName, $indexes, $options);
-        $server = select_server($this->manager, $options);
+        $server = select_server_for_write($this->manager, $options);
```
GromNaN, line 421: I don't know it that's relevant for search indexes. 
> These commands internally proxy the search index management commands to a separate process that runs alongside an Atlas cluster. As such, read concern and write concern are not relevant for the search index management commands.
https://github.com/mongodb/specifications/blob/db3114e957f7c0976a1af09882dbb46cb4a70049/source/index-management/index-management.rst#where-are-read-concern-and-write-concern

jmikola: I'm familiar with `mongot`, but does that mean `mongod` has absolutely not state when it comes to search indexes? Perhaps that's the case, as `listSearchIndexes()` is distinct from `listIndexes()`.

This also raises the question of what happens if the search index operations are executed within a transaction. Presumably they don't maintain state on `mongod` at all (and thus operate outside of a transaction), but it's not clear to me if `mongod` would actually restrict their usage during an active transaction.

That said, I don't see the harm in requiring a primary in this case since it is a write operation from the user's perspective.

alcaeus (approved):

PR #1179, opened 2023-09-28
PHPLIB-1260: Revamp evergreen config

alcaeus:
PHPLIB-1260

Similar to https://github.com/mongodb/mongo-php-driver/pull/1473, this PR completely changes how we run evergreen builds.

Instead of running tests on various platforms, this only installs the extension on all supported platforms, which translates to testing `pecl install mongodb` once a stable version is out. We then run tests mostly on Debian 11 and Debian 9.2, although the Atlas and CSFLE tests run on RHEL 8.

When starting a patch build for pull requests, only replica set configurations are tested automatically, while other tests can be enabled in the patch as required.


alcaeus:
Note that the AWS without credentials already are already failing. I've created PHPLIB-1263 to investigate this separately.

composer.json
```diff
@@ -13,7 +13,7 @@
         "php": "^7.4 || ^8.0",
         "ext-hash": "*",
         "ext-json": "*",
-        "ext-mongodb": "^1.17.0",
+        "ext-mongodb": "^1.16.0",
```
GromNaN, line 16: This was updated on-purpose by https://github.com/mongodb/mongo-php-library/commit/1acfab001810504214da8617e1500168086f7074

alcaeus: The commit was there for initial testing and has already been removed locally

.evergreen/config.yml
```diff
@@ -1,1031 +1,82 @@
...
-      params:
-        file: aws-expansion.yml
+# Protect ourselves against rogue test case that runs forever. Tasks are killed after 10 minutes, which shouldn't occur
+# under normal circumstances.
+exec_timeout_secs: 600
 
-pre:
+# These pre and post rules apply to all tasks not part of a task group, which should only ever be tests against local
+# MongoDB instances. All other tasks that require special preparation should be run from within a task group
+pre: &test_setup
```
jmikola, line 18: Are `test_setup` and `test_teardown` referenced anywhere?

.evergreen/compile-extension.sh
```diff
@@ -0,0 +1,42 @@
...
+      git submodule update --init
+      phpize
+      ./configure --enable-mongodb-developer-flags
+      make all -j20 > /dev/null
+      make install
+
+      cd ${PROJECT_DIRECTORY}
+   elif [ "${EXTENSION_VERSION}" != "" ]; then
+      echo "Installing driver version ${EXTENSION_VERSION} from PECL"
+      pecl install -f mongodb-${EXTENSION_VERSION}
```
jmikola, line 31: Prefixing this with `MAKEFLAGS=-j20` should speed things up (assuming you want to use the same parallelism as the manual build above).

Likewise for `pecl install` a few lines below.

.evergreen/config/build-variants.yml
```diff
@@ -0,0 +1,87 @@
+#
+# Build variants to build the driver - these are run for all operating systems and PHP versions we support
```
jmikola, line 2: Were all of these variants copied from PHPC despite the fact that we don't actually test on them for PHPLIB? Does it make sense to do so or would you rather strip this down to the minimum required?

.evergreen/config/functions.yml
```diff
@@ -0,0 +1,569 @@
...
+          rm -rf $DRIVERS_TOOLS
+          if [ "${project}" = "drivers-tools" ]; then
+            # If this was a patch build, doing a fresh clone would not actually test the patch
+            cp -R ${PROJECT_DIRECTORY}/ $DRIVERS_TOOLS
+          else
+            git clone https://github.com/mongodb-labs/drivers-evergreen-tools.git --depth 1 $DRIVERS_TOOLS
+          fi
+          echo "{ \"releases\": { \"default\": \"$MONGODB_BINARIES\" }}" > $MONGO_ORCHESTRATION_HOME/orchestration.config
+
+  "upload mo artifacts":
```
jmikola, line 129: IIRC you removed this from PHPC, along with a few other functions we had no use of. `upload working dir` might have been another.

.evergreen/config/functions.yml
```diff
@@ -0,0 +1,569 @@
...
+  "fix absolute paths":
+    - command: shell.exec
+      params:
+        script: |
+          ${PREPARE_SHELL}
+          for filename in $(find ${DRIVERS_TOOLS} -name \*.json); do
+            perl -p -i -e "s|ABSOLUTE_PATH_REPLACEMENT_TOKEN|${DRIVERS_TOOLS}|g" $filename
+          done
+
+  "windows fix":
```
jmikola, line 403: Irrelevant?

.evergreen/config/functions.yml
```diff
@@ -0,0 +1,569 @@
...
+        script: |
+          ${PREPARE_SHELL}
+          for i in $(find ${DRIVERS_TOOLS}/.evergreen ${PROJECT_DIRECTORY}/.evergreen -name \*.sh); do
+            cat $i | tr -d '\r' > $i.new
+            mv $i.new $i
+          done
+          # Copy client certificate because symlinks do not work on Windows.
+          cp ${DRIVERS_TOOLS}/.evergreen/x509gen/client.pem ${MONGO_ORCHESTRATION_HOME}/lib/client.pem
+
+  "make files executable":
```
jmikola, line 415: I think this was also removed from PHPC with the reasoning that we should fix this in `drivers-evergreen-tools` if needed.

.evergreen/config/functions.yml
```diff
@@ -0,0 +1,569 @@
...
+  "install composer":
+    - command: shell.exec
+      params:
+        add_expansions_to_env: true
+        working_dir: "src"
+        script: |
+          ${PREPARE_SHELL}
+          file="${PROJECT_DIRECTORY}/.evergreen/install-composer.sh"
+          # Don't use ${file} syntax here because evergreen treats it as an empty expansion.
+          [ -f "$file" ] && PHP_VERSION=${PHP_VERSION} EXTENSION_VERSION=${EXTENSION_VERSION} EXTENSION_REPO=${EXTENSION_REPO} EXTENSION_BRANCH=${EXTENSION_BRANCH} DEPENDENCIES=${DEPENDENCIES} sh $file || echo "$file not available, skipping"
```
jmikola, line 443: `DEPENDENCIES` is the only variable referenced in `install-composer.sh`. Do you want to strip this down?

.evergreen/config/functions.yml
```diff
@@ -0,0 +1,569 @@
...
+
+  "install dependencies":
+    - command: shell.exec
+      params:
+        working_dir: "src"
+        script: |
+          ${PREPARE_SHELL}
+          file="${PROJECT_DIRECTORY}/.evergreen/install-dependencies.sh"
+          # Don't use ${file} syntax here because evergreen treats it as an empty expansion.
+          [ -f "$file" ] && PHP_VERSION=${PHP_VERSION} EXTENSION_VERSION=${EXTENSION_VERSION} EXTENSION_REPO=${EXTENSION_REPO} EXTENSION_BRANCH=${EXTENSION_BRANCH} DEPENDENCIES=${DEPENDENCIES} sh $file || echo "$file not available, skipping"
```
jmikola, line 432: `install-dependencies.sh` only references `$DRIVERS_TOOLS`, so I don't think any of these variables are necessary.

.evergreen/config/templates/test/load-balanced.yml
```diff
@@ -0,0 +1,17 @@
...
+          TOPOLOGY: "sharded_cluster"
+          MONGODB_VERSION: "%mongodbVersion%"
+          LOAD_BALANCER: "true"
+          SSL: "yes"
+      - func: "start load balancer"
+      - func: "start kms servers"
+      - func: "set aws temp creds"
+      - func: "run tests"
+        vars:
+          MONGODB_URI: "${SINGLE_MONGOS_LB_URI}"
```
jmikola, line 16: Do you care about preserving the original comment?

```
# Note: loadBalanced=true should already be appended to SINGLE_MONGOS_LB_URI
```

I'm indifferent, just asking if it was intentionally removed.

.evergreen/config/test-task-groups.yml
```diff
@@ -0,0 +1,37 @@
...
+      - func: "install composer"
+      - command: subprocess.exec
+        params:
+          working_dir: src
+          binary: bash
+          add_expansions_to_env: true
+          env:
+            MONGODB_VERSION: '7.0'
+          args:
+            - ${DRIVERS_TOOLS}/.evergreen/atlas/setup-atlas-cluster.sh
```
jmikola, line 20: I noted this defaults `MONGODB_VERSION` to 6.0. Presumably 7.0 is required because this task group is used for search index testing (see: [Setup](https://github.com/mongodb/specifications/tree/master/source/index-management/tests#setup)).

.evergreen/config/test-variants.yml
```diff
@@ -0,0 +1,116 @@
...
+      FETCH_BUILD_VARIANT: "build-debian11"
+      FETCH_BUILD_TASK: "build-php-8.2"
+      PHP_VERSION: "8.2"
+      VARIANT: debian11 # Referenced by ADL build script for downloading MQLRun
+    depends_on:
+      - variant: "build-debian11"
+        name: "build-php-8.2"
+    tasks:
+      - ".standalone .local !.3.6 !.4.0 !.4.2 !.4.4 !.5.0"
+      - ".replicaset .local !.csfle !.3.6 !.4.0 !.4.2 !.4.4 !.5.0"
```
jmikola, line 20: Should all of these lines include `!.csfle`? As-is, this assumes that `templates/test/csfle.yml` will only apply to replica set topologies, which could change in the future.

.evergreen/config/functions.yml
```diff
@@ -0,0 +1,569 @@
...
+        # TODO: Use separate folder for the library once it exists
+#        remote_file: ${project}/${build_variant}/${revision}/${task_name}/${version}.tar.gz
+
+  "fetch extension":
+    - command: s3.get
+      params:
+        aws_key: ${aws_key}
+        aws_secret: ${aws_secret}
+        bucket: mciuploads
+        # TODO: Use separate folder for the library once it exists
```
jmikola, line 97: Is this TODO item tracked anywhere?

.evergreen/install-composer.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
+set -o errexit  # Exit the script with error if any of the commands fail
```
jmikola, line 2:
```suggestion
set -o errexit  # Exit the script with error if any of the commands fail

# Supported environment variables
DEPENDENCIES=${DEPENDENCIES:-} # Specify "lowest" to prefer lowest dependencies
COMPOSER_FLAGS="${COMPOSER_FLAGS:-}" # Optional, additional Composer flags
```

I suppose you could omit `COMPOSER_FLAGS` since it's not used anywhere and really just set below when we process `$DEPENDENCIES`.

I know `$PHP_PATH` and `$PROJECT_DIRECTORY` are also referenced below, but `$PROJECT_DIRECTORY` is always just assumed. I'm not sure where `$PHP_PATH` comes from as that was originally exported by `install-dependencies.sh` but now handled by the `locate PHP binaries` function.

tests/TestCase.php
```diff
@@ -70,7 +70,7 @@ public static function printConfiguration(): void
         $configuration = [
             '%uri%' => $uri,
             '%database%' => static::getDatabaseName(),
-            '%apiVersion%' => getenv('MONGODB_API_VERSION') ?: 'Not configured',
+            '%apiVersion%' => getenv('API_VERSION') ?: 'Not configured',
```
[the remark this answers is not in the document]
alcaeus: That's correct.

.evergreen/config/build-variants.yml
alcaeus: Yes - these effectively allows us to test that running `pecl install` works as expected. Tests are only run on a smaller set of variants.

.evergreen/config.yml
alcaeus: They aren't; removed.

.evergreen/compile-extension.sh
alcaeus: Ah, good point, I didn't know about `MAKEFLAGS`. Added 👍

.evergreen/config/functions.yml
alcaeus: Cleaned up along with the other irrelevant functions (windows fix, make files executable, etc.)

.evergreen/config/functions.yml
alcaeus: Cleaned up along with the other variables.

.evergreen/config/templates/test/load-balanced.yml
alcaeus: It was removed on purpose as it didn't add any value IMO.

.evergreen/config/test-variants.yml
alcaeus: Excluded this. As unlikely as changes are, this prevents surprises down the line.

.evergreen/install-composer.sh
alcaeus: `$PHP_PATH` is set through the `add_expansions_to_env` setting in the `install composer` function. `locate PHP binaries` is responsible for finding the path to PHP binaries and creates the evergreen expansion.

PR #1180, opened 2023-10-04
PHPLIB-1254 Create Yaml for all operators and stages

GromNaN:
Fix PHPLIB-1254

## Operators and stages specifications

To create Yaml for all the stages and operators (query, pipeline), I started by filling a  [Google Sheet](https://docs.google.com/spreadsheets/d/e/2PACX-1vROpGTJGXAKf2SVuSZaw16NwMVtzMVGH9b-YiMtddgZRZOjOO6jK2YLbTUZ0N_qe74nxGY9hYhUe-l2/pubhtml) manually, from [online documentation](https://www.mongodb.com/docs/manual/reference/operator/aggregation-pipeline/). Even if that took hours, it was a good learning. This document is temporary and will not be used after this PR. 

The Yaml files are generated with the [`scrape` command](https://github.com/mongodb/mongo-php-library/blob/27c9f21b67b1e8db7b9dcc00b38384556f25750b/generator/src/Command/ScrapeCommand.php), that will be removed before merging.

The `schema.json` applies to Yaml files.

All operators and stages have an "encoding" type, that describes how to convert the object into BSON.

https://github.com/mongodb/mongo-php-library/blob/27c9f21b67b1e8db7b9dcc00b38384556f25750b/generator/config/schema.json#L36-L40

## Types

The types will be reworked in [PHPLIB-1251](https://jira.mongodb.org/browse/PHPLIB-1251).

- Add `AccumulatorInterface` to type hint [Group Accumulator Operators](https://www.mongodb.com/docs/v3.4/reference/operator/aggregation-group/). It might be necessary to distinct the accumulators for each stage (`$group`, `$setWindowFields`...)
- Add `QueryInterface` distinct query operators from expressions marked with `ExpressionInterface`.

Add the `MongoDB\object` function to create `stdClass` from variadic named arguments.
https://github.com/mongodb/mongo-php-library/blob/27c9f21b67b1e8db7b9dcc00b38384556f25750b/src/functions.php#L62-L71


generator/src/Command/ScrapeCommand.php
```diff
@@ -0,0 +1,235 @@
...
+        $dirname = $this->configDir . '/' . $dirname;
+        if (! file_exists($dirname)) {
+            mkdir($dirname, 0755);
+        }
+
+        $name = str_replace('$', '', $data['name']) ?: 'positional';
+        $filename = $dirname . '/' . $name . '.yaml';
+
+        // Add a schema reference to the top of the file
+        $schema = '# $schema: ../schema.json' . PHP_EOL;
```
GromNaN, line 226: PHPStorm will [support](https://youtrack.jetbrains.com/issue/IDEA-323117) it soon.

> Will be available in 2023.3 Release.

generator/src/Command/ScrapeCommand.php
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    public function configure(): void
+    {
+        $this->setName('scrape');
+    }
+
+    public function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $index = file_get_contents('https://docs.google.com/spreadsheets/d/e/2PACX-1vROpGTJGXAKf2SVuSZaw16NwMVtzMVGH9b-YiMtddgZRZOjOO6jK2YLbTUZ0N_qe74nxGY9hYhUe-l2/pubhtml');
```
GromNaN, line 55: I published the Google Sheet publicly as an HTML document because csv and tsv doesn't have correct support for newline. Also, I can edit in Google and run my command without all the hassle of Google authentication.

GromNaN (commented):
Very WIP

generator/src/Command/ScrapeCommand.php
```diff
@@ -0,0 +1,246 @@
...
+use function str_contains;
+use function str_ends_with;
+use function str_replace;
+use function strip_tags;
+use function trim;
+use function var_export;
+
+use const PHP_EOL;
+
+final class ScrapeCommand extends Command
```
GromNaN, line 35: I'll probably remove this command before merging. The google sheet is a step for generating hundreds of Yaml file. But it is not suitable for day-to-day maintenance and versioning.

generator/src/Definition/OperatorDefinition.php
```diff
@@ -3,29 +3,48 @@
...
+        public ?string $description = null,
+        array $arguments = [],
     ) {
-        assert($encode || count($args) === 1, sprintf('Operator "%s" has %d arguments. The "encode" parameter must be specified.', $name, count($args)));
-        assert(in_array($encode, [null, 'array', 'object'], true), sprintf('Operator "%s" expect "encode" value to be "array" or "object". Got "%s".', $name, $encode));
+        $this->encode = match ($encode) {
+            'single' => Encode::Single,
+            'array' => Encode::Array,
+            'object' => Encode::Object,
+            'empty object' => Encode::Object,
```
GromNaN, line 31: To be removed. An empty object is a regular object without property.

generator/config/aggregation-stages/collStats.yaml
```diff
@@ -0,0 +1,15 @@
...
+type:
+    - Stage
+encode: single
+description: |
+    Returns statistics regarding a collection or view.
+arguments:
+    -
+        name: config
+        type:
+            - object
```
GromNaN, line 15: I did not modelize the full object. Let's use `object` function for that. People will have to read the documentation anyway.
I could have made it a variadic map.

generator/config/aggregation-stages/densify.yaml
```diff
@@ -0,0 +1,32 @@
...
+        name: partitionByFields
+        type:
+            - array<string>
+        optional: true
+        description: |
+            The field(s) that will be used as the partition keys.
+    -
+        name: range
+        type:
+            - Range
```
GromNaN, line 30: `range` model is a plain object for now. Not modelized.

generator/config/aggregation-operators/firstN.yaml
```diff
@@ -0,0 +1,23 @@
+# $schema: ../schema.json
+name: $firstN
```
GromNaN, line 2: I have a name conflict with `$firstN` that is both [aggregation accumulator](https://www.mongodb.com/docs/manual/reference/operator/aggregation/firstN/) and [array operator](https://www.mongodb.com/docs/manual/reference/operator/aggregation/firstN-array-element/). Same for `$lastN`.

.github/workflows/benchmark.yml
```diff
@@ -5,7 +5,7 @@ on:
     branches:
       - "v*.*"
       - "master"
-      - "feature/*"
+      #- "feature/*"
```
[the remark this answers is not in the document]
jmikola: > I'm also fine to not run benchmarks for pull requests into feature branches, in which case this can be removed.

Yes, please.

generator/src/Definition/OperatorDefinition.php
```diff
@@ -3,29 +3,55 @@
...
+        $this->encode = match ($encode) {
+            'single' => Encode::Single,
+            'array' => Encode::Array,
+            'object' => Encode::Object,
+            'group' => Encode::Group,
+            default => throw new UnexpectedValueException(sprintf('Unexpected "encode" value for operator "%s". Got "%s"', $name, $encode)),
+        };
+
+        // Convert arguments to ArgumentDefinition objects
+        // Optional arguments must be after required arguments
```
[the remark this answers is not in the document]
jmikola: > $slice have an optional argument in the middle or required arguments.



This sadly dates back to server version 3.2. I hope we've learned not to make such mistakes in the future.

generator/config/aggregation-stages/densify.yaml
jmikola: For cases where we are choosing not to model the object, can we incorporate links in the YAML so that we get `@see` references in the generated code? Then, if folks happen to view the generated operator file they can be directed to the server documentation.

generator/src/Definition/OperatorDefinition.php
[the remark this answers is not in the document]
GromNaN: Can we ask the server team to allow an object here?

generator/config/aggregation-stages/densify.yaml
GromNaN: I added a `@see` comment with the link to the doc on each operator class header.

generator/config/aggregation-operators/firstN.yaml
jmikola: Both of these fall under aggregation, correct? And not the query namespace.

The API for both of these operators looks similar, but I assume they are distinct and there's nothing to say other operators and accumulators with the same name might differ more significantly. In this case, would it make sense to organize accumulators under their own namespace?

I actually notice that there are two groupings for accumulators in the docs:

 * [Accumulators ($group, $bucket, $bucketAuto, $setWindowFields)](https://www.mongodb.com/docs/manual/reference/operator/aggregation/#accumulators---group---bucket---bucketauto---setwindowfields-)
 * [Accumulators (in Other Stages)(https://www.mongodb.com/docs/manual/reference/operator/aggregation/#accumulators--in-other-stages-)

It's hard to predict how we might encounter naming conflicts in the future, but do you think it'd be worth organizing the many aggregation operators by separate namespaces to group them logically?

We could also do the same for [Query and Projection](https://www.mongodb.com/docs/manual/reference/operator/query/#query-and-projection-operators) operators, which have far fewer categories, if you like.

generator/src/Command/ScrapeCommand.php
GromNaN: It's all done by hand. Unless you use AI, I don't see how you can recreate this file simply to detect changes.

The form in which things are documented varies greatly and some informations where extracted from the server source code. I'd better maintain Yaml files and get notified of changes.

generator/config/aggregation-operators/firstN.yaml
GromNaN: If we over-categorize, it could be confusing for developers, who will have to guess which category we've put the operator they want to use. 
Also, I don't want to duplicate Yaml files for operator what would be in multiple categories, like `$avg` or `$count`.

generator/src/Definition/OperatorDefinition.php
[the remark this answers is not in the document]
jmikola: They certainly can't change the array order due to BC reasons.

Argument lists for operators are typically lists, so I don't imagine using an object would be consistent. Even if such a change was made, we'd still have to support the legacy array format for older server versions, which in turns means we'd need to implement version checking here. AFAIK, all of the query syntax is modeled irrespective of the server version, so I'd rather not open that can of worms.

I expect `$slice` will just need to be handled specially.

generator/config/aggregation-operators/firstN.yaml
jmikola: Is this file modeling the array operator or accumulator? The `link` and `description` fields below suggest the array operator , but I see `type: [ Accumulator ]` as well.

jmikola: > Is this file modeling the array operator or accumulator? The `link` and `description` fields below suggest the array operator , but I see `type: [ Accumulator ]` as well.

I can see how over-categorization could be problematic (assuming users need to reference class names and don't use factory methods). What is your proposed solution?

src/Builder/BuilderEncoder.php
```diff
@@ -63,110 +66,118 @@ public function encode($value): stdClass|array|string
...
+            return $value;
+        }
+
+        if ($value instanceof stdClass) {
+            foreach (get_object_vars($value) as $key => $val) {
+                $value->{$key} = $this->recursiveEncode($val);
+            }
+        }
+
+        return $this->encodeIfSupported($value);
```
GromNaN, line 177: This function is all about the `ArrayCodec` and `ObjectCodec` that were described in the codec architecture doc. 
https://github.com/mongodb/mongo-php-library/pull/1125/files#diff-8bbcff6724826a3ca2b160e0718dbee5520987bbaa22e78eda96d0d448cad5e8R27

src/Builder/Stage/UnsetStage.php
```diff
@@ -0,0 +1,41 @@
...
+    /** @param list<FieldPath|non-empty-string> ...$field */
+    public array $field;
+
+    /**
+     * @param FieldPath|non-empty-string ...$field
+     * @no-named-arguments
+     */
+    public function __construct(FieldPath|string ...$field)
+    {
+        if (\count($field) < 1) {
```
jmikola, line 33: Is it expected that generated files reference global functions this way? It's inconsistent with what we do in the hand-written sources, so if it's easily configurable it may be worth changing.

src/Builder/Type/QueryInterface.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+/**
+ * THIS FILE IS AUTO-GENERATED. ANY CHANGES WILL BE LOST!
```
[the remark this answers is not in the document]
GromNaN: Oups, I removed the header from this interfaces. It was created by duplicating a generated file.

src/Builder/Stage/UnsetStage.php
GromNaN: Added `$namespace->addUseFunction` calls to add this. The code generator don't know about the function body, so it cannot be detected.

tests/Builder/BuilderEncoderTest.php
```diff
@@ -57,21 +58,24 @@ public function testSort(): void
...
+            ),
         );
 
         $expected = [
             [
                 '$match' => [
                     '$or' => [
-                        ['score' => ['$gt' => 70, '$lt' => 90]],
+                        // same as ['score' => ['$gt' => 70, '$lt' => 90]],
+                        ['score' => [['$gt' => 70], ['$lt' => 90]]],
```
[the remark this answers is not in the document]
GromNaN: It was too easy... to be correct. I reworked the implementation to generate the expected result.

examples/aggregation-builder.php
```diff
@@ -48,15 +50,15 @@ function toJSON(object $document): string
         totalCount: Aggregation::sum(1),
         evenCount: Aggregation::sum(
             Aggregation::mod(
-                Expression::fieldPath('randomValue'),
+                Expression::numberFieldPath('randomValue'),
```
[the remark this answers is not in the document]
GromNaN: That's a request from @alcaeus in the Technical Design doc.

> To complete the type experience, there are also typed fieldPath helpers (e.g. stringFieldPath) which communicate to the type system that the field path resolves to the given type. This can again be used by an ORM to verify that the field will actually resolve to the given type (if known).

PR #1184, opened 2023-10-18
Increase task timeouts

alcaeus:
Follow-Up to PHPLIB-1260.

The default `exec_timeout_secs` of 10 minutes is too short, as sharded tests run significantly longer. 20 minutes seem more appropriate before killing tasks. In addition to that, the Atlas and Serverless tests already run close to 20 minutes, so those tasks get a 30 minute timeout.

### DIFF
--- a/.evergreen/config.yml
+++ b/.evergreen/config.yml
@@ -11,7 +11,7 @@ pre_error_fails_task: true
 
 # Protect ourselves against rogue test case that runs forever. Tasks are killed after 10 minutes, which shouldn't occur
 # under normal circumstances.
-exec_timeout_secs: 600
+exec_timeout_secs: 1200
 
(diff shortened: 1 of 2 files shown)


.evergreen/config.yml
```diff
@@ -11,7 +11,7 @@ pre_error_fails_task: true
 
 # Protect ourselves against rogue test case that runs forever. Tasks are killed after 10 minutes, which shouldn't occur
```
GromNaN, line 12:
```suggestion
# Protect ourselves against rogue test case that runs forever. Tasks are killed after 20 minutes, which shouldn't occur
```

GromNaN (changes requested):

GromNaN (approved):

PR #1185, opened 2023-10-20
PHPLIB-1187: Run benchmark on Evergreen

alcaeus:
PHPLIB-1187

This pull request adds a task to run benchmarks on Evergreen. It also uses `perf.send` to store benchmark metrics in CI so we can analyse the performance.

I've also removed a couple of benchmarks in order to bring the benchmark time down. Most recent tests show a 33 minute runtime on Evergreen. I also had to skip the AMP subjects as I ran into issues with socket creation. Since AMP does not perform significantly different from forking, I think it's safe to exclude them from the metrics we collect on Evergreen.

### DIFF
--- a/.evergreen/config/php.ini
+++ b/.evergreen/config/php.ini
@@ -1,1 +1,2 @@
 extension=mongodb.so
+memory_limit=-1
--- a/benchmark/phpbench.json.dist
+++ b/benchmark/phpbench.json.dist
@@ -6,6 +6,5 @@
     "runner.file_pattern": "*Bench.php",
     "runner.path": "src",
     "runner.php_config": { "memory_limit": "1G" },
-    "runner.iterations": 3,
-    "runner.revs": 10
+    "runner.iterations": 3
 }
--- a/benchmark/src/DriverBench/ParallelMultiFileExportBench.php
+++ b/benchmark/src/DriverBench/ParallelMultiFileExportBench.php
@@ -72,15 +72,15 @@ final class ParallelMultiFileExportBench
      * Using a single thread to export multiple files.
      * By executing a single Find command for multiple files, we can reduce the number of roundtrips to the server.
      *
-     * @param array{chunk:int} $params
+     * @param array{chunkSize:int} $params
      */
     #[ParamProviders(['provideChunkParams'])]
     public function benchSequential(array $params): void
     {
-        foreach (array_chunk(self::getFileNames(), $params['chunk']) as $i => $files) {
+        foreach (array_chunk(self::getFileNames(), $params['chunkSize']) as $i => $files) {
             self::exportFile($files, [], [
-                'limit' => 5_000 * $params['chunk'],
-                'skip' => 5_000 * $params['chunk'] * $i,
+                'limit' => 5_000 * $params['chunkSize'],
+                'skip' => 5_000 * $params['chunkSize'] * $i,
             ]);
         }
     }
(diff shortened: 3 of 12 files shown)


benchmark/phpbench.json.dist
```diff
@@ -6,6 +6,5 @@
     "runner.file_pattern": "*Bench.php",
     "runner.path": "src",
     "runner.php_config": { "memory_limit": "1G" },
-    "runner.iterations": 3,
-    "runner.revs": 10
```
alcaeus, line 10: Removed this in favour of increasing revs only for benchmarks that run in the microsecond range. For anything that takes milliseconds, the precision is usually good enough with a single rev.

benchmark/src/DriverBench/ParallelMultiFileExportBench.php
```diff
@@ -74,15 +72,15 @@ public function afterIteration(): void
      * Using a single thread to export multiple files.
      * By executing a single Find command for multiple files, we can reduce the number of roundtrips to the server.
      *
-     * @param array{chunk:int} $params
+     * @param array{chunkSize:int} $params
```
alcaeus, line 75: Decided to rename this since I got confused as to what `chunk` meant: I thought `chunk: 1` meant a single chunk of files while it actually meant 100 chunks of 1 file each.

benchmark/src/DriverBench/ParallelMultiFileImportBench.php
```diff
@@ -73,20 +71,6 @@ public function beforeIteration(): void
...
     }
 
-    /**
-     * Using Driver's BulkWrite in a single thread.
-     * The number of files to import in each iteration is controlled by the "chunk" parameter.
-     *
-     * @param array{chunk:int} $params
-     */
-    #[ParamProviders(['provideChunkParams'])]
-    public function benchBulkWrite(array $params): void
```
alcaeus, line 83: This subject essentially tests the same as `benchInsertMany`, so I decided to skip it in order to shave ~6 minutes off the run time.

benchmark/src/Extension/EvergreenReport.php
```diff
@@ -0,0 +1,115 @@
...
+                        if ($variant->getParameterSet()->getName()) {
+                            $name .= '#' . $variant->getParameterSet()->getName();
+                        }
+
+                        $tests[] = [
+                            'info' => [
+                                'test_name' => $name,
+                            ],
+                            'created_at' => date(DATE_ATOM),
+                            'completed_at' => date(DATE_ATOM),
```
alcaeus, line 64: I don't think this information is relevant to us, so I didn't bother trying to extract the exact dates. The documentation also didn't mark this as optional, so I decided to just include the current date instead.

benchmark/src/DriverBench/ParallelMultiFileImportBench.php
GromNaN: We could keep this and remove the other that is a little less efficient

alcaeus: The reason I removed `bulkWrite` and kept `insertMany` is that benchmarking the latter also exposes performance regressions introduced in the library, while the other only tests the extension.

.evergreen/config/php.ini
```diff
@@ -1 +1,2 @@
 extension=mongodb.so
+memory_limit=-1
```
GromNaN, line 2: You should increase the limit. Otherwise we don't know how much memory is used if it's too much. What is the memory limit of the job runner?

benchmark/src/Extension/EvergreenReport.php
GromNaN: Well, `created_at` could take `$_SERVER['REQUEST_TIME']`.

GromNaN (approved):
LGTM.

.evergreen/config/php.ini
alcaeus: For some reason it aborted when it hit a previous 128M limit, apparently ignoring the 1G limit definer in the runner config. We can also report memory usage from the benchmarks using `perf.send`, which would be a better indicator than failing the build when it hits some limit. Want me to add those numbers?

GromNaN: `perf.send` is too late if the process crashes, no?

alcaeus: Correct, `perf.send` would not be executed in that case. With `-1` the memory would be unlimited, so the only case in which it would crash is if it used up all memory including the page file, which I'd consider highly unlikely.

PR #1186, opened 2023-10-21
Upgrade `rector/rector`

OskarStark:


src/Operation/Watch.php
```diff
@@ -391,15 +391,7 @@ private function getInitialResumeToken()
...
-        if (isset($this->changeStreamOptions['startAfter'])) {
-            return $this->changeStreamOptions['startAfter'];
-        }
-
-        if (isset($this->changeStreamOptions['resumeAfter'])) {
-            return $this->changeStreamOptions['resumeAfter'];
-        }
-
-        return null;
+        return $this->changeStreamOptions['startAfter'] ?? $this->changeStreamOptions['resumeAfter'] ?? null;
```
OskarStark, line 394: Change created by `IfIssetToCoalescingRector`, can be deactivated if desired.

GromNaN: We try not to use ternaries like this too much, because they make it harder to identify code paths not covered by the tests.

GromNaN (changes requested):
Super. I'm glad to see that the bugs we were encountering have been fixed by rector.

Can you revert the `IfIssetToCoalescingRector` please.

src/Operation/Watch.php
OskarStark: Done

OskarStark:
Rebased ✅

GromNaN (approved):

rector.php
```diff
@@ -20,18 +18,8 @@
...
-            __DIR__ . '/src/Operation/',
-        ],
-        // Assigns wrong type due to outdated PHPStan stubs
-        TypedPropertyFromAssignsRector::class => [
-            __DIR__ . '/src/Model/BSONIterator.php',
-        ],
+        // Do not use ternaries extensively
+        IfIssetToCoalescingRector::class,
         // Not necessary in documentation examples
         JsonThrowOnErrorRector::class => [
```
GromNaN, line 24: False CS error: "Key specified for array entry; first entry has no key".
We can disable this rule for this file.

jmikola: Working on this now.

OskarStark: Done ✅

PR #1189, opened 2023-11-07
PHPLIB-1143: Document Atlas Search index helpers and fix aggregate "collation" docs

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1143
https://jira.mongodb.org/browse/PHPLIB-1306

PR #1190, opened 2023-11-08
PHPLIB-1305 Support "codec" option in listSearchIndexes()

GromNaN:
Fix PHPLIB-1305

It's definitely not a place where I would expect a codec to be used, but why not.

I don't think we need to test all the options.

### DIFF
--- a/tests/Operation/ListSearchIndexesTest.php
+++ b/tests/Operation/ListSearchIndexesTest.php
@@ -28,6 +28,8 @@ class ListSearchIndexesTest extends TestCase
             $options[][] = ['batchSize' => $value];
         }
 
+        $options[][] = ['codec' => 'foo'];
+
         return $options;
     }
 }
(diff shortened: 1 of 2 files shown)


jmikola (approved):
This issue has no fixVersion, but I have no qualms about throwing it into `master` before we branch v1.17.

In the meantime, I'll add it into https://github.com/mongodb/mongo-php-library/pull/1189.

GromNaN:
> This issue has no fixVersion, but I have no qualms about throwing it into `master` before we branch v1.17.

The codec is a new feature for v1.17. I don't see what alternative there would be to merging it into master before the 1.17.0 release.

jmikola:
> I don't see what alternative there would be to merging it into master before the 1.17.0 release.

I wasn't sure if this was something we intended to fix in 1.17 or defer to 1.18. If 1.18, then it would only be merged to master after branching v1.17.

alcaeus:
I'm fine tossing it into 1.17 - however there should be at least a single test. For all I care, it could be a test passing an invalid codec, expecting an `InvalidArgumentException` to ensure that the `codec` option is handled somewhere.

tests/Operation/ListSearchIndexesTest.php
```diff
@@ -28,6 +28,8 @@ public function provideInvalidConstructorOptions(): array
             $options[][] = ['batchSize' => $value];
         }
 
+        $options[][] = ['codec' => 'foo'];
```
GromNaN, line 31: @alcaeus test added.

alcaeus (approved):

PR #1191, opened 2023-11-09
PHPLIB-1243: Document add/remove logger functions

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1243

### DIFF
--- a/docs/includes/apiargs-function-add_logger-param.yaml
+++ b/docs/includes/apiargs-function-add_logger-param.yaml
@@ -1,0 +1,11 @@
+arg_name: param
+name: $logger
+type: Psr\\Log\\LoggerInterface
+description: |
+  A logger to register.
+
+  If the logger is already registered, the method will have no effect.
+interface: phpmethod
+operation: ~
+optional: false
+...
--- a/docs/reference/function/add_logger.txt
+++ b/docs/reference/function/add_logger.txt
@@ -35,16 +35,16 @@ This function allows the application to register one or more
 objects to receive log messages from libmongoc and the driver. Each registered
 logger will receive messages for *all* clients.
 
-Messages originating from libmongoc will have their log level translated to an
-equivalent `PSR log level <https://www.php-fig.org/psr/psr-3/#5-psrlogloglevel>`__.
-For performance reasons, trace-level messages from libmongoc are *not* forwarded
-to PSR loggers. The extension's
+Messages originating from the extension will have their log level translated to
+an equivalent `PSR log level <https://www.php-fig.org/psr/psr-3/#5-psrlogloglevel>`__.
+For performance reasons, trace-level messages from the extension are *not*
+forwarded to PSR loggers. The extension's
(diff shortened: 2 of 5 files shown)


docs/includes/apiargs-function-add_logger-param.yaml
```diff
@@ -0,0 +1,11 @@
+arg_name: param
+name: $logger
+type: Psr\\Log\\LoggerInterface
```
jmikola, line 3: For "reasons", we can't use a URL here. I tried:

```
`Psr\Log\LoggerInterface <https://www.php-fig.org/psr/psr-3/#3-psrlogloggerinterface>`__
```

Giza complains about a leading backtick character. Text roles (e.g. `:php:`) work, but we don't have one for the PSR website.

I _was_ able to specify a URL when also using multi-line string syntax (e.g. `type: |`), but that appends a trailing " or" to the output. I expect there's something in the build system that supports multiple types on separate lines -- which we've never used since we typically write "array|object".

Anyway, the PSR website is linked from the main doc page so this isn't a huge problem.

docs/reference/function/add_logger.txt
```diff
@@ -0,0 +1,59 @@
...
+
+Behavior
+--------
+
+This function allows the application to register one or more
+`Psr\\Log\\LoggerInterface <https://www.php-fig.org/psr/psr-3/#3-psrlogloggerinterface>`__
+objects to receive log messages from libmongoc and the driver. Each registered
+logger will receive messages for *all* clients.
+
+Messages originating from libmongoc will have their log level translated to an
```
GromNaN, line 38: "libmongoc" is something for insiders. What do you think of the more abstract word "driver" or "extension".

```suggestion
Messages originating from the mongodb extension will have their log level translated to an
```
Reading this, I'd be interested to know what the mapping is, so that I know which types of messages I'm going to receive on which levels.

jmikola: I'm actually linking to the libmongoc docs below as that's the canonical reference for its log levels. Users are going to be exposed to libmongoc either way, as that's where the domain strings come from.

No objection changing this to the extension, though. This also relates to [PHPLIB-1279](https://jira.mongodb.org/browse/PHPLIB-1279).

This all makes we wish we could require users to complete a quick modal popup before landing on the docs and have them acknowledge that the "driver" consists of a PHP library, PHP extension, and libmongoc. It would use a cookie and attempt to be a one-time thing -- but that would probably land us in a quandary of then needing some silly GDPR cookie interstitial, which in my experience happens _every time_. Alas.

alcaeus (approved):
LGTM

docs/reference/function/add_logger.txt
GromNaN: Ok, let's not confront European legislation and let readers learn at their own pace.

Thanks for the update.

GromNaN (approved):
LGTM

PR #1194, opened 2023-11-20
Fix wrong data type in codec tutorial

alcaeus:
As is, the example would cause a type error. Thanks @OskarStark for bringing this to our attention!

### DIFF
--- a/docs/examples/codecs/handling-data-types/Person.php
+++ b/docs/examples/codecs/handling-data-types/Person.php
@@ -6,7 +6,7 @@ final class Person
 {
     public function __construct(
         public string $name,
-        public readonly DateTime $createdAt = new DateTime(),
+        public readonly DateTimeImmutable $createdAt = new DateTimeImmutable(),
         public readonly ObjectId $id = new ObjectId(),
     ) {
     }
--- a/docs/tutorial/codecs.txt
+++ b/docs/tutorial/codecs.txt
@@ -74,7 +74,12 @@ BSON date and accompanying timezone string. Those same embedded documents can th
 .. literalinclude:: /examples/codecs/handling-data-types/DateTimeCodec.php
    :language: php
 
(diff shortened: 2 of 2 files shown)


GromNaN (approved):

GromNaN:
@alcaeus the phpcs job should run on this PHP file.

OskarStark:
Should it be renamed then?
```diff
- DateTimeCodec
- DateTimeImmutableCodec
```

GromNaN:
> Should it be renamed then?
> 
> ```diff
> - DateTimeCodec
> - DateTimeImmutableCodec
> ```

Immutability should be the default and `DateTime` should be prohibed. The codec name seems correct. It could be more permissive and accept any `DateTimeInterface`, but always output a `DateTimeImmutable`.

OskarStark:
Well if I have to deal with external objects which are using DateTime, I would end up having a DateTimeCodec and a DateTimeImmuatableCodec. I agree that immutability is recommended and would always use it, but maybe I am not in charge of this 🤷‍♂️

GromNaN:
It's a tutorial, I like that it teaches good practices more broadly than the subject covered.

OskarStark:
I totally agree, but would change the name anyway to `DateTimeImmutable`, but its not my decision

alcaeus:
> the phpcs job should run on this PHP file.

I deferred this on purpose so that I can use PHP 8.1 syntax in the codec tutorial. If we want to run phpcs, I'll have to change it back to PHP 7.4 syntax.

I decided to ignore the naming discussion as this is just supposed to be inspiration for people. I did however add a note about the codec accepting any `DateTimeInterface` when encoding to BSON but always decoding to `DateTimeImmutable`.

GromNaN (approved):

OskarStark (approved):
❤️

PR #1196, opened 2023-11-21
PHPLIB-1248 Add examples on GridFS

GromNaN:
Fix [PHPLIB-1248](https://jira.mongodb.org/browse/PHPLIB-1248)


examples/gridfs-stream.php
```diff
@@ -0,0 +1,47 @@
...
+// Open a stream for reading, similar to fopen with mode 'r'
+$stream = $bucket->openDownloadStreamByName('hello.txt');
+
+$size = 0;
+while (! feof($stream)) {
+    $data = fread($stream, 2 ** 10);
+    $size += strlen($data);
+}
+
+echo 'Read a total of ' . $size . ' bytes' . PHP_EOL;
```
jmikola, line 47: Can these be added to ExamplesTest? The output for both scripts looks deterministic enough to assert.

examples/gridfs-upload.php
```diff
@@ -0,0 +1,47 @@
...
+require __DIR__ . '/../vendor/autoload.php';
+
+$client = new Client(getenv('MONGODB_URI') ?: 'mongodb://127.0.0.1/');
+
+$gridfs = $client->selectDatabase('test')->selectGridFSBucket();
+
+// Create an in-memory stream, this can be any stream source like STDIN or php://input for web requests
+$stream = fopen('php://temp', 'w+');
+fwrite($stream, 'Hello world!');
+rewind($stream);
```
jmikola, line 33: Is `rewind()` necessary? `Bucket::uploadFromStream()` uses [`stream_copy_to_stream()`](https://www.php.net/stream_copy_to_stream), the docs for which are actually a bit ambiguous because the `$offset` parameter defaults to `0`.

> Makes a copy of up to length bytes of data from the current position (or from the offset position, if specified) in from to to. If length is null, all remaining content in from will be copied.

Looking at the [implementation](https://github.com/php/php-src/blob/php-8.3.1/ext/standard/streamsfuncs.c#L517), it appears that `$offset=0` actually has no effect. The offset is only used to seek the `$from` stream when the value is > 0. That suggests that we can't even explicitly pass `0` to copy from the beginning of `$from`.

To that end, I suppose this `rewind()` _is_ necessary.

examples/gridfs-stream.php
```diff
@@ -0,0 +1,47 @@
...
+
+// Open a stream for writing, similar to fopen with mode 'w'
+$stream = $bucket->openUploadStream('hello.txt');
+
+for ($i = 0; $i < 1_000_000; $i++) {
+    fwrite($stream, 'Hello line ' . $i . PHP_EOL);
+}
+
+// Last data are flushed to the server when the stream is closed
+fclose($stream);
```
jmikola, line 36: Do you want to delete this stream at the end of the script as you do in `gridfs-upload.php`?

examples/gridfs-stream.php
GromNaN: I added the tests. But I had to update the upload example to use echo, the contents written in `STDOUT` isn't caught by `ob_get_contents` for the tests.

examples/gridfs-upload.php
GromNaN: Streams are generally consumed from the current cursor location. I have to rewind to get the file contents from the beginning.

examples/gridfs-stream.php
GromNaN: This gave me the opportunity to use the `getFileIdForStream` method.

GromNaN:
I renamed the example files to use underscores as it was previously done for `command_logger.php`.

examples/gridfs_upload.php
```diff
@@ -0,0 +1,50 @@
...
+
+// Create an in-memory stream, this can be any stream source like STDIN or php://input for web requests
+$stream = fopen('php://temp', 'w+');
+fwrite($stream, 'Hello world!');
+rewind($stream);
+
+// Upload to GridFS from the stream
+$id = $gridfs->uploadFromStream('hello.txt', $stream);
+assert($id instanceof ObjectId);
+echo 'Inserted file with ID: ' . $id . PHP_EOL;
```
jmikola, line 38: `gridfs_stream_wrapper.php` appears to be the first file that used `PHP_EOL`. Every other examples uses newline characters. Since the expected output in ExamplesTest uses Unix-style line endings, I think `"\n"` is the optimal thing to use in output.

jmikola (approved):
One comment about line endings, but LGTM otherwise.

examples/gridfs_upload.php
GromNaN: Ok, fixed for the 3 gridfs examples.

alcaeus (approved):

examples/gridfs_stream.php
```diff
@@ -0,0 +1,54 @@
...
+use function feof;
+use function fread;
+use function fwrite;
+use function getenv;
+use function strlen;
+
+require __DIR__ . '/../vendor/autoload.php';
+
+$client = new Client(getenv('MONGODB_URI') ?: 'mongodb://127.0.0.1/');
+// Disable MD5 computation for faster uploads, this feature is deprecated
```
GromNaN, line 25: I explicitly disabled MD5, that's what developers should do in their applications.

PR #1197, opened 2023-11-22
PHPLIB-1313 Ensure the GridFS stream is saved when the script ends

GromNaN:
Fix PHPLIB-1313

By forcing the writable stream to be properly closed, we ensure the file is sent to the server when the process stops.


src/GridFS/StreamWrapper.php
```diff
@@ -53,10 +53,11 @@ class StreamWrapper
 
     public function __destruct()
     {
-        /* This destructor is a workaround for PHP trying to use the stream well
-         * after all objects have been destructed. This can cause autoloading
-         * issues and possibly segmentation faults during PHP shutdown. */
-        $this->stream = null;
```
GromNaN, line 59: I could remove this line introduced by PHPLIB-345 because even if PHP tries to close the stream again, this is no-op.

GromNaN: Note that the new test reproduces also the issue described in PHPLIB-345.

alcaeus (approved):

src/GridFS/StreamWrapper.php
alcaeus: I'd have to go back and check, but PHP may also have fixed this issue by shutting down streams earlier in the shutdown process. Either way, manually closing the stream definitely fixes the write issue, but if we want to be absolutely sure we can also reset the `stream` property as was done previously to ensure the issue does not re-appear.

GromNaN: I was able to reproduce the "autoloading" issue by removing the `__destruct()` function. So I don't see what would have changed in PHP core for this.

alcaeus: Ah, I see. Since you've confirmed that the current code works around the issue, no objections to merging this as is 👍

GromNaN:
`testDanglingOpenWritableStream` is failing because now it executes a MongoDB command and the server URI is not read from the config. I updated the test instead of creating a new one.
```
[2023/11/23 14:49:49.753] FAILURE:  (PHPUnit\Framework\ExpectationFailedException)
[2023/11/23 14:49:49.753] MongoDB\Tests\GridFS\BucketFunctionalTest::testDanglingOpenWritableStream
[2023/11/23 14:49:49.753] Failed asserting that 255 is identical to 0.
[2023/11/23 14:49:49.753] /data/mci/06e79c8f4d646cf1e20e31e5d62575cb/src/tests/GridFS/BucketFunctionalTest.php:867
```

alcaeus (approved):

PR #1200, opened 2023-11-27
Fix evergreen failures

alcaeus:
This PR fixes a couple of failures that occur due to the test client being created not having the required API version set. To avoid issues in the future, the PR also changes the names of those tasks to match the schema used elsewhere so they can be run in PR builds.


GromNaN (approved):
Smart 👍🏻

PR #1202, opened 2023-11-27
PHPLIB-1304: Pull mongohouse image from ECR repo

alcaeus:
PHPLIB-1304

ADL tests have been scheduled as part of the PR patch on Evergreen.

### DIFF
--- a/.evergreen/config/functions.yml
+++ b/.evergreen/config/functions.yml
@@ -158,12 +158,17 @@ functions:
     - command: shell.exec
       params:
         script: |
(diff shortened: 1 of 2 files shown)


.evergreen/config/functions.yml
```diff
@@ -158,12 +158,17 @@ functions:
     - command: shell.exec
       params:
         script: |
-          VARIANT=${VARIANT} DRIVERS_TOOLS="${DRIVERS_TOOLS}" sh ${DRIVERS_TOOLS}/.evergreen/atlas_data_lake/build-mongohouse-local.sh
```
GromNaN, line 161: The `VARIANT` variable is not necessary, or is it exported?

GromNaN (approved):

.evergreen/config/functions.yml
alcaeus: It was necessary to compile on the correct platform. It's no longer necessary so we can do without :)

PR #1205, opened 2023-12-15
PHPLIB-1325 Deprecate setting `disableMD5` to `false` when using GridFS

GromNaN:
Fix PHPLIB-1325

### DIFF
--- a/src/GridFS/Bucket.php
+++ b/src/GridFS/Bucket.php
@@ -61,7 +61,10 @@ use function stream_context_create;
 use function stream_copy_to_stream;
 use function stream_get_meta_data;
 use function stream_get_wrappers;
+use function trigger_error;
 use function urlencode;
+
+use const E_USER_DEPRECATED;
 
 /**
  * Bucket provides a public API for interacting with the GridFS files and chunks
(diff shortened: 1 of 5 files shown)


src/GridFS/Bucket.php
```diff
@@ -158,6 +162,10 @@ public function __construct(Manager $manager, string $databaseName, array $optio
             throw InvalidArgumentException::invalidType('"disableMD5" option', $options['disableMD5'], 'boolean');
         }
 
+        if (! $options['disableMD5']) {
+            @trigger_error('Not setting "disableMD5" option to "true" is deprecated. Checksum is deprecated by GridFS spec and will be removed in mongodb/mongodb 2.0', E_USER_DEPRECATED);
```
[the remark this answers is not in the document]
GromNaN: Technically, it's a breaking change. For people that relies on the md5, they will get an incompatible version.
This process is typically what we do in Symfony to change the default value of an option. But Yes, I agree that it force people to add an option that they will remove later. Or ignore the deprecation warning.

GromNaN:
Updated the PR to raise a deprecation notice if user explicitly passes false.

src/GridFS/Bucket.php
```diff
@@ -132,6 +133,10 @@ class Bucket
      */
     public function __construct(Manager $manager, string $databaseName, array $options = [])
     {
+        if (isset($options['disableMD5']) && $options['disableMD5'] === false) {
+            trigger_deprecation('mongodb/mongodb', '1.10', 'Setting GridFS "disableMD5" option to "false" is deprecated and will not be supported in version 2.0.');
```
jmikola, line 137: "1.10" looks incorrect here. Should it be "1.18", since that will be the version where this change is implemented?

GromNaN: Fixed. Copilot need to be smarter with package versions.

jmikola (approved):

alcaeus (approved):

PR #1206, opened 2023-12-15
PHPLIB-1323 Implement `unlink` for GridFS stream wrapper

GromNaN:
Fix PHPLIB-1323

This is not optimized as it creates a `WritableStream` instance that is not really used.
The context resolver mechanism introduces in #1138 doesn't give direct access to the `CollectionWrapper` which is required for this feature.

### DIFF
--- a/docs/reference/method/MongoDBGridFSBucket-registerGlobalStreamWrapperAlias.txt
+++ b/docs/reference/method/MongoDBGridFSBucket-registerGlobalStreamWrapperAlias.txt
@@ -48,8 +48,8 @@ Supported stream functions:
 - :php:`filesize() <filesize>`
 - :php:`file() <file>`
 - :php:`fopen() <fopen>` with "r", "rb", "w", and "wb" modes
-- :php:`rename() <rename>` rename all revisions of a file in the same bucket
-- :php:`unlink() <unlink>` remove all revisions of a file in the same bucket
+- :php:`rename() <rename>`
+- :php:`unlink() <unlink>`
 
 In read mode, the stream context can contain the option ``gridfs['revision']``
 to specify the revision number of the file to read. If omitted, the most recent
@@ -57,6 +57,9 @@ revision is read (revision ``-1``).
 
 In write mode, the stream context can contain the option ``gridfs['chunkSizeBytes']``.
 If omitted, the defaults are inherited from the ``Bucket`` instance option.
+
+The functions `rename` and `unlink` will rename or remove all revisions of a
+filename. If the filename does not exist, these functions will return ``false``.
 
 Example
 -------
--- a/docs/reference/method/MongoDBGridFSBucket-registerGlobalStreamWrapperAlias.txt
+++ b/docs/reference/method/MongoDBGridFSBucket-registerGlobalStreamWrapperAlias.txt
@@ -49,7 +49,7 @@ Supported stream functions:
 - :php:`file() <file>`
 - :php:`fopen() <fopen>` with "r", "rb", "w", and "wb" modes
 - :php:`rename() <rename>` rename all revisions of a file in the same bucket
-- :php:`unlink() <unlink>` (remove all revisions of a file)
+- :php:`unlink() <unlink>` remove all revisions of a file in the same bucket
 
 In read mode, the stream context can contain the option ``gridfs['revision']``
 to specify the revision number of the file to read. If omitted, the most recent
--- a/docs/reference/method/MongoDBGridFSBucket-registerGlobalStreamWrapperAlias.txt
+++ b/docs/reference/method/MongoDBGridFSBucket-registerGlobalStreamWrapperAlias.txt
@@ -49,6 +49,7 @@ Supported stream functions:
 - :php:`file() <file>`
 - :php:`fopen() <fopen>` with "r", "rb", "w", and "wb" modes
 - :php:`rename() <rename>` rename all revisions of a file in the same bucket
+- :php:`unlink() <unlink>` (remove all revisions of a file)
 
 In read mode, the stream context can contain the option ``gridfs['revision']``
 to specify the revision number of the file to read. If omitted, the most recent
@@ -85,6 +86,8 @@ Each call to these functions makes a request to the server.
    var_dump(file_exists('gridfs://mybucket/hello.txt'));
 
    echo file_get_contents('gridfs://mybucket/hello.txt');
+
+   unlink('gridfs://mybucket/hello.txt');
 
 The output would then resemble:
 
--- a/docs/reference/method/MongoDBGridFSBucket-registerGlobalStreamWrapperAlias.txt
+++ b/docs/reference/method/MongoDBGridFSBucket-registerGlobalStreamWrapperAlias.txt
@@ -59,7 +59,7 @@ In write mode, the stream context can contain the option ``gridfs['chunkSizeByte
 If omitted, the defaults are inherited from the ``Bucket`` instance option.
 
 The functions `rename` and `unlink` will rename or remove all revisions of a
-filename. If the filename does not exist, these functions will return ``false``.
+filename. If the filename does not exist, these functions throw a ``FileNotFoundException``.
 
 Example
 -------
--- a/src/GridFS/CollectionWrapper.php
+++ b/src/GridFS/CollectionWrapper.php
@@ -284,7 +284,7 @@ class CollectionWrapper
         return $this->filesCollection->updateMany(
             ['filename' => $filename],
             ['$set' => ['filename' => $newFilename]],
-        )->getModifiedCount();
+        )->getMatchedCount();
     }
 
     /**
--- a/src/GridFS/CollectionWrapper.php
+++ b/src/GridFS/CollectionWrapper.php
@@ -81,16 +81,8 @@ class CollectionWrapper
     }
 
     /**
-     * Deletes a GridFS file and related chunks by ID.
-     *
-     * @param mixed $id
+     * Delete all GridFS files and chunks for a given filename.
      */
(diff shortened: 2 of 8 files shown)


src/GridFS/WritableStream.php
```diff
@@ -173,6 +174,33 @@ public function close(): void
...
+            'projection' => ['_id' => 1],
+        ]);
+
+        $count = 0;
+
+        foreach ($revisions as $file) {
+            $this->collectionWrapper->findFileById($file->get('_id'));
+            $this->collectionWrapper->deleteFileAndChunksById($file->get('_id'));
+            $count++;
+        }
```
alcaeus, line 197: I'm wondering if we can improve on this by collecting all IDs, then deleting all files with the given filename, followed by removing all chunks for the IDs we previously found.

While there's always a chance to end up in an undesirable state, e.g. by having older revisions of files remaining, the suggestion above would reduce this by instructing the server to delete all file documents with a given name, ensuring that future requests for such files would not produce any result. Thus, if there was an error while deleting any chunks, we wouldn't end with older revisions left over.

GromNaN: I will implement a solution consistent with `rename` https://github.com/mongodb/mongo-php-library/pull/1207#discussion_r1430167458

jmikola: This would definitely benefit from some bulk strategy to remove the `fs.files` documents first (likely using `$in` on multiple `_id` values), and _then_ cleaning up the orphan chunks. I think it'd also benefit from being addressed in the GridFS spec, since we'd probably want to have a consistent way to address this in any driver that wishes to do so. And codifying that in a spec would ensure we're not overlooking some edge case.

docs/reference/method/MongoDBGridFSBucket-registerGlobalStreamWrapperAlias.txt
```diff
@@ -49,6 +49,7 @@ Supported stream functions:
 - :php:`file() <file>`
 - :php:`fopen() <fopen>` with "r", "rb", "w", and "wb" modes
 - :php:`rename() <rename>` rename all revisions of a file in the same bucket
+- :php:`unlink() <unlink>` (remove all revisions of a file)
```
jmikola, line 52: Should "in the same bucket" be mentioned here, as it is for `rename()` above?

I'd consider "in the same bucket" to be assumed, but can understand why you might want to be explicit here.

src/GridFS/WritableStream.php
```diff
@@ -173,6 +173,21 @@ public function close(): void
...
 
+    /**
+     * Delete all files and chunks associated with this stream filename.
+     *
+     * @return int The number of deleted files
+     */
+    public function delete(): int
+    {
+        try {
+            return $this->collectionWrapper->deleteFileAndChunksByFilename($this->file['filename']);
```
jmikola, line 184: I realize this is an internal class, so we're not actually introducing a public API here, but a `delete()` method that affects multiple revisions seems odd, which is otherwise an abstraction for writing a single revision.

In the PR description, you said:

> The context resolver mechanism introduced in https://github.com/mongodb/mongo-php-library/pull/1138 doesn't give direct access to the CollectionWrapper which is required for this feature.

If we look at `stream_open`, it accesses a CollectionWrapper after resolving the path to an array of context options. Couldn't the same approach be used here to access a CollectionWrapper directly?

src/GridFS/CollectionWrapper.php
```diff
@@ -80,6 +80,30 @@ public function deleteChunksByFilesId($id): void
...
+    public function deleteFileAndChunksByFilename(string $filename): int
+    {
+        /** @var iterable<array{_id: mixed}> $files */
+        $files = $this->findFiles(['filename' => $filename], [
+            'codec' => null,
+            'typeMap' => ['root' => 'array'],
+            'projection' => ['_id' => 1],
+        ]);
+
+        /** @var list<mixed> $ids */
```
jmikola, line 95: It's a shame [`array_column()`](https://www.php.net/array_column) doesn't support iterables (despite allowing elements _within_ the input array to be either arrays or objects).

src/GridFS/CollectionWrapper.php
```diff
@@ -80,6 +80,30 @@ public function deleteChunksByFilesId($id): void
...
     }
 
+    /**
+     * Delete all GridFS files and chunks for a given filename.
+     */
+    public function deleteFileAndChunksByFilename(string $filename): int
+    {
+        /** @var iterable<array{_id: mixed}> $files */
+        $files = $this->findFiles(['filename' => $filename], [
+            'codec' => null,
```
jmikola, line 90: This shouldn't be necessary.

Operations such as Find already raise exceptions if both a `codec` _and_ `typeMap` are specified, and `Collection::find()` relies on `inheritCodecOrTypeMap()`, which avoids inheriting `codec` if an explicit `typeMap` option is specified (and vice versa).

src/GridFS/WritableStream.php
GromNaN: I felt deep down that using `ReadableStream` and `WritableStream` wasn't correct. I refactored by adding a private `getContext` method to use `CollectionWrapper` directly for `rename` and `unlink`.

src/GridFS/CollectionWrapper.php
GromNaN: Yes, I tried. We need the function `iterable_column`

src/GridFS/StreamWrapper.php
```diff
@@ -324,6 +295,14 @@ public function stream_write(string $data): int
         return $this->stream->writeBytes($data);
     }
 
+    public function unlink(string $path): bool
+    {
+        $context = $this->getContext($path, 'r');
```
jmikola, line 300: `Bucket::resolveStreamContext()` executes `findFileByFilenameAndRevision()` for read modes in order to fetch the file document. Would it make sense to use `'w'` here and avoid the unnecessary query?

I understand `rename()` needs to use `'r'` since it does read the filename from the file document; however, there may be a refactoring opportunity there if you'd just like to trust the `$fromPath` string. I suppose it's a matter of intentionally NOP-ing on a FileNotFoundException or executing an update query that may not match anything -- so I'd be find leaving `rename()` as-is.

GromNaN: The idea of using `r` mode was to check that the file exists before doing operations, to throw a `FileNotFoundException`. This could be optimized by checking the number of updated documents after the update, but that's not feasible if `Write Concern = 0`.

GromNaN: I reverted to throw a `FileNotFoundException` in `rename` and `unlink`, instead of silently returning `false`. This behavior is consistent with S3 stream wrapper.

docs/reference/method/MongoDBGridFSBucket-registerGlobalStreamWrapperAlias.txt
```diff
@@ -57,6 +58,9 @@ revision is read (revision ``-1``).
 In write mode, the stream context can contain the option ``gridfs['chunkSizeBytes']``.
 If omitted, the defaults are inherited from the ``Bucket`` instance option.
 
+The functions `rename` and `unlink` will rename or remove all revisions of a
+filename. If the filename does not exist, these functions throw a ``FileNotFoundException``.
```
GromNaN, line 62: @jmikola removing or renaming a single revision is not supported. Even if the option `gridfs[revision]` is set.

src/GridFS/StreamWrapper.php
```diff
@@ -105,16 +106,17 @@ public function rename(string $fromPath, string $toPath): bool
...
 
-        try {
-            $this->stream_open($fromPath, 'r', 0, $openedPath);
-        } catch (FileNotFoundException $e) {
-            return false;
-        }
+        $context = $this->getContext($fromPath, 'w');
+
+        $newFilename = explode('/', $toPath, 4)[3] ?? '';
+        $count = $context['collectionWrapper']->updateFilenameForFilename($context['filename'], $newFilename);
```
jmikola, line 112: Edge case: if `$fromPath` and `$toPath` are identical, the file exists but you'll throw FileNotFoundException. That may warrant reverting to an "r" mode when selecting the context to throw eagerly.

Alternatively, you could have `updateFilenameForFilename()` return the matched count and assume that the update was successful or a NOP. The only normal scenario where the matched and modified counts wouldn't be equals is if the paths were the same, in which case you could probably NOP earlier and avoid calling `updateFilenameForFilename()` altogether (FWIW, PHP's `rename()` returns `true` in this case, but I dunno if there's some internal NOP behavior).

src/GridFS/StreamWrapper.php
```diff
@@ -338,6 +330,42 @@ public function url_stat(string $path, int $flags)
         return $this->stream_stat();
     }
 
+    /** @psalm-return ($mode == 'r' ? array{collectionWrapper: CollectionWrapper, file: object} : array{collectionWrapper: CollectionWrapper, filename: string, options: array}) */
```
jmikola, line 333: Not sure if this is relevant, but the ternary here only checks for "r" and `stream_open()` may call this with "rb". Can you invoke functions here, or are you restricted on syntax here?

src/GridFS/StreamWrapper.php
GromNaN: Good catch. I updated the code to return the matched count, and added a test for renaming with the same filename (existing and not existing).

src/GridFS/StreamWrapper.php
GromNaN: The logical `or` works. I also added the `@return` annotation so PHPStorm understands it.

jmikola (approved):

PR #1207, opened 2023-12-15
PHPLIB-1324 Implement `rename` for GridFS stream wrapper 

GromNaN:
Fix PHPLIB-1324

Rename all the revisions as [defined in the spec](https://github.com/mongodb/specifications/blob/0b47194538aa817978fae0f77f684f6d5e62ebab/source/gridfs/gridfs-spec.rst#renaming-stored-files)

> To rename multiple revisions of the same filename, users must retrieve the full list of files collection documents for a given filename and execute “rename” on each corresponding “_id”.

### Example

```php
file_put_contents('gridfs://bucket/filename', 'Hello, GridFS!');
rename('gridfs://bucket/filename', 'gridfs://bucket/newname');
echo file_get_contents('gridfs://bucket/newname');
```

### DIFF
--- a/src/GridFS/CollectionWrapper.php
+++ b/src/GridFS/CollectionWrapper.php
@@ -254,6 +254,17 @@ class CollectionWrapper
     }
 
     /**
+     * Updates the filename field in the file document for all the files with a given filename.
+     */
+    public function updateFilenameForFilename(string $filename, string $newFilename): UpdateResult
+    {
+        return $this->filesCollection->updateMany(
+            ['filename' => $filename],
+            ['$set' => ['filename' => $newFilename]],
+        );
+    }
+
+    /**
      * Updates the filename field in the file document for a given ID.
      *
      * @param mixed $id
--- a/src/GridFS/ReadableStream.php
+++ b/src/GridFS/ReadableStream.php
@@ -19,16 +19,17 @@ namespace MongoDB\GridFS;
 
 use Iterator;
 use MongoDB\BSON\Binary;
-use MongoDB\BSON\Document;
 use MongoDB\Driver\CursorInterface;
 use MongoDB\Exception\InvalidArgumentException;
 use MongoDB\GridFS\Exception\CorruptFileException;
+use MongoDB\GridFS\Exception\LogicException;
 
 use function assert;
 use function ceil;
 use function floor;
 use function is_integer;
 use function is_object;
+use function is_string;
 use function property_exists;
 use function sprintf;
 use function strlen;
@@ -179,24 +180,16 @@ class ReadableStream
 
     /**
      * Rename all revisions of the file.
-     *
-     * @return int The number of revisions renamed
      */
-    public function rename(string $newFilename): int
+    public function rename(string $newFilename): bool
     {
-        /** @var Document[] $revisions */
-        $revisions = $this->collectionWrapper->findFiles(
-            ['filename' => $this->file->filename],
-            ['typeMap' => ['root' => 'bson'], 'projection' => ['_id' => 1]],
-        );
-        $count = 0;
-
-        foreach ($revisions as $revision) {
-            $this->collectionWrapper->updateFilenameForId($revision->get('_id'), $newFilename);
-            $count++;
+        if (! isset($this->file->filename) || ! is_string($this->file->filename)) {
+            throw new LogicException('Cannot rename file without a filename');
         }
 
-        return $count;
+        $this->collectionWrapper->updateFilenameForFilename($this->file->filename, $newFilename);
+
+        return true;
     }
 
     /**
--- a/src/GridFS/ReadableStream.php
+++ b/src/GridFS/ReadableStream.php
@@ -19,6 +19,7 @@ namespace MongoDB\GridFS;
 
 use Iterator;
 use MongoDB\BSON\Binary;
+use MongoDB\BSON\Document;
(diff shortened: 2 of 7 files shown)


src/GridFS/ReadableStream.php
```diff
@@ -176,6 +177,28 @@ public function readBytes(int $length): string
...
+            ['typeMap' => ['root' => 'bson'], 'projection' => ['_id' => 1]],
+        );
+        $count = 0;
+
+        foreach ($revisions as $revision) {
+            $this->collectionWrapper->updateFilenameForId($revision->get('_id'), $newFilename);
+            $count++;
+        }
+
+        return $count;
```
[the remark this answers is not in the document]
GromNaN: It's explicitly written in the [spec](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst#renaming-stored-files) to perform the update one-by-one.

> Drivers construct and execute an update_one command on the files collection using `{ _id: @id }` as the filter and `{ $set : { filename : "new_filename" } }` as the update parameter.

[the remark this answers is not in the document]
jmikola: This may be worth bringing up with the wider DBX team, as that GridFS spec language definitely dates back a while and could stand to be updated. Re-reading it, it's clear to me that one of the priorities was raising an error if a file doesn't exist for a given `_id`; however, I see no technical reason why an `updateMany()` with all `_id` values expressed in a single filter (e.g. using `$in`) couldn't be used, provided the method also asserted that the expected number of documents were modified by the bulk update.

What you certainly don't want to do is attempt to update many documents based on the filename directly, since that could end up hitting additional file records that we didn't expect. That's certainly not happening here, but I wanted to point it out regardless.

src/GridFS/StreamWrapper.php
```diff
@@ -90,6 +93,30 @@ public static function register(string $protocol = 'gridfs'): void
         stream_wrapper_register($protocol, static::class, STREAM_IS_URL);
     }
 
+    /**
+     * Rename all revisions of a filename.
+     *
+     * @return bool True on success or false on failure.
+     */
+    public function rename(string $path_from, string $path_to): bool
```
jmikola, line 101: I realize these parameter names were taken from https://www.php.net/manual/en/streamwrapper.rename.php but do we want to use something more internally consistent with the library? This would match the format used in RenameCollection:

### 
```suggestion
    public function rename(string $fromPath, string $toPath): bool
```

src/GridFS/StreamWrapper.php
```diff
@@ -90,6 +93,30 @@ public static function register(string $protocol = 'gridfs'): void
         stream_wrapper_register($protocol, static::class, STREAM_IS_URL);
     }
 
+    /**
+     * Rename all revisions of a filename.
```
jmikola, line 97: I assume this only supports the bucket alias API, since we still only have a single `Bucket::rename()` method. Is this something that Drupal requires?

GromNaN: I don't know if Drupal needs `rename`.

src/GridFS/StreamWrapper.php
GromNaN: So simple, I hadn't thought of it. Fixed, and the phpcs rule exclusion can be removed.

src/GridFS/ReadableStream.php
```diff
@@ -179,24 +180,16 @@ public function readBytes(int $length): string
...
-            $this->collectionWrapper->updateFilenameForId($revision->get('_id'), $newFilename);
-            $count++;
+        if (! isset($this->file->filename) || ! is_string($this->file->filename)) {
+            throw new LogicException('Cannot rename file without a filename');
         }
 
-        return $count;
+        $result = $this->collectionWrapper->updateFilenameForFilename($this->file->filename, $newFilename);
+
+        return $result->isAcknowledged();
```
alcaeus, line 192: This would typically be `true`, unless a user specified a `writeConcern` of 0. In other cases there would be an exception. You can keep the previous return value by returning `$result->getModifiedCount()` (which can be `null` in case of a `writeConcern` of 0), or always return true as the exception will tell people if it didn't succeed.

src/GridFS/CollectionWrapper.php
```diff
@@ -253,6 +253,17 @@ public function insertFile($file): void
         $this->filesCollection->insertOne($file);
     }
 
+    /**
+     * Updates the filename field in the file document for all the files with a given filename.
+     */
+    public function updateFilenameForFilename(string $filename, string $newFilename): UpdateResult
+    {
+        return $this->filesCollection->updateMany(
```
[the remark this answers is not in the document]
alcaeus: Ack - I think since renaming files by filename is not part of the spec and the spec only contains a suggested implementation **for users**, we should be fine adding this implementation. I'd suggest we amend the spec to add APIs to work on all revisions of a file (i.e. remove and rename by filename instead of by ID).

alcaeus (approved):

src/GridFS/CollectionWrapper.php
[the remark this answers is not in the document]
jmikola: > renaming files by filename is not part of the spec and the spec only contains a suggested implementation for users

@alcaeus: Good point. I overlooked that the instructions were telling _users_ to run the `rename()` operation for each ID, instead of discussing how a driver should implement filename-based renames.

PR #1209, opened 2023-12-18
Fix psalm errors

alcaeus:
This fixes a couple of errors that appeared due to a psalm update which updated the PHPC method signatures.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="5.16.0@2897ba636551a8cb61601cc26f6ccfbba6c36591">
+<files psalm-version="5.18.0@b113f3ed0259fd6e212d87c3df80eec95a6abf19">
   <file src="docs/examples/encryption/csfle-explicit_encryption.php">
     <MixedArgument>
       <code><![CDATA[$clientEncryption->decrypt($document->encryptedField)]]></code>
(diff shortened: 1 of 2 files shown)


psalm-baseline.xml
```diff
@@ -114,10 +114,16 @@
...
+    <LessSpecificReturnStatement>
+      <code><![CDATA[(object) $this->getArrayCopy()]]></code>
+    </LessSpecificReturnStatement>
     <MixedAssignment>
       <code>$this[$key]</code>
       <code>$value</code>
     </MixedAssignment>
+    <MoreSpecificReturnType>
+      <code>stdClass</code>
+    </MoreSpecificReturnType>
```
alcaeus, line 126: The `LessSpecificReturnStatement` and `MoreSpecificReturnType` are due to https://github.com/vimeo/psalm/issues/8187. In essence, Psalm takes the return value of `(object) []` to be `object`, even though it is `stdClass`.

psalm-baseline.xml
```diff
@@ -509,6 +515,9 @@
     </MixedAssignment>
   </file>
   <file src="src/Operation/FindOneAndReplace.php">
+    <MixedArgument>
+      <code><![CDATA[$options['codec'] ?? null]]></code>
+    </MixedArgument>
```
alcaeus, line 520: Not sure why this errors in that one specific file - it works in all other files where we have the same checks on the codec option and then use it in a typed context 🤷

GromNaN (approved):
Perhaps we should set a stronger constraint on the psalm version, since there is no `compose.lock`.

alcaeus:
> Perhaps we should set a stronger constraint on the psalm version, since there is no `compose.lock`.

Yes, perhaps paired with dependabot updating the version constraint range when an update was released. I'll give it some thought.

psalm-baseline.xml
GromNaN: I couldn't reproduce.
https://psalm.dev/r/609a35bce9

PR #1215, opened 2024-01-12
Fix CI issues

alcaeus:
This PR takes care of a number of test failures:
* no-aws-cred tests were failing due to exception class changing. Instead of the expected `AuthenticationException`, it now throws an `EncryptionException`. I don't think the exception class matters for the tests (as the spec only suggests checking for failures, not a specific failure), but I can take a closer look at the reason for the change if necessary.
* PHPLIB-1327: Index management tests fail on latest due to recent changes. This PR syncs the tests with mongodb/specifications@e54cb7c3da226cb14285a14fa7219cdc5c8ed997.
* PHPLIB-1304: changes for ADL tests were only merged to the main branch, I backported these changes to v1.17 to avoid the system failure. Tests still fail however as we can't connect to the local ADL instance. This is a failure I've seen with other drivers as well and I'm still investigating this for a later fix.

This leaves the serverless tests to fail, which I'm investigating as part of PHPLIB-1303.

The patch build for this PR includes the no-aws-creds tasks that currently fail.

### DIFF
--- a/.evergreen/config/functions.yml
+++ b/.evergreen/config/functions.yml
@@ -158,12 +158,17 @@ functions:
     - command: shell.exec
       params:
         script: |
-          VARIANT=${VARIANT} DRIVERS_TOOLS="${DRIVERS_TOOLS}" ${DRIVERS_TOOLS}/.evergreen/atlas_data_lake/build-mongohouse-local.sh
+          cd ${DRIVERS_TOOLS}/.evergreen/atlas_data_lake
+
+          DRIVERS_TOOLS="${DRIVERS_TOOLS}" \
+          ./pull-mongohouse-image.sh
     - command: shell.exec
       params:
-        background: true
         script: |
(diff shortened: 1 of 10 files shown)


alcaeus:
After some investigation, I've reopened PHPLIB-1307 and fixed the issue where `listIndexes` sends a `readConcern` option, which was revealed by the tests.

GromNaN:
> fixed the issue where `listIndexes` sends a `readConcern` option, which was revealed by the tests.

I don't see this change in the PR, but tests are green.

alcaeus:
> I don't see this change in the PR, but tests are green.

Upon further investigation, I found that libmongoc is always adding the `readConcern`, even though the `Collection` class doesn't. I'm skipping that test for the time being until libmongoc fixes the issue or the test is changed.

GromNaN (approved):

PR #1218, opened 2024-01-22
PHPLIB-1372 Fix `RiskyTruthyFalsyComparison`

GromNaN:
Fix PHPLIB-1372

The new [RiskyTruthyFalsyComparison](https://psalm.dev/docs/running_psalm/issues/RiskyTruthyFalsyComparison/) errors are introduced by psalm [5.20.0](https://github.com/vimeo/psalm/releases/tag/5.20.0).

- Disable on example files. Using `getenv('MONGODB_URI') ?: 'mongodb://127.0.0.1/'` is safe as any falsy value could be replaced by the default URI.
- Replace `!empty($filter)` by `$filter !== []` as `$filter` is an array or an object, and even `empty(new stdClass)` is true. This might be a necessary bug fix.

### DIFF
--- a/src/ChangeStream.php
+++ b/src/ChangeStream.php
@@ -262,7 +262,7 @@ class ChangeStream implements Iterator
      */
     private function resume(): void
     {
-        if (null === $this->resumeCallable) {
+        if ($this->resumeCallable === null) {
             throw new BadMethodCallException('Cannot resume a closed change stream.');
         }
 
--- a/src/ChangeStream.php
+++ b/src/ChangeStream.php
@@ -262,7 +262,7 @@ class ChangeStream implements Iterator
      */
     private function resume(): void
     {
-        if (! $this->resumeCallable) {
+        if (null === $this->resumeCallable) {
             throw new BadMethodCallException('Cannot resume a closed change stream.');
         }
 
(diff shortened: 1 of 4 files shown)


src/ChangeStream.php
```diff
@@ -262,7 +262,7 @@ private function onIteration(bool $incrementKey): void
      */
     private function resume(): void
     {
-        if (! $this->resumeCallable) {
+        if (null === $this->resumeCallable) {
```
alcaeus, line 265: Not sure this is strictly necessary, but if we're going to not use the shorthand then let's not use Yoda style ;)

```suggestion
        if ($this->resumeCallable === null) {
```

src/Operation/Count.php
```diff
@@ -187,7 +187,7 @@ private function createCommandDocument(): array
     {
         $cmd = ['count' => $this->collectionName];
 
-        if (! empty($this->filter)) {
+        if ($this->filter !== []) {
```
jmikola, line 190: > `$filter` is an array or an object and even `empty(new stdClass)` is true. This might be a necessary bug fix.

The [count](https://www.mongodb.com/docs/manual/reference/command/count/) and [distinct](https://www.mongodb.com/docs/manual/reference/command/distinct/) commands both use an empty filter by default, so including `new stdClass` in the outgoing command was inconsistent (and redundant) but had no functional impact on the result.

I don't think we need to go the extra mile to detect empty objects, so if this is the most concise change to silence Psalm that works for me.

jmikola (approved):
I'll defer you on @alcaeus' suggestion.

Issue #608, opened 2019-04-23
Issue with error handling for GridFS readable stream


karptonite:
This is loosely related to issue #579.

When using `openDownloadableStream()` in gridFS, then feeding the resource to `fread`, the GridFS StreamWrapper catches exceptions then triggers errors in `stream_read()`. Then rather than re-throwing the error, it returns `false`.

That means that something like
```php
while ( !feof( $stream ) )
  echo fread( $stream, 8192 );
```
seems to create an infinite loop in the case of an exception such as ` MongoDB\GridFS\Exception\CorruptFileException: Chunk not found for index "0"`. (Setting aside the separate issue of why that exception was hitting in the first place.)

The error is logged, but the execution goes right back to the loop. Now that I know about this, I suppose I could fix it by assigning the results of `fread()` then breaking out of the loop if the result is ever empty while `feof()` is still false, but this was surprising to me. Is there some better way to handle this? I might have expected for the exception to be re-thrown after the error was logged, in which case I would have caught it and handled it myself.

karptonite:
Actually, I'm having trouble actually simulating this error. I have the 500,000 error logs for one request to suggest something is going wrong, but I'm not sure the issue is exactly what I've described.

karptonite:
Here was my code to work around this issue for now, but it obviously isn't ideal:
```php
while ( !feof( $stream ) ) {
  $result = fread( $stream, 8192 );
  if( !$result && !feof( $stream ) ) {
    throw new Exception( "Error reading data from store" );
  }
  echo $result;
}
```

jmikola:
> The error is logged, but the execution goes right back to the loop.

When you say the error is logged, are you referring to your application logging the `E_USER_WARNING` emitted by `trigger_error()` (see: [`StreamWrapper::stream_read()`](https://github.com/mongodb/mongo-php-library/blob/1.4.2/src/GridFS/StreamWrapper.php#L126) for context)?

When implementing GridFS, I tried to follow PHP's [StreamWrapper](https://www.php.net/manual/en/class.streamwrapper.php) guidelines to the letter. Unfortunately, it's doesn't provide guidance on exposing exceptions, since standard IO functions (e.g. `fopen`, `fread`) have their own error-reporting convention (returning boolean `false`) and generally aren't expected to throw.

[...]

PR #1224, opened 2024-01-29
PHPLIB-427 Rethrow the exception in EOF if it occured in fread

GromNaN:
Try to fix PHPLIB-427
Related to https://github.com/mongodb/mongo-php-library/issues/608

Remember when an exception is thrown when reading a stream, so that this exception keep being thrown when we try to reuse the stream.

### DIFF
--- a/src/GridFS/ReadableStream.php
+++ b/src/GridFS/ReadableStream.php
@@ -22,6 +22,7 @@ use MongoDB\BSON\Binary;
 use MongoDB\Driver\CursorInterface;
 use MongoDB\Exception\InvalidArgumentException;
 use MongoDB\GridFS\Exception\CorruptFileException;
+use Throwable;
 
 use function assert;
 use function ceil;
@@ -60,6 +61,8 @@ class ReadableStream
     private int $length;
 
     private int $numChunks = 0;
+
+    private Throwable $error;
 
     /**
      * Constructs a readable GridFS stream.
@@ -129,6 +132,10 @@ class ReadableStream
      */
     public function isEOF(): bool
     {
+        if (isset($this->error)) {
+            throw $this->error;
+        }
+
         if ($this->chunkOffset === $this->numChunks - 1) {
             return $this->bufferOffset >= $this->expectedLastChunkSize;
         }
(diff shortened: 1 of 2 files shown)


src/GridFS/ReadableStream.php
```diff
@@ -252,18 +263,24 @@ private function initBufferFromCurrentChunk(): bool
         }
 
         if (! $this->chunksIterator->valid()) {
-            throw CorruptFileException::missingChunk($this->chunkOffset);
+            $this->chunksIterator = null;
+
+            throw $this->error = CorruptFileException::missingChunk($this->chunkOffset);
```
jmikola, line 268: The change in this PR only captures a CorruptFileException. I assume any other exceptions, such as a socket error, would still propagate from the `initBufferFromCurrentChunk()` and `initBufferFromNextChunk()` methods and _not_ be captured.

That seems inconsistent, but before trying to address this I think we should reconsider whether _any_ changes are needed.

Per my [comment in PHPLIB-427](https://jira.mongodb.org/browse/PHPLIB-427?focusedCommentId=6050056&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-6050056), mongodb/mongo-php-library#608 was reported before the GridFS StreamWrapper was changed to throw. The OP encountered an infinite loop precisely because the PHP warning was not interrupting execution (despite being logged).

I expect that [PHPLIB-592](https://jira.mongodb.org/browse/PHPLIB-592) actually would have fixed their issue, and I see no reason to go out of our way to cater to users that may be _ignoring_ exceptions and still interacting with the stream.

If there is still a desire to invalidate the stream after some exception is thrown, I think we'd need some more thorough handling to catch anything that may be potentially thrown -- not just in this class but any of its own method calls. And in that case, preserving the original exception shouldn't be necessary -- just marking that an exception occurred and that future methods should do nothing but report failure seems sufficient. But I'm not convinced that's necessary.

GromNaN: Thanks for the analysis. I agree that retriability is incompatible with what I'm proposing here. 

Let's close.

PR #1225, opened 2024-01-31
PHPLIB-1321: Use correct method in createSearchIndexes() example

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1321

I opted not to pull in a more complex example from https://www.mongodb.com/docs/atlas/atlas-search/define-field-mappings that creates multiple indexes at once, since we already link to that page. Demonstrating the argument and return type difference seems sufficient.

### DIFF
--- a/docs/reference/method/MongoDBCollection-createSearchIndexes.txt
+++ b/docs/reference/method/MongoDBCollection-createSearchIndexes.txt
@@ -91,12 +91,14 @@ to index all document fields containing
 
    $collection = (new MongoDB\Client)->selectCollection('test', 'articles');
 
-   $indexNames = $collection->createSearchIndexes([
+   $indexNames = $collection->createSearchIndexes(
        [
-           'name' => 'test-search-index',
-           'definition' => ['mappings' => ['dynamic' => true]],
-       ],
-   ]);
+           [
+               'name' => 'test-search-index',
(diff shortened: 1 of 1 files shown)


docs/reference/method/MongoDBCollection-createSearchIndexes.txt
```diff
@@ -91,18 +91,23 @@ to index all document fields containing
 
    $collection = (new MongoDB\Client)->selectCollection('test', 'articles');
 
-   $indexName = $collection->createSearchIndex(
-      ['mappings' => ['dynamic' => true]],
-      ['name' => 'test-search-index']
-   );
+   $indexNames = $collection->createSearchIndexes([
```
GromNaN, line 94: You can put this first square bracket on a new line to make it more evident that it's a list of indexes.

GromNaN (approved):
LGTM, keeping the example simple.

docs/reference/method/MongoDBCollection-createSearchIndexes.txt
jmikola: This was consistent with what we already had in `createIndexes()` but I won't object.

GromNaN: Perfect

PR #1227, opened 2024-02-02
Stop testing on RHEL 7.6

alcaeus:
RHEL 7.6 currently fails due to it not being able to connect to pecl.php.net. Since the OS is EOL'd and doesn't play a big role for the PHP driver, there's no harm in removing it from the build pipeline. We're still testing RHEL 9.0 and 8.0 for x64 platforms, as well as 8.3 for Z, 8.2 for ARM64, and 8.1 for PPC.

### DIFF
--- a/.evergreen/config/build-variants.yml
+++ b/.evergreen/config/build-variants.yml
@@ -53,12 +53,6 @@ buildvariants:
     run_on: rhel80-small
     tasks:
       - name: "build-all-php"
-  - name: build-rhel76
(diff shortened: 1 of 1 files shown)


jmikola (approved):

alcaeus:
Note: I inadvertently merged this PR to the main branch even though it was supposed to go to 1.17. I've cherry-picked the merged commit to 1.17 and merged 1.17 back to main. This unfortunately now shows the commit twice, but at least we won't be bothered by failing builds anymore.

PR #1229, opened 2024-02-08
DOCSP-36027: Shorten class headings

norareidy:
This PR includes all the tickets in this [JIRA epic](https://jira.mongodb.org/browse/DOCSP-36027), which shorten the titles of API documentation pages.

Note: I also restructured the TOC a little by adding new drawers with class methods. This might be slightly out of scope for DOCSP-36027, but it seemed like the best way to reduce page titles enough so that they fit in the sidebar but don't omit necessary context (i.e., which class they belong to).  

Staging: https://preview-mongodbnorareidy.gatsbyjs.io/php-library/DOCSP-36027/reference/


docs/reference/class/MongoDBBulkWriteResult.txt
```diff
@@ -0,0 +1,26 @@
...
+   /reference/method/MongoDBBulkWriteResult-getDeletedCount
+   /reference/method/MongoDBBulkWriteResult-getInsertedCount
+   /reference/method/MongoDBBulkWriteResult-getInsertedIds
+   /reference/method/MongoDBBulkWriteResult-getMatchedCount
+   /reference/method/MongoDBBulkWriteResult-getModifiedCount
+   /reference/method/MongoDBBulkWriteResult-getUpsertedCount
+   /reference/method/MongoDBBulkWriteResult-getUpsertedIds
+   /reference/method/MongoDBBulkWriteResult-isAcknowledged
+
+- :phpmethod:`MongoDB\\BulkWriteResult::getDeletedCount()`
```
GromNaN, line 19: The the double backslashes are usually used for escaping but it looks like it's not necessary. On staging, I see double backslashes in all namespaces. A single backslash should be displayed.

image

docs/reference/class/MongoDBModelIndexInfo.txt
```diff
@@ -0,0 +1,30 @@
+.. _php-model-index-info:
+
+===============================
+MongoDB\\Model\\IndexInfo Class
```
GromNaN, line 4: How can I access this page? When I click on this class name, it opens the list of method but doesn't open this page.

image

docs/reference/method/MongoDBClient-addSubscriber.txt
GromNaN, line 1: Adding the comment here for the thread. Not related to this file.

Not a regression but a bug that you might fix here: the page on the `MongoDB\Client` class have lost the list of methods
https://www.mongodb.com/docs/php-library/current/reference/class/MongoDBClient/

image

GromNaN (commented):
While reviewing the preview page, I spotted some issues that might be related to your change.

> Staging: https://preview-mongodbnorareidy.gatsbyjs.io/php-library/DOCSP-36027/reference/

Why is this particular page empty?

mongoKart (approved):
shortened headings LGTM, assuming DBX is on board!

docs/reference/class/MongoDBBulkWriteResult.txt
norareidy: Just to clarify: every namespace mentioned in the docs should use one instead of two backslashes? If so, I'm happy to fix the backslashes in this PR  - or make a new ticket, since it seems like all the namespaces currently have two backslashes and many files would need to be modified

docs/reference/class/MongoDBModelIndexInfo.txt
norareidy: I just updated the snooty.toml file to make the "Enumeration Classes" drawer clickable (similar to the "Result Classes" and "Write Result Classes" drawers). Now you can access this page by going to "Enumeration Classes" -> scrolling down to "MongoDB\\Model\\IndexInfo"

docs/reference/method/MongoDBClient-addSubscriber.txt
norareidy: I actually noticed this about all the "Methods" sections (on MongoDB\Client, MongoDB\Database, MongoDB\Collection, etc). I'll fix the rendering issues in this PR

norareidy: Alright fixed!

docs/reference/class/MongoDBBulkWriteResult.txt
GromNaN: It's always one backslash. The double backslash doesn't exist in PHP namespaces.

norareidy:
> While reviewing the preview page, I spotted some issues that might be related to your change. Could you fix them directly in this PR, or create jira issues for later?
> 
> > Staging: https://preview-mongodbnorareidy.gatsbyjs.io/php-library/DOCSP-36027/reference/
> 
> Why is this particular page empty?

Hi Jerome, this page isn't actually accessible from the docs site (other than by editing the URL directly) as it's just a drawer for all the API documentation and doesn't have any content. I just added it as the staging link since all the edited pages fall under this heading. Sorry for the confusion!

docs/reference/class/MongoDBBulkWriteResult.txt
jmikola: The double backslashes were necessary for escaping syntax in RST; however, I think that was only relevant for actual headings and not cases where text is monospace-formatted.

docs/reference/class/MongoDBBulkWriteResult.txt
```diff
@@ -0,0 +1,26 @@
+.. _php-bulk-write-result:
+
+==============================
+MongoDB\\BulkWriteResult Class
```
jmikola, line 4: If all of these pages are nestled under "API Reference", I suggest we just get rid of the " Class" suffix to conserve space in the menus. It's redundant and tends to add extra word wrapping to the menu.

I'll defer to @alcaeus and @GromNaN, though.

docs/reference/write-result-classes.txt
```diff
@@ -29,15 +29,10 @@ Methods
...
-   /reference/method/MongoDBBulkWriteResult-getInsertedCount
-   /reference/method/MongoDBBulkWriteResult-getInsertedIds
-   /reference/method/MongoDBBulkWriteResult-getMatchedCount
-   /reference/method/MongoDBBulkWriteResult-getModifiedCount
-   /reference/method/MongoDBBulkWriteResult-getUpsertedCount
-   /reference/method/MongoDBBulkWriteResult-getUpsertedIds
-   /reference/method/MongoDBBulkWriteResult-isAcknowledged
+   /reference/class/MongoDBBulkWriteResult
 
+:ref:`BulkWriteResult Methods <php-bulk-write-result>`
```
jmikola, line 34: What is the reasoning for eliminating the list of linked methods from `write-result-classes.txt` and `result-classes.txt`? The original point of these pages were to avoid cluttering the navigation.

So instead of a concrete landing page for each result class (like we have with `MongoDBClient.txt`), we consolidated all of the content on these two pages.

I see this PR adds concrete landing pages for the result classes, but the only purpose seems to be including the list of methods. The "Description" headings that describe each class are still _here_, which is inconsistent with other class pages like `MongoDBClient.txt`.

I'd suggest either of the following:

 * Move "Definition" (the only remaining content on these pages) to the individual result class pages, and eliminate `write-result-classes.txt` and `result-classes.txt` altogether. In this case, we can lump all of the classes under a single "Result Classes" heading (no real need to differentiate write results).
 * Move the list of methods back to `write-result-classes.txt` and `result-classes.txt` and eliminate the need for individual pages for these classes (only keeping their methods).

docs/reference/method/MongoDBClient__construct.txt
```diff
@@ -1,6 +1,6 @@
-==============================
-MongoDB\\Client::__construct()
-==============================
+=============
+__construct()
```
jmikola, line 2: I assume this was done to consolidate entries in the nav menu, but it means the page title itself is much more vague. Is there no way to preserve `MongoDB\Client::__construct()` here and only trim the navigation entry? One of my original suggestions (perhaps in Slack) was to modify the logic to generate menu content and apply some string manipulation. That would certainly require some tooling changes, but I think it's preferable to leaving the destination pages themselves without identifying titles.

As is, this page is going to show up in search results as "__construct() -- PHP Library Manual", which looks pretty bad.

docs/reference/functions.txt
```diff
@@ -16,3 +16,7 @@ Functions
    /reference/function/add_logger
    /reference/function/remove_logger
    /reference/function/with_transaction
+
+- :phpmethod:`MongoDB\\add_logger()`
+- :phpmethod:`MongoDB\\remove_logger()`
+- :phpmethod:`MongoDB\\with_transaction()`
```
jmikola, line 22: What does `toctree` actually do if it's necessary to duplicate the list of pages in an RST here? Previously, `toctree` was visible in the rendered page output. After the Snooty migration, does it now _only_ control the rendering of the left-hand navigation?

docs/reference/class/MongoDBBulkWriteResult.txt
GromNaN: I don't think it's a problem since it's on the right. And it's a bit more explicit.

docs/reference/class/MongoDBBulkWriteResult.txt
GromNaN: Tracked by @norareidy in DOCSP-36627

docs/reference/functions.txt
norareidy: Yes I believe it just affects the organization of the left-hand navigation

docs/reference/write-result-classes.txt
norareidy: Okay, fixed according to your first suggestion (moved the "Definition" sections to the individual class pages)

docs/reference/class/MongoDBBulkWriteResult.txt
```diff
@@ -0,0 +1,39 @@
+.. _php-bulk-write-result:
```
jmikola, line 1: Are these anchors still needed if the result classes now have their own pages? No need to clean this up now, as it can be addressed in a subsequent PR. It would entail both removing this and updating any links that refer to it (for all such redundant anchors).

docs/reference/class/MongoDBChangeStream.txt
```diff
@@ -0,0 +1,44 @@
+.. _php-change-stream:
```
jmikola, line 1: Same comment as `_php-bulk-write-result` above.

docs/reference/class/MongoDBModelDatabaseInfoIterator.txt
```diff
@@ -0,0 +1,22 @@
+.. _php-model-database-info-iterator:
+
+==========================================
+MongoDB\\Model\\DatabaseInfoIterator Class
```
jmikola, line 4: I can't say what's responsible for this, but I noticed some of the links under "Enumeration Classes" leave the UI in a strange state. For instance, https://preview-mongodbnorareidy.gatsbyjs.io/php-library/DOCSP-36027/reference/class/MongoDBModelDatabaseInfoIterator/ scrolls past all of the main content.

This could be due to the long titles. In this case, I'd be amenable to stripping the full namespace from the class page titles and preserving the fully qualified class name (FQCN) below in the Definition section.

So this can just be "DatabaseInfoIterator Class" if that helps with navigation. I think that would still be fine for SEO purposes.

docs/reference/enumeration-classes.txt
```diff
@@ -4,190 +4,20 @@ Enumeration Classes
...
+   /reference/class/MongoDBModelIndexInfoIterator
 
-Methods
-~~~~~~~
+- :ref:`CollectionInfo <php-model-collection-info>`
+- :ref:`CollectionInfoIterator <php-model-collection-info-iterator>`
+- :ref:`DatabaseInfo <php-model-database-info>`
+- :ref:`DatabaseInfoIterator <php-model-database-info-iterator>`
+- :ref:`IndexInfo <php-model-index-info>`
+- :ref:`IndexInfoIterator <php-model-index-info-iterator>`
```
jmikola, line 22: Since the "Enumeration Classes" drawer is now clickable and expands (per https://github.com/mongodb/mongo-php-library/pull/1229/files#r1484800559), is this content even viewable anymore? I found no way to navigate to this page directly in the current staging deployment.

jmikola (commented):
Note: https://github.com/mongodb/mongo-php-library/pull/1229#discussion_r1486444603 is still outstanding.

docs/reference/method/MongoDBClient__construct.txt
jmikola: I think we need some input from the docs platform team to see what's possible here. If there's not an immediate solution available, I think reducing the page titles to just the base class name and method (e.g. `MongoDB\Client::__construct()` -> `Client::__construct()`) would be the best compromise to allow meaningful page titles.

norareidy: I reached out to docs platform and it seems like there is a way to do this! I'll change all the page titles back to the full namespace and just trim the navigation entry

docs/reference/class/MongoDBModelDatabaseInfoIterator.txt
norareidy: Seems like shortening the title fixed the scrolling issue

docs/reference/enumeration-classes.txt
norareidy: It's not viewable right now (other than by editing the URL or using the right/left arrow buttons at the bottom of each page). I can make it viewable, or leave as is and remove the content on this page

docs/reference/class/MongoDBBulkWriteResult.txt
norareidy: If you'd prefer I remove the content from "Enumeration Classes", "Result Classes", etc (mentioned in [this comment](https://github.com/mongodb/mongo-php-library/pull/1229/files#r1488489592)) then yes these anchors aren't necessary and I'll remove them. Otherwise, I'll keep the anchors if I make those landing pages clickable.

docs/reference/class/MongoDBModelDatabaseInfoIterator.txt
jmikola: This does still happen with some other pages: https://preview-mongodbnorareidy.gatsbyjs.io/php-library/DOCSP-36027/reference/method/MongoDBModelCollectionInfo-isCapped/

I imagine this isn't specific to PHPLIB, though, and is probably a more general issue with the docs theme. Perhaps worthy of a DOCSP ticket to be addressed later.

jmikola:
Noted that @schmalliso pointed you to a work-around for preserving the long page titles while still shortening their appearance in the menu (see: [Slack thread](https://mongodb.slack.com/archives/C018UTGP23T/p1707924824236599)).

docs/reference/method/MongoDBClient__construct.txt
jmikola: I saw the [Slack thread](https://mongodb.slack.com/archives/C018UTGP23T/p1707924824236599). Glad there's an existing solution for this, and thanks @schmalliso for the assist.

jmikola:
@norareidy: One last request on my end. Can we just eliminate "Enumeration Classes" and merge that with the "Result Classes" heading? I think the differentiation only made sense when there was an intermediary page showing all classes at once. Now that it's just a menu element and each class has its own page, I can't think of any reason to keep them separated.

Hopefully this also makes the docs configuration a bit simpler.

docs/reference/class/MongoDBModelDatabaseInfoIterator.txt
norareidy: Ah, yeah I think this issue has come up before. I'll follow up with Docs platform.

jmikola (approved):

norareidy:
> @norareidy: Is this ready to merge on your end? If so, I can do so manually since the current repository rules are preventing you from doing so (see: #1231).

Yes it is! I was waiting for the checks to finish, but if you can bypass them that would be great.

PR #1239, opened 2024-02-29
DOCSP-36627: Fix double backslashes

norareidy:
JIRA - https://jira.mongodb.org/browse/DOCSP-36627
Staging - https://preview-mongodbnorareidy.gatsbyjs.io/php-library/DOCSP-36627-staging/

PR #1243, opened 2024-03-04
DOCSP-36627: Additional double backslash fixes for v1.17

jmikola:
Merges #1239 from v1.16 up to v1.17 and applies additional fixes.


docs/reference/method/MongoDBCollection-explain.txt
```diff
@@ -29,7 +29,7 @@ Definition
 Parameters
 ----------
 
-``$explainable`` : :phpclass:`MongoDB\\Operation\\Explainable`
```
jmikola, line 32: Note: `\\Explainable` was missed in the original v1.16 PR: https://github.com/mongodb/mongo-php-library/pull/1239#discussion_r1511407703

docs/reference/function/add_logger.txt
```diff
@@ -26,7 +26,7 @@ Definition
 Parameters
 ----------
 
-``$logger`` : Psr\\Log\\LoggerInterface
+``$logger`` : Psr\Log\LoggerInterface
```
jmikola, line 29: @norareidy: Correct me if I'm wrong, but `\\` should only exist in title headings. I assume the plain text here should be changed to remove the escaping.

norareidy: After staging the docs for this PR, it looks like the `\\` should stay in this case as the single backslash doesn't show up -[ here's the staging link](https://preview-mongodbnorareidy.gatsbyjs.io/php-library/DOCSP-36627-v1.17/reference/function/add_logger/) to this page.

docs/reference/function/add_logger.txt
```diff
@@ -35,7 +35,7 @@ Behavior
 --------
 
 This function allows the application to register one or more
-`Psr\\Log\\LoggerInterface <https://www.php-fig.org/psr/psr-3/#3-psrlogloggerinterface>`__
+`Psr\Log\LoggerInterface <https://www.php-fig.org/psr/psr-3/#3-psrlogloggerinterface>`__
```
jmikola, line 38: @norareidy: I assume the escaping here also needs to state since this is a basic link? Similar to what you had to do in https://github.com/mongodb/mongo-php-library/pull/1244/files#diff-0494d8195d629119e6de652538a76bd23063d2234cf08cad50eeebc94e6dee99R34

docs/reference/function/add_logger.txt
```diff
@@ -15,7 +15,7 @@ MongoDB\\add_logger()
 Definition
 ----------
 
-.. phpmethod:: MongoDB\\add_logger()
+.. phpmethod:: MongoDB\add_logger()
```
jmikola, line 18: @norareidy: The "Behavior" section later in this file has:

```rst
`Psr\\Log\\LoggerInterface <https://www.php-fig.org/psr/psr-3/#3-psrlogloggerinterface>`__
```

I assume the escaping there needs to stay since it's basic link syntax? We only need to remove escaping within directives.

Similar to what you had to do in https://github.com/mongodb/mongo-php-library/pull/1244/files#diff-0494d8195d629119e6de652538a76bd23063d2234cf08cad50eeebc94e6dee99R34

norareidy: Yeah, the `\\` should stay in that case since it's not a directive

norareidy (approved):
lgtm!

PR #1247, opened 2024-03-06
PHPLIB-1404: Convert retryable reads spec tests to unified test format 

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1404

POC for https://github.com/mongodb/specifications/pull/1536/

Patch build: https://spruce.mongodb.com/version/65ea17691e2d178a7a47bae9/tasks


tests/UnifiedSpecTests/retryable-reads/gridfs-download-serverErrors.json
```diff
@@ -0,0 +1,1092 @@
...
+        {
+          "object": "bucket0",
+          "name": "download",
+          "arguments": {
+            "id": {
+              "$oid": "000000000000000000000001"
+            }
+          },
+          "expectResult": {
+            "$$matchesHexBytes": "11"
```
jmikola, line 117: Related to https://github.com/mongodb/specifications/pull/1536/#discussion_r1515092985

In [`findChunksByFileId()`](https://github.com/mongodb/mongo-php-library/blob/1.17.0/src/GridFS/CollectionWrapper.php#L115), we add additional criteria to the chunks query beyond the `files_id` expected in the original tests. The legacy test format permitted extra fields at all levels, so this was never checked.

I could have worked around this with `$$matchAsRoot`, but PHPLIB hasn't implemented that yet. Instead, I opted to relax the commandStartedEvent query to ignore `filter` altogether and instead assert the result of the successful operation (which was never done in the original tests).

tests/SpecTests/RetryableReadsSpecTest.php
```diff
@@ -1,135 +0,0 @@
...
-use function is_object;
-use function strpos;
-
-/**
- * Retryable reads spec tests.
- *
- * @see https://github.com/mongodb/specifications/tree/master/source/retryable-reads
- * @group serverless
- */
-class RetryableReadsSpecTest extends FunctionalTestCase
```
jmikola, line 19: Note: PHP has yet to implement any of the retryable reads prose tests. The CMAP prose test doesn't apply, and the mongos-related tests are waiting on [PHPC-1911](https://jira.mongodb.org/browse/PHPC-1911).

As such, there was nothing else in this file that needed to be kept.

tests/UnifiedSpecTests/Operation.php
```diff
@@ -517,14 +517,14 @@ private function executeForCollection(Collection $collection)
                 assertArrayHasKey('out', $args);
                 assertInstanceOf(Javascript::class, $args['map']);
                 assertInstanceOf(Javascript::class, $args['reduce']);
-                assertIsString($args['out']);
+                assertThat($args['out'], logicalOr(new IsType('string'), new IsType('array'), new IsType('object')));
 
-                return $collection->mapReduce(
+                return iterator_to_array($collection->mapReduce(
```
jmikola, line 522: Without this, the test runner attempted to match a MapReduceResult directly and timed out (likely due to some loop when dumping the comparison failure). This is consistent with eager iteration for other cursor-returning operations.

While looking into this, I wondered by we're still handling write result classes in ExpectedResult instead of immediately after an Operation. Opened [PHPLIB-1406](https://jira.mongodb.org/browse/PHPLIB-1406) to revisit that.

tests/UnifiedSpecTests/UnifiedTestRunner.php
```diff
@@ -485,7 +485,11 @@ private function preventStaleDbVersionError(array $operations, Context $context)
             switch ($operation->name) {
                 case 'distinct':
                     $hasDistinct = true;
-                    $collection = $context->getEntityMap()[$operation->object];
+                    /* TODO: If this operation references an entity that would
+                     * be created by a createEntities test runner operation, the
+                     * assertion below will fail; however, there is no need to
+                     * address this until such a transaction test is created. */
+                    $collection = $context->getEntityMap()[$operation->object] ?? null;
```
jmikola, line 492: Working around this would require running the `createEntities` operation ahead of time, which is a can of worms.

alcaeus (approved):
LGTM. Thanks for calling out the potential issue with entities created in `createEntities`. Fine leaving this as is for the time being.

GromNaN (approved):

PR #1253, opened 2024-03-13
Exclude read-write-concern tests from serverless testing

alcaeus:
Previously we were fine running all unified tests on serverless, but the read-write-concern tests fail and are not required to be tested. I've moved the `@group` annotation to all individual tests and only included the read-write-concern tests.

### DIFF
--- a/tests/UnifiedSpecTests/UnifiedSpecTest.php
+++ b/tests/UnifiedSpecTests/UnifiedSpecTest.php
@@ -16,7 +16,6 @@ use function glob;
  * Unified test format spec tests.
  *
  * @see https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst
- * @group serverless
  */
 class UnifiedSpecTest extends FunctionalTestCase
 {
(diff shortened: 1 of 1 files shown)


alcaeus:
Note: serverless tasks have been added to the PR patch.

tests/UnifiedSpecTests/UnifiedSpecTest.php
```diff
@@ -16,7 +16,6 @@
  * Unified test format spec tests.
  *
  * @see https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst
- * @group serverless
```
jmikola, line 19: I was considering doing this while porting tests over the past few weeks. Looks like a good change.

I did overlook the conflict with the `$out` tests and serverless. I do think it'd be reasonable to go back and add `serverless: forbid` to that spec test's `runOnRequirements`.

jmikola (approved):

tests/UnifiedSpecTests/UnifiedSpecTest.php
jmikola: Opened https://github.com/mongodb/specifications/pull/1550 and https://github.com/mongodb/mongo-php-library/pull/1254 to modify the spec tests.

PR #1260, opened 2024-03-14
PHPLIB-1399: Docs examples for agg expr projection

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1399


tests/DocumentationExamplesTest.php
```diff
@@ -728,6 +729,45 @@ public function testExample_42_50(): void
             $this->assertObjectNotHasAttribute('size', $document);
             $this->assertCount(1, $document->instock);
         }
+
+        // Start Aggregation Projection Example 1
```
jmikola, line 733: The examples are rendered within [Project Fields with Aggregation Expressions](https://www.mongodb.com/docs/manual/tutorial/project-fields-from-query-results/#project-fields-with-aggregation-expressions), which follows the earlier examples in this method and shares its data fixtures.

tests/DocumentationExamplesTest.php
```diff
@@ -730,6 +731,71 @@ public function testExample_42_50(): void
         }
     }
 
+    public function testAggregationProjectionExample_1(): void
+    {
+        $this->skipIfServerVersion('<', '4.4.0', '$project syntax for find and findAndModify is not supported');
```
jmikola, line 736: Just realized that the projection syntax requires 4.4+, so I've broken this example out into a separate method and pulled in some data fixtures from the related test method.

PR #1262, opened 2024-03-19
PHPLIB-1302: Use Composer\InstalledVersions

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1302

There is no impact on Composer requirements since jean85/pretty-package-versions 2.0+ previously required composer-runtime-api 2.0+.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -17,8 +17,8 @@
 
 namespace MongoDB;
 
+use Composer\InstalledVersions;
 use Iterator;
-use Jean85\PrettyVersions;
 use MongoDB\Driver\ClientEncryption;
 use MongoDB\Driver\Exception\InvalidArgumentException as DriverInvalidArgumentException;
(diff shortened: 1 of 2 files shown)


src/Client.php
```diff
@@ -396,9 +396,9 @@ private static function getVersion(): string
     {
         if (self::$version === null) {
             try {
-                self::$version = PrettyVersions::getVersion('mongodb/mongodb')->getPrettyVersion();
+                self::$version = InstalledVersions::getPrettyVersion('mongodb/mongodb') ?? 'unknown';
```
jmikola, line 399: Quoting the [Composer docs](https://getcomposer.org/doc/07-runtime.md#knowing-the-version-of-package-x):

> Returns the original version (e.g. v1.2.3) if vendor/package is installed, or null if it is provided/replaced, or throws OutOfBoundsException if the package is not installed at all

GromNaN (approved):
LGTM

PR #1274, opened 2024-04-08
Leverage PHP 8.0 string functions

GromNaN:
We use symfony/php80-polyfill that provides `str_starts_with`, `str_ends_with` and `str_contains`.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -45,6 +45,7 @@ use function is_object;
 use function is_string;
 use function MongoDB\BSON\fromPHP;
 use function MongoDB\BSON\toPHP;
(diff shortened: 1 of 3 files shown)


tools/connect.php
```diff
@@ -2,7 +2,7 @@
 
 function getHosts(string $uri): array
 {
-    if (strpos($uri, '://') === false) {
+    if (!str_contains($uri, '://')) {
```
GromNaN, line 5: This files doesn't include composer autoload. I'll revert this change.

jmikola (approved):

PR #1285, opened 2024-04-23
Fix syntax error in docs

### LINKED ISSUE #1284: Docs: missing `]` character in database->aggregate documentation
### Bug Report
Hi 👋 ,

Was going through the documentation for the MongoDB PHP Library and came across [this example](https://www.mongodb.com/docs/php-library/current/reference/method/MongoDBDatabase-aggregate/#examples),
which seems to result in a syntax error when taken verbatim due to a missing `]` character.

If so, it looks like [this](https://github.com/mongodb/mongo-php-library/blob/e3a588407120d0ac9f3cf7b23fb739fc79d95340/docs/reference/method/MongoDBDatabase-aggregate.txt#L163-L168) is the corresponding block of code/text that needs to be updated.

Let me know if this is the incorrect place/avenue to report this.

Thanks! 

### Environment
`n/a`

### Test Script
`n/a`

### Expected and Actual Behavior
`n/a`

### Debug Log
`n/a`

alcaeus:
Fixes #1284.

### DIFF
--- a/docs/reference/method/MongoDBDatabase-aggregate.txt
+++ b/docs/reference/method/MongoDBDatabase-aggregate.txt
@@ -163,7 +163,7 @@ running command operations.
    $cursor = $database->aggregate(
        [
            ['$currentOp' => []],
-           ['$match' => ['op' => 'command'],
+           ['$match' => ['op' => 'command']],
        ]
    );
 


GromNaN (approved):
What's stopping us from extracting all the examples into PHP files?

jmikola (approved):

PR #1287, opened 2024-04-29
PHPLIB-1415 Add tests for PHP 8.4

GromNaN:
Fix PHPLIB-1415

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,6 +36,7 @@ jobs:
           - "8.1"
           - "8.2"
           - "8.3"
+          - "8.4"
         mongodb-version:
           - "4.4"
         topology:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -104,7 +80,7 @@ jobs:
       - name: "Install dependencies with Composer"
         uses: "ramsey/composer-install@3.0.0"
         with:
-          composer-options: "--no-suggest"
+          composer-options: "--no-suggest ${{ matrix.php-version == '8.4' && '--ignore-platform-req=php+' || '' }}"
```
jmikola, line 83: TIL about [ternary syntax](https://docs.github.com/en/actions/learn-github-actions/expressions#example) in GitHub Action expressions.

Also, [`--ignore-platform-req=php+`](https://getcomposer.org/doc/03-cli.md#install-i).

What necessitates removing the upper bound for the PHP version? PHPLIB itself allows `^8.0`, so is this for some dev dependency?

.github/workflows/tests.yml
```diff
@@ -32,35 +32,11 @@ jobs:
           - "ubuntu-20.04"
         php-version:
           - "7.4"
-          - "8.0"
-          - "8.1"
-          - "8.2"
-          - "8.3"
+          - "8.4"
```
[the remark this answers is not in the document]
GromNaN: That was temporary to debug the CI config.

.github/workflows/tests.yml
GromNaN: [psalm](https://github.com/vimeo/psalm/blob/5.x/composer.json#L18) have strict PHP constraint. I added a comment.

PR #1288, opened 2024-05-07
PHPLIB-583: Automate release process

alcaeus:
PHPLIB-583 (Release process automation)
PHPLIB-584 (Signing releases)
PHPLIB-1435 (Running static analysis for tags)

This PR builds on the work done for the MongoDB extension. It uses the same flow to trigger a release, but there is no packaging step involved. Since we don't have a file to track the version, this workflow runs different consistency checks: it ensures that the version being released doesn't exist yet, and that it is created from the correct branch (i.e. version `X.Y.Z` has to be released from the `vX.Y` branch). The tag created for the release is signed with the same key that we use for the PHP extension.

The static analysis workflow has been changed to also run for tags. The resulting SARIF report is uploaded to GitHub's Code Scanning feature where we can then handle it further.

Todo list:
- [x] Wait for https://github.com/mongodb/mongo-php-driver/pull/1544 to be merged
- [x] Extract garasign actions to separate repository?
- [ ] Add information on verifying release integrity

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -80,24 +80,21 @@ jobs:
           git config user.name "${GIT_AUTHOR_NAME}"
           git config user.email "${GIT_AUTHOR_EMAIL}"
 
-      # Create a draft release with a changelog
-      # TODO: Consider using the API to generate changelog
-      - name: "Create draft release with generated changelog"
-        run: gh release create ${{ inputs.version }} --target ${{ github.ref_name }} --generate-notes --draft
-
-      - name: "Read changelog from draft release"
-        run: gh release view ${{ inputs.version }} --json body --template '{{ .body }}' >> changelog
-
-      - name: "Prepare tag message"
+      # Create a draft release with release message filled in
+      - name: "Prepare release message"
         run: |
-          echo -e "Release ${PACKAGE_VERSION}\n" > tag-message
-          cat changelog >> tag-message
+          cat > release-message <<'EOL'
+          ${{ format(env.default-release-message, inputs.version, inputs.jira-version-number) }}
+          EOL
+
+      - name: "Create draft release"
+        run: echo "RELEASE_URL=$(gh release create ${{ inputs.version }} --target ${{ github.ref_name }} --title "${{ inputs.version }}" --notes-file release-message --draft)" >> "$GITHUB_ENV"
 
       # This step creates the signed release tag
       - name: "Create release tag"
         uses: mongodb-labs/drivers-github-tools/garasign/git-sign@main
         with:
-          command: "git tag -F tag-message -s --local-user=${{ vars.GPG_KEY_ID }} ${{ inputs.version }}"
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,129 @@
...
+
+      - name: "Read changelog from draft release"
+        run: gh release view ${{ inputs.version }} --json body --template '{{ .body }}' >> changelog
+
+      # This step creates the signed release tag
+      - name: "Create release tag"
+        uses: ./.github/actions/garasign/git-sign
+        with:
+          command: "git tag -F changelog -s --local-user=${{ vars.GPG_KEY_ID }} ${{ inputs.version }}"
+          garasign_username: ${{ secrets.GRS_CONFIG_USER1_USERNAME }}
```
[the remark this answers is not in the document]
alcaeus: Secrets added.

.github/workflows/release.yml
```diff
@@ -0,0 +1,134 @@
...
+        required: true
+        type: "string"
+      jira-version-number:
+        description: "JIRA version ID (e.g. 54321)"
+        required: true
+        type: "string"
+
+env:
+  # TODO: Use different token
+  GH_TOKEN: ${{ secrets.MERGE_UP_TOKEN }}
```
jmikola, line 18: I noticed this TODO is still present in PHPC (https://github.com/mongodb/mongo-php-driver/blob/0ced119f9d95da5d1c1a06fe412f9d26bce982e8/.github/workflows/release.yml#L18). Should there be a JIRA ticket to track this?

.github/workflows/release.yml
```diff
@@ -0,0 +1,134 @@
...
+
+      # Create a draft release with a changelog
+      # TODO: Consider using the API to generate changelog
+      - name: "Create draft release with generated changelog"
+        run: gh release create ${{ inputs.version }} --target ${{ github.ref_name }} --generate-notes --draft
+
+      - name: "Read changelog from draft release"
+        run: gh release view ${{ inputs.version }} --json body --template '{{ .body }}' >> changelog
+
+      - name: "Prepare tag message"
```
jmikola, line 91: Going back to PHPC's 1.19 release, did we reach a conclusion about not wanting to embed the generated changelog in the tag message? I know we ended up pulling it out of the release notes.

.github/workflows/release.yml
alcaeus: The idea is to use a token for a GitHub app, which makes it easier to allow bypassing branch protection rules. Tracking in PHPC-2383

.github/workflows/release.yml
alcaeus: I haven't had a chance to update this pull request due to Skunkworks. I've removed the automated changelog, so the release workflow now creates the release with our release message and only uses `Release X.Y.Z` as the tag message. I'll apply those changes to PHPC as well, once I've figured out how to best handle the changelog needed for PECL packaging.

jmikola (approved):

PR #1289, opened 2024-05-07
DOCSP-37027: Fix build errors

norareidy:
JIRA - https://jira.mongodb.org/browse/DOCSP-37027
Staging - https://preview-mongodbnorareidy.gatsbyjs.io/php-library/DOCSP-37027-build-errors/
Build log - https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=663a527fb67dbf804ac387bb


docs/reference/method/MongoDBCollection-dropIndexes.txt
```diff
@@ -27,8 +27,7 @@ Parameters
 
 ``$indexName`` : string| :phpclass:`MongoDB\Model\IndexInfo`
   The name or model object of the index to drop. View the existing indexes on
-  the collection using the :phpmethod:`listIndexes()
-  <MongoDB\Collection::listIndexes>` method.
+  the collection using the :phpmethod:`MongoDB\Collection::listIndexes()` method.
```
jordan-smith721, line 30:
```suggestion
  the collection by using the :phpmethod:`MongoDB\Collection::listIndexes()` method.
```

jordan-smith721 (approved):
LGTM!

docs/includes/extracts-option-requires.yaml
```diff
@@ -1,48 +1,43 @@
 ---
 ref: option-requires-4.2
-source:
-  ref: option-requires-version
-  file: extracts-option-requires.yaml
```
jmikola, line 5: I was under the impression that both extracts and replacements were supported in Snooty. See slide 49 of [Marking up: Tech Writer Training on rST and YAML](https://docs.google.com/presentation/d/125-MvuDr66EIUgb3bVWCX3Uh2Jj2MkM8kM0iM_JoOuA/edit). Is this not the case, or is there some other limitation here (e.g. a file referencing itself)?

If you're going to remove the inheritance here, I see no reason to preserve the `replacement` block. The version number might as well just be hard-coded into the content. But before doing that, I would like to address my first question.

docs/reference/method/MongoDBCollection-explain.txt
```diff
@@ -29,7 +29,7 @@ Definition
 Parameters
 ----------
 
-``$explainable`` : :phpclass:`MongoDB\Operation\Explainable`
+``$explainable`` : ``MongoDB\Operation\Explainable``
```
jmikola, line 32: Noted that we don't have a stub for the Explainable interface in the docs (nor do we need one), so code formatting is the right call here.

docs/reference/method/MongoDBDatabase-createCollection.txt
```diff
@@ -296,7 +296,7 @@ Parameters
          rules or expressions. You can specify the expressions using the same
          operators as MongoDB's
          :manual:`query operators </reference/operator/query>` with the
-         exception of :query:`$geoNear`, :query:`$near`, :query:`$nearSphere`,
+         exception of :query:`$near`, :query:`$nearSphere`,
```
jmikola, line 299: I assume this dates back to some earlier version of the docs (e.g. 3.2), but I confirmed `$geoNear` is no longer mentioned in the [create](https://www.mongodb.com/docs/manual/reference/command/create/) docs.

docs/tutorial/encryption.txt
```diff
@@ -19,16 +19,15 @@ To get started using in-use encryption in your project, the
 to be compiled with `libmongocrypt <https://github.com/mongodb/libmongocrypt>`_
 (enabled by default).
 
-Additionally, either `crypt_shared`_ or `mongocryptd`_ are required in order to
```
jmikola, line 22: I believe these might have been links at some point, which would explain the use of a trailing underscore. In any event, they're no longer defined.

I will point out that this docs page inconsistently uses code-formatting for crypt_shared and mongocryptd. I'm not sure what the server manual does, but we could probably stand to make this internally consistent. Feel free to make this consistent with whatever you do in docs beyond PHPLIB. I also have no objections to just removing all of the backticks for these terms.

docs/tutorial/gridfs.txt
```diff
@@ -35,7 +35,7 @@ The bucket can be constructed with various options:
   needed. The default size is ``261120`` (i.e. 255 KiB).
 - ``readConcern``, ``readPreference`` and ``writeConcern`` options can be used
   to specify defaults for read and write operations, much like the
-  :phpclass:`MongoDB\GridFS\Collection` options.
```
jmikola, line 38: Good catch.

jmikola (commented):
No big concerns, although I have a question about `extracts-options-requires.yaml`.

I noticed this was opened against the `master` branch. Should it apply further back, or do you plan to manually backport to earlier, deployed branches?

docs/includes/extracts-option-requires.yaml
```diff
@@ -1,43 +1,48 @@
...
-content: |
-  This option is available since MongoDB {{version}} and will result in an
-  exception at execution time if specified for an older server version.
+source:
+  file: extracts-option-requires.yaml
+  ref: option-requires-version
 replacement:
   version: "7.0"
-...
+...
```
jmikola, line 48: Not sure if this is needed, but the original file (and most) have a trailing newline character.

docs/includes/extracts-option-requires.yaml
norareidy: Ok I switched it back to using inheritance - I changed "source:" to "inherit:" and that fixed the build issues. I think that rST / YAML powerpoint might need to be updated, I can investigate why "inherit" fixes the problem.

docs/tutorial/encryption.txt
norareidy: Looks like the Server manual formats them in monospace ([like on this page](https://www.mongodb.com/docs/v6.2/core/queryable-encryption/reference/mongocryptd/)), so I monospaced all instances of crypt_shared and mongocryptd

docs/includes/extracts-option-requires.yaml
norareidy: Never mind, seems like the issue was the missing _ before "option-requires-version"

norareidy:
> No big concerns, although I have a question about `extracts-options-requires.yaml`.
> 
> I noticed this was opened against the `master` branch. Should it apply further back, or do you plan to manually backport to earlier, deployed branches?

My mistake, I meant to target this against the v1.16 branch. I'll open another PR on that branch.

jmikola (approved):
The corresponding v1.16 PR (#1290) is blocked on CI, so let's wait a bit. If we can merge that first (and up to v1.17 and master), it'd be preferable to rebase this down to a smaller diff.

PR #1291, opened 2024-05-08
Fix build

alcaeus:
This PR makes some changes to the build system to ensure pull requests to older branches can be merged.


jmikola (approved):
Looking at https://github.com/mongodb/docs-php-library/commit/f6762025d8c57184652139be83f5ce4135017693, I think the Snooty migration was applied to v1.16.

The v1.17 branch still includes https://github.com/mongodb/mongo-php-library/commit/4d2132672922582082550adc64e2dd355f714eef, which was meant to allow running Giza through ubuntu-20.04. Correct me if I'm wrong, but that's probably obsolete as well and we should remove `docs.yml` from all branches above v1.16 when merging this up.

alcaeus:
> we should remove docs.yml from all branches above v1.16 when merging this up.

Yes, my plan was to merge this up manually until the merge-up action picks it up.

PR #1292, opened 2024-05-08
PHPLIB-1447: Add SBOM lite

alcaeus:
PHPLIB-1447

Since we don't have any bundled dependencies, we can add an empty SBOM file

### DIFF
--- a/sbom.json
+++ b/sbom.json
@@ -1,0 +1,85 @@
+{
+    "$schema": "http://cyclonedx.org/schema/bom-1.5.schema.json",
+    "bomFormat": "CycloneDX",
+    "specVersion": "1.5",
+    "serialNumber": "urn:uuid:dc42a43b-4ace-4c42-9a6e-0b9e28fdd100",
+    "version": 1,
+    "metadata": {
+        "timestamp": "2024-05-08T09:51:01Z",
+        "tools": [
+            {
+                "name": "composer",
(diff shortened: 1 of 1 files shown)


sbom.json
```diff
@@ -0,0 +1,85 @@
...
+    "metadata": {
+        "timestamp": "2024-05-08T09:51:01Z",
+        "tools": [
+            {
+                "name": "composer",
+                "version": "2.7.6"
+            },
+            {
+                "vendor": "cyclonedx",
+                "name": "cyclonedx-php-composer",
```
jmikola, line 16: This file looks generated. Where did you get it from?

jmikola (approved):
I'll trust this is the correct format for an empty SBOM file, but I'm curious how it was generated. Ideally, we should make a note of that either in this PR or PHPLIB-1447 in case we ever need to refer back to this.

sbom.json
alcaeus: The file was generated using the cyclonedx/cyclonedx-php-composer library to generate an SBOM from composer.lock, then manually removing all components and dependencies as we only report bundled dependencies:
```shell
composer require --dev cyclonedx/cyclonedx-php-composer
composer CycloneDX:make-sbom --output-file=sbom.json --output-format=JSON --omit=dev --omit=plugin
```

PR #1335, opened 2024-06-04
PHPLIB-1279: Revise driver/extension terminology

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1279
https://jira.mongodb.org/browse/PHPLIB-1466

This updates references where we're discussing something specific to the extension. I've left "driver" in place for more general discussion about the PHP driver or for special topics such as "driver options" on Manager and Client objects.

I've targeted v1.17 for this as it's the oldest branch that doesn't conflict. I think it'd be reasonable to drop v1.16 entirely from https://www.mongodb.com/docs/php-library/current/ since we've had new releases following the original Giza -> Snooty port.

### DIFF
--- a/docs/reference/exception-classes.txt
+++ b/docs/reference/exception-classes.txt
@@ -53,7 +53,7 @@ MongoDB\\Exception\\InvalidArgumentException
    Thrown for errors related to the parsing of parameters or options within the
    library.
 
-   This class extends the driver's :php:`InvalidArgumentException
+   This class extends the extension's :php:`InvalidArgumentException
    <mongodb-driver-exception-invalidargumentexception>` class and implements the
    library's :phpclass:`Exception <MongoDB\Exception\Exception>` interface.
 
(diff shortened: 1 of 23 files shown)


docs/reference/exception-classes.txt
```diff
@@ -67,10 +67,9 @@ MongoDB\\Exception\\UnexpectedValueException
    This exception is thrown when a command response from the server is
    malformed or not what the library expected. This exception means that an
    assertion in some operation, which abstracts a database command, has failed.
-   It may indicate a corrupted BSON response or bug in the server, driver, or
-   library.
+   It may indicate a corrupted BSON response or bug in the server or driver.
```
bisht2050, line 70: Is driver referring to the extension, library or extension+library in this line?

bisht2050 (approved):
LGTM, with one minor clarification.

alcaeus (approved):
LGTM!

> I've targeted v1.17 for this as it's the oldest branch that doesn't conflict. I think it'd be reasonable to drop v1.16 entirely

I'd be very much in favour to drop old versions from the docs, especially if they conflict or need manual merges.

docs/reference/exception-classes.txt
jmikola: Any remaining references to "driver" should be taken to mean extension and/or library, where there is no need to differentiate.

norareidy (approved):
LGTM

alcaeus:
We can ignore rector in this branch.

PR #1342, opened 2024-06-10
PHPLIB-1434: Publish SSDLC assets upon release

alcaeus:
PHPLIB-1434

This PR changes the release workflow to publish all SSDLC assets. Since we need to produce a SARIF export of all code scanning alert, the "Static Analysis" workflow is manually invoked ([see docs](https://docs.github.com/en/actions/using-workflows/reusing-workflows)) instead of being automatically run for tags. This ensures that when we get to the `publish-ssdlc-assets` job the psalm run will have completed and analysis of the SARIF report uploaded to Code Scanning is done. This couldn't be guaranteed if static analysis was run on the tag directly.

In the `publish-ssdlc-assets` job, we generate the document for authorized publication, download the SBOM file from silk, create a SARIF export of all open code scanning alerts, generate the SSDLC compliance report, and upload the whole thing to our S3 release assets bucket. Merging this pull request will conclude the SSDLC effort in PHPLIB.

This PR depends on the following pull requests being merged:
- [ ] https://github.com/mongodb-labs/drivers-github-tools/pull/25
- [ ] https://github.com/mongodb-labs/drivers-github-tools/pull/29
- [ ] https://github.com/mongodb-labs/drivers-github-tools/pull/30

Furthermore, there currently is an issue with Silk that causes the SBOM download to fail. Once that issue is resolved, I can test the workflow again with the SBOM steps enabled. The rest was already tested and confirmed to work.


.github/workflows/release.yml
```diff
@@ -119,11 +117,88 @@ jobs:
...
 
       - name: "Set summary"
         run: |
           echo '🚀 Created tag and drafted release for version [${{ inputs.version }}](${{ env.RELEASE_URL }})' >> $GITHUB_STEP_SUMMARY
           echo '✍️ You may now update the release notes and publish the release when ready' >> $GITHUB_STEP_SUMMARY
+
+  static-analysis:
+    needs: prepare-release
+    name: "Run Static Analysis"
+    uses: ./.github/workflows/static-analysis.yml
```
jmikola, line 131: By default, I assume this would invoke all jobs in the referenced file and that's why you added a conditional on the Rector job.

I noted that this is re-using an entire workflow ([`jobs.<job_id>.uses`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_iduses)), as opposed to an action being imported into a step ([`jobs.<job_id>.steps[*].uses`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsuses)).

I'm a bit confused about how this single `static-analysis` job would map to multiple jobs in `static-analysis.yml` if you hadn't specified a condition to exclude Rector. Would they just be run in sequence? [Reusing Workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows) doesn't address this, and the example therein only has a single job defined in the import file. To that end, I wonder if it'd make more sense to split the Psalm and Rector jobs into separate files (e.g. `static-analysis-psalm.yml` and `static-analysis-rector.yml`) if there isn't much commonality between them anyway.

.github/workflows/release.yml
```diff
@@ -119,11 +117,88 @@ jobs:
...
+
+      - name: "Upload SBOM as release artifact"
+        run: gh release upload ${{ inputs.version }} ${{ env.S3_ASSETS }}/cyclonedx.sbom.json
+        continue-on-error: true
+
+      - name: "Generate SARIF report from code scanning alerts"
+        uses: mongodb-labs/drivers-github-tools/code-scanning-export@v2
+        with:
+          ref: ${{ inputs.version }}
+          output-file: ${{ env.S3_ASSETS }}/code-scanning-alerts.json
```
jmikola, line 193: It's not clear to me where `env.S3_ASSETS` gets set. Does publishing here depend on the "Set up drivers-github-tools" task above, which is configured with the AWS secrets?

.github/workflows/static-analysis.yml
```diff
@@ -12,8 +12,12 @@ on:
       - "v*.*"
       - "master"
       - "feature/*"
-    tags:
-      - "*"
+  workflow_call:
+    inputs:
+      ref:
+        description: "The git reference to check"
```
jmikola, line 18: Nit: in this case, I think "git ref" is the more readable name, even though it's an abbreviation.

.github/workflows/release.yml
alcaeus: Yes - unlike actions that contain a set of functionality to be included in a job, this allows to reuse an entire workflow as a job in a different workflow, without sharing data with any other job.

As for what this looks like, I wasn't sure at first either but it actually looks rather nice: https://github.com/alcaeus/mongo-php-library/actions/runs/9447195881. I'll split the workflow up into separate workflows for psalm and rector.

.github/workflows/release.yml
alcaeus: This is set by the setup action above. I'll add a comment for this

jmikola (approved):

PR #1344, opened 2024-06-13
Use checkout and tag actions from drivers-github-tools

alcaeus:
This PR makes use of the `secure-checkout` and `tag-version` actions from drivers-github-tools to reduce the amount of logic in our release workflow.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -48,21 +48,11 @@ jobs:
       - name: "Create release output"
         run: echo '🎬 Release process for version ${{ inputs.version }} started by @${{ github.triggering_actor }}' >> $GITHUB_STEP_SUMMARY
 
-      - name: "Create temporary app token"
-        uses: actions/create-github-app-token@v1
-        id: app-token
+      - name: "Generate token and checkout repository"
+        uses: mongodb-labs/drivers-github-tools/secure-checkout@v2
         with:
-          app-id: ${{ vars.APP_ID }}
-          private-key: ${{ secrets.APP_PRIVATE_KEY }}
-
-      - name: "Store GitHub token in environment"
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -114,12 +105,6 @@ jobs:
       # 3. push next branch
       # 4. switch back to release branch, then push
 
-      - name: "Push changes from release branch"
-        run: git push
-
-      - name: "Push release tag"
-        run: git push origin ${{ inputs.version }}
```
jmikola, line 121: I assume these two steps are handled entirely by the `git push origin --tags` line in [tag-version/action.yml](https://github.com/mongodb-labs/drivers-github-tools/blob/7edc545fcf00ff1da4910079b7488c81d7c583b0/tag-version/action.yml#L54).

.github/workflows/release.yml
```diff
@@ -146,21 +131,12 @@ jobs:
...
-        id: app-token
-        with:
-          app-id: ${{ vars.APP_ID }}
-          private-key: ${{ secrets.APP_PRIVATE_KEY }}
-
-      - name: "Store GitHub token in environment"
-        run: echo "GH_TOKEN=${{ steps.app-token.outputs.token }}" >> "$GITHUB_ENV"
-        shell: bash
-
-      - uses: actions/checkout@v4
```
jmikola, line 160: Noted that this lines up with [secure-checkout/action.yml](https://github.com/mongodb-labs/drivers-github-tools/blob/7edc545fcf00ff1da4910079b7488c81d7c583b0/secure-checkout/action.yml).

jmikola (approved):

.github/workflows/release.yml
alcaeus: That is correct

PR #1346, opened 2024-06-14
Fix refs used for code scanning results

alcaeus:
There are two changes here:
- When invoking the static analysis job from the release job, the `github.ref` variable is set to the branch that the release workflow was triggered in. This is changed to the tag reference we just created.
- The code scanning export action was invoked with the same `github.ref`, which was also changed to using the ref of the newly created tag.

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -34,6 +34,11 @@ jobs:
         with:
           ref: ${{ github.event_name == 'workflow_dispatch' && inputs.ref || github.ref }}
 
+      - name: "Get SHA hash of checked out ref"
+        if: ${{ github.event_name == 'workflow_dispatch' }}
+        run: |
+          echo CHECKED_OUT_SHA=$(git rev-parse HEAD) >> $GITHUB_ENV
+
       - name: "Setup"
         uses: "./.github/actions/setup"
         with:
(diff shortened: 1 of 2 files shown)


jmikola (approved):

.github/workflows/static-analysis.yml
```diff
@@ -47,3 +52,19 @@ jobs:
         uses: "github/codeql-action/upload-sarif@v3"
         with:
           sarif_file: psalm.sarif
+          ref: ${{ github.event_name == 'workflow_dispatch' && inputs.ref || github.ref }}
+          sha: ${{ github.event_name == 'workflow_dispatch' && env.CHECKED_OUT_SHA || github.sha }}
+
+      - name: "Upload SARIF report"
+        if: ${{ github.event_name != 'workflow_dispatch' }}
```
alcaeus, line 59: I needed to split this, as the upload-sarif action contains logic to fix the `refs/pulls/<number>/merge` ref, but doesn't do so when it was provided explicitly. Thus the only way to retain current logic for pull requests is to invoke the action without `ref` and `sha` arguments, and then explicitly pass them when the workflow was dispatched manually.

jmikola (approved):

alcaeus:
Note: ignored failing evergreen variants, as they're related to a newer PHPC version.

PR #1357, opened 2024-07-02
PHPLIB-1450: Test with MonogDB 8.0 on Debian 12

alcaeus:
PHPLIB-1450

This adds Debian 12 to the build matrix to test PHP 8.0 and restricts testing with newer PHPC versions to the latest PHP version, similar to how it's handled with PHPC testing newer libmongoc versions.

The skipped spec tests depend on PHPLIB-1458 which is currently in progress but may require changes to libmongoc to be fully compatible, depending on the approach we choose.


alcaeus:
evergreen keep-definitions

alcaeus:
Note: the PR patch on Evergreen contains all 8.0 configurations that are available to ensure that everything passes.

GromNaN (approved):
Noted the split of `modernPhpVersions` and `legacyPhpVersions`.

PR #1358, opened 2024-07-04
PHPLIB-1375: Test KMIP "delegated" protocol

alcaeus:
PHPLIB-1375

I also included the changes for PHPLIB-1401, which make no difference as both test are skipped as we don't implement CSOT.

The patch definition for this pull request includes CSFLE tests with mongocryptd and crypt_shared for all supported server versions.


alcaeus:
evergreen keep-definitions

GromNaN (approved):

PR #1360, opened 2024-07-16
PHPLIB-1485: Drop support for MongoDB 3.6

alcaeus:
PHPLIB-1485

With the extension dropping support for MongoDB 3.6 by bumping its `minWireVersion` (see https://github.com/mongodb/mongo-php-driver/pull/1610), this PR removes MongoDB 3.6 from the test matrix and removes wire version checks related to MongoDB 3.6 as we know we'll never be connected to a 3.6 server.


GromNaN (commented):
What is the behavior of the driver if someone tries to use it with MongoDB 3.4?

GromNaN (approved):
LGTM. You only remove exceptions and let the connection error happen.

alcaeus:
> What is the behavior of the driver if someone tries to use it with MongoDB 3.4?

Note that 3.4 has been dropped a while ago, this PR drops support for 3.6. libmongoc will throw an error about the wire version not being supported:
> Server at <host> reports wire version <version>, but this version of libmongoc requires at least 7 (MongoDB 4.0)

PR #1365, opened 2024-07-23
Fix build

alcaeus:
This PR removes serverless tests on Debian 11 as our serverless test clusters run MongoDB 8.0. It also removes MongoDB rapid and latest from testing on Debian 11 which was left in when I refactored the build configuration.

This removes all test failures on master, except for the serverless proxy tests where a single retryable reads test fails. I'll investigate that separately as we may end up removing those tests entirely.


GromNaN (approved):
So serverless is no longer tested?

alcaeus:
> So serverless is no longer tested?

Serverless will continue to be tested, but PHPLIB-1484 calls for the serverless proxy tests to be removed.

PR #1366, opened 2024-07-25
PHPLIB-1489: Handle CursorId deprecations

alcaeus:
PHPLIB-1489

This PR complements https://github.com/mongodb/mongo-php-driver/pull/1616 by introducing the same `$asInt64` argument to all classes that implement `CursorInterface::getId` or wrap a cursor class. Not providing the argument or providing `false` as a value will trigger a deprecation warning to inform the user they need to update. In 2.0, we can deprecate the parameter and change the return type of the methods to `Int64`.

### DIFF
--- a/src/Model/CodecCursor.php
+++ b/src/Model/CodecCursor.php
@@ -19,17 +19,20 @@ namespace MongoDB\Model;
 
 use Iterator;
 use MongoDB\BSON\Document;
+use MongoDB\BSON\Int64;
 use MongoDB\Codec\DocumentCodec;
 use MongoDB\Driver\Cursor;
 use MongoDB\Driver\CursorId;
 use MongoDB\Driver\CursorInterface;
 use MongoDB\Driver\Server;
+use ReturnTypeWillChange;
 
 use function assert;
 use function iterator_to_array;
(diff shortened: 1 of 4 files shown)


src/Model/CodecCursor.php
```diff
@@ -73,9 +76,23 @@ public static function fromCursor(Cursor $cursor, DocumentCodec $codec): self
         return new self($cursor, $codec);
     }
 
-    public function getId(): CursorId
+    /** @return CursorId|Int64 */
+    #[ReturnTypeWillChange]
+    public function getId(bool $asInt64 = false)
```
GromNaN, line 81: For BC layer, adding an argument should be virtual (`func_get_arg`).

GromNaN (dismissed):

src/Model/CodecCursor.php
alcaeus: I thought the argument with a default value would be fine, can you explain why `func_get_args` works better?

GromNaN: Although I'm not sure why or how anyone would extend this class.

GromNaN (approved):

PR #1370, opened 2024-08-11
DOP-4904: Remove docs/ from v1.18

i80and:
This is a companion PR to https://github.com/mongodb/docs-php-library/pull/96.


alcaeus:
Closing as we won't be removing docs from old branches.

PR #1374, opened 2024-09-04
PHPLIB-955: Require PHP 8.1

alcaeus:
PHPLIB-955

The first commit updates the requirement in composer.json and updates CI configs. In the second commit, I applied all automated fixes from phpcbf. This mainly affects the use of union types as well as the `mixed` type.

A couple of notes on phpcs rules:
* The "DisallowedShortNullable" rule was disabled, as it would change all `?Type` types to `Type|null`. I prefer the short syntax over the consistency with other union types (e.g. `Type|Other|null`)
* The rule requiring constructor property promotion was disabled as it caused some invalid syntax, for example by adding a default value to a constructor parameter when other parameters after it don't have a default value. I'll investigate and create a bug ticket for this.


src/Codec/DecodeIfSupported.php
```diff
@@ -25,27 +25,22 @@
...
 
     /**
-     * @param mixed $value
      * @psalm-param BSONType $value
      * @return mixed
      * @psalm-return NativeType
      * @throws UnsupportedValueException if the decoder does not support the value
      */
-    abstract public function decode($value);
+    abstract public function decode(mixed $value);
```
GromNaN, line 37: Confirmed this is *not* a breaking change for classes that override this abstract method. https://3v4l.org/donQG

I don't get why Symfony BC promise prohibit adding typehint to arguments in classes. https://symfony.com/doc/current/contributing/code/bc.html#changing-classes

src/GridFS/StreamWrapper.php
```diff
@@ -54,8 +54,7 @@ class StreamWrapper
     /** @var resource|null Stream context (set by PHP) */
     public $context;
 
-    /** @var ReadableStream|WritableStream|null */
-    private $stream;
+    private ReadableStream|WritableStream|null $stream = null;
```
GromNaN, line 57: Adding the type on properties produces an error in case of access to uninitialized property. That's why you added the default null value here.

src/Exception/UnsupportedValueException.php
```diff
@@ -24,29 +24,25 @@
 
 class UnsupportedValueException extends InvalidArgumentException implements Exception
 {
-    /** @var mixed */
-    private $value;
+    private mixed $value;
```
GromNaN, line 27: You should add CPP for this one as you are already changing the property definition and the contructor.

alcaeus:
Marking this ready for review. I'll revisit the failing tests shortly, as I suspect those will be fixed by #1373

tests/TestCase.php
```diff
@@ -250,7 +243,8 @@ protected function getInvalidBooleanValues(bool $includeNull = false): array
      */
     protected function getInvalidDocumentValues(bool $includeNull = false): array
     {
-        return array_merge([123, 3.14, 'foo', true, PackedArray::fromPHP([])], $includeNull ? [null] : []);
+        return array_merge([123, 3.14, 'foo', true], $includeNull ? [null] : []);
+//        return array_merge([123, 3.14, 'foo', true, PackedArray::fromPHP([])], $includeNull ? [null] : []);
```
GromNaN, line 247: You left this commented.

tests/UnifiedSpecTests/EventObserver.php
```diff
@@ -227,7 +226,7 @@ private function assertEvent($actual, stdClass $expected, string $message)
     {
         assertIsObject($actual);
 
-        switch (get_class($actual)) {
+        switch ($actual::class) {
```
GromNaN, line 229: Use `match` operator here.

tests/Operation/MapReduceTest.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
```
alcaeus, line 3: This directive is necessary to disable type coercion in `testConstructorOutArgumentTypeCheck`; all of the invalid values provided there can be coerced to strings.

GromNaN: This means that the validation is weaker than before, since the values were not cast. Do we consider that it's the user's responsibility to send the right type and use strict mode if they wish to avoid this kind of error? In any case, it won't work with an int casted to string.

src/Codec/DecodeIfSupported.php
alcaeus: Correct. Without a type specified for an argument, people can't add their own, narrower types. When we add a type, people are still allowed to widen types or omit the type entirely, which is why we can add parameter types in classes and interfaces in minor versions.

src/GridFS/StreamWrapper.php
alcaeus: That's correct 👍

src/Exception/UnsupportedValueException.php
alcaeus: I enabled the CPP rule and fixed the two mistakes that the automatic fixer made. CPP is now used wherever possible.

tests/TestCase.php
alcaeus: Whoops, removed.

For context, with the `object|array` type now enforced for documents, `PackedArray` instances no longer cause an error. I could bring back the `PackedArray` check if we want to be more specific about what objects are allowed, but I defer to @jmikola on this.

tests/UnifiedSpecTests/EventObserver.php
alcaeus: Updated. I also removed the `return` statements as the methods being called don't return a value. Still, the `match` operator has a much more concise syntax than `switch`, so I think this is an improvement.

psalm-baseline.xml
```diff
@@ -123,19 +123,24 @@
...
+      <code><![CDATA[$value]]></code>
     </MixedAssignment>
     <MoreSpecificReturnType>
-      <code>stdClass</code>
+      <code><![CDATA[stdClass]]></code>
     </MoreSpecificReturnType>
   </file>
+  <file src="src/Model/BSONIterator.php">
+    <MixedArgument>
+      <code><![CDATA[$this->options['typeMap']]]></code>
```
GromNaN, line 135: Should we define array shape for the options?

GromNaN (approved):
LGTM

tests/Operation/MapReduceTest.php
alcaeus: Before this change, users could pass an int which we'd have to guard against. With the `string` type enforced by PHP, there are two scenarios: if the calling file has `strict_types` disabled, an int would be coerced to string; with `strict_types` enabled a `TypeError` is thrown. The end result is the same: the method will not be called with a non-string parameter.

psalm-baseline.xml
alcaeus: We had that. However, in the `param` declaration, it would have to be `array{typeMap?: array, ...}`, whereas for the property it's `array{typeMap: array, ...}`. With promoted properties, the same type is used for the parameter and property, leading to this error. Instead of adding custom logic to work around this, I decided to add the error to the baseline.

src/GridFS/StreamWrapper.php
jmikola: When does PHP raise an error for the uninitialized property? I'm curious if the error is only triggered by explicitly assigning `null` in the property declaration. If so, would you rather do without that and omit `|null` from the type?

src/Client.php
```diff
@@ -397,7 +397,7 @@ private static function getVersion(): string
         if (self::$version === null) {
             try {
                 self::$version = InstalledVersions::getPrettyVersion('mongodb/mongodb') ?? 'unknown';
-            } catch (Throwable $t) {
+            } catch (Throwable) {
```
jmikola, line 400: Noted that the variable is no longer required as of PHP 8.0 (see: [Exceptions[(https://www.php.net/manual/en/language.exceptions.php)).

tests/TestCase.php
jmikola: Isn't the PackedArray check still in place? I saw you removed various conditionals when adding `array|object` type hints, but `is_document()` still rejects PackedArray instances. I would expect that that `InvalidArgumentException::expectedDocumentType()` still gets thrown for PackedArray instances.

I assume changing `InvalidArgumentException` expectations to `TypeError` is what conflicts here. I don't feel strongly about this, but I'd argue that parameter and option type checks shouldn't necessarily be expected to raise `TypeError`. There are still various other contexts where we raise `InvalidArgumentException` for validation errors (e.g. option value is out of range, or a string value doesn't satisfy our homemade enum).

src/GridFS/StreamWrapper.php
alcaeus: If the type was only `ReadableStream|WritableStream` without a default value, accessing `$this->stream` would cause a fatal error. I believe we could theoretically use that instead of `assert` in a number of places, as even with disabled assertion handling any access to the stream would result in an object access on `null`, which again yields a fatal error.

I'll note that `isset` does not trigger the uninitialised property error, so we would still be able to check if a stream was initialised. I can change this if y'all prefer to leave `null` out of the type.

tests/TestCase.php
alcaeus: It is indeed. I've updated the type assertions based on the value given to expect a `TypeError` or `InvalidArgumentException` and restored the `PackedArray` instance as an invalid document, as it adds confidence that we have the right checks in place.

With `TypeError` being a thing, I wonder if we should consider throwing `TypeError` throwables instead of `InvalidArgumentException` on type mismatches.

src/GridFS/StreamWrapper.php
jmikola: Feel free to leave as-is. Seems like the path of least resistance.

tests/TestCase.php
jmikola: > With TypeError being a thing, I wonder if we should consider throwing TypeError throwables instead of InvalidArgumentException on type mismatches.

That sounds like a great idea, now that we can reliably expect TypeErrors from PHP itself. Might be something to consider for both PHPC and PHPLIB, and then we can also remove the vapid documentation for throwing InvalidArgumentException from most things.

jmikola (approved):

PR #1381, opened 2024-09-09
PHPLIB-1506 Import `mongodb/builder`

GromNaN:
Fix PHPLIB-1506

- Import files from https://github.com/mongodb/mongo-php-builder
- Add `replace` rule in `composer.json`
- Add a GitHub Action to validate the generated files are up-to-date


composer.json
```diff
@@ -26,6 +26,9 @@
         "symfony/phpunit-bridge": "^5.2",
         "vimeo/psalm": "^5.13"
     },
+    "replace": {
+        "mongodb/builder": ">= 0.2"
```
alcaeus, line 30: Since the library includes all version of the builder and we never really advertised it, I think we should be able to replace all versions here. WDYT?
```suggestion
        "mongodb/builder": "*"
```

.github/workflows/generator.yml
```diff
@@ -0,0 +1,50 @@
...
+    branches:
+      - "v*.*"
+      - "master"
+      - "feature/*"
+  workflow_call:
+    inputs:
+      ref:
+        description: "The git ref to check"
+        type: string
+        required: true
```
alcaeus, line 20: This is only necessary if we're calling this from a different workflow (e.g. in the release workflow). Since we're not doing that, I suggest removing this section.
```suggestion
```

.github/workflows/generator.yml
```diff
@@ -0,0 +1,50 @@
...
+jobs:
+  psalm:
+    name: "Diff check"
+    runs-on: "ubuntu-22.04"
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v4"
+        with:
+          ref: ${{ github.event_name == 'workflow_dispatch' && inputs.ref || github.ref }}
```
alcaeus, line 37: If you remove the `workflow_call` trigger above, you can also remove this part:
```suggestion
```

alcaeus (commented):
Reviewed the tooling changes, but did not review all the config and generator logic. I trust you copied that over from the builder project without changes. Just have a couple of minor comments, but in general this looks good to go.

composer.json
GromNaN: Ok, that's fine.

.github/workflows/generator.yml
```diff
@@ -0,0 +1,44 @@
...
+jobs:
+  psalm:
+    name: "Diff check"
+    runs-on: "ubuntu-22.04"
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v4"
+        with:
+          fetch-depth: 2
```
alcaeus, line 31: Note that fetch-depth 2 isn't necessary - the default value of 1 to fetch the single commit being tested is sufficient for this action. We use `fetch-depth: 0` in some of the release or merge actions so that we have access to the full history, but since that's not necessary here you can most likely remove this.
```suggestion
```

alcaeus (approved):
Minor comment regarding `fetch-depth`, but LGTM!

PR #1382, opened 2024-09-09
PHPLIB-1505 Add driver option `builderEncoder`

GromNaN:
Fix PHPLIB-1505

The new option `builderEncoder` contains a `MongoDB\Codec\Encoder` object that will be used to encode queries and pipelines. When not set, it defaults to a new instance of `MongoDB\Builder\BuilderEncoder`.


tests/ClientTest.php
```diff
@@ -129,11 +138,19 @@ public function testGetSelectsDatabaseAndInheritsOptions(): void
     {
         $uriOptions = ['w' => WriteConcern::MAJORITY];
 
-        $client = new Client(static::getUri(), $uriOptions);
+        $driverOptions = [
+            'builderEncoder' => $builderEncoder = $this->createMock(Encoder::class),
+            'typeMap' => ['root' => 'array'],
```
GromNaN, line 143: typeMap option was not tested here.

tests/TestCase.php
```diff
@@ -209,23 +208,28 @@ protected function getCollectionName(): string
      */
     protected function getInvalidArrayValues(bool $includeNull = false): array
     {
-        return array_merge([123, 3.14, 'foo', true, new stdClass()], $includeNull ? [null] : []);
+        return [123, 3.14, 'foo', true, new stdClass(), ...($includeNull ? [null] : [])];
```
GromNaN, line 211: Reformatting data generator using array unpacking.

src/Client.php
```diff
@@ -111,9 +123,10 @@ public function __construct(?string $uri = null, array $uriOptions = [], array $
         $driverOptions['driver'] = $this->mergeDriverInfo($driverOptions['driver'] ?? []);
 
         $this->uri = $uri ?? self::DEFAULT_URI;
+        $this->builderEncoder = $driverOptions['builderEncoder'] ?? new BuilderEncoder();
```
GromNaN, line 126: Static analysis could be better at detecting the type here. Line 107, we assert the value is not set, `null` or an instance of `Encoder`.

src/Client.php
```diff
@@ -80,6 +85,9 @@ class Client
      *
      *  * typeMap (array): Default type map for cursors and BSON documents.
      *
+     *  * builderEncoder(MongoDB\Codec\Encoder): Encoder for converting aggregation
+     *    builder stages and operators into BSON.
```
GromNaN, line 89: Any suggestion to improve the description is welcome.

src/Client.php
```diff
@@ -78,6 +83,8 @@ class Client
      *
      * Supported driver-specific options:
      *
+     *  * builderEncoder (MongoDB\Builder\Encoder): Encoder for query builder
```
alcaeus, line 86: Suggestion to call out there's a default value being used:
```suggestion
     *  * builderEncoder (MongoDB\Builder\Encoder): Encoder for query and aggregation builders. If not given, the default encoder will be used.
```

If you accept this, it would make sense to apply this to the Database and Collection classes as well.

src/Client.php
```diff
@@ -67,6 +70,8 @@ class Client
 
     private array $typeMap;
 
+    private Encoder $builderEncoder;
```
alcaeus, line 73: PHP 8.1 allows us to make this readonly:
```suggestion
    private readonly Encoder $builderEncoder;
```

src/Collection.php
```diff
@@ -84,6 +86,8 @@ class Collection
 
     private const WIRE_VERSION_FOR_READ_CONCERN_WITH_WRITE_STAGE = 8;
 
+    private Encoder $builderEncoder;
```
alcaeus, line 89:
```suggestion
    private readonly Encoder $builderEncoder;
```

src/Database.php
```diff
@@ -61,6 +63,8 @@ class Database
 
     private const WIRE_VERSION_FOR_READ_CONCERN_WITH_WRITE_STAGE = 8;
 
+    private Encoder $builderEncoder;
```
alcaeus, line 66:
```suggestion
    private readonly Encoder $builderEncoder;
```

tests/TestCase.php
alcaeus: I feel bad for forgetting PHP has this feature now 😅

alcaeus (changes requested):
I'd suggest making the `encoder` property `readonly`, but LGTM other than that.

src/Client.php
alcaeus: Also, just thought of this after seeing the Psalm errors in the follow-up PR, but specifying a template for the Encoder might be helpful:
```suggestion
    /** @psalm-var Encoder<array|stdClass|Document|PackedArray, mixed> */
    private readonly Encoder $builderEncoder;
```

alcaeus (approved):

PR #1383, opened 2024-09-10
PHPLIB-1419 Encode Agg builder objects in Collection methods

GromNaN:
Fix PHPLIB-1419

~Requires https://github.com/mongodb/mongo-php-library/pull/1382~

The encoding is done in the `Collection` methods, not in the various `Operation` classes:
- No option to validate in each Operation class
- Very little code duplicate in each method, whereas handling in operation classes would require the documentation and validation of the new option in each class.
- In the future, it is still possible to move the encoding and add an option to the methods without breaking change.

Only the `BulkWrite` operation has the option, because the processing of each operation is already organised in this operation class.

The Collection methods does not accept the `builderEncoder` option for now. If a specific encoder is necessary for an operation, if can be called just before the operation:

```php
$pipeline = new Pipeline(...);

// Required for now, as we could not change the argument type. In 2.0 this will not be necessary.
$pipeline = iterator_to_array($pipeline);

$pipeline = (new CustomBuilderEncoder)->encode($pipeline);
$collection->aggregate($pipeline);
```

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -22,6 +22,7 @@ use Iterator;
 use MongoDB\BSON\Document;
 use MongoDB\BSON\PackedArray;
 use MongoDB\Builder\BuilderEncoder;
+use MongoDB\Builder\Pipeline;
 use MongoDB\Codec\Encoder;
 use MongoDB\Driver\ClientEncryption;
 use MongoDB\Driver\Exception\InvalidArgumentException as DriverInvalidArgumentException;
@@ -391,6 +392,12 @@ class Client
      */
     public function watch(array $pipeline = [], array $options = [])
     {
+        if (is_builder_pipeline($pipeline)) {
+            $pipeline = new Pipeline(...$pipeline);
+        }
+
+        $pipeline = $this->builderEncoder->encodeIfSupported($pipeline);
+
         if (! isset($options['readPreference']) && ! is_in_transaction($options)) {
             $options['readPreference'] = $this->readPreference;
         }
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -1110,6 +1110,7 @@ class Collection
         }
 
         $pipeline = $this->builderEncoder->encodeIfSupported($pipeline);
+
         $options = $this->inheritReadOptions($options);
         $options = $this->inheritCodecOrTypeMap($options);
 
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -23,6 +23,7 @@ use MongoDB\BSON\Document;
 use MongoDB\BSON\JavascriptInterface;
 use MongoDB\BSON\PackedArray;
 use MongoDB\Builder\BuilderEncoder;
+use MongoDB\Builder\Pipeline;
 use MongoDB\Codec\DocumentCodec;
 use MongoDB\Codec\Encoder;
 use MongoDB\Driver\CursorInterface;
@@ -223,6 +224,12 @@ class Collection
      */
     public function aggregate(array $pipeline, array $options = [])
     {
+        if (is_builder_pipeline($pipeline)) {
+            $pipeline = new Pipeline(...$pipeline);
+        }
+
+        $pipeline = $this->builderEncoder->encodeIfSupported($pipeline);
+
         $hasWriteStage = is_last_pipeline_operator_write($pipeline);
 
         $options = $this->inheritReadPreference($options);
@@ -1098,6 +1105,11 @@ class Collection
      */
     public function watch(array $pipeline = [], array $options = [])
     {
+        if (is_builder_pipeline($pipeline)) {
+            $pipeline = new Pipeline(...$pipeline);
+        }
+
+        $pipeline = $this->builderEncoder->encodeIfSupported($pipeline);
         $options = $this->inheritReadOptions($options);
         $options = $this->inheritCodecOrTypeMap($options);
 
--- a/src/functions.php
+++ b/src/functions.php
@@ -336,28 +336,17 @@ function is_pipeline(array|object $pipeline, bool $allowEmpty = false): bool
  */
 function is_builder_pipeline(array $pipeline): bool
 {
-    if (! $pipeline) {
+    if (! $pipeline || ! array_is_list($pipeline)) {
         return false;
     }
 
-    if (! array_is_list($pipeline)) {
-        return false;
-    }
-
-    $result = false;
     foreach ($pipeline as $stage) {
-        if (! is_array($stage) && ! is_object($stage)) {
-            return false;
-        }
-
-        if ($stage instanceof StageInterface) {
-            $result = true;
-        } elseif (! is_first_key_operator($stage)) {
-            return false;
+        if (is_object($stage) && $stage instanceof StageInterface) {
+            return true;
         }
     }
 
-    return $result;
+    return false;
 }
 
 /**
--- a/src/functions.php
+++ b/src/functions.php
@@ -21,6 +21,7 @@ use Exception;
 use MongoDB\BSON\Document;
 use MongoDB\BSON\PackedArray;
 use MongoDB\BSON\Serializable;
+use MongoDB\Builder\Type\StageInterface;
 use MongoDB\Driver\Exception\RuntimeException as DriverRuntimeException;
 use MongoDB\Driver\Manager;
 use MongoDB\Driver\ReadPreference;
@@ -325,6 +326,38 @@ function is_pipeline(array|object $pipeline, bool $allowEmpty = false): bool
     }
 
     return true;
+}
+
+/**
+ * Returns whether the argument is a list that contains at least one
+ * {@see StageInterface} object.
+ *
+ * @internal
+ */
+function is_builder_pipeline(array $pipeline): bool
+{
+    if (! $pipeline) {
+        return false;
+    }
+
+    if (! array_is_list($pipeline)) {
+        return false;
+    }
+
+    $result = false;
+    foreach ($pipeline as $stage) {
+        if (! is_array($stage) && ! is_object($stage)) {
+            return false;
+        }
+
+        if ($stage instanceof StageInterface) {
+            $result = true;
+        } elseif (! is_first_key_operator($stage)) {
+            return false;
+        }
+    }
+
+    return $result;
 }
 
 /**
--- a/tests/ClientFunctionalTest.php
+++ b/tests/ClientFunctionalTest.php
@@ -2,6 +2,9 @@
 
 namespace MongoDB\Tests;
 
+use MongoDB\Builder\Pipeline;
+use MongoDB\Builder\Query;
+use MongoDB\Builder\Stage;
 use MongoDB\Client;
 use MongoDB\Driver\BulkWrite;
 use MongoDB\Driver\Command;
@@ -13,6 +16,7 @@ use MongoDB\Model\DatabaseInfoIterator;
 
 use function call_user_func;
 use function is_callable;
+use function iterator_to_array;
 use function sprintf;
 
 /**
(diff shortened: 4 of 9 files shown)


tests/Collection/BuilderCollectionFunctionalTest.php
```diff
@@ -0,0 +1,270 @@
...
+    {
+        $result = $this->collection->updateOne(
+            Query::query(x: Query::lt(2)),
+            new Pipeline(
+                Stage::set(x: 3),
+            ),
+        );
+
+        $this->assertEquals(1, $result->getModifiedCount());
+    }
```
GromNaN, line 220: This test is failing in evergreen:

```
[2024/09/10 23:21:34.399] ERROR:  (MongoDB\Driver\Exception\BulkWriteException)
[2024/09/10 23:21:34.399] MongoDB\Tests\Collection\BuilderCollectionFunctionalTest::testUpdateWithPipeline
[2024/09/10 23:21:34.399] MongoDB\Driver\Exception\BulkWriteException: BSON field 'update.updates.u' is the wrong type 'array', expected type 'object'
[2024/09/10 23:21:34.399] /data/mci/b14552e6624c454125590fb9d49d9cc2/src/src/Operation/Update.php:191
[2024/09/10 23:21:34.399] /data/mci/b14552e6624c454125590fb9d49d9cc2/src/src/Operation/UpdateOne.php:107
[2024/09/10 23:21:34.399] /data/mci/b14552e6624c454125590fb9d49d9cc2/src/src/Collection.php:1070
[2024/09/10 23:21:34.399] /data/mci/b14552e6624c454125590fb9d49d9cc2/src/tests/Collection/BuilderCollectionFunctionalTest.php:212
```

tests/Collection/BuilderCollectionFunctionalTest.php
```diff
@@ -0,0 +1,270 @@
...
+            new Pipeline(
+                Stage::set(x: 3),
+            ),
+        );
+        $this->assertEquals(2, $result->getModifiedCount());
+
+        $result = $this->collection->find(Query::query(x: Query::eq(3)))->toArray();
+        $this->assertCount(2, $result);
+        $this->assertEquals(3, $result[0]->x);
+    }
```
GromNaN, line 249: This test is failing in evergreen:

```
[2024/09/10 23:21:34.461] ERROR:  (MongoDB\Driver\Exception\BulkWriteException)
[2024/09/10 23:21:34.461] MongoDB\Tests\Collection\BuilderCollectionFunctionalTest::testUpdateManyWithPipeline
[2024/09/10 23:21:34.461] MongoDB\Driver\Exception\BulkWriteException: BSON field 'update.updates.u' is the wrong type 'array', expected type 'object'
[2024/09/10 23:21:34.461] /data/mci/b14552e6624c454125590fb9d49d9cc2/src/src/Operation/Update.php:191
[2024/09/10 23:21:34.461] /data/mci/b14552e6624c454125590fb9d49d9cc2/src/src/Operation/UpdateMany.php:107
[2024/09/10 23:21:34.461] /data/mci/b14552e6624c454125590fb9d49d9cc2/src/src/Collection.php:1046
[2024/09/10 23:21:34.461] /data/mci/b14552e6624c454125590fb9d49d9cc2/src/tests/Collection/BuilderCollectionFunctionalTest.php:238
```

GromNaN (commented):
Tests failing only in evergreen ["test-mongodb-4.0-replicaset-noauth-nossl"](https://spruce.mongodb.com/task/mongo_php_library_test_debian92_php_8.3_local_test_mongodb_4.0_replicaset_noauth_nossl_patch_be2df01eb8fa3d3071d4dc1889ea75a14ede93bc_66e0b461a84baa000785b674_24_09_10_21_04_34/tests?execution=0&page=0&sortBy=STATUS&sortDir=ASC)

src/Collection.php
```diff
@@ -221,8 +222,9 @@ public function __toString()
      * @throws InvalidArgumentException for parameter/option parsing errors
      * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
      */
-    public function aggregate(array $pipeline, array $options = [])
+    public function aggregate(array|Pipeline $pipeline, array $options = [])
```
alcaeus, line 225: I don't know how I missed this while we were planning things, but this is a BC break, as any class extending `Collection` and overriding `aggregate` will be incompatible due to a narrower parameter type. I don't know if there's a way to solve this in 1.x, as I'd hate to defer this feature to 2.x.

src/Collection.php
```diff
@@ -1112,6 +1131,13 @@ public function withOptions(array $options = [])
         return new Collection($this->manager, $this->databaseName, $this->collectionName, $options);
     }
 
+    private function inheritBuilderEncoder(array $options): array
+    {
+        $options['builderEncoder'] = $this->builderEncoder;
+
+        return $options;
```
alcaeus, line 1138: You could make this a little more concise:
```suggestion
        return $options + ['builderEncoder' => $this->builderEncoder];
```

I'm aware this would allow passing a `builderEncoder` option to any operation that needs to handle encoding itself, so feel free to ignore if you'd rather not do this.

tests/Collection/BuilderCollectionFunctionalTest.php
alcaeus: What versions is this failing on? Support for update pipelines was only added in 4.2.0, so you'll have to [skip this on older versions](https://github.com/mongodb/mongo-php-library/blob/5ebc5dc0bc3c5643b42d68e001bb797eebd36acc/tests/Operation/BulkWriteFunctionalTest.php#L204).

alcaeus (changes requested):
We missed a BC break while planning this, so we'll have to think about how we can accomplish this.

The changes made to `aggregate` and `watch` will also have to be applied to the `Client::watch`, `Database::watch`, and `Database::aggregate` methods.

src/Collection.php
GromNaN: > I'm aware this would allow passing a `builderEncoder` option to any operation that needs to handle encoding itself, so feel free to ignore if you'd rather not do this.

That's a feature we don't want to support now.

The equivalent code to what I've done would therefore be:
```suggestion
        return ['builderEncoder' => $this->builderEncoder] + $options;
```

src/Collection.php
GromNaN: The compromise we can make for the moment is to accept an array of stages.

```php
$pipeline = new Pipeline(...);
$results = $collection->aggregate(iterator_to_array($pipeline));
```

tests/Database/BuilderDatabaseFunctionalTest.php
```diff
@@ -0,0 +1,61 @@
...
+                groupBy: Expression::intFieldPath('x'),
+                buckets: 2,
+            ),
+        );
+        // Extract the list of stages for arg type restriction
+        $pipeline = iterator_to_array($pipeline);
+
+        $results = $this->database->aggregate($pipeline)->toArray();
+        $this->assertCount(2, $results);
+    }
```
GromNaN: Failing test with `test-mongodb-5.0-replicaset-noauth-nossl`.
```
[2024/09/11 12:21:08.760] ERROR:  (MongoDB\Driver\Exception\CommandException)
[2024/09/11 12:21:08.760] MongoDB\Tests\Database\BuilderDatabaseFunctionalTest::testAggregate
[2024/09/11 12:21:08.760] MongoDB\Driver\Exception\CommandException: $documents' is not allowed in user requests
[2024/09/11 12:21:08.760] /data/mci/4db91d1c8fc44dbdea82e2266907468c/src/src/Operation/Aggregate.php:334
[2024/09/11 12:21:08.760] /data/mci/4db91d1c8fc44dbdea82e2266907468c/src/src/Operation/Aggregate.php:241
[2024/09/11 12:21:08.760] /data/mci/4db91d1c8fc44dbdea82e2266907468c/src/src/Database.php:245
[2024/09/11 12:21:08.760] /data/mci/4db91d1c8fc44dbdea82e2266907468c/src/tests/Database/BuilderDatabaseFunctionalTest.php:37
```

alcaeus: `$documents` was added in MongoDB 6.0. I'm fine with limiting this test to 6.0+ instead of finding a different pipeline stage to run on earlier versions.

src/functions.php
```diff
@@ -327,6 +328,38 @@ function is_pipeline(array|object $pipeline, bool $allowEmpty = false): bool
...
+        }
+
+        if ($stage instanceof StageInterface) {
+            $result = true;
+        } elseif (! is_first_key_operator($stage)) {
+            return false;
+        }
+    }
+
+    return $result;
```
alcaeus: There's no need for a result variable:

```suggestion
    foreach ($pipeline as $stage) {
        if (! is_array($stage) && ! is_object($stage)) {
            return false;
        }

        if (! $stage instanceof StageInterface && ! is_first_key_operator($stage)) {
            return false;
        }
    }

    return true;
```

src/Client.php
```diff
@@ -391,6 +392,12 @@ public function startSession(array $options = [])
      */
     public function watch(array $pipeline = [], array $options = [])
     {
+        if (is_builder_pipeline($pipeline)) {
```
alcaeus, line 395: This should allow us to change the typehint to `array|Pipeline` in 2.0, is that correct? Do you know of a way to prepare users that are extending this class for the change?

GromNaN: If this method is overridden, they can remove or widen the type hint. The `Pipeline` objects can be accepted in subclasses.

```php
use MongoDB\Builder\Pipeline;
use MongoDB\Collection;

class CustomCollection extends Collection
{
    /** @param array|Pipeline $pipeline */
    public function watch(array|Pipeline $pipeline = [], array $options = [])
    {
        // Enable behavior of version 2.0: Pipeline objects are accepted
        if ($pipeline instanceof Pipeline) {
            $pipeline = iterator_to_array($pipeline);
        }

        return parent::watch($pipeline, $options);
    }
}
```

We can also provide this class, flagged as "internal" and add an option to change the `Collection` class and enable the feature. But that seems to me to be over-engineering.

src/functions.php
GromNaN: Your implementation would return `true` for the following pipeline, which is not what we want:
```php
[['$match' => ['x' => 1]]]
```

This highlights the fact that the encoder is currently unable to encode `Operator` objects nested in arrays, it's not possible to mix array and builder syntaxes.

alcaeus: Ah, I see, we want to specifically exclude these items. Got it, sorry for the confusion.

src/Client.php
alcaeus: I don't think that's necessary - widening the parameter type in 2.0 is good enough.

alcaeus (approved):

tests/ClientFunctionalTest.php
```diff
@@ -137,4 +141,25 @@ public function testAddAndRemoveSubscriber(): void
...
         $client->getManager()->executeCommand('admin', new Command(['ping' => 1]));
     }
+
+    public function testWatchWithBuilderPipeline(): void
+    {
+        $this->skipIfChangeStreamIsNotSupported();
+
+        if ($this->isShardedCluster()) {
+            $this->markTestSkipped('Test does not apply on sharded clusters: need more than a single getMore call on the change stream.');
+        }
```
jmikola, line 151: I assume this was copied from `testChangeStreamExample_1_4` in DocumentationExamplesTest. I think this skip was only done for documentation examples to avoid complicating the code snippets.

https://github.com/mongodb/mongo-php-library/commit/32bade4780e432251e448a34ccf39b39baea0e33 also introduced [`advanceCursorUntilValid()`](https://github.com/mongodb/mongo-php-library/commit/32bade4780e432251e448a34ccf39b39baea0e33#diff-e684dc1dab05b372a1d383b95d0fde73fdc3a79ddc8559aeda884d644b3d2caeR1673) to fix change stream iteration for sharded clusters in our other tests.

I won't push for changing this, but wanted to provide some context.

GromNaN: Thanks for the context. In fact, I've simplified the documentation example you pointed. Given what's being tested, I don't think it's essential for the test to run on all server typologies.

PR #1385, opened 2024-09-11
PHPLIB-1420 Integrate query builder for non-aggregation APIs

GromNaN:
Fix PHPLIB-1420

Extracted from https://github.com/mongodb/mongo-php-library/pull/1383 without the [breaking changes on the methods Collection::aggregate() and Collection::watch()](https://github.com/mongodb/mongo-php-library/pull/1383#discussion_r1753314570).

### DIFF
--- a/tests/Collection/BuilderCollectionFunctionalTest.php
+++ b/tests/Collection/BuilderCollectionFunctionalTest.php
@@ -2,7 +2,6 @@
 
 namespace MongoDB\Tests\Collection;
 
-use MongoDB\Builder\Expression;
 use MongoDB\Builder\Pipeline;
 use MongoDB\Builder\Query;
 use MongoDB\Builder\Stage;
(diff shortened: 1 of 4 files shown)


tests/Collection/BuilderCollectionFunctionalTest.php
```diff
@@ -0,0 +1,250 @@
...
+    {
+        $this->markTestSkipped('Not supported yet');
+    }
+
+    public function testBulkWriteDeleteMany(): void
+    {
+        $result = $this->collection->bulkWrite([
+            [
+                'deleteMany' => [
+                    Query::query(x: Query::gt(1)),
```
alcaeus, line 28: Not necessary for this PR, but I just came to think that the following might be a little easier on the eyes:
```suggestion
                    new Query(x: Query::gt(1)),
```

Alternatively, if we have a generic builder class:
```suggestion
                    new Builder::query(x: Query::gt(1)),
```

For some reason, the `Query::query(...)` part irks me a bit 🤔

alcaeus (approved):
LGTM.

I left a comment about the syntax to create a query, we can tackle that separately (if at all)

tests/Collection/BuilderCollectionFunctionalTest.php
GromNaN: Sure, if we can create functions for all the factories that would shorten the syntax.

PR #1391, opened 2024-09-12
PHPLIB-954: Add return types to all methods

alcaeus:
PHPLIB-954

This PR:
* adds testing with the 2.0 branch for PHPC,
* adds native return types for all methods, removing `ReturnTypeWillChange` attributes where present
* replaces all `Cursor` types with `CursorInterface`
* removes the compatibility layer introduced for the `CursorId` deprecation

### DIFF
--- a/src/BulkWriteResult.php
+++ b/src/BulkWriteResult.php
@@ -38,10 +38,9 @@ class BulkWriteResult
      * This method should only be called if the write was acknowledged.
      *
      * @see BulkWriteResult::isAcknowledged()
-     * @return integer|null
      * @throws BadMethodCallException if the write result is unacknowledged
      */
-    public function getDeletedCount()
+    public function getDeletedCount(): ?int
     {
         if ($this->isAcknowledged) {
             return $this->writeResult->getDeletedCount();
@@ -56,10 +55,9 @@ class BulkWriteResult
      * This method should only be called if the write was acknowledged.
      *
      * @see BulkWriteResult::isAcknowledged()
-     * @return integer|null
      * @throws BadMethodCallException if the write result is unacknowledged
      */
-    public function getInsertedCount()
+    public function getInsertedCount(): ?int
     {
         if ($this->isAcknowledged) {
             return $this->writeResult->getInsertedCount();
@@ -76,10 +74,8 @@ class BulkWriteResult
      * the driver did not generate an ID), the index will contain its "_id"
      * field value. Any driver-generated ID will be a MongoDB\BSON\ObjectId
      * instance.
-     *
-     * @return array
      */
-    public function getInsertedIds()
+    public function getInsertedIds(): array
     {
         return $this->insertedIds;
     }
@@ -90,10 +86,9 @@ class BulkWriteResult
      * This method should only be called if the write was acknowledged.
      *
      * @see BulkWriteResult::isAcknowledged()
-     * @return integer|null
      * @throws BadMethodCallException if the write result is unacknowledged
      */
-    public function getMatchedCount()
+    public function getMatchedCount(): ?int
     {
         if ($this->isAcknowledged) {
             return $this->writeResult->getMatchedCount();
@@ -111,10 +106,9 @@ class BulkWriteResult
      * This method should only be called if the write was acknowledged.
      *
      * @see BulkWriteResult::isAcknowledged()
-     * @return integer|null
      * @throws BadMethodCallException if the write result is unacknowledged
      */
-    public function getModifiedCount()
+    public function getModifiedCount(): ?int
     {
         if ($this->isAcknowledged) {
             return $this->writeResult->getModifiedCount();
@@ -129,10 +123,9 @@ class BulkWriteResult
      * This method should only be called if the write was acknowledged.
      *
      * @see BulkWriteResult::isAcknowledged()
-     * @return integer|null
      * @throws BadMethodCallException if the write result is unacknowledged
      */
-    public function getUpsertedCount()
+    public function getUpsertedCount(): ?int
     {
         if ($this->isAcknowledged) {
             return $this->writeResult->getUpsertedCount();
@@ -152,10 +145,9 @@ class BulkWriteResult
      * This method should only be called if the write was acknowledged.
      *
      * @see BulkWriteResult::isAcknowledged()
-     * @return array
      * @throws BadMethodCallException if the write result is unacknowledged
      */
-    public function getUpsertedIds()
+    public function getUpsertedIds(): array
     {
         if ($this->isAcknowledged) {
             return $this->writeResult->getUpsertedIds();
@@ -169,10 +161,8 @@ class BulkWriteResult
      *
      * If the update was not acknowledged, other fields from the WriteResult
      * (e.g. matchedCount) will be undefined.
-     *
-     * @return boolean
      */
-    public function isAcknowledged()
+    public function isAcknowledged(): bool
     {
         return $this->isAcknowledged;
     }
--- a/src/ChangeStream.php
+++ b/src/ChangeStream.php
@@ -27,7 +27,6 @@ use MongoDB\Driver\Exception\ServerException;
 use MongoDB\Exception\BadMethodCallException;
 use MongoDB\Exception\ResumeTokenException;
 use MongoDB\Model\ChangeStreamIterator;
-use ReturnTypeWillChange;
 
 use function assert;
 use function call_user_func;
@@ -108,20 +107,14 @@ class ChangeStream implements Iterator
      * Null may be returned if no change documents have been iterated and the
      * server did not include a postBatchResumeToken in its aggregate or getMore
      * command response.
-     *
-     * @return array|object|null
      */
-    public function getResumeToken()
+    public function getResumeToken(): array|object|null
     {
         return $this->iterator->getResumeToken();
     }
 
-    /**
-     * @see https://php.net/iterator.key
-     * @return int|null
-     */
-    #[ReturnTypeWillChange]
-    public function key()
+    /** @see https://php.net/iterator.key */
+    public function key(): ?int
     {
         if ($this->valid()) {
             return $this->key;
@@ -132,11 +125,9 @@ class ChangeStream implements Iterator
 
     /**
      * @see https://php.net/iterator.next
-     * @return void
      * @throws ResumeTokenException
      */
-    #[ReturnTypeWillChange]
-    public function next()
+    public function next(): void
     {
         try {
             $this->iterator->next();
@@ -148,11 +139,9 @@ class ChangeStream implements Iterator
 
     /**
      * @see https://php.net/iterator.rewind
-     * @return void
      * @throws ResumeTokenException
      */
-    #[ReturnTypeWillChange]
-    public function rewind()
+    public function rewind(): void
     {
         try {
             $this->iterator->rewind();
@@ -165,12 +154,8 @@ class ChangeStream implements Iterator
         }
     }
 
-    /**
-     * @see https://php.net/iterator.valid
-     * @return boolean
-     */
-    #[ReturnTypeWillChange]
-    public function valid()
+    /** @see https://php.net/iterator.valid */
+    public function valid(): bool
     {
         return $this->iterator->valid();
     }
--- a/src/ChangeStream.php
+++ b/src/ChangeStream.php
@@ -21,7 +21,6 @@ use Iterator;
 use MongoDB\BSON\Document;
 use MongoDB\BSON\Int64;
 use MongoDB\Codec\DocumentCodec;
-use MongoDB\Driver\CursorId;
 use MongoDB\Driver\Exception\ConnectionException;
 use MongoDB\Driver\Exception\RuntimeException;
 use MongoDB\Driver\Exception\ServerException;
@@ -33,10 +32,6 @@ use ReturnTypeWillChange;
 use function assert;
 use function call_user_func;
 use function in_array;
-use function sprintf;
-use function trigger_error;
-
-use const E_USER_DEPRECATED;
 
 /**
  * Iterator for a change stream.
@@ -88,12 +83,8 @@ class ChangeStream implements Iterator
      */
     private bool $hasAdvanced = false;
 
-    /**
-     * @see https://php.net/iterator.current
-     * @return array|object|null
-     */
-    #[ReturnTypeWillChange]
-    public function current()
+    /** @see https://php.net/iterator.current */
+    public function current(): array|object|null
     {
         $value = $this->iterator->current();
 
@@ -106,26 +97,9 @@ class ChangeStream implements Iterator
         return $this->codec->decode($value);
     }
 
-    /**
-     * @return CursorId|Int64
-     * @psalm-return ($asInt64 is true ? Int64 : CursorId)
-     */
-    #[ReturnTypeWillChange]
-    public function getCursorId(bool $asInt64 = false)
+    public function getCursorId(): Int64
     {
-        if (! $asInt64) {
-            @trigger_error(
-                sprintf(
-                    'The method "%s" will no longer return a "%s" instance in the future. Pass "true" as argument to change to the new behavior and receive a "%s" instance instead.',
-                    __METHOD__,
-                    CursorId::class,
-                    Int64::class,
-                ),
-                E_USER_DEPRECATED,
-            );
-        }
-
-        return $this->iterator->getInnerIterator()->getId($asInt64);
+        return $this->iterator->getInnerIterator()->getId();
     }
 
     /**
@@ -255,7 +229,8 @@ class ChangeStream implements Iterator
          * have been received in the last response. Therefore, we can unset the
          * resumeCallable. This will free any reference to Watch as well as the
          * only reference to any implicit session created therein. */
-        if ((string) $this->getCursorId(true) === '0') {
+        // Use a type-unsafe comparison to compare with Int64 instances
+        if ($this->getCursorId() == 0) {
             $this->resumeCallable = null;
         }
 
--- a/src/Client.php
+++ b/src/Client.php
@@ -39,6 +39,7 @@ use MongoDB\Exception\UnsupportedException;
 use MongoDB\Model\BSONArray;
 use MongoDB\Model\BSONDocument;
 use MongoDB\Model\DatabaseInfo;
+use MongoDB\Model\DatabaseInfoIterator;
 use MongoDB\Operation\DropDatabase;
 use MongoDB\Operation\ListDatabaseNames;
 use MongoDB\Operation\ListDatabases;
@@ -143,9 +144,8 @@ class Client
      * Return internal properties for debugging purposes.
      *
      * @see https://php.net/manual/en/language.oop5.magic.php#language.oop5.magic.debuginfo
-     * @return array
      */
-    public function __debugInfo()
+    public function __debugInfo(): array
     {
         return [
             'manager' => $this->manager,
@@ -166,19 +166,16 @@ class Client
      * @see https://php.net/oop5.overloading#object.get
      * @see https://php.net/types.string#language.types.string.parsing.complex
      * @param string $databaseName Name of the database to select
-     * @return Database
      */
-    public function __get(string $databaseName)
+    public function __get(string $databaseName): Database
     {
         return $this->selectDatabase($databaseName);
     }
 
     /**
      * Return the connection string (i.e. URI).
-     *
-     * @return string
      */
-    public function __toString()
+    public function __toString(): string
     {
         return $this->uri;
     }
@@ -197,10 +194,8 @@ class Client
      * Returns a ClientEncryption instance for explicit encryption and decryption
      *
      * @param array $options Encryption options
-     *
-     * @return ClientEncryption
      */
-    public function createClientEncryption(array $options)
+    public function createClientEncryption(array $options): ClientEncryption
     {
         if (isset($options['keyVaultClient'])) {
             if ($options['keyVaultClient'] instanceof self) {
@@ -224,7 +219,7 @@ class Client
      * @throws InvalidArgumentException for parameter/option parsing errors
      * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
      */
-    public function dropDatabase(string $databaseName, array $options = [])
+    public function dropDatabase(string $databaseName, array $options = []): array|object
     {
         if (! isset($options['typeMap'])) {
             $options['typeMap'] = $this->typeMap;
@@ -243,10 +238,8 @@ class Client
 
     /**
      * Return the Manager.
-     *
-     * @return Manager
      */
-    public function getManager()
+    public function getManager(): Manager
     {
         return $this->manager;
     }
@@ -255,29 +248,24 @@ class Client
      * Return the read concern for this client.
      *
      * @see https://php.net/manual/en/mongodb-driver-readconcern.isdefault.php
-     * @return ReadConcern
      */
-    public function getReadConcern()
+    public function getReadConcern(): ReadConcern
     {
         return $this->readConcern;
     }
 
     /**
      * Return the read preference for this client.
-     *
-     * @return ReadPreference
      */
-    public function getReadPreference()
+    public function getReadPreference(): ReadPreference
     {
         return $this->readPreference;
     }
 
     /**
      * Return the type map for this client.
-     *
-     * @return array
      */
-    public function getTypeMap()
+    public function getTypeMap(): array
     {
         return $this->typeMap;
     }
@@ -286,9 +274,8 @@ class Client
      * Return the write concern for this client.
      *
      * @see https://php.net/manual/en/mongodb-driver-writeconcern.isdefault.php
-     * @return WriteConcern
      */
-    public function getWriteConcern()
+    public function getWriteConcern(): WriteConcern
     {
         return $this->writeConcern;
     }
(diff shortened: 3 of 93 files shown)


src/BulkWriteResult.php
```diff
@@ -38,10 +38,9 @@ public function __construct(private WriteResult $writeResult, private array $ins
      * This method should only be called if the write was acknowledged.
      *
      * @see BulkWriteResult::isAcknowledged()
-     * @return integer|null
      * @throws BadMethodCallException if the write result is unacknowledged
      */
-    public function getDeletedCount()
+    public function getDeletedCount(): ?int
```
alcaeus, line 43: These methods can be refactored when PHPC-2144 is implemented. At that point, the return type of `WriteResult` methods will change to `int` and they will always throw when accessing any result counts for an unacknowledged write.

src/Collection.php
```diff
@@ -216,13 +214,12 @@ public function __toString()
...
      * @param array $pipeline Aggregation pipeline
      * @param array $options  Command options
-     * @return CursorInterface&Iterator
      * @throws UnexpectedValueException if the command response was malformed
      * @throws UnsupportedException if options are not supported by the selected server
      * @throws InvalidArgumentException for parameter/option parsing errors
      * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
      */
-    public function aggregate(array $pipeline, array $options = [])
+    public function aggregate(array $pipeline, array $options = []): CursorInterface&Iterator
```
alcaeus: We don't have any plans for this yet, but I propose having `CursorInterface` extend `Iterator` to avoid the need for an intersection type here. Currently, only the `Cursor` class implements `Iterator` (added in PHPC-1691, but not added to `CursorInterface` due to BC constraints).

src/Operation/DropSearchIndex.php
```diff
@@ -53,11 +54,13 @@ public function __construct(private string $databaseName, private string $collec
...
      * Execute the operation.
      *
+     * TODO PHP 8.2: Change return type to null
      * @see Executable::execute()
      * @throws UnsupportedException if write concern is used and unsupported
      * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     * @return null
      */
-    public function execute(Server $server): void
+    public function execute(Server $server): mixed
```
alcaeus: The void type hint is invalid, as the interface has a `mixed` type. Since `mixed` is a type that can't combined with other types, but also doesn't include `void`, we have to use `mixed` for now. We can change this when upgrading to PHP 8.2, as that version introduces `null` as a standalone type. If we were to make this class final, we could make the change in an upcoming minor version, especially as the method only ever returns `null`.

src/Operation/Explainable.php
```diff
@@ -27,8 +27,6 @@ interface Explainable extends Executable
 {
     /**
      * Returns the command document for this operation.
-     *
-     * @return array
      */
-    public function getCommandDocument();
+    public function getCommandDocument(): array;
```
alcaeus, line 31: We currently return arrays for all commands - this major version would be a chance to create objects instead if we wanted to. I don't have any strong feelings about this, but wanted to mention it.

src/Operation/UpdateSearchIndex.php
```diff
@@ -65,7 +66,7 @@ public function __construct(private string $databaseName, private string $collec
      * @throws UnsupportedException if write concern is used and unsupported
      * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
      */
-    public function execute(Server $server): void
+    public function execute(Server $server): Cursor
```
alcaeus: I decided to return the `Cursor` instance returned from `executeCommand` here. We could also return `null` as was done in `DropSearchIndex` and change the return type in an upcoming minor version.

alcaeus (commented):
Left some explanations. A number of the psalm errors are due to https://github.com/vimeo/psalm/issues/10019, which doesn't correctly infer the return type or `(object) ['foo' => 'bar']` as `stdClass` but instead assumes `object`. I'll take a look at the other psalm errors regarding codec generics, but feel free to leave feedback in the meantime.

src/Client.php
```diff
@@ -143,9 +143,8 @@ public function __construct(?string $uri = null, array $uriOptions = [], array $
      * Return internal properties for debugging purposes.
      *
      * @see https://php.net/manual/en/language.oop5.magic.php#language.oop5.magic.debuginfo
-     * @return array
```
GromNaN, line 146: We could describe the return type in detail with the types of each field.

src/Collection.php
GromNaN: `CursorInterface` already extends `Traversable`, which is better IMO. We should not require the `Iterator` interface for custom cursor classes.
In case people really need the `Iterator` API, they can wrap it in an `IteratorIterator`.
```suggestion
    public function aggregate(array $pipeline, array $options = []): CursorInterface
```

src/Database.php
```diff
@@ -201,7 +197,7 @@ public function __toString()
      * @throws InvalidArgumentException for parameter/option parsing errors
      * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
      */
-    public function aggregate(array $pipeline, array $options = [])
+    public function aggregate(array $pipeline, array $options = []): Traversable
```
GromNaN: I would expect a `CursorInterface` like `Collection::aggregate()`

src/Database.php
```diff
@@ -251,11 +247,10 @@ public function aggregate(array $pipeline, array $options = [])
      * @see DatabaseCommand::__construct() for supported options
      * @param array|object $command Command document
      * @param array        $options Options for command execution
-     * @return Cursor
      * @throws InvalidArgumentException for parameter/option parsing errors
      * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
      */
-    public function command(array|object $command, array $options = [])
+    public function command(array|object $command, array $options = []): Cursor
```
GromNaN: Not a `CursorInterface`?
```suggestion
    public function command(array|object $command, array $options = []): CursorInterface
```

src/Operation/UpdateSearchIndex.php
GromNaN: We discussed returning `void` for all non-meaningful results: [PHPLIB-1227](https://jira.mongodb.org/browse/PHPLIB-1227)

src/Database.php
alcaeus: Good point, will update.

src/Operation/UpdateSearchIndex.php
alcaeus: This won't work - best we can do is to return `null` as we can't change a return type to `void` if the interface specifies the method as returning a value.

src/Collection.php
alcaeus: A `CursorInterface` type without an intersection with `Iterator` won't allow you to call `$cursor->next();` - that's an issue we faced in ODM. Ideally, we'd fix this in PHPC by having `CursorInterface` extend `Iterator`, but until then we're stuck with this solution.

src/Database.php
alcaeus: Updated

src/Operation/UpdateSearchIndex.php
GromNaN: This interface is useless. It is never used to type, and in any case it is not capable of indicating the return type. At best, it guarantees a certain consistency between classes.

At the least, I'd remove the return type, but I'd be more inclined to remove this interface.

src/Client.php
alcaeus: Technically, yes. However, since this method is only used when calling `var_dump` and we don't want users to calling it manually, I don't see the benefit of properly defining the array shape for static analysis.

src/Collection.php
GromNaN: There is always solutions to wrap a Transversable that is not an `Iterator`.
With a Generator (see [UnrewindableIterator](https://github.com/doctrine/mongodb-odm/blob/8e0deb186540a89111181c3be89924a3d791045c/lib/Doctrine/ODM/MongoDB/Iterator/UnrewindableIterator.php#L126-L127]), `IteratorIterator` or `IteratorAggregate::getIterator()`

jmikola: Providing some additional context and throwing in my two cents:

 * We added Iterator to Cursor in [PHPC-1691](https://jira.mongodb.org/browse/PHPC-1691) in order to comply with PHP 8.
 * CursorInterface was introduced in [PHPC-1123](https://jira.mongodb.org/browse/PHPC-1123) to allow other classes to fill in for a Cursor, which cannot be extended. AFAIK, the only existing use case for this within our own driver is CodecCursor.
 * CodecCursor currently implements both CursorInterface and Iterator, so that would not be negatively impacted by this change.

While conditionally wrapping a CursorInterface to get access to Iterator methods is always possible, it may be tedious. Given that CursorInterface is intended to be used for classes that compose a Cursor, it seems a small ask for the wrapping class to also implement Iterator.

I'm not sure what the use case would be for a CursorInterface implementation to rely on IteratorAggregate, as it would then have no way to influence the inner Cursor's iteration. Hypothetically, it could proxy other methods (e.g. `toArray()`), but I've never encountered such an implementation in the wild.

If having CursorInterface extend Iterator doesn't negatively impact PHPLIB, Doctrine ODM, and the Laravel integration, I wouldn't be too concerned about other use cases. Worst case, we're asking users to implement a implement a few extra methods (easily accomplished with a trait if needed).

src/Operation/DropSearchIndex.php
jmikola: Would it be preferable to just remove the return type from `Executable::execute()` interface? This interface has always been internal and doesn't really serve any purpose within the library.

It is extended by Explainable, but that also seems unnecessary since the Explain operation only depends on `Explainable::getCommandDocument()`.

For context: Executable was created in a different time where I was adamant about everything having an interface ([PHPC-378](https://jira.mongodb.org/browse/PHPC-378)). Nine years later, I think we'd have done better to embrace the [YAGNI](https://en.wikipedia.org/wiki/You_aren't_gonna_need_it) philosophy earlier on.

src/Operation/Explainable.php
jmikola: The pedantic choice would probably be `array|object` for consistency with [`MongoDB\Driver\Command`](https://www.php.net/manual/en/class.mongodb-driver-command.php).

**/me glances at the wall** ([yes, really](https://x.com/jmikola/status/1701971766137004529))

[Covariance](https://www.php.net/manual/en/language.oop5.variance.php) would still allow us to maintain `array` type hints in the operation class implementations: https://3v4l.org/AvFKa

And I don't think the Explain operation would require any changes one way or another.

src/Operation/UpdateSearchIndex.php
jmikola: Saw this after replying in https://github.com/mongodb/mongo-php-library/pull/1391#discussion_r1759155384.

Removing the interface's return type hint is an easy solution. I also have no objections to removing the interface entirely, as it's internal and doesn't provide any real value. Operations are only ever used interchangeably via the Explainable interface (which really doesn't need to extend Executable).

src/Database.php
jmikola: For context, when `Database::aggregate()` was implemented in https://github.com/mongodb/mongo-php-library/commit/c2922d30f2122c285b4a75db3b8b114b6e233350, the [Aggregate operation](https://github.com/mongodb/mongo-php-library/blob/c2922d30f2122c285b4a75db3b8b114b6e233350/src/Operation/Aggregate.php) conditionally returned non-Cursor Iterator implementations crafted from inline responses. That is certainly no longer relevant.

src/BulkWriteResult.php
jmikola: Do you want to create a separate PHPLIB ticket related to [PHPLIB-954](https://jira.mongodb.org/browse/PHPLIB-954) and depending on [PHPC-2144](https://jira.mongodb.org/browse/PHPC-2144) to track this?

src/ChangeStream.php
```diff
@@ -106,12 +101,8 @@ public function current()
...
     }
 
-    /**
-     * @return CursorId|Int64
-     * @psalm-return ($asInt64 is true ? Int64 : CursorId)
-     */
-    #[ReturnTypeWillChange]
-    public function getCursorId(bool $asInt64 = false)
+    /** @psalm-return ($asInt64 is true ? Int64 : CursorId) */
+    public function getCursorId(bool $asInt64 = false): CursorId|Int64
```
jmikola: This is temporary, correct? I assume this changes to `getCursorId(): Int64` once [PHPC-2351](https://jira.mongodb.org/browse/PHPC-2351) is implemented.

src/GridFS/Bucket.php
```diff
@@ -203,9 +203,8 @@ public function __construct(private Manager $manager, private string $databaseNa
      * Return internal properties for debugging purposes.
      *
      * @see https://php.net/manual/en/language.oop5.magic.php#language.oop5.magic.debuginfo
-     * @return array
      */
-    public function __debugInfo()
+    public function __debugInfo(): array
```
jmikola, line 187: Noted that we cannot `resource` return type hints, so the `@return resource` annotations remain in place for those methods that open a stream.

src/InsertManyResult.php
```diff
@@ -38,10 +38,9 @@ public function __construct(private WriteResult $writeResult, private array $ins
      * This method should only be called if the write was acknowledged.
      *
      * @see InsertManyResult::isAcknowledged()
-     * @return integer|null
      * @throws BadMethodCallException if the write result is unacknowledged
      */
-    public function getInsertedCount()
+    public function getInsertedCount(): ?int
```
jmikola, line 43: I assume https://github.com/mongodb/mongo-php-library/pull/1391/files#r1758279785 applies here as well, yes? Along with any other result classes that proxy a WriteResult?

src/Operation/DatabaseCommand.php
```diff
@@ -82,9 +82,8 @@ public function __construct(private string $databaseName, array|object $command,
      * Execute the operation.
      *
      * @see Executable::execute()
-     * @return Cursor
      */
-    public function execute(Server $server)
+    public function execute(Server $server): Cursor
```
jmikola: Should this be CursorInterface, or is the narrow return type sensible given we know exactly what the method does? I saw @GromNaN pointed something similar out elsewhere, so I'm not sure if it's the same issue (in which case there could also be others that were missed).

src/Operation/Executable.php
```diff
@@ -31,8 +31,6 @@ interface Executable
 {
     /**
      * Execute the operation.
-     *
-     * @return mixed
      */
-    public function execute(Server $server);
+    public function execute(Server $server): mixed;
```
jmikola: Consider leaving this untyped or removing the interface entirely (that should warrant a PHPLIB ticket for our internal tracking).

src/Operation/ListCollections.php
```diff
@@ -68,10 +68,9 @@ public function __construct(private string $databaseName, array $options = [])
      * Execute the operation.
      *
      * @see Executable::execute()
-     * @return CollectionInfoIterator
      * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
      */
-    public function execute(Server $server)
+    public function execute(Server $server): CollectionInfoIterator
```
jmikola, line 76: I started a [Slack thread](https://mongodb.slack.com/archives/C026UA4GF0B/p1726247537366849) to discuss whether these iterators and the corresponding model classes provide any value. Even if we keep the model classes, I suppose we could remove the iterators and rely on Psalm to type them as an iterator of a particular model class.

src/ChangeStream.php
alcaeus: That assumption is correct 👍

src/InsertManyResult.php
alcaeus: Correct

src/BulkWriteResult.php
alcaeus: Tracking in PHPLIB-1518.

src/Collection.php
alcaeus: Sounds good. Added PHPC-2447 to track this.

I'll note that we shouldn't implement `IteratorAggregate`, as that would suggest one can obtain multiple iterators from a single cursor. As a cursor can't be rewound and may only ever yield a single iterator, I don't think we should consider implementing `IteratorAggregate`.

src/Operation/DropSearchIndex.php
alcaeus: Sounds good. I've created PHPLIB-1519 to track the deprecation and PHPLIB-1520 for the removal of the `Executable` interface. In the meantime, I've removed the native type from `Executable::execute` in favour of a `mixed|void` type hint in the docblock. The `execute` methods of operations that would previously return `void` have been changed back to that type instead of trying to returning something semi-useful.

src/Operation/Explainable.php
alcaeus: In that case, I'd rather keep it as `array` for the time being - no need to add a less than useful BC break for the sake of adding one.

src/Operation/UpdateSearchIndex.php
alcaeus: This was changed back to `void`.

src/Operation/DatabaseCommand.php
alcaeus: While we technically know what return type is coming from the extension, using `CursorInterface` allows us to change the actual cursor implementation without causing a BC break. I've changed this and other types to `CursorInterface&Iterator`, which we can then replace with `CursorInterface` after PHPC-2447.

src/Operation/Executable.php
alcaeus: Removed for the moment and added tracking tickets.

src/Operation/ListCollections.php
jmikola: Noted the following JIRA tickets were created:

 * [PHPLIB-1522](https://jira.mongodb.org/browse/PHPLIB-1522) to deprecate the iterators and interfaces in 1.x
 *  * [PHPLIB-1523](https://jira.mongodb.org/browse/PHPLIB-1523) to remove them in 2.0 (replaced with templated iterator return types)

We agreed to leave the model classes in place.

src/Model/CodecCursor.php
```diff
@@ -61,22 +59,19 @@ public function current(): ?object
 
     /**
      * @template NativeClass of Object
-     * @param DocumentCodec<NativeClass> $codec
+     * @param CursorInterface<int, NativeClass> &Iterator<int, NativeClass> $cursor
+     * @param DocumentCodec<NativeClass>        $codec
      * @return self<NativeClass>
      */
-    public static function fromCursor(Cursor $cursor, DocumentCodec $codec): self
+    public static function fromCursor(CursorInterface&Iterator $cursor, DocumentCodec $codec): self
```
jmikola: Offhand, would Psalm actually complain about a redundant type hint once its stubs are updated for CursorInterface extending Iterator?

src/Operation/Find.php
```diff
@@ -314,6 +313,7 @@ public function execute(Server $server)
         $cursor = $server->executeQuery($this->databaseName . '.' . $this->collectionName, new Query($this->filter, $this->createQueryOptions()), $this->createExecuteOptions());
 
         if (isset($this->options['codec'])) {
+            /** @var CursorInterface<int, object>&Iterator<int, object> $cursor */
```
jmikola: `$cursor` is assigned a few lines above, so I'm curious why phpcs is unable to find it. Should this warning be suppressed?

tests/Database/DatabaseFunctionalTest.php
```diff
@@ -79,7 +79,7 @@ public function testCommand(): void
 
         $cursor = $this->database->command($command, $options);
 
-        $this->assertInstanceOf(Cursor::class, $cursor);
+        $this->assertInstanceOf(CursorInterface::class, $cursor);
```
jmikola: Noted we're not bothering with additional checks for `Iterator::class` since it'd just become irrelevant after CursorInterface extends it in 2.0.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -1382,7 +1382,7 @@ public function testOriginalReadPreferenceIsPreservedOnResume(): void
             ChangeStream::class,
         );
         $cursor = $getCursor();
-        assert($cursor instanceof Cursor);
```
jmikola: I'm not sure why this was using `assert()` instead of a PHPUnit assertion. Apart from this, I only see `assert()` used in a couple of spec prose tests and TestFileCodec.

Consider replacing with `self::assertInstanceOf(CursorInterface::class, $cursor)` if you like.

jmikola (dismissed):
Optional suggestions.

Decide how to handle the CI warnings, but LGTM.

alcaeus:
Updated the PR. This currently includes changes from #1427, but I'll wait until that PR has been merged up to 2.x and rebasing this PR to preserve proper history.

src/Collection.php
alcaeus: `CursorInterface` was changed to extend `Iterator`, and this PR now includes changes for PHPLIB-1114, using `CursorInterface` where possible.

src/Operation/DatabaseCommand.php
alcaeus: This was changed to `CursorInterface`

src/Operation/Find.php
alcaeus: This has now been solved.

src/Model/CodecCursor.php
alcaeus: Not sure - this went out the door with PHPC-2447 being implemented before I got done with these changes.

src/Operation/Watch.php
```diff
@@ -353,10 +351,8 @@ private function createChangeStreamIterator(Server $server): ChangeStreamIterato
      *
      * The command will be executed using APM so that we can capture data from
      * its response (e.g. firstBatch size, postBatchResumeToken).
-     *
-     * @return CursorInterface&Iterator
      */
-    private function executeAggregate(Server $server)
+    private function executeAggregate(Server $server): CursorInterface
     {
         addSubscriber($this);
```
jmikola, line 357: Nothing relevant to this PR, but I want to point out that it's a bit awkward that we can assign subscribers to Managers but not individual Servers. That forces us to use a global subscriber here.

I suppose we could add `Server::getManager()` to work around that :man_shrugging:

tests/Operation/WatchFunctionalTest.php
```diff
@@ -1370,11 +1370,11 @@ public function testOriginalReadPreferenceIsPreservedOnResume(): void
         }
 
         $changeStream = $operation->execute($secondary);
-        $previousCursorId = $changeStream->getCursorId(true);
+        $previousCursorId = $changeStream->getCursorId();
         $this->forceChangeStreamResume();
 
         $changeStream->next();
-        $this->assertNotSame($previousCursorId, $changeStream->getCursorId(true));
+        $this->assertNotSame($previousCursorId, $changeStream->getCursorId());
```
jmikola: I don't think `assertNotSame()` was ever correct here, as it previously would have been comparing two CursorId objects. I suggest switching this to `assertNotEquals()`.

Also, please double check any other cursor ID comparisons if it's not too much trouble.

The original line here seems to date back to https://github.com/mongodb/mongo-php-library/commit/aac3c6bd84ed39708c156d72dcfde7e97212254b

alcaeus: Ah, excellent point. I've looked through all usages of `CursorInterface::getId` and `ChangeStream::getCursorId` and confirmed that they only use `assertNotEquals` as well as loose comparisons (`==`) throughout. 👍

tests/Operation/WatchFunctionalTest.php
```diff
@@ -720,7 +720,7 @@ public function testInitialCursorIsNotClosed(): void
          * reports the cursor as alive. While the cursor ID is accessed through
          * ChangeStream, we'll need to use reflection to access the internal
          * Cursor and call isDead(). */
-        $this->assertNotEquals('0', (string) $changeStream->getCursorId(true));
+        $this->assertNotEquals(0, $changeStream->getCursorId());
```
GromNaN, line 724: Would it make the test more robust if you don't rely on int cast?
```suggestion
        $this->assertNotEquals(new Int64(0), $changeStream->getCursorId());
```

alcaeus: It doesn't make a difference, as the custom `Int64Comparator` will always handle these assertions, and it leverages the comparison behaviour implemented in `Int64` directly.

tests/Operation/WatchFunctionalTest.php
```diff
@@ -1370,11 +1371,11 @@ public function testOriginalReadPreferenceIsPreservedOnResume(): void
         }
 
         $changeStream = $operation->execute($secondary);
-        $previousCursorId = $changeStream->getCursorId(true);
-        $this->forceChangeStreamResume();
+        $previousCursorId = $changeStream->getCursorId();
+        $this->forceChangeStreamResume($secondary);
```
alcaeus, line 1375: Note: this method previously created the fail point on the primary server, instead of on the secondary server where the changestream was created.  This meant that the change stream was never resumed and the cursor ID was the same as before. We just didn't catch this mistake as the `assertNotSame` comparison always saw the objects as different as they didn't have the same identity. Lessons learned...

GromNaN: It should be fixed on the 1.x branch, right?

alcaeus: Extracted to #1434. Will rebase when that PR is merged.

jmikola (approved):

GromNaN (approved):

PR #1392, opened 2024-09-12
PHPLIB-953 Make internal classes and Operation classes final

GromNaN:
Fix PHPLIB-953

### DIFF
--- a/UPGRADE-2.0.md
+++ b/UPGRADE-2.0.md
@@ -1,0 +1,4 @@
+UPGRADE FROM 1.x to 2.0
+========================
+
+ * Classes in the namespace `MongoDB\Operation\` are `final`.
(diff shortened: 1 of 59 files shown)


alcaeus:
This only includes a part of the classes in the `Operation` namespace. @jmikola should we make all of those classes internal and make those final that we don't extend for our own purposes?

GromNaN:
`@internal` I'm not sure because the operation classes can be used with `explain`. `@final` for sure, they are all composable.

jmikola:
I suppose all Operation classes can be `@final` if you want to go that route. Those that implement Explainable should not be considered internal, as @GromNaN stated.

jmikola (approved):
I'll defer to you if this warrants some communication in `UPGRADE-2.0.md`.

jmikola (approved):

PR #1393, opened 2024-09-12
Use `match` instead of `switch` when a simple value is returned

GromNaN:


tests/SpecTests/ClientSideEncryption/Prose22_RangeExplicitEncryptionTest.php
```diff
@@ -451,26 +451,13 @@ private function assertMultipleDocumentsMatch(array $expectedDocuments, Iterator
...
-                throw new LogicException('Unsupported type: ' . $type);
-        }
+        return match ($type) {
+            'DecimalNoPrecision', 'DecimalPrecision' => fn (int $value) => new Decimal128((string) $value),
+            'DoubleNoPrecision', 'DoublePrecision' => fn (int $value) => (double) $value,
+            'Date' => fn (int $value) => new UTCDateTime($value),
+            'Int' => fn (int $value) => $value,
+            'Long' => fn (int $value) => new Int64($value),
+            default => throw new LogicException('Unsupported type: ' . $type),
+        };
```
GromNaN, line 461: I think this function could be simplified to cast the value directly instead of returning a closure.


```php
private static function cast(string $type, int $value): callable
{
    return match ($type) {
        'DecimalNoPrecision', 'DecimalPrecision' => new Decimal128((string) $value),
        'DoubleNoPrecision', 'DoublePrecision' => (double) $value,
        'Date' => new UTCDateTime($value),
        'Int' => $value,
        'Long' => new Int64($value),
        default => throw new LogicException('Unsupported type: ' . $type),
    };
}
```

GromNaN: Outdated class: https://github.com/mongodb/mongo-php-library/pull/1394

src/Builder/Encoder/OperatorEncoder.php
```diff
@@ -111,8 +98,10 @@ private function encodeAsDollarObject(OperatorInterface $value): stdClass
     /**
      * $group stage have a specific encoding because the _id argument is required and others are variadic
      */
-    private function encodeAsGroup(GroupStage $value): stdClass
+    private function encodeAsGroup(OperatorInterface $value): stdClass
     {
+        assert($value instanceof GroupStage);
```
alcaeus, line 103: Noted that this assert was moved here, it was previously handled in the `switch` statement.

tests/UnifiedSpecTests/Constraint/IsBsonType.php
```diff
@@ -90,77 +90,32 @@ public static function anyOf(string ...$types): Constraint
...
-                return $other instanceof MaxKeyInterface;
-
-            case 'number':
-                return is_int($other) || $other instanceof Int64 || is_float($other) || $other instanceof Decimal128Interface;
-
-            default:
-                // This should already have been caught in the constructor
-                throw new LogicException('Unsupported type: ' . $this->type);
-        }
+        return match ($this->type) {
```
alcaeus, line 93: This particular instance highlights just how much more concise the `match` syntax is compared to a `switch`.

alcaeus (approved):

tests/UnifiedSpecTests/Constraint/IsBsonType.php
GromNaN: All the `switch` `cases` could have been wrote on a single line. That's a CS rule that enforce newline.

src/Builder/Encoder/OperatorEncoder.php
GromNaN: Only `switch` that I converted manually.

PR #1395, opened 2024-09-12
Remove PHPUnit functions polyfill

GromNaN:
The polyfill of PHPUnit assert function was added to support both PHPUnit 8 and 9: https://github.com/mongodb/mongo-php-library/pull/793. Because the namespace was added in version 9.

Now that we require PHP 8.1+, the Symfony PHPUnit bridge only installs [PHPUnit 9.6](https://github.com/symfony/phpunit-bridge/blob/e876eb90e32a8fc4c4911d458e09f88d65877d1c/bin/simple-phpunit.php#L100-L101). We can therefore be sure that we always have these namespaced functions. The polyfill is useless.


alcaeus (approved):
Woohoo!

PR #1398, opened 2024-09-13
PHPLIB-1218 Remove deprecated fields from GridFS files

GromNaN:
Fix PHPLIB-1218

Remove fields `md5`, `contentType` and `aliases`.


tests/UnifiedSpecTests/Util.php
```diff
@@ -134,8 +134,8 @@ final class Util
             'delete' => ['id'],
             'downloadByName' => ['filename', 'revision'],
             'download' => ['id'],
-            'uploadWithId' => ['id', 'filename', 'source', 'chunkSizeBytes', 'disableMD5', 'contentType', 'metadata'],
-            'upload' => ['filename', 'source', 'chunkSizeBytes', 'disableMD5', 'contentType', 'metadata'],
```
GromNaN, line 138: Spec tests should be updated to remove this options.

src/GridFS/Bucket.php
```diff
@@ -129,14 +120,9 @@ class Bucket
      */
     public function __construct(private Manager $manager, private string $databaseName, array $options = [])
     {
-        if (isset($options['disableMD5']) && $options['disableMD5'] === false) {
-            @trigger_error('Setting GridFS "disableMD5" option to "false" is deprecated since mongodb/mongodb 1.18 and will not be supported in version 2.0.', E_USER_DEPRECATED);
-        }
```
GromNaN, line 134: We can throw an exception when the value is set and different than `true`.

jmikola: That presumably only helps users that were ignoring this deprecation notice in 1.x. I'm not convinced it's worth keeping that logic around given there are far more impactful changes where we can do nothing (e.g. removing interfaces or changing signatures).

This is something we should note in an upgrade document, which @alcaeus proposed we start doing (see: [Slack thread](https://mongodb.slack.com/archives/C026UA4GF0B/p1726146411850819) from yesterday). I suppose `UPGRADE-2.0.md` works as a filename (after Symfony).

src/GridFS/WritableStream.php
```diff
@@ -25,14 +25,8 @@
 use MongoDB\Exception\InvalidArgumentException;
 
 use function array_intersect_key;
-use function hash_final;
```
jmikola, line 28: You can additionally remove:

 * `private ?HashContext $hashCtx = null`
 * The corresponding `use` statement
 * The `ext-hash` requirement in `composer.json`

tests/UnifiedSpecTests/Util.php
jmikola: Removing them here presumably makes the spec tests fail due to unrecognized fields. Should those tests be skipped in the meantime? That can be done by additions to `UnifiedSpecTest::$incompleteTests` (using the test file's base name and test name).

You can open a DRIVERS ticket to propose updating/deleting the spec tests. Additionally, it looks like many of the spec tests fill in the corresponding deprecated fields when populating `fs.files`. The ticket can also propose removing those.

tests/UnifiedSpecTests/gridfs/upload.json
```diff
@@ -1,616 +0,0 @@
-{
```
jmikola: I assume this was unintentionally deleted, as `upload-disableMD5.json` also uses deprecated fields and wasn't touched.

jmikola (changes requested):

src/GridFS/WritableStream.php
GromNaN: It's great that we can eliminate a dependency, even if it's installed by default.

tests/UnifiedSpecTests/Util.php
GromNaN: Thanks for the directions. The failing test is now skipped and I created [DRIVERS-2976](https://jira.mongodb.org/browse/DRIVERS-2976)

tests/UnifiedSpecTests/UnifiedSpecTest.php
```diff
@@ -187,6 +187,8 @@ class UnifiedSpecTest extends FunctionalTestCase
         'crud/db-aggregate-write-readPreference: Database-level aggregate with $out omits read preference for pre-5.0 server' => 'PHPLIB-1458',
         'crud/db-aggregate-write-readPreference: Database-level aggregate with $merge includes read preference for 5.0+ server' => 'PHPLIB-1458',
         'crud/db-aggregate-write-readPreference: Database-level aggregate with $merge omits read preference for pre-5.0 server' => 'PHPLIB-1458',
+        // GridFS deprecated fields are removed
+        'gridfs/gridfs-upload: upload when contentType is provided' => 'Deprecated fields are removed',
```
jmikola, line 193: Did you also mean to disable the tests in `gridfs-upload-disableMD5.json`?

Also, can `disableMD5` and `contentType` be removed from the `uploadWithId` and `upload` arrays in `MongoDB\Tests\UnifiedSpecTests\Util::$args`? If you're skipping all tests including those deprecated options, I expect so.

UPGRADE-2.0.md
```diff
@@ -0,0 +1,28 @@
...
+GridFS
+------
+
+ * The `md5` is no longer calculated when a file is uploaded to GridFS.
+   Applications that require a file digest should implement it outside GridFS
+   and store in metadata.
+
+   ```php
+   $hash = hash_file('sha256', $filename);
+   $bucket->openDownloadStream($fileId, ['metadata' => ['hash' => $hash]]);
```
jmikola: I think all of these examples should be using `openUploadStream()` since they're demonstrating uploads. `openDownloadStream()` only takes a single `$id` argument.

Applies to the before/after examples in the subsequent bullet, too.

tests/UnifiedSpecTests/UnifiedSpecTest.php
GromNaN: It wasn't clear to me. There's no need for these tests, which now duplicate others.

UPGRADE-2.0.md
GromNaN: Updated.

jmikola (approved):

PR #1403, opened 2024-09-13
Replace `arrayHasKey` with `assertArrayHasKey` in tests

GromNaN:
The function `Assert::arrayHasKey()` constructs a Constraint object that need be applied to a value like this:
```php
Assert::arrayHasKey('key')->matches($array);
```

This PR switch to the correct assertion method.

### DIFF
--- a/tests/Operation/WatchFunctionalTest.php
+++ b/tests/Operation/WatchFunctionalTest.php
@@ -307,15 +307,15 @@ class WatchFunctionalTest extends FunctionalTestCase
         $this->assertCount(3, $events);
 
         $this->assertSame('getMore', $events[0]['started']->getCommandName());
-        $this->arrayHasKey('failed', $events[0]);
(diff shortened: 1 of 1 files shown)


alcaeus (approved):

tests/Operation/WatchFunctionalTest.php
```diff
@@ -307,15 +307,15 @@ function (array $event) use (&$events): void {
         $this->assertCount(3, $events);
 
         $this->assertSame('getMore', $events[0]['started']->getCommandName());
-        $this->arrayHasKey('failed', $events[0]);
```
jmikola, line 310: My mistake: https://github.com/mongodb/mongo-php-library/blame/3724afc329619018b7ff0c283c6374e4ce7a2cd5/tests/Operation/WatchFunctionalTest.php#L310

PR #1412, opened 2024-09-17
PHPLIB-1369 Upgrade to PHPUnit 10

GromNaN:
Fix PHPLIB-1369

- Migration from `@annotations` to `#[Attributes]`
- Replace `hasFailed` with the method `onNotSuccessfulTest` in `FunctionalTestCase`
- User `createStub` in `getInvalidDocumentCodecValues`
- Update custom constraints (`TestCase::exporter()` is deprecated and `ComparisonFailure` 5th argument was removed)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -75,7 +75,6 @@ jobs:
           php-ini-values: "zend.assertions=1"
 
       - name: "Run PHPUnit"
-        run: "vendor/bin/phpunit -v"
+        run: "vendor/bin/phpunit"
         env:
-          SYMFONY_DEPRECATIONS_HELPER: 999999
           MONGODB_URI: ${{ steps.setup-mongodb.outputs.cluster-uri }}
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
     "require-dev": {
         "doctrine/coding-standard": "^12.0",
         "phpunit/phpunit": "^9.6.11",
-        "rector/rector": "^1.1",
+        "rector/rector": "^1.2",
         "squizlabs/php_codesniffer": "^3.7",
         "vimeo/psalm": "^5.13"
     },
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
     },
     "require-dev": {
         "doctrine/coding-standard": "^12.0",
-        "phpunit/phpunit": "^9.6.11",
+        "phpunit/phpunit": "^10.5",
         "rector/rector": "^1.2",
         "squizlabs/php_codesniffer": "^3.7",
         "vimeo/psalm": "^5.13"
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
     },
     "require-dev": {
         "doctrine/coding-standard": "^12.0",
-        "phpunit/phpunit": "^10.5",
+        "phpunit/phpunit": "10.5.3",
         "rector/rector": "^1.2",
         "squizlabs/php_codesniffer": "^3.7",
         "vimeo/psalm": "^5.13"
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
     },
     "require-dev": {
         "doctrine/coding-standard": "^12.0",
-        "phpunit/phpunit": "10.5.3",
+        "phpunit/phpunit": "^10.5.35",
         "rector/rector": "^1.2",
         "squizlabs/php_codesniffer": "^3.7",
         "vimeo/psalm": "^5.13"
--- a/rector.php
+++ b/rector.php
@@ -5,6 +5,7 @@ use Rector\DeadCode\Rector\ClassLike\RemoveAnnotationRector;
 use Rector\Php70\Rector\StmtsAwareInterface\IfIssetToCoalescingRector;
 use Rector\Php80\Rector\Switch_\ChangeSwitchToMatchRector;
 use Rector\PHPUnit\PHPUnit100\Rector\Class_\StaticDataProviderClassMethodRector;
+use Rector\PHPUnit\Set\PHPUnitSetList;
 use Rector\Set\ValueObject\LevelSetList;
 
 return static function (RectorConfig $rectorConfig): void {
@@ -16,7 +17,10 @@ return static function (RectorConfig $rectorConfig): void {
     ]);
 
     // Modernize code
-    $rectorConfig->sets([LevelSetList::UP_TO_PHP_74]);
+    $rectorConfig->sets([
+        LevelSetList::UP_TO_PHP_74,
+        PHPUnitSetList::PHPUNIT_100,
+    ]);
 
     $rectorConfig->rule(ChangeSwitchToMatchRector::class);
     $rectorConfig->rule(StaticDataProviderClassMethodRector::class);
--- a/rector.php
+++ b/rector.php
@@ -3,6 +3,7 @@
 use Rector\Config\RectorConfig;
 use Rector\DeadCode\Rector\ClassLike\RemoveAnnotationRector;
 use Rector\Php70\Rector\StmtsAwareInterface\IfIssetToCoalescingRector;
+use Rector\Php71\Rector\FuncCall\RemoveExtraParametersRector;
 use Rector\Php80\Rector\Switch_\ChangeSwitchToMatchRector;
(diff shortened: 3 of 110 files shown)


composer.json
```diff
@@ -21,8 +21,8 @@
     },
     "require-dev": {
         "doctrine/coding-standard": "^12.0",
-        "phpunit/phpunit": "^9.6",
-        "rector/rector": "^1.1",
+        "phpunit/phpunit": "10.5.3",
```
GromNaN: Version locked, waiting for fix of https://jira.mongodb.org/browse/PHPC-2456

tests/FunctionalTestCase.php
```diff
@@ -69,12 +70,15 @@ public function setUp(): void
         $this->configuredFailPoints = [];
     }
 
-    public function tearDown(): void
+    protected function onNotSuccessfulTest(Throwable $t): never
```
alcaeus, line 73: Oh, I like this way of handling failed tests 💪

.github/workflows/tests.yml
```diff
@@ -77,7 +77,6 @@ jobs:
           php-ini-values: "zend.assertions=1"
 
       - name: "Run PHPUnit"
-        run: "vendor/bin/phpunit -v"
+        run: "vendor/bin/phpunit"
```
alcaeus, line 78: I was thinking about removing the output of skipped tests, so I agree with this change 👍

tests/Builder/BuilderEncoderTest.php
```diff
@@ -153,9 +154,8 @@ public function testPerformCount(): void
      *
      * @param list<int>          $limit
      * @param array<string, int> $expectedLimit
-     *
-     * @dataProvider provideExpressionFilterLimit
```
alcaeus, line 157: We've excluded the `ForbiddenAnnotations` sniff due to a different issue, but if you'd like we can try to re-enable it and add all of the old PHPUnit annotations to the list. Not sure if PHPUnit 10 still supports annotations, but if it does it may be worth doing to make sure muscle memory doesn't kick in when writing tests.

tests/TestCase.php
```diff
@@ -202,7 +200,7 @@ protected function getCollectionName(): string
     {
         $class = new ReflectionClass($this);
 
-        return sprintf('%s.%s', $class->getShortName(), hash('crc32b', $this->getName()));
+        return sprintf('%s.%s', $class->getShortName(), hash('crc32b', $this->name()));
```
GromNaN: I'll update to a faster algorithm
```suggestion
        return sprintf('%s.%s', $class->getShortName(), hash('xxh3', $this->name()));
```

.github/workflows/tests.yml
GromNaN: It's also, the `-v` option isn't available on PHPUnit 10

tests/Builder/BuilderEncoderTest.php
GromNaN: PHPUnit 10 still support annotations. But the rector rule will force usage of attributes.

composer.json
GromNaN: Reverted since https://github.com/mongodb/mongo-php-driver/pull/1683 is merged.

rector.php
```diff
@@ -16,13 +18,17 @@
     ]);
 
     // Modernize code
-    $rectorConfig->sets([LevelSetList::UP_TO_PHP_74]);
+    $rectorConfig->sets([
+        LevelSetList::UP_TO_PHP_74,
```
jmikola, line 22: Should this have been bumped to PHP 8.1 in https://github.com/mongodb/mongo-php-library/commit/24214b6e5a0380716357e5616c162d7025c974e6?

tests/Comparator/Int64Comparator.php
```diff
@@ -24,16 +25,17 @@ public function assertEquals($expected, $actual, $delta = 0.0, $canonicalize = f
             return;
         }
 
+        $exporter = new Exporter();
```
jmikola, line 28: Noted that the internal property was removed from the Comparator base class in https://github.com/sebastianbergmann/comparator/commit/bf22fc1b035ecf7a4e989e2a5ace1ceb06dd9bcf

tests/FunctionalTestCase.php
jmikola: Available since PHPUnit 3.4 ([sebastianbergmann/phpunit@8e6c10c](https://github.com/sebastianbergmann/phpunit/commit/8e6c10ced83a26394e8a1b5167c7ffa89aa34a99#diff-d7442c71b93b9933979f85ce8c9eaa035c469f6dd7c8fab3e32386b5ddfbad19R1452)), yet there's no trace of this in the documentation :shrug:

tests/Operation/FindFunctionalTest.php
```diff
@@ -36,11 +38,16 @@ function (array $event) use ($expectedQuery): void {
         );
     }
 
-    /** @dataProvider provideModifierDocuments */
+    #[DataProvider('provideModifierDocuments')]
     public function testModifierDocuments($modifiers, stdClass $expectedSort): void
     {
         (new CommandObserver())->observe(
             function () use ($modifiers): void {
+                // @todo revert this lines after PHPC-2457
```
jmikola, line 46: Technically, this should become irrelevant after https://github.com/mongodb/mongo-php-library/pull/1439

tests/PedantryTest.php
```diff
@@ -3,6 +3,7 @@
 namespace MongoDB\Tests;
 
 use MongoDB;
+use PHPUnit\Framework\Attributes\DataProvider;
```
jmikola, line 6: Unrelated to this PR, but I'm wondering what `use MongoDB;` above is used for. Looks like this dates back to https://github.com/mongodb/mongo-php-library/commit/eafad0580dc4638000c195e8c0f231964616beeb.

tests/TestCase.php
```diff
@@ -205,7 +208,7 @@ protected function getCollectionName(): string
     {
         $class = new ReflectionClass($this);
 
-        return sprintf('%s.%s', $class->getShortName(), hash('crc32b', $this->getName()));
+        return sprintf('%s.%s', $class->getShortName(), hash('xxh3', $this->name()));
```
jmikola, line 211: Noted: https://xxhash.com/ available since PHP 8.1 (see: [`hash_algos()`](https://www.php.net/manual/en/function.hash-algos.php)).

tests/UnifiedSpecTests/Constraint/Matches.php
```diff
@@ -256,7 +257,7 @@ private function assertMatchesOperator(BSONDocument $operator, $actual, string $
             $constraint = IsBsonType::anyOf(...(array) $operator['$$type']);
 
             if (! $constraint->evaluate($actual, '', true)) {
-                self::failAt(sprintf('%s is not an expected BSON type: %s', $this->exporter()->shortenedExport($actual), implode(', ', $types)), $keyPath);
```
jmikola, line 259: Noted that `$types` was never defined. This dates back to [4608a73](https://github.com/mongodb/mongo-php-library/commit/4608a734993be0b0aa1e0e838fa52e4caa9bfff2#diff-5911ac585872dc4a55ff2fe77af552d46e160b4d5636699c9cf2c1211217e25eR250). Curious that it wasn't caught by static analysis, unless this snuck into the baseline.

jmikola (approved):
One question about the Rector config, but everything else LGTM.

alcaeus (approved):
LGTM

tests/Comparator/Int64Comparator.php
GromNaN: The PHP 8.1 rules have to be tuned. Some like `ReadOnlyPropertyRector` and `ClassPropertyAssignToConstructorPromotionRector` need to be challenged.
https://jira.mongodb.org/browse/PHPLIB-1536

tests/PedantryTest.php
GromNaN: This prevent duplication of class names as `use` statement and in the `SKIPPED_CLASSES` list.

rector.php
GromNaN: The PHP 8.1 rules have to be tuned. Some like ReadOnlyPropertyRector and ClassPropertyAssignToConstructorPromotionRector need to be challenged.
https://jira.mongodb.org/browse/PHPLIB-1536

PR #1413, opened 2024-09-17
PHPLIB-1525 Removes dependency to Symfony PHPUnit bridge

GromNaN:
Fix PHPLIB-1525

> The PHPUnit Bridge provides utilities to report legacy tests and usage of deprecated code and helpers for mocking native functions related to time, DNS and class existence.

We don't use any of this features.

The bridge also helps in maintaining custom constraints compatible with [multiple versions of PHPUnit](https://github.com/symfony/phpunit-bridge/blob/7.0/ConstraintTrait.php), due to type changes between versions 7, 8 and 9. This should longer be required as the [Constraint class](https://github.com/sebastianbergmann/phpunit/blob/9.6/src/Framework/Constraint/Constraint.php#L43) is fully typed.

drivers-atlas-testing will require an update:
- https://github.com/mongodb-labs/drivers-atlas-testing/blob/784b9e5e4656e207003cac7d6bc9dd7a551b800f/integrations/php/install-driver.sh#L85-L86
- https://github.com/mongodb-labs/drivers-atlas-testing/blob/784b9e5e4656e207003cac7d6bc9dd7a551b800f/integrations/php/workload-executor.php#L11

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,12 +43,6 @@ credentials in the connection string (i.e. `MONGODB_URI`) or set the
 Note that `MONGODB_USERNAME` and `MONGODB_PASSWORD` will override any
 credentials present in the connection string.
 
-To run tests use the `phpunit` executable installed by Composer:
-
-```console
-$ vendor/bin/phpunit
-```
-
 ### Environment Variables
 
 The test suite references the following environment variables:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,12 +43,10 @@ credentials in the connection string (i.e. `MONGODB_URI`) or set the
 Note that `MONGODB_USERNAME` and `MONGODB_PASSWORD` will override any
 credentials present in the connection string.
 
-By default, the `simple-phpunit` binary chooses the correct PHPUnit version for
-the PHP version you are running. To run tests against a specific PHPUnit
-version, use the `SYMFONY_PHPUNIT_VERSION` environment variable:
+To run tests use the `phpunit` executable installed by Composer:
 
 ```console
-$ SYMFONY_PHPUNIT_VERSION=8.5 vendor/bin/simple-phpunit
+$ vendor/bin/phpunit
 ```
 
 ### Environment Variables
--- a/tests/ExamplesTest.php
+++ b/tests/ExamplesTest.php
@@ -5,10 +5,14 @@ namespace MongoDB\Tests;
 use Generator;
 
 use function bin2hex;
+use function error_reporting;
 use function getenv;
 use function putenv;
 use function random_bytes;
 use function sprintf;
+
+use const E_ALL;
+use const E_DEPRECATED;
 
 /** @runTestsInSeparateProcesses */
 final class ExamplesTest extends FunctionalTestCase
@@ -304,7 +308,14 @@ OUTPUT;
 
     private function assertExampleOutput(string $file, string $expectedOutput): void
     {
-        require $file;
+        // Hide deprecation messages
+        $level = error_reporting(E_ALL ^ E_DEPRECATED);
+
+        try {
+            require $file;
+        } finally {
+            error_reporting($level);
+        }
 
         $this->assertStringMatchesFormat($expectedOutput, $this->getActualOutputForAssertion());
     }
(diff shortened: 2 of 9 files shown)


tests/GridFS/StreamWrapperFunctionalTest.php
```diff
@@ -46,7 +46,7 @@ public function setUp(): void
         parent::setUp();
 
         $this->filesCollection->insertMany([
-            ['_id' => 'length-10', 'length' => 10, 'chunkSize' => 4, 'uploadDate' => new UTCDateTime('1484202200000')],
```
GromNaN: > PHP Deprecated:  MongoDB\BSON\UTCDateTime::__construct(): Creating a MongoDB\BSON\UTCDateTime instance with a string is deprecated and will be removed in ext-mongodb 2.0

tests/ExamplesTest.php
```diff
@@ -304,7 +308,14 @@ public function testWithTransaction(): void
 
     private function assertExampleOutput(string $file, string $expectedOutput): void
     {
-        require $file;
+        // Hide deprecation messages
+        $level = error_reporting(E_ALL ^ E_DEPRECATED);
```
GromNaN: > PHP Deprecated:  Method MongoDB\Driver\Monitoring\CommandSucceededEvent::getServer() is deprecated in [src/Operation/Watch.php on line 320](https://github.com/mongodb/mongo-php-library/blob/524c8585a08314050fe7e83153108becfb314181/src/Operation/Watch.php#L320)

> PHP Deprecated:  MongoDB\Driver\Cursor::getId(): The method "MongoDB\Driver\Cursor::getId" will no longer return a "MongoDB\Driver\CursorId" instance in the future. Pass "true" as argument to change to the new behavior and receive a "MongoDB\BSON\Int64" instance instead. in [src/ChangeStream.php on line 128](https://github.com/mongodb/mongo-php-library/blob/524c8585a08314050fe7e83153108becfb314181/src/ChangeStream.php#L128)

GromNaN: See [PHPLIB-1529](https://jira.mongodb.org/browse/PHPLIB-1529)

alcaeus: Ah, thanks for filing this. Looks like this could've been removed when I dropped support for MongoDB 3.6: https://github.com/mongodb/mongo-php-library/commit/de3aaf881dcf8a4807529bf9b1ad4d609d101107#diff-97711093784103ece28740e725cc6a5f44bba1a469d5dcb78ef94a85778b3549L473

I'm a bit surprised that we don't have a sniff that alerts us to unused arguments 🤔

GromNaN: `psalm --find-dead-code` returns only false positive.

CONTRIBUTING.md
```diff
@@ -43,12 +43,10 @@ credentials in the connection string (i.e. `MONGODB_URI`) or set the
...
 credentials present in the connection string.
 
-By default, the `simple-phpunit` binary chooses the correct PHPUnit version for
-the PHP version you are running. To run tests against a specific PHPUnit
-version, use the `SYMFONY_PHPUNIT_VERSION` environment variable:
+To run tests use the `phpunit` executable installed by Composer:
 
 ```console
-$ SYMFONY_PHPUNIT_VERSION=8.5 vendor/bin/simple-phpunit
+$ vendor/bin/phpunit
```
alcaeus: We document running the test suite with `composer run test` a few lines above (can't add a comment to the corresponding context line). We could document the direct way of calling the `phpunit` binary up there and remove this entire paragraph.

tests/ExamplesTest.php
alcaeus: Is this still necessary after the PRs handling deprecations have been merged?

tests/Model/CodecCursorFunctionalTest.php
```diff
@@ -55,12 +55,15 @@ function (...$args) use (&$previousErrorHandler, &$deprecations) {
                 E_USER_DEPRECATED | E_DEPRECATED,
             );
 
-            $cursorId = $codecCursor->getId();
+            $cursorId = $codecCursor->getId(true);
```
alcaeus: With this change the error handler should no longer be necessary, and there should not be any deprecations to assert against.

GromNaN: Yes, I released this should be reverted if we want to validate the deprecation message.

alcaeus (approved):

PR #1414, opened 2024-09-18
PHPLIB-1528 Add traits to implement Command and SDAM subscriber methods

GromNaN:
Fix PHPLIB-1528

Make it easier to write partial implementations of interfaces `CommandSubscriber` and `SDAMSubscriber`.

Not adding a trait for `LogSubscriber`, as this interface contains only 1 method.


alcaeus (approved):
LGTM!

GromNaN:
Waiting investigation if we implement this in C in the mongodb extension.

PR #1421, opened 2024-09-18
Performance: Keep collections and indexes between GridFS tests

GromNaN:
Performance impact on my Mac M1 Pro, PHP 8.3, local replica set on MongoDB 8.

Before:

```
Time: 01:42.729, Memory: 36.00 MB

OK (319 tests, 3164 assertions)
```

After:
```
Time: 00:17.596, Memory: 36.00 MB

OK (319 tests, 3164 assertions)
```


tests/GridFS/BucketFunctionalTest.php
```diff
@@ -814,6 +814,10 @@ public function testUploadingFirstFileCreatesIndexes(): void
 
     public function testExistingIndexIsReused(): void
     {
+        // The collections may exist from other tests, ensure they are removed before and after the test
+        $this->dropCollection($this->getDatabaseName(), 'fs.chunks');
+        $this->dropCollection($this->getDatabaseName(), 'fs.files');
```
GromNaN, line 820: By calling `dropCollection`, the collection is also removed on `tearDown`.

tests/GridFS/BucketFunctionalTest.php
```diff
@@ -814,6 +814,10 @@ public function testUploadingFirstFileCreatesIndexes(): void
...
+
```
jmikola, line 821: I'm curious about the mixed use of `1.0` and `1` when creating the indexes below.

@alcaeus: it looks like you added this in [79c2993](https://github.com/mongodb/mongo-php-library/commit/79c299361f1e23e4cf2706af66bbf813ea148de2#diff-784bd4066ea3d6bdb4a937fa2e99c9a64efe89c4edb28944d5e640348e43355e) for [PHPLIB-522](https://jira.mongodb.org/browse/PHPLIB-522) ("loosening index equality checks"). Assuming this is still relevant, is there a short comment we can add here explaining what's being tested? I don't think `testExistingIndexIsReused()` really captures what's going on here.

Maybe a multi-line comment:

> Create indexes with different numeric types before interacting with GridFS to assert that PHPLIB respects the existing indexes and does not attempt to create its own.

I'm also curious if there should have been `assertIndexExists()` calls for the two indexes we _did_ create, although I assume we're relying on a successful command response from the server.

tests/GridFS/FunctionalTestCase.php
```diff
@@ -28,10 +29,26 @@ public function setUp(): void
...
+
+    public function tearDown(): void
+    {
+        $this->chunksCollection->deleteMany([]);
+        $this->filesCollection->deleteMany([]);
+
+        parent::tearDown();
+    }
+
+    public static function tearDownAfterClass(): void
```
jmikola: Would it be prudent to add a `setUpBeforeClass()` method to ensure the collections are dropped and empty going into the tests?

tests/GridFS/FunctionalTestCase.php
```diff
@@ -28,10 +29,26 @@ public function setUp(): void
...
 
-        $this->chunksCollection = $this->createCollection($this->getDatabaseName(), 'fs.chunks');
-        $this->filesCollection = $this->createCollection($this->getDatabaseName(), 'fs.files');
+        $this->chunksCollection = new Collection($this->manager, $this->getDatabaseName(), 'fs.chunks');
+        $this->filesCollection = new Collection($this->manager, $this->getDatabaseName(), 'fs.files');
+    }
+
+    public function tearDown(): void
+    {
+        $this->chunksCollection->deleteMany([]);
```
jmikola, line 39: Could we add a comment here explaining that we're leaving the collections in place as an optimization to avoid repeated index creation?

Offhand, I don't know if the real performance improvement here is from reusing collections or indexes (or a mix of both). If collection re-use is a big contributor, I expect much of PHPLIB's test suite could be further optimized by re-using the same collection name. That might also hold true for spec tests, although some other drivers might rely on different names to execute tests in parallel (we've yet to explore that in PHPLIB).

tests/GridFS/BucketFunctionalTest.php
jmikola: My suggestion to explain the optimization in a comment within `tearDown()` would pair nicely with this comment.

tests/GridFS/BucketFunctionalTest.php
jmikola: Note: this is admittedly unrelated from this PR, so feel free to disregard. I can also address this in a separate PR myself, but wanted to ask for some context first.

GromNaN: I added a comment and assertions on index existence.

tests/GridFS/FunctionalTestCase.php
GromNaN: Added a comment to the new static function `dropCollectionsBeforeAfterClass`.

jmikola (approved):

tests/GridFS/BucketFunctionalTest.php
alcaeus: Oof, trying to think back to it. I believe the issue was that the shell used slightly different index settings, resulting in valid indexes that weren't recognised. It may have been an issue with using a double for the index direction instead of an int. I don't remember precisely though, indicating that I should've left a comment at the time.

PR #1423, opened 2024-09-19
Fix `CursorId` and `Event::getServer()` deprecation in tests

GromNaN:
Related to https://github.com/mongodb/mongo-php-library/pull/1366

### DIFF
--- a/tests/Model/CodecCursorFunctionalTest.php
+++ b/tests/Model/CodecCursorFunctionalTest.php
@@ -55,12 +55,15 @@ class CodecCursorFunctionalTest extends FunctionalTestCase
                 E_USER_DEPRECATED | E_DEPRECATED,
             );
 
-            $cursorId = $codecCursor->getId();
+            $cursorId = $codecCursor->getId(true);
         } finally {
             restore_error_handler();
         }
 
-        self::assertInstanceOf(CursorId::class, $cursorId);
+        self::logicalOr(
+            self::isInstanceOf(Int64::class),
+            self::isInstanceOf(CursorId::class),
+        )->matches($cursorId);
 
         // Expect 2 deprecations: 1 from CodecCursor, one from Cursor
         self::assertCount(2, $deprecations);
--- a/tests/Model/CodecCursorFunctionalTest.php
+++ b/tests/Model/CodecCursorFunctionalTest.php
@@ -55,15 +55,12 @@ class CodecCursorFunctionalTest extends FunctionalTestCase
                 E_USER_DEPRECATED | E_DEPRECATED,
             );
 
-            $cursorId = $codecCursor->getId(true);
+            $cursorId = $codecCursor->getId();
         } finally {
             restore_error_handler();
         }
 
-        self::logicalOr(
-            self::isInstanceOf(Int64::class),
-            self::isInstanceOf(CursorId::class),
-        )->matches($cursorId);
+        self::assertInstanceOf(CursorId::class, $cursorId);
 
         // Expect 2 deprecations: 1 from CodecCursor, one from Cursor
         self::assertCount(2, $deprecations);
--- a/tests/SpecTests/RetryableReads/Prose2_RetryOnMongosTest.php
+++ b/tests/SpecTests/RetryableReads/Prose2_RetryOnMongosTest.php
@@ -65,7 +65,8 @@ class Prose2_RetryOnMongosTest extends FunctionalTestCase
 
         // Step 4: Enable failed command event monitoring for client
         $subscriber = new class implements CommandSubscriber {
-            public $commandFailedServers = [];
+            /** @var int[] */
+            public array $commandFailedServers = [];
 
             public function commandStarted(CommandStartedEvent $event): void
             {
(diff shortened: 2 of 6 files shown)


tests/Model/CodecCursorFunctionalTest.php
```diff
@@ -55,12 +55,15 @@ function (...$args) use (&$previousErrorHandler, &$deprecations) {
...
+            $cursorId = $codecCursor->getId(true);
         } finally {
             restore_error_handler();
         }
 
-        self::assertInstanceOf(CursorId::class, $cursorId);
+        self::logicalOr(
+            self::isInstanceOf(Int64::class),
+            self::isInstanceOf(CursorId::class),
+        )->matches($cursorId);
```
GromNaN: This assertion can be removed in 2.0 when return type is enforced.

alcaeus (approved):

GromNaN:
I added fixes for calls to `Event::getServer()`, using `getServerConnectionId()` instead.
But for the `FailPointsOberver`, I need an actual `Server` instance. This requires a larger refactoring.

https://github.com/mongodb/mongo-php-library/blob/b27a61d7c13d4353e8b2acee23beddb1d5539dd6/tests/UnifiedSpecTests/FailPointObserver.php#L36

https://github.com/mongodb/mongo-php-library/blob/b27a61d7c13d4353e8b2acee23beddb1d5539dd6/tests/UnifiedSpecTests/FailPointObserver.php#L46-L49

Tracked in PHPLIB-1531

tests/UnifiedSpecTests/EventCollector.php
```diff
@@ -120,7 +120,7 @@ private function handleCommandMonitoringEvent($event): void
             'name' => self::getEventName($event),
             'observedAt' => microtime(true),
             'commandName' => $event->getCommandName(),
-            'connectionId' => self::getConnectionId($event),
+            'connectionId' => $event->getServerConnectionId(),
```
GromNaN, line 122: This logs and the `$eventList` property are never used. They can be useful for debugging.

tests/SpecTests/RetryableReads/Prose2_RetryOnMongosTest.php
```diff
@@ -65,7 +64,8 @@ public function testRetryOnDifferentMongos(): void
 
         // Step 4: Enable failed command event monitoring for client
         $subscriber = new class implements CommandSubscriber {
-            public $commandFailedServers = [];
+            /** @var int[] */
+            public array $commandFailedServers = [];
```
jmikola, line 68: Later, this was used for the following assertion: 

```php
/* Step 6: Assert that two failed command events occurred. Assert that
 * the failed command events occurred on different mongoses. */
```

Per [`getServerConnectionId()`](https://www.php.net/manual/en/mongodb-driver-monitoring-commandstartedevent.getserverconnectionid.php), the connection ID is unique within the context of a single server. It is also only reported by MongoDB 4.2+. I think the correct change here would have been to replace `getServer()` with a concatenation of the server host/port. I'll make the change in a subsequent PR.

As-is, I think this might sometimes fail if both mongoses reported the same connection ID or if we test on mongos 4.0.

tests/SpecTests/RetryableReads/Prose2_RetryOnMongosTest.php
```diff
@@ -130,21 +130,21 @@ public function testRetryOnSameMongos(): void
 
         // Step 4: Enable succeeded and failed command event monitoring
         $subscriber = new class implements CommandSubscriber {
-            public Server $commandSucceededServer;
-            public Server $commandFailedServer;
+            public ?int $commandSucceededServer = null;
+            public ?int $commandFailedServer = null;
```
jmikola, line 134: This relates the following assertion:

```php
/* Step 6: Assert that exactly one failed command event and one
 * succeeded command event occurred. Assert that both events occurred on
 * the same mongos. */
```

Those rely on `assertNotNull()`, so this could also fail on a 4.0 server where connection IDs are not reported.

tests/UnifiedSpecTests/EventCollector.php
jmikola: Quoting the [Workload Executor Specification](https://drivers-atlas-testing.readthedocs.io/en/latest/spec-workload-executor/) from [drivers-atlas-testing](https://github.com/mongodb-labs/drivers-atlas-testing):

> each object is expected to have a name string field and an observedAt numeric field, in addition to any other fields specific to the event's type.

The only other field here that corresponds to an existing field in the Unified Test Format spec's [expectedEvent](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.md#expectedevent) section is `commandName`. That said, I think the original intention of `connectionId` here was from the _client's_ perspective, so this should also change to a host/port combination. I'd also propose renaming this to `server` and adding `serverConnectionId` (which may not have existed at the time we implemented Atlas testing). I'll address that in my forthcoming PR to address my other feedback.

jmikola: See: https://github.com/mongodb/mongo-php-library/pull/1428

PR #1425, opened 2024-09-19
Remove compatibility layer for CursorId deprecation

alcaeus:
This pull request removes the compatibility layer introduced to handle the deprecation of `MongoDB\Driver\CursorId`. To work around psalm not having updated stubs, we introduce stubs temporarily until the stubs are properly updated.

Note that due to an issue in psalm, we still need a workaround for `ChangeStreamIterator::getInnerIterator`. I've created https://github.com/vimeo/psalm/pull/11100 to fix this issue, and the errors are suppressed in the baseline until a fix is released.

### DIFF
--- a/stubs/Driver/Cursor.stub.php
+++ b/stubs/Driver/Cursor.stub.php
@@ -1,0 +1,41 @@
+<?php
+
+namespace MongoDB\Driver;
+
+/**
+ * @template-covariant TValue of array|object
+ *
+ * @template-implements CursorInterface<TValue>
+ */
+final class Cursor implements CursorInterface
+{
+    /**
+     * @return TValue|null
+     * @psalm-ignore-nullable-return
+     */
+    public function current(): array|object|null
+    {
+    }
+
+    public function next(): void
+    {
+    }
+
+    /** @psalm-ignore-nullable-return */
+    public function key(): ?int
+    {
+    }
+
+    public function valid(): bool
+    {
+    }
+
+    public function rewind(): void
+    {
+    }
+
+    /** @return array<TValue> */
+    public function toArray(): array
+    {
+    }
+}
--- a/stubs/Driver/CursorInterface.stub.php
+++ b/stubs/Driver/CursorInterface.stub.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace MongoDB\Driver;
+
+use Iterator;
+
+/**
+ * @template TValue of array|object
+ * @template-implements Iterator<int, TValue>
+ */
+interface CursorInterface extends Iterator
+{
+    /**
+     * @return TValue|null
+     * @psalm-ignore-nullable-return
+     */
+    public function current(): array|object|null;
+
+    public function getId(): \MongoDB\BSON\Int64;
+
+    public function getServer(): Server;
+
+    public function isDead(): bool;
+
+    /** @psalm-ignore-nullable-return */
+    public function key(): ?int;
+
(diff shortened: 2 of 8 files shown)


alcaeus:
The CI build fails as we're not testing with the 2.0 version yet. This depends on https://github.com/mongodb/mongo-php-driver/pull/1672. It may be necessary to wrap this work into #1391.

stubs/Driver/Cursor.stub.php
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * @template-covariant TValue of array|object
+ *
+ * @template-implements CursorInterface<TValue>
+ */
+final class Cursor implements CursorInterface
+{
+    /**
+     * @return TValue|null
+     * @psalm-ignore-nullable-return
```
jmikola, line 14: IIUC, [`@psalm-ignore-nullable-return`](https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-ignore-nullable-return) tells Psalm to never consider a null return value. But `current()` and `key()` can certainly return `null` if `valid()` is `false`. Why needn't we consider that case?

stubs/Driver/CursorInterface.stub.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+interface CursorInterface extends Iterator
+{
+    /**
+     * @return TValue|null
+     * @psalm-ignore-nullable-return
+     */
+    public function current(): array|object|null;
+
+    public function getId(): \MongoDB\BSON\Int64;
```
jmikola, line 19: I imagine this was copied from PHPC which doesn't support `use` statements in its stubs, but would Psalm normally prefer a `use` statement for this?

jmikola (approved):
Some questions, but I don't think I have any actionable feedback.

stubs/Driver/Cursor.stub.php
alcaeus: This was blatantly copied over from [psalm's iterator stubs](https://github.com/vimeo/psalm/blob/d747f6500b38ac4f7dfc5edbcae6e4b637d7add0/stubs/CoreGenericIterators.phpstub#L27). The way I see it, this is a shortcoming of the original interface. The interface doesn't specify whether `current` and `key` should throw when on an invalid position, and most iterators return `null` instead. IMO, we should throw, but I imagine a lot of code isn't exactly built to handle throwing in the methods. At the same time, we do expect people to not call `current` and `key` without checking that the iterator is on a valid position, hence that annotation.

For immutable classes, the `psalm-assert-if-true` annotation would solve this, but since `next` has to reset a previously memoized result from `valid` I'm not sure if psalm can handle it. It would definitely be a worthwhile improvement.

stubs/Driver/CursorInterface.stub.php
alcaeus: Yep, good point. Will update this

alcaeus:
Folded into #1391

stubs/Driver/Cursor.stub.php
jmikola: Thanks for the explanation.

I imagine throwing would be incredibly annoying, as manual iteration is already a complicated subject. I reckon the average PHP developer isn't familiar with the sequence of method calls to mimic a `foreach` loop.

Not so much the case with our write result getters relying on `isAcknowledged()`, as that's easier to wrap one's head around.

PR #1427, opened 2024-09-19
PHPLIB-1531: Replace FailPointObserver with different logic

alcaeus:
PHPLIB-1531

`FailPointObserver` relies on APM events exposing a server instance to record servers with enabled fail points. Since the method is deprecated, the context will now store all fail points created during the execution of a test, and the test runner will disable all fail points recorded in the context. This will achieve the same effect while allowing us to remove `getServer` from APM classes.


tests/UnifiedSpecTests/Context.php
```diff
@@ -232,6 +236,20 @@ public function stopEventCollectors(): void
         }
     }
 
+    public function registerFailPoint(stdClass $failPoint, Server $server): void
+    {
+        $this->failPoints[] = [
+            'failPoint' => $failPoint,
+            'server' => $server,
+        ];
```
GromNaN: Do you want to adopt Laravel' Style?
```suggestion
        $this->failPoints[] = compact('failPoint', 'server');
```

jmikola: This is Laravel style? I was using `compact()` back when Taylor was still a C# developer.

jmikola: This is Laravel style? I was using `compact()` back when Taylor was still writing C#.



----

On a serious note, I've considered using this in PHPLIB but opted to avoid the function call. It'd be less a concern in `tests/` than `src/`, though. I also wonder if the tooling is smart enough to not complain about variables being assigned and never referenced.

GromNaN: > This is Laravel style?

Yep https://github.com/search?q=repo%3Alaravel%2Fframework%20compact&type=code

That was more a joke. I find it hard to read.

tests/UnifiedSpecTests/UnifiedTestRunner.php
```diff
@@ -235,6 +231,15 @@ private function doTestCase(stdClass $test, string $schemaVersion, ?array $runOn
         }
     }
 
+    /** @param list<array{failPoint: stdClass, server: Server}> $failPoints */
+    private function disableFailPoints(array $failPoints): void
+    {
+        foreach ($failPoints as ['failPoint' => $failPoint, 'server' => $server]) {
+            $operation = new DatabaseCommand('admin', ['configureFailPoint' => $failPoint, 'mode' => 'off']);
```
jmikola:
```suggestion
            $operation = new DatabaseCommand('admin', ['configureFailPoint' => $failPoint->configureFailPoint, 'mode' => 'off']);
```

@alcaeus: This should resolve the runtime error you encountered.

That said, I came up with another PR that consolidates this logic into a ManagesFailPointsTrait, which is closer to what we had with FailPointObserver. These files are already quite complex, so I wanted to minimize additional clutter. I'll defer to you, though.

See: https://github.com/alcaeus/mongo-php-library/pull/1

tests/UnifiedSpecTests/Context.php
alcaeus: Glad I waited with my response, that way I can keep my true feelings about `compact` to myself 🤫

jmikola (approved):

PR #1429, opened 2024-09-20
PHPLIB-1541: Include specs repository as a submodule

alcaeus:
This is a proof of concept that I'd like to try out for a few weeks. The idea is to do away with manual spec test updates and automate the process in a simple manner. To achieve this, we include the specifications as a git submodule and run the tests directly from there. Using dependabot, we get a weekly PR to update the submodule if any changes have been merged, and as long as the build is green we can merge the PR and close any PHP tickets associated to the linked drivers tickets.

The downside to this is that all our specs will be at the same versions. It should be possible to leverage test skips to work around any issues, but only time will tell. Note that undoing this will be relatively straightforward: we can copy the last set of tests that passed to the original test locations, then revert all of the commits from this PR.


.evergreen/config/functions.yml
```diff
@@ -4,6 +4,11 @@ functions:
     - command: git.get_project
       params:
         directory: "src"
+    - command: shell.exec
+      params:
+        working_dir: "src"
+        script: |
+          git submodule update --init
```
GromNaN: I've just confirmed that this command checkout the submodule on the version committed in the main project. To update the submodule, I go to the directory and checkout another commit, then commit the reference to the main project.

GromNaN (commented):
Could you add a pre-install/pre-update hook in composer.json to update the submodule? So that we don't have to think about it.

alcaeus:
> Could you add a pre-install/pre-update hook in composer.json to update the submodule? So that we don't have to think about it.

Added this command. I've used `pre-install` and `pre-update` for hooks, which relies on `git` information to be present. If this is not the case, `composer install` and `composer update` will fail.

CONTRIBUTING.md
```diff
@@ -108,6 +109,35 @@ The following environment variables are used for [CSFLE testing](https://github.
...
+   test runner, skip the corresponding test with a reference to the ticket that
+   covers implementing the new features
+ * If the test failure points to a bug in the spec, consider the effort required
+   to fix the failure. If it's a small change, commit and push the fix directly
+   to the pull request. Otherwise, skip the test with a reference to a ticket to
+   fix the failing test.
+
+The goal is that the library passes tests with the latest spec version at all
+times, either by implementing small changes quickly, or by skipping tests as
+necessary.
```
alcaeus, line 131: @jmikola this is a first shot at covering potential failures in a dependabot PR. Let me know if I missed something or if we should change the guidance. Either way, this would be a learning experience from us and might require some care when implementing spec changes (e.g. don't modify existing test cases when adding functionality).

alcaeus:
Updated to include changes for PHPLIB-1458.

GromNaN (approved):

CONTRIBUTING.md
jmikola: The three bullets above seem sufficient.

I'm not sure what you mean by "don't modify existing test cases when adding functionality". I assume you're not talking about making custom modifications to the JSON files, as I've never done that (although I recall libmongoc has).

CONTRIBUTING.md
```diff
@@ -13,7 +13,8 @@ $ composer update
 
 In addition to installing project dependencies, Composer will check that the
 required extension version is installed. Directions for installing the extension
-may be found [here](https://php.net/manual/en/mongodb.installation.php).
+may be found [here](https://php.net/manual/en/mongodb.installation.php). Composer will also install the submodule required for
+running spec tests.
```
jmikola:
```suggestion
may be found [here](https://php.net/manual/en/mongodb.installation.php).
Composer will also install the submodule required for running spec tests.
```

Consistent wrapping.

tests/SpecTests/ClientSideEncryption/Prose22_RangeExplicitEncryptionTest.php
```diff
@@ -38,6 +38,8 @@ class Prose22_RangeExplicitEncryptionTest extends FunctionalTestCase
     private ?Client $encryptedClient = null;
     private $key1Id;
 
+    private static string $specDir = __DIR__ . '/../../specifications/source/client-side-encryption';
```
jmikola, line 41: Any reason not to use a `const` or `readonly` here? Doesn't make much difference -- just curious. The important thing is that it's concise to reference later.

CONTRIBUTING.md
alcaeus: I'm referring to when changes are made in the specs repository. For example, new functionality should rely on new tests as much as possible, without changing existing tests. This allows skipping new tests without sacrificing code coverage because we'd have to skip a test that we were previously able to run.

tests/UnifiedSpecTests/Operation.php
```diff
@@ -235,6 +235,10 @@ private function executeForChangeStream(ChangeStream $changeStream)
 
     private function executeForClient(Client $client)
     {
+        if ($this->name === 'clientBulkWrite') {
+            Assert::markTestSkipped('clientBulkWrite operation is not implemented');
+        }
```
jmikola: I was going to ask why you implemented the check here, but then realized that `Util::assertArgumentsBySchema()` asserts that that the operation is supported. That actually makes the `switch` statement's default case unreachable; although I'm not sure if tooling is smart enough to not complain about that.

I do think we should revisit how we track skipped tests, as splitting logic between the Operation class and test runner will ultimately make things harder to track. I've long wanted to support regex patterns to avoid enumerating individual test files and cases. This kind of feature detection also makes sense, although it would ideally be consolidated in a single place. Opened [PHPLIB-1540](https://jira.mongodb.org/browse/PHPLIB-1540) to track that idea.

tests/UnifiedSpecTests/UnifiedSpecTest.php
```diff
@@ -178,19 +179,22 @@ class UnifiedSpecTest extends FunctionalTestCase
...
-        'crud/aggregate-write-readPreference: Aggregate with $out omits read preference for pre-5.0 server' => 'PHPLIB-1458',
-        'crud/aggregate-write-readPreference: Aggregate with $merge includes read preference for 5.0+ server' => 'PHPLIB-1458',
-        'crud/aggregate-write-readPreference: Aggregate with $merge omits read preference for pre-5.0 server' => 'PHPLIB-1458',
-        'crud/db-aggregate-write-readPreference: Database-level aggregate with $out includes read preference for 5.0+ server' => 'PHPLIB-1458',
-        'crud/db-aggregate-write-readPreference: Database-level aggregate with $out omits read preference for pre-5.0 server' => 'PHPLIB-1458',
-        'crud/db-aggregate-write-readPreference: Database-level aggregate with $merge includes read preference for 5.0+ server' => 'PHPLIB-1458',
-        'crud/db-aggregate-write-readPreference: Database-level aggregate with $merge omits read preference for pre-5.0 server' => 'PHPLIB-1458',
     ];
 
+    private static array $duplicateTests = ['crud/client bulkWrite partial results: partialResult is set when first operation fails during an unordered bulk write (summary)'];
```
jmikola: I assume this is related to a non-unique test name. Consider adding a comment here, as this should ultimately be temporary and something we can revisit after enforcing uniqueness in the specs repo.

jmikola (approved):
Some suggestions but LGTM. Please create a ticket for this (and amend the commit message) before merging.

tests/UnifiedSpecTests/Operation.php
```diff
@@ -166,6 +184,7 @@ private function execute()
         $object = $this->entityMap[$this->object];
         assertIsObject($object);
 
+        $this->skipIfOperationIsNotSupported($object::class);
```
jmikola, line 187: I realize `Util::assertArgumentsBySchema()` accepts either a class name or `Operation::OBJECT_TEST_RUNNER`; however, an internal skip method could solicit the object and operation names as strings (i.e. `$o->object` and `$o->name`). If so, it could be a `static` method that's called immediately after the type assertions in the constructor:

```php
assertIsString($o->name);
$this->name = $o->name;

assertIsString($o->object);
$this->isTestRunnerOperation = $o->object === self::OBJECT_TEST_RUNNER;
$this->object = $this->isTestRunnerOperation ? null : $o->object;

self::skipIfOperationIsNotSupported($o->object, $o->name);
```

Just a suggestion. Feel free to disregard if you prefer the current approach.

tests/UnifiedSpecTests/UnifiedSpecTest.php
```diff
@@ -221,6 +99,12 @@ public function setUp(): void
         if ($this->isLoadBalanced() && isset(self::$incompleteLoadBalancerTests[$this->dataDescription()])) {
             $this->markTestIncomplete(self::$incompleteLoadBalancerTests[$this->dataDescription()]);
         }
+
+        foreach (self::$incompleteTestGroups as $testGroup => $reason) {
+            if (preg_match(sprintf('#^%s#', $testGroup), $this->dataDescription())) {
```
jmikola: Consider just using [`str_starts_with()`](https://www.php.net/str_starts_with).

I think a regex makes sense if we actually want to store pattern strings in the static array, but in that case I wouldn't apply _any_ processing here and rely on those strings including an anchor or trailing `.*`.

tests/UnifiedSpecTests/UnifiedSpecTest.php
```diff
@@ -484,16 +368,26 @@ public function testIndexManagement(UnifiedTestCase $test): void
...
+
         foreach (glob($pattern) as $filename) {
-            $group = basename(dirname($filename));
+            $testGroup = $group ?? dirname(basename($filename));
 
             foreach (UnifiedTestCase::fromFile($filename) as $name => $test) {
-                yield $group . '/' . $name => [$test];
+                $testKey = $testGroup . '/' . $name;
+
+                if (isset($duplicateTests[$testKey])) {
```
jmikola, line 395: Do I understand correctly that the duplicate test name is only a problem because PHPUnit asserts that the data provider yields unique keys?

tests/UnifiedSpecTests/UnifiedSpecTest.php
```diff
@@ -484,16 +368,26 @@ public function testIndexManagement(UnifiedTestCase $test): void
...
-    private static function provideTests(string $pattern): Generator
+    private static function provideTests(string $directory, ?string $group = null): Generator
     {
+        $pattern = self::$testDir . '/' . $directory . '/*.json';
+
+        $duplicateTests = array_flip(self::$duplicateTests);
+
         foreach (glob($pattern) as $filename) {
-            $group = basename(dirname($filename));
+            $testGroup = $group ?? dirname(basename($filename));
```
jmikola: Why was the order of `basename()` and `dirname()` swapped? Is that code path even used if every caller is passing a non-null `$group` string?

I suppose the explicit `$group` is only helpful for the `valid-pass` and `valid-fail` tests, since those groups are derived from the immediate parent directory of the test files instead of a grandparent; however, there's also nothing in the specs repo to enforce this path convention so it makes sense to be explicit here.

tests/UnifiedSpecTests/UnifiedSpecTest.php
```diff
@@ -484,16 +368,26 @@ public function testIndexManagement(UnifiedTestCase $test): void
 
     public static function provideIndexManagementTests(): Generator
     {
-        yield from self::provideTests(__DIR__ . '/index-management/*.json');
+        yield from self::provideTests('index-management/tests', 'index-management');
     }
 
-    private static function provideTests(string $pattern): Generator
+    private static function provideTests(string $directory, ?string $group = null): Generator
```
jmikola: It looks like an explicit value is always passed for `$group`, so consider removing the nullable attribute.

tests/UnifiedSpecTests/UnifiedSpecTest.php
jmikola: Asked about the offending test in https://github.com/mongodb/specifications/pull/1665/files#r1790278528

tests/UnifiedSpecTests/UnifiedSpecTest.php
```diff
@@ -21,145 +24,24 @@
  */
 class UnifiedSpecTest extends FunctionalTestCase
 {
+    /**
+     * Incomplete test groups are listed here. These are checked using a left-
+     * anchored regex. Note that regex placeholder can't be used.
```
jmikola: What would be the point of a placeholder? Based on how these strings are used below, it looks like one could use other regex operators.

I still think `str_starts_with()` is probably a simpler approach for what you actually need to do.

jmikola (commented):
Some more suggestions/questions.

tests/UnifiedSpecTests/UnifiedSpecTest.php
alcaeus: That's a good point. I've changed this to use `str_starts_with`, we can always change this to a regex syntax if we find the simpler method lacking in functionality.

tests/UnifiedSpecTests/UnifiedSpecTest.php
alcaeus: That is correct. PHPUnit asserts that each key is unique. While this was always true for arrays as data providers, using generators would allow one to yield the same key multiple times. I'm not sure why PHPUnit enforces this, but here we are 🤷

tests/UnifiedSpecTests/UnifiedSpecTest.php
alcaeus: Indeed. I used this as the directory for some of the tests changed, so I didn't want to enforce a convention for naming tests. I've removed the default value, so all tests now have to provide a group name for consistency.

tests/UnifiedSpecTests/UnifiedSpecTest.php
alcaeus: Thank you! This was fixed in the meantime, but I've kept the `duplicateTests` variable as a quick way to skip tests in future.

tests/UnifiedSpecTests/Operation.php
alcaeus: I started implementing this approach, unfortunately the object we have here is not an actual object, but a name of an object in the entity map. Since this object may not even exist at the time the operation instance is created (e.g. if it's only created later using `createEntities`), I also can't get a class name for the object being used. I'm afraid we'll have to skip when we get to `execute`.

CONTRIBUTING.md
alcaeus: Lost track of this. I was also wondering why I didn't notice this, until I realised that PhpStorm had the link target collapsed, suggesting that the text will fit into the 80-character limit we impose 🤦‍♂️

tests/SpecTests/ClientSideEncryption/Prose22_RangeExplicitEncryptionTest.php
alcaeus: No real reason other than using `private static` in other places. Note that `readonly` wouldn't work, as `readonly` properties cannot have a default value but have to be initialised in the constructor.

I suppose we could change this property, as well as a number of other properties in UnifiedSpecTest private constants.

tests/UnifiedSpecTests/UnifiedSpecTest.php
```diff
@@ -14,8 +14,7 @@
 use function basename;
 use function dirname;
```
jmikola: Note the `phpcs` warnings for `basename` and `dirname`.

jmikola (approved):
LGTM with `phpcs` warnings addressed.

PR #1436, opened 2024-09-23
PHPLIB-797: Remove unused methods in UnsupportedException

alcaeus:
PHPLIB-797

In addition to the ask in the ticket, I've marked the remaining factory methods in the class as internal, meaning that we can remove them in a future minor version once they are no longer necessary. The class itself is covered by the BC promise, but the factories are only for internal use and shouldn't be called by users.

I can revert that change if there's disagreement.

### DIFF
--- a/UPGRADE-2.0.md
+++ b/UPGRADE-2.0.md
@@ -28,3 +28,18 @@ GridFS
    ```php
    $bucket->openUploadStream($fileId, ['metadata' => ['contentType' => 'image/png']]);
    ```
+
+UnsupportedException method removals
+------------------------------------
+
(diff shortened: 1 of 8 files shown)


GromNaN:
You also should add the `@internal` annotation to other methods of other exception classes.

alcaeus:
> You also should add the `@internal` annotation to other methods of other exception classes.

Done. The only exception is the `UnsupportedValueException` class, as that's expected to be used by users in their own codecs when they encounter an unsupported value.

GromNaN (approved):
SGTM

PR #1438, opened 2024-09-23
PHPLIB-1522: Deprecate info iterators

alcaeus:
PHPLIB-1522

This PR deprecates the `CollectionInfoIterator`, `DatabaseInfoIterator`, and `IndexInfoIterator` interfaces and implementing classes. These will be replaced with generic iterators in 2.0.

In addition to this change, this PR also deprecates `IndexInfo::getNamespace`. The `ns` field that feeds this method has been removed from output in MongoDB 4.4, and keeping the necessary code added in PHPLIB-499 around makes no sense going forward.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -33,7 +33,6 @@ use MongoDB\Exception\UnexpectedValueException;
 use MongoDB\Exception\UnsupportedException;
 use MongoDB\Model\BSONArray;
 use MongoDB\Model\BSONDocument;
-use MongoDB\Model\DatabaseInfoIterator;
 use MongoDB\Operation\DropDatabase;
 use MongoDB\Operation\ListDatabaseNames;
 use MongoDB\Operation\ListDatabases;
@@ -294,7 +293,7 @@ class Client
      * List databases.
      *
      * @see ListDatabases::__construct() for supported options
-     * @return DatabaseInfoIterator
+     * @return Iterator
      * @throws UnexpectedValueException if the command response was malformed
      * @throws InvalidArgumentException for parameter/option parsing errors
      * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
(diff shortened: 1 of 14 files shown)


src/Client.php
```diff
@@ -294,7 +293,7 @@ public function listDatabaseNames(array $options = []): Iterator
      * List databases.
      *
      * @see ListDatabases::__construct() for supported options
-     * @return DatabaseInfoIterator
+     * @return Iterator
```
GromNaN: Can't you use the generics syntax here? Otherwise this change is a regression for typing.
```suggestion
     * @return Iterator<DatabaseInfo>
```

src/Collection.php
```diff
@@ -882,7 +881,7 @@ public function insertOne($document, array $options = [])
      * Returns information for all indexes for the collection.
      *
      * @see ListIndexes::__construct() for supported options
-     * @return IndexInfoIterator
+     * @return Iterator
```
GromNaN:
```suggestion
     * @return Iterator<IndexInfo>
```

src/Database.php
```diff
@@ -473,7 +472,7 @@ public function listCollectionNames(array $options = []): Iterator
      * Returns information for all collections in this database.
      *
      * @see ListCollections::__construct() for supported options
-     * @return CollectionInfoIterator
+     * @return Iterator
```
GromNaN:
```suggestion
     * @return Iterator<CollectionInfo>
```

GromNaN (commented):
I would change the assertion on tests also:
https://github.com/mongodb/mongo-php-library/blob/49677cbc93dceb40035877978ab17d57035ec8e3/tests/ClientFunctionalTest.php#L61

alcaeus:
> I would change the assertion on tests also:
> 
> https://github.com/mongodb/mongo-php-library/blob/49677cbc93dceb40035877978ab17d57035ec8e3/tests/ClientFunctionalTest.php#L61

For now I'd leave in the assertion as a safe guard. We can update this in 2.x. I have added the templated return types for Psalm's benefit and updated the baseline.

GromNaN (approved):
LGTM

PR #1439, opened 2024-09-23
Remove deprecated functionality

alcaeus:
This PR includes:
* PHPLIB-360: make `ChangeStream::CURSOR_NOT_FOUND` private
* PHPLIB-818: remove `Watch::FULL_DOCUMENT_DEFAULT`
* PHPLIB-1512: remove the deprecated `IndexInfo::isGeoHaystack` method
* PHPLIB-1511: remove the deprecated `maxScan`, `modifiers`, `oplogReplay`, and `snapshot` query options for `find` and `findOne`
* PHPLIB-1517: remove the `MonogDB\Operation\Executable` interface
* PHPLIB-1523: remove iterators for databases, collections, and indexes
* PHPLIB-1513: remove mapReduce functionality


src/Operation/ListIndexes.php
```diff
@@ -128,17 +97,39 @@ private function executeCommand(Server $server): IndexInfoIteratorIterator
...
              */
             if ($e->getCode() === self::ERROR_CODE_NAMESPACE_NOT_FOUND || $e->getCode() === self::ERROR_CODE_DATABASE_NOT_FOUND) {
-                return new IndexInfoIteratorIterator(new EmptyIterator());
+                return new EmptyIterator();
             }
 
             throw $e;
         }
 
+        /** @var CursorInterface<array> $cursor */
```
GromNaN: This comment should be line 93, just before the variable is assigned.

psalm-baseline.xml
GromNaN, line 1: The diffs on this file proves that you are doing the right thing, improving typing.

UPGRADE-2.0.md
GromNaN, line 1: It looks like I forgot to update this file in other PRs.

GromNaN (approved):

tests/Operation/ListCollectionsFunctionalTest.php
```diff
@@ -52,12 +52,12 @@ public function testIdIndexAndInfo(): void
...
         $collections = $operation->execute($server);
 
-        $this->assertInstanceOf(CollectionInfoIterator::class, $collections);
+        $this->assertInstanceOf(Iterator::class, $collections);
 
         foreach ($collections as $collection) {
             $this->assertInstanceOf(CollectionInfo::class, $collection);
             $this->assertArrayHasKey('readOnly', $collection['info']);
-            $this->assertEquals(['v' => 2, 'key' => ['_id' => 1], 'name' => '_id_', 'ns' => $this->getNamespace()], $collection['idIndex']);
+            $this->assertMatchesDocument(['v' => 2, 'key' => ['_id' => 1], 'name' => '_id_'], $collection['idIndex']);
```
alcaeus, line 61: This change was necessary as MongoDB 4.0 and 4.2 still return the `ns` field, while newer versions don't.

tests/UnifiedSpecTests/Operation.php
```diff
@@ -507,19 +506,8 @@ private function executeForCollection(Collection $collection)
...
-                assertInstanceOf(Javascript::class, $args['reduce']);
-                assertThat($args['out'], logicalOr(new IsType('string'), new IsType('array'), new IsType('object')));
-
-                return iterator_to_array($collection->mapReduce(
-                    $args['map'],
-                    $args['reduce'],
-                    $args['out'],
-                    array_diff_key($args, ['map' => 1, 'reduce' => 1, 'out' => 1]),
-                ));
+                Assert::markTestSkipped('mapReduce operation is not supported');
```
alcaeus, line 509: Unfortunately, there's no `markTestSkipped` helper function like the various `assert` helpers, hence the static call in the assert class. The end result wouldn't differ, as this method only throws an exception. I figured that marking a test as skipped when it uses `mapReduce` makes more sense than explicitly listing all tests. We currently list some tests explicitly when we don't support the tested operations, and we can consider moving those to the operation class as well to reduce the size of our skip list.

jmikola: My only concern with skipping at this point is that we're midway into the test. I trust this is safer now that we have a `finally` block for some cleanup around operation execution: https://github.com/mongodb/mongo-php-library/commit/90dcf186c8664a2cdd57314122d46b16234555c2#diff-fab05380d19ea3d7a426128a38cde0e262291e90b82294c085a7397d949216a8R216

UPGRADE-2.0.md
```diff
@@ -3,6 +3,24 @@ UPGRADE FROM 1.x to 2.0
 
  * Classes in the namespace `MongoDB\Operation\` are `final`.
  * All methods in interfaces and classes now define a return type.
+ * The `MongoDB\ChangeStream::CURSOR_NOT_FOUND` constant is now private.
+ * The `MongoDB\Operation\Watch::FULL_DOCUMENT_DEFAULT` constant has been
+   removed.
+ * The `getNamespace` and `isGeoHaystack` methods hav been removed from the
```
jmikola:
```suggestion
 * The `getNamespace` and `isGeoHaystack` methods have been removed from the
```

UPGRADE-2.0.md
```diff
@@ -3,6 +3,24 @@ UPGRADE FROM 1.x to 2.0
...
+   `MongoDB\Model\IndexInfo` class.
+ * The `maxScan`, `modifiers`, `oplogReplay`, and `snapshot` options for `find`
+   and `findOne` operations have been removed.
+ * The `MongoDB\Collection::mapReduce` method has been removed. Use aggregation
```
jmikola: Consider linking to https://www.mongodb.com/docs/manual/reference/map-reduce-to-aggregation-pipeline/ if you want to be helpful.

src/Client.php
```diff
@@ -305,7 +304,7 @@ public function listDatabaseNames(array $options = []): Iterator
      * @throws InvalidArgumentException for parameter/option parsing errors
      * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
      */
-    public function listDatabases(array $options = []): DatabaseInfoIterator
+    public function listDatabases(array $options = []): Iterator
```
jmikola, line 308: Noted that these methods already have a templated return type, as do the Operation class `execute()` methods.

src/Command/ListCollections.php
```diff
@@ -93,17 +91,16 @@ public function __construct(private string $databaseName, private array $options
...
      */
-    public function execute(Server $server): CachingIterator
+    public function execute(Server $server): CursorInterface
     {
-        /** @var Cursor<array> $cursor */
+        /** @var CursorInterface<array> $cursor */
         $cursor = $server->executeReadCommand($this->databaseName, $this->createCommand(), $this->createOptions());
         $cursor->setTypeMap(['root' => 'array', 'document' => 'array']);
 
-        return new CachingIterator($cursor);
```
jmikola, line 106: Noted that you consolidated this logic in the corresponding Operation class.

src/Operation/ListCollectionNames.php
```diff
@@ -67,15 +68,16 @@ public function __construct(string $databaseName, array $options = [])
...
-     * @see Executable::execute()
      * @return Iterator
      * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
      */
     public function execute(Server $server): Iterator
     {
-        return new CallbackIterator(
-            $this->listCollections->execute($server),
-            fn (array $collectionInfo): string => (string) $collectionInfo['name'],
+        return new CachingIterator(
```
jmikola, line 76: Should the return type of this method be typed as an iterator of strings?

Likewise for `ListDatabaseNames::execute()`. I'm not sure what the corresponding helper methods use, but I'll leave that for you to double-check.

src/Operation/ListCollections.php
```diff
@@ -61,20 +61,27 @@ final class ListCollections implements Executable
      * @param array  $options      Command options
      * @throws InvalidArgumentException for parameter/option parsing errors
      */
-    public function __construct(private string $databaseName, array $options = [])
```
jmikola, line 64: I assume this was inadvertently added in a previous PR. Was this the only case?

src/Operation/ListIndexes.php
```diff
@@ -128,17 +97,39 @@
...
+     * the command be executed on the primary.
+     *
+     * @see https://php.net/manual/en/mongodb-driver-server.executecommand.php
+     */
+    private function createOptions(): array
+    {
+        $options = [];
+
+        if (isset($this->options['session'])) {
+            $options['session'] = $this->options['session'];
```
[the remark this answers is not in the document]
jmikola: Does this require a suppression?

src/functions.php
```diff
@@ -387,24 +387,6 @@ function is_last_pipeline_operator_write(array $pipeline): bool
...
- * @param string|array|object $out Output specification
- */
-function is_mapreduce_output_inline(string|array|object $out): bool
-{
-    if (! is_array($out) && ! is_object($out)) {
-        return false;
-    }
-
-    return array_key_first(document_to_array($out)) === 'inline';
-}
```
jmikola, line 406: I'll miss you, little utility function.

stubs/Driver/CursorInterface.stub.php
```diff
@@ -4,6 +4,7 @@
 
 use Iterator;
 use MongoDB\BSON\Int64;
+use Traversable;
```
jmikola: I don't see this referenced anywhere below.

tests/Model/IndexInfoTest.php
```diff
@@ -166,40 +142,12 @@ public function testIs2dSphere(): void
             'v' => 2,
             'key' => ['pos' => '2dsphere'],
             'name' => 'pos_2dsphere',
-            'ns' => 'foo.bar',
         ]);
 
         $this->assertSame(2, $info->getVersion());
         $this->assertSame(['pos' => '2dsphere'], $info->getKey());
         $this->assertSame('pos_2dsphere', $info->getName());
-        $this->assertSame('foo.bar', $info->getNamespace());
```
jmikola, line 175: How did this work previously if `getNamespace()` was already emitting `E_USER_DEPRECATED`?

tests/Operation/ListCollectionsFunctionalTest.php
jmikola: I suggest capturing this in a code comment so we know to revisit after dropping support for MongoDB 4.2.

tests/SpecTests/Operation.php
```diff
@@ -411,12 +407,8 @@ private function executeForCollection(Collection $collection, Context $context):
...
 
             case 'mapReduce':
-                return $collection->mapReduce(
-                    $args['map'],
-                    $args['reduce'],
-                    $args['out'],
-                    array_diff_key($args, ['map' => 1, 'reduce' => 1, 'out' => 1]),
-                );
+                Assert::markTestSkipped('mapReduce is not supported');
+                break;
```
jmikola, line 411: I suppose there's no harm in it, but is this `break` actually necessary? Can we rely on Psalm to assert that this line is unreachable?

tests/UnifiedSpecTests/Operation.php
jmikola: In the past, my only concern with skipping at this point would be that we're already midway into the test. I trust this is safer now that we have a `finally` block for some cleanup around operation execution: https://github.com/mongodb/mongo-php-library/commit/90dcf186c8664a2cdd57314122d46b16234555c2#diff-fab05380d19ea3d7a426128a38cde0e262291e90b82294c085a7397d949216a8R216

I suppose there's some wasted effort running the test until this point, but if that makes the skip list easier to digest I suppose it's worth the trade-off.

src/Operation/ListCollections.php
alcaeus: The database name was previously used in the `execute` method when constructing the iterator. This is no longer necessary, hence why we're no longer storing it as a class property.

src/Operation/ListIndexes.php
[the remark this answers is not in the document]
alcaeus: This alert is there because it is a new suppression. I've dismissed the alert.

tests/Operation/ListCollectionsFunctionalTest.php
alcaeus: Added

tests/UnifiedSpecTests/Operation.php
alcaeus: Yeah, ideally all kinds of cleanup happens in `finally` blocks to make sure we clean up. That said, I currently don't see any negative side effects, so it looks like we've done things properly, as I'd expect 😉

UPGRADE-2.0.md
alcaeus: Whoops, thanks!

UPGRADE-2.0.md
alcaeus: Ah, yes, exactly the information people will need. Added a link to this 👍

src/Operation/ListCollectionNames.php
alcaeus: Changed this for `listCollections` and `listDatabases`.

stubs/Driver/CursorInterface.stub.php
alcaeus: Nice catch, removed

tests/Model/IndexInfoTest.php
alcaeus: That's an excellent question 🤷‍♂️

tests/SpecTests/Operation.php
alcaeus: `phpcs` was complaining about this. While psalm is able to infer that this won't be reached due to a `@return never`, `phpcs` doesn't and can only use a `return` within a case to infer that a `break` statement is not needed.

tests/Operation/FindFunctionalTest.php
GromNaN, line 1: Please revert this change after rebase: https://github.com/mongodb/mongo-php-library/pull/1412#discussion_r1777852091

alcaeus: It has been removed along with the entire test surrounding it.

jmikola (approved):

GromNaN (approved):

PR #1441, opened 2024-09-23
Deprecate functionality to be removed

alcaeus:
This PR deprecates the `Collection::mapReduce` method (PHPLIB-1533). Duplicated deprecation notices for deprecated query options have been removed.

### DIFF
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -956,7 +956,7 @@ class Collection
      */
     public function mapReduce(JavascriptInterface $map, JavascriptInterface $reduce, string|array|object $out, array $options = [])
     {
-        @trigger_error(sprintf('The %s method is deprecated and will be removed in a future release.', __METHOD__), E_USER_DEPRECATED);
+        @trigger_error(sprintf('The %s method is deprecated and will be removed in a version 2.0.', __METHOD__), E_USER_DEPRECATED);
 
         $hasOutputCollection = ! is_mapreduce_output_inline($out);
 
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -77,7 +77,11 @@ use function array_intersect_key;
 use function array_key_exists;
 use function current;
 use function is_array;
+use function sprintf;
 use function strlen;
+use function trigger_error;
+
+use const E_USER_DEPRECATED;
 
 class Collection
 {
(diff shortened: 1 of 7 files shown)


src/Collection.php
```diff
@@ -927,6 +931,8 @@ public function listSearchIndexes(array $options = []): Iterator
      */
     public function mapReduce(JavascriptInterface $map, JavascriptInterface $reduce, $out, array $options = [])
     {
+        @trigger_error(sprintf('The %s method is deprecated and will be removed in a future release.', __METHOD__), E_USER_DEPRECATED);
```
GromNaN: Will you remove it in version 2.0? In that case, you can mention the version here.
```suggestion
        @trigger_error(sprintf('The %s method is deprecated and will be removed in version 2.0.', __METHOD__), E_USER_DEPRECATED);
```

GromNaN (approved):
LGTM

src/Operation/Find.php
```diff
@@ -292,12 +292,20 @@ public function __construct(string $databaseName, string $collectionName, $filte
...
 
-        if (isset($options['maxScan'])) {
-            trigger_error('The "maxScan" option is deprecated and will be removed in a future release', E_USER_DEPRECATED);
+        if (isset($options['modifiers'])) {
+            @trigger_error('The "modifiers" option is deprecated and will be removed in a future release', E_USER_DEPRECATED);
+        }
+
+        if (isset($options['oplogReplay'])) {
+            @trigger_error('The "oplogReplay" option is deprecated and will be removed in a future release', E_USER_DEPRECATED);
+        }
```
jmikola: Excluding `modifiers`, PHPC raises its own `E_DEPRECATED` notice these options in the Query constructor. Does that mean users would see two deprecation notices?

As for `modifiers`, that seems like something we should add in PHPC 1.x before we remove the option entirely in 2.0 ([PHPC-2440](https://jira.mongodb.org/browse/PHPC-2440)). I created [PHPC-2452](https://jira.mongodb.org/browse/PHPC-2453) and opened https://github.com/mongodb/mongo-php-driver/pull/1680 to address that.

alcaeus: Yes, that would cause duplicate deprecation notices, so we probably should remove the extra notices thrown in PHPLIB. I'll comment on https://github.com/mongodb/mongo-php-driver/pull/1680 regarding the `modifier` deprecations, but I'll remove the query option deprecations entirely in this PR.

src/Collection.php
alcaeus: Updated all occurrences of "in a future release" to refer to 2.0.

src/Operation/Find.php
alcaeus: I repurposed this PR to remove the existing deprecation notices for query options instead of adding new ones, as PHPC already emits deprecation notices for deprecated options. The mapReduce deprecation was left in place.

jmikola (approved):

PR #1445, opened 2024-09-24
PHPLIB-1531: Retry once when disabling fail points

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1531

This is an attempt to resolve flaky test failures following https://github.com/mongodb/mongo-php-library/pull/1427

### DIFF
--- a/tests/UnifiedSpecTests/ManagesFailPointsTrait.php
+++ b/tests/UnifiedSpecTests/ManagesFailPointsTrait.php
@@ -2,6 +2,7 @@
 
 namespace MongoDB\Tests\UnifiedSpecTests;
 
+use MongoDB\Driver\Exception\ConnectionException;
 use MongoDB\Driver\Server;
 use MongoDB\Operation\DatabaseCommand;
 use stdClass;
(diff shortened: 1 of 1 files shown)


GromNaN (approved):

tests/UnifiedSpecTests/ManagesFailPointsTrait.php
```diff
@@ -31,8 +32,14 @@ public function configureFailPoint(stdClass $failPoint, Server $server): void
     public function disableFailPoints(): void
     {
         foreach ($this->failPointsAndServers as [$failPoint, $server]) {
-            $operation = new DatabaseCommand('admin', ['configureFailPoint' => $failPoint, 'mode' => 'off']);
-            $operation->execute($server);
+            try {
+                $operation = new DatabaseCommand('admin', ['configureFailPoint' => $failPoint, 'mode' => 'off']);
+                $operation->execute($server);
+            } catch (ConnectionException $e) {
+                // Retry once in case the connection was dropped by the last operation
```
jmikola, line 39: I'm still at a loss for why this wasn't an issue with the original FailPointObserver. There, the Server object was instantiated from the CommandStartedEvent; however, it's just an integer `server_id` and Manager reference. That's not much different from the object we're capturing here from `configureFailPoint()`. In both cases, the command to disable a fail point is still being executed via `phongo_execute_command()` with the same underlying `mongoc_client_t` and `server_id`.

The only difference I can identify is _when_ we're disabling fail points. Previously, FailPointObserver disabled fail points during our equivalent of a `tearDown()` method. Now, we're doing so immediately after executing operations and before evaluating `expectEvents` (note: this is what the spec itself suggests). It's possible the previous deferral allowed some opportunity to reestablish a dropped connection.

This may be related to [CDRIVER-4532](https://jira.mongodb.org/browse/CDRIVER-4532), which highlights an outstanding libmongoc deficiency in the code paths utilized by PHPC.

PR #1451, opened 2024-09-25
Assert that the modifiers option is indeed deprecated

alcaeus:
This PR adds an assertion that the `modifiers` query option is deprecated. While this commit is not necessary in 1.20 due to our usage of Symfony's PHPUnit Bridge which handles deprecations, when running phpunit directly the deprecation creates test output which causes a test failure in later versions. For this reason, I figured it's best to include this in 1.20.

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -36,6 +36,7 @@ use function set_error_handler;
 use function sprintf;
 use function strtr;
 
+use const E_DEPRECATED;
 use const E_USER_DEPRECATED;
 
 abstract class TestCase extends BaseTestCase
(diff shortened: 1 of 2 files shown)


tests/TestCase.php
```diff
@@ -171,7 +171,7 @@ protected function assertDeprecated(callable $execution): void
 
         set_error_handler(function ($errno, $errstr) use (&$errors): void {
             $errors[] = $errstr;
-        }, E_USER_DEPRECATED);
+        }, E_USER_DEPRECATED | E_DEPRECATED);
```
GromNaN, line 175: `use const E_DEPRECATED` missing.

alcaeus: Whoops, added.

GromNaN (approved):

PR #1454, opened 2024-09-25
PHPLIB-1518 `WriteResult::get*Count()` always return an int on acknowledged result

GromNaN:
Fix PHPLIB-1518

I chose to rely on the driver exception thrown by https://github.com/mongodb/mongo-php-driver/pull/1687

### DIFF
--- a/src/GridFS/Bucket.php
+++ b/src/GridFS/Bucket.php
@@ -583,16 +583,9 @@ class Bucket
             return;
         }
 
-        /* If the update resulted in no modification, it's possible that the
-         * file did not exist, in which case we must raise an error. Checking
-         * the write result's matched count will be most efficient, but fall
-         * back to a findOne operation if necessary (i.e. legacy writes).
-         */
-        $found = $updateResult->getMatchedCount() !== null
-            ? $updateResult->getMatchedCount() === 1
-            : $this->collectionWrapper->findFileById($id) !== null;
-
-        if (! $found) {
+        // If the update resulted in no modification, it's possible that the
+        // file did not exist, in which case we must raise an error.
+        if ($updateResult->getMatchedCount() !== 1) {
             throw FileNotFoundException::byId($id, $this->getFilesNamespace());
         }
     }
--- a/psalm.xml.dist
+++ b/psalm.xml.dist
@@ -23,6 +23,7 @@
         <file name="stubs/BSON/PackedArray.stub.php"/>
         <file name="stubs/Driver/Cursor.stub.php"/>
         <file name="stubs/Driver/CursorInterface.stub.php"/>
+        <file name="stubs/Driver/WriteResult.stub.php"/>
     </stubs>
 
     <issueHandlers>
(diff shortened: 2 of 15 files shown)


src/GridFS/Bucket.php
```diff
@@ -583,16 +583,9 @@ public function rename(mixed $id, string $newFilename): void
...
-         * back to a findOne operation if necessary (i.e. legacy writes).
-         */
-        $found = $updateResult->getMatchedCount() !== null
-            ? $updateResult->getMatchedCount() === 1
-            : $this->collectionWrapper->findFileById($id) !== null;
-
-        if (! $found) {
+        // If the update resulted in no modification, it's possible that the
+        // file did not exist, in which case we must raise an error.
+        if ($updateResult->getMatchedCount() !== 1) {
```
alcaeus, line 588: I'm not sure we should consider using unacknowledged writes a valid use case, but in case we do there's a functional change here. If we want to preserve the option to use unacknowledged writes with GridFS, we should check `isAcknowledged` instead of relying on the `null` value from `getMatchedCount`:

```suggestion
        /* If the update resulted in no modification, it's possible that the
         * file did not exist, in which case we must raise an error. Checking
         * the write result's matched count will be most efficient, but fall
         * back to a findOne operation if necessary (i.e. legacy writes).
         */
        $found = $updateResult->isAcknowledged()
            ? $updateResult->getMatchedCount() === 1
            : $this->collectionWrapper->findFileById($id) !== null;

        if (! $found) {
```

alcaeus (commented):
LGTM, but we may want to consider unacknowledged writes in `rename`.

src/GridFS/Bucket.php
GromNaN: In that case, that's a fix to backport to v1.x, as it already throws an exception when there is an unacknowledged operation.

alcaeus: In that case, let's defer to a separate PR, as this does PR not change existing behaviour.

alcaeus (approved):

GromNaN:
I did not update the UPGRADE file, there is nothing to upgrade.

src/GridFS/Bucket.php
jmikola: > let's defer to a separate PR

I assume a ticket needs to be created for this. Please cross-reference it here after doing so.

GromNaN: I just realized that this second `if ($updateResult->getMatchedCount() !== 1)` is always true after `if ($updateResult->getModifiedCount() === 1) return;`.

Do you know what are the "legacy writes" that are mentioned in the comment?

jmikola: Legacy opcodes (e.g. `OP_UPDATE`), which were [removed in MongoDB 6.0](https://www.mongodb.com/docs/manual/release-notes/6.0-compatibility/#legacy-opcodes-removed). At this point, drivers _may_ still use them for unacknowledged writes but all acknowledged writes are guaranteed to use `OP_MSG` and the corresponding write commands (e.g. [update](https://www.mongodb.com/docs/manual/reference/command/update/)).

PR #1455, opened 2024-09-25
Expect BulkWriteException

alcaeus:
With the `WriteException` class removed in PHPC 2.0, we need to specifically expect a `BulkWriteException` in spec tests.


GromNaN (approved):

GromNaN:
Good catch.

alcaeus:
Note: re-targeted to 1.x. It doesn't really make a difference since we're only using this class in tests, but it helps me work around a circular dependency build failure issue.

PR #1458, opened 2024-09-25
Fix deprecations in tests

GromNaN:


tests/GridFS/BucketFunctionalTest.php
```diff
@@ -951,8 +951,6 @@ public function testResolveStreamContextForRead(): void
         fclose($stream);
 
         $method = new ReflectionMethod($this->bucket, 'resolveStreamContext');
-        $method->setAccessible(true);
```
GromNaN, line 954: `ReflectionMethod::setAccessible` is useless

> As of PHP 8.1.0, calling this method has no effect; all properties are accessible by default.
> https://www.php.net/manual/en/reflectionproperty.setaccessible.php

tests/Operation/WatchFunctionalTest.php
```diff
@@ -1282,19 +1279,19 @@ function (array $event) use (&$aggregateCommands): void {
             $aggregateCommands[0]['pipeline'][0]->{'$changeStream'},
             $this->logicalNot(
                 $this->logicalOr(
-                    $this->objectHasAttribute('resumeAfter'),
-                    $this->objectHasAttribute('startAfter'),
-                    $this->objectHasAttribute('startAtOperationTime'),
+                    new ObjectHasProperty('resumeAfter'),
+                    new ObjectHasProperty('startAfter'),
+                    new ObjectHasProperty('startAtOperationTime'),
```
GromNaN, line 1284: `ObjectHasAttribute` is deprecated and there is no helper for `ObjectHasProperty`. I don't want to add the helper, to avoid conflict if they decide to add it later.

alcaeus (approved):

PR #1468, opened 2024-09-26
PHPLIB-1227 Use void return types for operations without meaningful result document

GromNaN:
Fix PHPLIB-1227

This introduces major breaking changes that we did not deprecate.

### DIFF
--- a/UPGRADE-2.0.md
+++ b/UPGRADE-2.0.md
@@ -31,10 +31,11 @@ type changed to `void`. In case of an error, an exception is thrown.
 
  * `MongoDB\Client`: `dropDatabase`
  * `MongoDB\Collection`: `drop`, `dropIndex`, `dropIndexes`, `dropSearchIndex`, `rename`
- * `MongoDB\Database`: `createCollection`, `drop`, `dropCollection`, `modifyCollection`, `renameCollection`
+ * `MongoDB\Database`: `createCollection`, `drop`, `dropCollection`, `renameCollection`
  * `MongoDB\Database::createEncryptedCollection()` returns the list of encrypted fields
 
-If you still need to access the raw command result, you can use a `CommandSubscriber`.
+If you still need to access the raw command result, you can use a
+[`CommandSubscriber`](https://www.php.net/manual/en/class.mongodb-driver-monitoring-commandsubscriber.php).
 
 GridFS
 ------
--- a/UPGRADE-2.0.md
+++ b/UPGRADE-2.0.md
@@ -23,6 +23,19 @@ UPGRADE FROM 1.x to 2.0
    * `MongoDB\Model\IndexInfoIteratorIterator`
    * `MongoDB\Operation\Executable`
 
+Operations with no result
+-------------------------
+
+The following operations no longer return the raw command result. The return
+type changed to `void`. In case of an error, an exception is thrown.
+
+ * `MongoDB\Client`: `dropDatabase`
+ * `MongoDB\Collection`: `drop`, `dropIndex`, `dropIndexes`, `dropSearchIndex`, `rename`
+ * `MongoDB\Database`: `createCollection`, `drop`, `dropCollection`, `modifyCollection`, `renameCollection`
+ * `MongoDB\Database::createEncryptedCollection()` returns the list of encrypted fields
+
+If you still need to access the raw command result, you can use a `CommandSubscriber`.
+
 GridFS
 ------
 
--- a/src/Client.php
+++ b/src/Client.php
@@ -49,10 +49,6 @@ use Throwable;
 use function array_diff_key;
 use function is_array;
 use function is_string;
-use function sprintf;
-use function trigger_error;
-
-use const E_USER_DEPRECATED;
 
 class Client
 {
@@ -217,19 +213,12 @@ class Client
      * @see DropDatabase::__construct() for supported options
      * @param string $databaseName Database name
      * @param array  $options      Additional options
-     * @return array|object Command result document
      * @throws UnsupportedException if options are unsupported on the selected server
      * @throws InvalidArgumentException for parameter/option parsing errors
      * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
      */
-    public function dropDatabase(string $databaseName, array $options = []): array|object
+    public function dropDatabase(string $databaseName, array $options = []): void
     {
-        if (! isset($options['typeMap'])) {
-            $options['typeMap'] = $this->typeMap;
-        } else {
-            @trigger_error(sprintf('The function %s() will return nothing in mongodb/mongodb v2.0, the "typeMap" option is deprecated', __FUNCTION__), E_USER_DEPRECATED);
-        }
-
         $server = select_server_for_write($this->manager, $options);
 
         if (! isset($options['writeConcern']) && ! is_in_transaction($options)) {
@@ -238,7 +227,7 @@ class Client
 
         $operation = new DropDatabase($databaseName, $options);
 
-        return $operation->execute($server);
+        $operation->execute($server);
     }
 
     /**
--- a/src/Database.php
+++ b/src/Database.php
@@ -478,8 +478,12 @@ class Database
      * @throws InvalidArgumentException for parameter/option parsing errors
      * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
      */
-    public function modifyCollection(string $collectionName, array $collectionOptions, array $options = []): void
+    public function modifyCollection(string $collectionName, array $collectionOptions, array $options = []): array|object
     {
+        if (! isset($options['typeMap'])) {
+            $options['typeMap'] = $this->typeMap;
+        }
+
         $server = select_server_for_write($this->manager, $options);
 
         if (! isset($options['writeConcern']) && ! is_in_transaction($options)) {
@@ -488,7 +492,7 @@ class Database
 
         $operation = new ModifyCollection($this->databaseName, $collectionName, $collectionOptions, $options);
 
-        $operation->execute($server);
+        return $operation->execute($server);
     }
 
     /**
--- a/src/Database.php
+++ b/src/Database.php
@@ -54,11 +54,7 @@ use stdClass;
 use Throwable;
 
 use function is_array;
-use function sprintf;
 use function strlen;
-use function trigger_error;
-
-use const E_USER_DEPRECATED;
 
 class Database
 {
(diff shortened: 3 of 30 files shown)


src/Database.php
```diff
@@ -314,17 +310,13 @@ public function createCollection(string $collectionName, array $options = []): a
      * getPrevious() and getEncryptedFields() methods, respectively.
      *
      * @see CreateCollection::__construct() for supported options
-     * @return array A tuple containing the command result document from creating the collection and the modified "encryptedFields" option
+     * @return array The modified "encryptedFields" option
```
GromNaN, line 311: Instead of a tuple, we return the 2nd element (the `encryptedFields`), but the return type is not modified.

tests/FunctionalTestCase.php
```diff
@@ -189,6 +196,45 @@ protected function assertCollectionExists(string $collectionName, ?string $datab
         }
     }
 
+    /** @param callable():void $closure */
+    protected static function commandResult(callable $closure): object
```
GromNaN: Read command result from the events. That allows to keep the `assertCommandSucceeded`. But it might not be necessary if the operations throw an exception when the command fails.

src/Operation/CreateCollection.php
```diff
@@ -112,9 +111,6 @@ final class CreateCollection implements Executable
      *
      *    This is not supported for servers versions < 5.0.
      *
-     *  * typeMap (array): Type map for BSON deserialization. This will only be
-     *    used for the returned command result document.
-     *
```
GromNaN, line 117: The typeMap option can be removed since we don't return the command results.

src/Operation/CreateEncryptedCollection.php
```diff
@@ -97,21 +97,21 @@ public function __construct(private string $databaseName, private string $collec
...
-     * data keys have been created.
+     * Returns the data keys that have been created.
      *
      * @see \MongoDB\Database::createEncryptedCollection()
      * @see https://www.php.net/manual/en/mongodb-driver-clientencryption.createdatakey.php
+     * @return array
      * @throws DriverRuntimeException for errors creating a data key
      */
-    public function createDataKeys(ClientEncryption $clientEncryption, string $kmsProvider, ?array $masterKey, ?array &$encryptedFields = null): void
+    public function createDataKeys(ClientEncryption $clientEncryption, string $kmsProvider, ?array $masterKey): array
```
GromNaN, line 106: The function return null, we don't need to have a by-reference parameter to output the data keys. Using the returned value is more common.
This change can be backported to 1.x, with a deprecation message if `$encryptedFields` parameter is passed.

jmikola: > This change can be backported to 1.x

FWIW, I'm OK skipping that and just communicating this in the upgrade notes. I think it's very unlikely that anyone was actually using this API.

Defer to you, though.

src/Operation/DropCollection.php
```diff
@@ -83,36 +82,28 @@ public function __construct(private string $databaseName, private string $collec
...
         try {
-            $cursor = $server->executeWriteCommand($this->databaseName, $this->createCommand(), $this->createOptions());
+            $server->executeWriteCommand($this->databaseName, $this->createCommand(), $this->createOptions());
         } catch (CommandException $e) {
             /* The server may return an error if the collection does not exist.
              * Check for an error code and return the command reply instead of
              * throwing. */
             if ($e->getCode() === self::ERROR_CODE_NAMESPACE_NOT_FOUND) {
-                return $e->getResultDocument();
+                return;
```
jmikola, line 92: "return the command reply instead of throwing" in the comment above should probably be changed to "return instead of throwing" or "suppress the error".

src/Operation/CreateCollection.php
jmikola: Thinking of your comment below on `createDataKeys()`, I noted you didn't suggest deprecating the `typeMap` option in 1.x. I also don't think that's necessary and feel we can just rely on upgrade notes for 2.0.

tests/FunctionalTestCase.php
jmikola: I concur. There should be no reason to maintain `assertCommandSucceeded()`, although removing it entirely might leave some tests without assertions (easy fix if so).

GromNaN: I removed all that assertions. `assertCommandSucceeded` is still used when running raw commands.

UPGRADE-2.0.md
```diff
@@ -4,6 +4,19 @@ UPGRADE FROM 1.x to 2.0
...
+Operations with no result
+-------------------------
+
+The following operations no longer return the raw command result. The return
+type changed to `void`. In case of an error, an exception is thrown.
+
+ * `MongoDB\Client`: `dropDatabase`
+ * `MongoDB\Collection`: `drop`, `dropIndex`, `dropIndexes`, `dropSearchIndex`, `rename`
+ * `MongoDB\Database`: `createCollection`, `drop`, `dropCollection`, `modifyCollection`, `renameCollection`
+ * `MongoDB\Database::createEncryptedCollection()` return the list of encrypted fields
```
alcaeus:
```suggestion
 * `MongoDB\Database::createEncryptedCollection()` returns the list of encrypted fields
```

src/Client.php
```diff
@@ -214,17 +214,12 @@ public function createClientEncryption(array $options): ClientEncryption
      * @see DropDatabase::__construct() for supported options
      * @param string $databaseName Database name
      * @param array  $options      Additional options
-     * @return array|object Command result document
      * @throws UnsupportedException if options are unsupported on the selected server
      * @throws InvalidArgumentException for parameter/option parsing errors
      * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
      */
-    public function dropDatabase(string $databaseName, array $options = []): array|object
+    public function dropDatabase(string $databaseName, array $options = []): void
```
alcaeus, line 220: Should these changes come with a `ReturnTypeWillChange` attribute in 1.x or is that attribute only designed to suppress notifications from PHP core?

src/Operation/CreateCollection.php
alcaeus: I agree with not needing a deprecation notice for removing typeMap, as we won't throw an error when the option is present, it just won't do anything.

I would point out however that a `typeMap` option here is a pretty good indication that the caller intends to use the result, so it could be a way to notify users of the changing return type. Since it doesn't catch all cases, I'm not sure it's warranted though.

src/Operation/CreateEncryptedCollection.php
alcaeus: If we consider the operation (and thus the API) internal in 1.x, I agree with not adding a deprecation layer in 1.x.

alcaeus (commented):
Left a few comments, and I agree with the list of operations that no longer return a result. I'll defer to @jmikola for the detailed review.

src/Client.php
GromNaN: The [`ReturnTypeWillChange ` attribute](https://www.php.net/manual/en/class.returntypewillchange.php) doesn't work for this use-case: https://3v4l.org/SP9O8

src/Operation/DropCollection.php
```diff
@@ -83,36 +73,27 @@ public function __construct(private string $databaseName, private string $collec
...
         }
 
         try {
-            $cursor = $server->executeWriteCommand($this->databaseName, $this->createCommand(), $this->createOptions());
+            $server->executeWriteCommand($this->databaseName, $this->createCommand(), $this->createOptions());
         } catch (CommandException $e) {
             /* The server may return an error if the collection does not exist.
-             * Check for an error code and return the command reply instead of
-             * throwing. */
+             * Ignore the exception to make the drop operation is idempotent */
```
GromNaN:
```suggestion
             * Ignore the exception to make the drop operation idempotent */
```

src/Client.php
alcaeus: Shame, but good to know.

src/Operation/CreateCollection.php
GromNaN: Rebased on https://github.com/mongodb/mongo-php-library/pull/1473

UPGRADE-2.0.md
```diff
@@ -4,6 +4,19 @@ UPGRADE FROM 1.x to 2.0
...
+
+The following operations no longer return the raw command result. The return
+type changed to `void`. In case of an error, an exception is thrown.
+
+ * `MongoDB\Client`: `dropDatabase`
+ * `MongoDB\Collection`: `drop`, `dropIndex`, `dropIndexes`, `dropSearchIndex`, `rename`
+ * `MongoDB\Database`: `createCollection`, `drop`, `dropCollection`, `modifyCollection`, `renameCollection`
+ * `MongoDB\Database::createEncryptedCollection()` returns the list of encrypted fields
+
+If you still need to access the raw command result, you can use a `CommandSubscriber`.
```
jmikola: Consider either linking this to the [class docs](https://www.php.net/manual/en/class.mongodb-driver-monitoring-commandsubscriber.php) or referencing the FQCN: `MongoDB\Driver\Monitoring\CommandSubscriber`.

GromNaN:
Rebased. Ready for review @jmikola @alcaeus

tests/Operation/ModifyCollectionFunctionalTest.php
```diff
@@ -29,9 +33,23 @@ public function testCollMod(): void
             ['index' => ['keyPattern' => ['lastAccess' => 1], 'expireAfterSeconds' => 1000]],
             ['typeMap' => ['root' => 'array', 'document' => 'array']],
         );
-        $result = $modifyCollection->execute($this->getPrimaryServer());
+        $modifyCollection->execute($this->getPrimaryServer());
```
jmikola: I'm not sure that the return value of `collMod` is entirely irrelevant, as the [server docs include a note](https://www.mongodb.com/docs/manual/reference/command/collMod/#hide-an-index-from-the-query-planner) about being able to infer whether a value actually changed server-side by inspecting the command result:

> If the operation is successful but the hidden value has not changed (specifically, hiding an already hidden index or unhiding an already unhidden index), the command omits the `hidden_old` and `hidden_new` fields from the output.

I didn't consider that until seeing this test change.

If you agree, I'd suggest just reverting the changes to ModifyCollection. Otherwise, feel free to leave as-is.

jmikola (approved):
Two suggestions (your call). LGTM with whatever you decide.

alcaeus (approved):

tests/Operation/ModifyCollectionFunctionalTest.php
GromNaN: I reverted the change on this operation. This was missing from https://github.com/mongodb/mongo-php-library/pull/1473, so nothing to revert on v1.x

UPGRADE-2.0.md
GromNaN: Link added.

PR #1480, opened 2024-10-14
Fix array shape for `Collection::listSearchIndex($options)`

GromNaN:
The `@param` annotation only accept `name` option. Which is wrong since all the `aggregate` options are allowed. 

See https://github.com/doctrine/mongodb-odm/pull/2689#issuecomment-2410880675

### DIFF
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -896,7 +896,7 @@ class Collection
      * Returns information for all Atlas Search indexes for the collection.
      * Only available when used against a 7.0+ Atlas cluster.
      *
-     * @param array{name?: string} $options Command options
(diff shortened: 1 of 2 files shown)


alcaeus (commented):
Change LGTM, but I'm not sure what phpcs is complaining about.

GromNaN:
> Change LGTM, but I'm not sure what phpcs is complaining about.

This is a known issue https://github.com/squizlabs/PHP_CodeSniffer/issues/3734. The false-positive error is reported even if I use `@phpstan-param` or `@psalm-param` instead of `@param` So I removed the array shape from the phpdoc.

alcaeus (approved):

PR #1482, opened 2024-10-14
PHPLIB-1548 Inherit `typeMap` option in `Collection::listSearchIndexes()`

GromNaN:
Fix PHPLIB-1548

### DIFF
--- a/tests/Collection/CollectionFunctionalTest.php
+++ b/tests/Collection/CollectionFunctionalTest.php
@@ -8,6 +8,7 @@ use MongoDB\Codec\Encoder;
 use MongoDB\Collection;
 use MongoDB\Database;
 use MongoDB\Driver\BulkWrite;
+use MongoDB\Driver\Exception\CommandException;
 use MongoDB\Driver\ReadConcern;
 use MongoDB\Driver\ReadPreference;
 use MongoDB\Driver\WriteConcern;
@@ -23,6 +24,7 @@ use TypeError;
 use function array_filter;
 use function call_user_func;
 use function is_scalar;
+use function iterator_to_array;
 use function json_encode;
 use function str_contains;
 use function usort;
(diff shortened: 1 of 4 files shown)


alcaeus (approved):

tests/Collection/CollectionFunctionalTest.php
```diff
@@ -800,6 +802,32 @@ public function testMethodInTransactionWithReadConcernOption($method): void
...
+    {
+        $this->skipIfAtlasSearchIndexIsNotSupported();
+
+        $collection = new Collection($this->manager, $this->getDatabaseName(), $this->getCollectionName(), ['typeMap' => ['root' => 'array']]);
+
+        // Insert a document to create the collection
+        $collection->insertOne(['_id' => 1]);
+
+        try {
+            $collection->createSearchIndex(['mappings' => ['dynamic' => false]], ['name' => 'test-search-index']);
```
jmikola, line 815: Does `dynamic: false` make sense here? Quoting [`createSearchIndexes` command docs](https://www.mongodb.com/docs/manual/reference/command/createSearchIndexes/):

> If set to false, you must specify individual fields to index using `mappings.fields`.

Presumably, this creates a search index that does nothing. I'm curious as to why the server would even allow that, but wonder if future validation might break this. `dynamic: true` seems more correct (if only as a valid demonstration).

GromNaN: I used this mapping because it is in the spec test: https://github.com/mongodb/specifications/blob/b746fcc71ad7c1b376482852c3c9cbba4eed7d26/source/index-management/tests/README.md#case-1-driver-can-successfully-create-and-list-search-indexes

This requires a spec change if we want a client error in case of `dynamic: false AND fields: empty`

jmikola: Thanks for digging that up. Since spec test reads the definition back from the server, I think this goes beyond anything drivers should address themselves. I'll share a link to this thread with the Atlas Search team but OK to leave this PR as-is.

PR #1484, opened 2024-10-15
PHPLIB-1502: Test with PHP 8.4

alcaeus:
PHPLIB-1502

This adds testing with PHP 8.4 on evergreen. Due to an implicit nullable parameter in PHPUnit, I had to skip a single test. This test can be re-enabled when we merge this up to v1.x.

### DIFF
--- a/.evergreen/config/functions.yml
+++ b/.evergreen/config/functions.yml
@@ -325,9 +325,9 @@ functions:
         working_dir: "src"
         script: |
           ${PREPARE_SHELL}
-          file="${PROJECT_DIRECTORY}/.evergreen/install-composer.sh"
-          # Don't use ${file} syntax here because evergreen treats it as an empty expansion.
-          [ -f "$file" ] && DEPENDENCIES=${DEPENDENCIES} bash $file || echo "$file not available, skipping"
+          DEPENDENCIES=${DEPENDENCIES} \
+          PHP_VERSION=${PHP_VERSION} \
+          bash ${PROJECT_DIRECTORY}/.evergreen/install-composer.sh
 
   "start load balancer":
     - command: shell.exec
--- a/.evergreen/config/generate-config.php
+++ b/.evergreen/config/generate-config.php
@@ -3,6 +3,7 @@
 
 // Supported PHP versions. Add new versions to the beginning of the list
 $modernPhpVersions = [
+    '8.4',
     '8.3',
     '8.2',
     '8.1',
@@ -13,7 +14,9 @@ $legacyPhpVersions = [
 ];
 $supportedPhpVersions = array_merge($modernPhpVersions, $legacyPhpVersions);
 
-$latestPhpVersion = max($supportedPhpVersions);
+// TODO: use max() once PHP 8.4 is stable
+//$latestPhpVersion = max($supportedPhpVersions);
+$latestPhpVersion = '8.3';
 $lowestPhpVersion = min($supportedPhpVersions);
 
 // Supported MongoDB versions. Add new versions after "rapid"
--- a/.evergreen/install-composer.sh
+++ b/.evergreen/install-composer.sh
@@ -38,4 +38,9 @@ php --ri mongodb
 
 install_composer
 
+# Remove psalm as it's not compatible with PHP 8.4: https://github.com/vimeo/psalm/pull/10928
+if [ "$PHP_VERSION" == "8.4" ]; then
+  php composer.phar remove --no-update --dev vimeo/psalm
+fi
+
 php composer.phar update $COMPOSER_FLAGS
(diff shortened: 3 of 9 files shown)


.evergreen/config/functions.yml
```diff
@@ -325,9 +325,9 @@ functions:
         working_dir: "src"
         script: |
           ${PREPARE_SHELL}
-          file="${PROJECT_DIRECTORY}/.evergreen/install-composer.sh"
-          # Don't use ${file} syntax here because evergreen treats it as an empty expansion.
-          [ -f "$file" ] && DEPENDENCIES=${DEPENDENCIES} bash $file || echo "$file not available, skipping"
```
alcaeus, line 330: This line was responsible for Evergreen not exiting with a setup failure when something went wrong. With this change, any error while installing dependencies will cause a setup failure.

.evergreen/config/generate-config.php
```diff
@@ -13,7 +14,9 @@
 ];
 $supportedPhpVersions = array_merge($modernPhpVersions, $legacyPhpVersions);
 
-$latestPhpVersion = max($supportedPhpVersions);
+// TODO: use max() once PHP 8.4 is stable
+//$latestPhpVersion = max($supportedPhpVersions);
+$latestPhpVersion = '8.3';
```
alcaeus, line 19: Since PHP 8.4 isn't stable yet, I figured that the extra tests that run on the latest PHP version should continue to run on 8.3

.evergreen/install-composer.sh
```diff
@@ -38,4 +38,9 @@ php --ri mongodb
 
 install_composer
 
+# Remove psalm as it's not compatible with PHP 8.4: https://github.com/vimeo/psalm/pull/10928
+if [ "$PHP_VERSION" == "8.4" ]; then
+  php composer.phar remove --no-update --dev vimeo/psalm
```
alcaeus, line 43: Since we only run PHPUnit, I decided to explicitly remove the psalm dependency instead of ignoring any upper bounds for PHP version constraints, as that may result in actually incompatible packages being pulled in.

tests/GridFS/BucketFunctionalTest.php
```diff
@@ -860,6 +860,9 @@ public function testDanglingOpenWritableStream(): void
         $code = <<<'PHP'
             require '%s';
             require '%s';
+            // Don't report deprecations - if the issue exists this code will
+            // result in a fatal error
+            error_reporting(E_ALL & ~E_DEPRECATED & ~E_USER_DEPRECATED);
```
jmikola, line 865: Was this due to `disableMD5` being used below? Is that even relevant to what's being tested here? Why not remove it and allow deprecation notices to fail the test?

jmikola (approved):
I'm not sure how the GridFS tests relates to an "implicit nullable parameter", so I'll defer to you if I'm mistaken in thinking that `disableMD5` was to blame.

tests/GridFS/BucketFunctionalTest.php
alcaeus: The deprecation in question is about implicit nullable args in a PHPUnit class, which is loaded when we call `createTestClient` as it's in the inheritance chain. The issue is that checking for no output is the only way to test this behaviour. It goes back to #779 which fixed PHPLIB-345 - without the fix PHP fails with an autoloading failure as it already started unloading things when open resources (including streams with custom stream wrappers) are closed, leading to an autoload failure that doesn't find a class.

tests/UnifiedSpecTests/Constraint/MatchesTest.php
```diff
@@ -69,6 +71,10 @@ public function testOperatorExists(): void
 
     public function testOperatorType(): void
     {
+        if (version_compare(phpversion(), '8.4', '>=')) {
+            $this->markTestIncomplete('Test fails on PHP 8.4 due to deprecations');
+        }
+
```
GromNaN, line 77: Nit: I prefer using the `PHP_VERSION_ID` for performance reason.

```suggestion
        if (PHP_VERSION_ID >= 8_04_00) {
            $this->markTestIncomplete('Test fails on PHP 8.4 due to deprecations');
        }

```

But the best is to use the [`#[RequiresPhp('>= 8.4')]` attribute](https://docs.phpunit.de/en/10.5/attributes.html#requiresphp).

GromNaN (approved):
LGTM.

My comment is irrelevant since the skip code will be removed in the next minor 1.x release with the PHPUnit upgrade.

tests/UnifiedSpecTests/Constraint/MatchesTest.php
alcaeus: Ah, good point. There's also `#[RequiresPhpExtension]` that we can use in another tests that depends on the extension version. I wonder if we can also leverage this mechanism for server version checks - I'll look into it.

jmikola: Might be tricky for server version checks as it'd need to use a persistent client connection. But there may be an opportunity there to streamline everything with a service that captures all of that information when PHPUnit is first loaded (fetching the URI from environment variables).

PR #1494, opened 2024-10-23
PHPLIB-1185: Add new methods to get database and collection instances

alcaeus:
PHPLIB-1185

This PR adds the following new methods:
* `Client::getDatabase`, changed `Client::selectDatabase` and `Client::__get` to proxy to this method
* `Client::getCollection`, changed `Client::selectCollection` to proxy to this method
* `Database::getCollection`, changed `Database::selectCollection` and `Database::__get` to proxy to this method

Since they are widely used, I did not add any deprecation warnings, nor did I add any language to the documentation to indicate that they are deprecated. I left the tests as they are, as the old methods were changed to proxy to the new methods, so we are implicitly testing them. I can duplicate the existing tests, but I don't think this is necessary.

As a follow-up to this PR, we should update the documentation to use the new `get(Database|Collection)` methods, starting in 2.0 (as we only maintain a single documentation for 1.x, so users would have to figure out which method is correct for their version).

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -174,7 +174,7 @@ class Client
      */
     public function __get(string $databaseName)
     {
-        return $this->selectDatabase($databaseName);
+        return $this->getDatabase($databaseName);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


GromNaN (approved):
LGTM

src/Client.php
```diff
@@ -247,6 +247,43 @@ public function dropDatabase(string $databaseName, array $options = [])
...
+     * Returns a collection instance.
+     *
+     * If the collection does not exist in the database, it is not created when
+     * invoking this method.
+     *
+     * @see Collection::__construct() for supported options
+     * @param string $databaseName   Name of the database containing the collection
+     * @param string $collectionName Name of the collection to select
+     * @param array  $options        Collection constructor options
+     * @throws InvalidArgumentException for parameter/option parsing errors
```
jmikola, line 257: I suppose you want the doc block for the note about "does not exist", but the parameter annotations here have no value. The code is self-documenting.

src/Client.php
```diff
@@ -247,6 +247,43 @@ public function dropDatabase(string $databaseName, array $options = [])
...
+
+    /**
+     * Returns a database instance.
+     *
+     * If the database does not exist on the server, it is not created when
+     * invoking this method.
+     *
+     * @param string $databaseName Name of the database to select
+     * @param array  $options      Database constructor options
+     * @throws InvalidArgumentException for parameter/option parsing errors
```
jmikola: Likewise here, I don't think you need anything beyond the `@see` below.

src/Database.php
```diff
@@ -422,6 +422,30 @@ public function dropCollection(string $collectionName, array $options = [])
...
 
+    /**
+     * Returns a collection instance.
+     *
+     * If the collection does not exist in the database, it is not created when
+     * invoking this method.
+     *
+     * @param string $collectionName Name of the collection to select
+     * @param array  $options        Collection constructor options
+     * @throws InvalidArgumentException for parameter/option parsing errors
```
jmikola: `@param` and `@throws` can be dropped.

jmikola (commented):
A few suggestions to doc blocks.

src/Client.php
GromNaN: Does it helps static analysis to tell which kind of exception can be thrown?

jmikola: Perhaps, but since there is no `throws` within this method I would assume that tooling would inherit from doc blocks of any referenced methods, which would have `@throws`. I'll defer to you guys, though.

alcaeus: I left `@throws` in for the time being, since they are included in all other methods as well. I'm on the fence about the usefulness of them, I can see them being useful when coupled with static analysis, but since we don't have anything to enforce correctness of `@throws` statements, we may reconsider having them for methods that don't directly throw any exceptions.

That said, I did remove the `@param` tags as the parameter names are self-explanatory.

PR #1502, opened 2024-10-29
Ignore `disableMD5` option as `md5` field is removed from the spec

GromNaN:
The `disableMD5` option is ignored. The behavior is to never write the `md5` field, which is [supported by upload tests](https://github.com/mongodb/specifications/blob/04560fc74c823732c693e89a63027bc1453e6dc3/source/gridfs/tests/upload.yml#L55-L56). 

Related to
- https://github.com/mongodb/specifications/pull/1683 Removed test on `contentType` option
- https://github.com/mongodb/mongo-php-library/pull/1398 Removed `md5` field from the library

### DIFF
--- a/tests/UnifiedSpecTests/Context.php
+++ b/tests/UnifiedSpecTests/Context.php
@@ -463,7 +463,7 @@ final class Context
 
     private function createBucket(string $id, stdClass $o): void
     {
-        Util::assertHasOnlyKeys($o, ['id', 'database', 'bucketOptions', 'disableMD5']);
+        Util::assertHasOnlyKeys($o, ['id', 'database', 'bucketOptions']);
 
         $databaseId = $o->database ?? null;
         assertIsString($databaseId);
--- a/tests/UnifiedSpecTests/Context.php
+++ b/tests/UnifiedSpecTests/Context.php
@@ -463,7 +463,7 @@ final class Context
 
     private function createBucket(string $id, stdClass $o): void
     {
-        Util::assertHasOnlyKeys($o, ['id', 'database', 'bucketOptions']);
+        Util::assertHasOnlyKeys($o, ['id', 'database', 'bucketOptions', 'disableMD5']);
 
         $databaseId = $o->database ?? null;
         assertIsString($databaseId);
@@ -499,7 +499,7 @@ final class Context
 
     private static function prepareBucketOptions(array $options): array
     {
-        Util::assertHasOnlyKeys($options, ['bucketName', 'chunkSizeBytes', 'readConcern', 'readPreference', 'writeConcern']);
+        Util::assertHasOnlyKeys($options, ['bucketName', 'chunkSizeBytes', 'disableMD5', 'readConcern', 'readPreference', 'writeConcern']);
 
         if (array_key_exists('bucketName', $options)) {
             assertIsString($options['bucketName']);
--- a/tests/UnifiedSpecTests/Operation.php
+++ b/tests/UnifiedSpecTests/Operation.php
@@ -769,10 +769,6 @@ final class Operation
     private function executeForBucket(Bucket $bucket)
     {
         $args = $this->prepareArguments();
-
-        // "md5" field is removed from the spec, option "disableMD5" is ignored
-        unset($args['disableMD5']);
-
         Util::assertArgumentsBySchema(Bucket::class, $this->name, $args);
 
         switch ($this->name) {
--- a/tests/UnifiedSpecTests/Operation.php
+++ b/tests/UnifiedSpecTests/Operation.php
@@ -769,6 +769,10 @@ final class Operation
     private function executeForBucket(Bucket $bucket)
     {
         $args = $this->prepareArguments();
+
+        // "md5" field is removed from the spec, option "disableMD5" is ignored
+        unset($args['disableMD5']);
+
         Util::assertArgumentsBySchema(Bucket::class, $this->name, $args);
 
         switch ($this->name) {
--- a/tests/UnifiedSpecTests/UnifiedSpecTest.php
+++ b/tests/UnifiedSpecTests/UnifiedSpecTest.php
@@ -64,10 +64,6 @@ class UnifiedSpecTest extends FunctionalTestCase
         'index-management/search index operations ignore read and write concern: listSearchIndexes ignores read and write concern' => 'libmongoc appends readConcern to aggregate command',
         // Uses an invalid object name
         'run-command/runCursorCommand: does not close the cursor when receiving an empty batch' => 'Uses an invalid object name',
-        // GridFS deprecated fields are removed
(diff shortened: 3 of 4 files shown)


alcaeus (approved):
While it feels hacky to just unset the option, the alternative would be to re-add it to the schema but ignore it when actually executing commands. I don't think that is feasible.

tests/UnifiedSpecTests/UnifiedSpecTest.php
```diff
@@ -64,10 +64,6 @@ class UnifiedSpecTest extends FunctionalTestCase
         'index-management/search index operations ignore read and write concern: listSearchIndexes ignores read and write concern' => 'libmongoc appends readConcern to aggregate command',
         // Uses an invalid object name
         'run-command/runCursorCommand: does not close the cursor when receiving an empty batch' => 'Uses an invalid object name',
-        // GridFS deprecated fields are removed
-        'gridfs/gridfs-upload-disableMD5: upload when length is 0 sans MD5' => 'Deprecated fields are removed',
-        'gridfs/gridfs-upload-disableMD5: upload when length is 1 sans MD5' => 'Deprecated fields are removed',
-        'gridfs/gridfs-upload: upload when contentType is provided' => 'Deprecated fields are removed',
```
jmikola, line 72: Noted that this test was removed in https://github.com/mongodb/specifications/commit/12be2dfab99bdad54c3499174b0519ddb4588332.

tests/UnifiedSpecTests/Operation.php
```diff
@@ -769,6 +769,10 @@ private function executeForSession(Session $session)
     private function executeForBucket(Bucket $bucket)
     {
         $args = $this->prepareArguments();
+
+        // "md5" field is removed from the spec, option "disableMD5" is ignored
+        unset($args['disableMD5']);
```
jmikola: Is this needed to allow running `upload-disableMD5.yml`, which still exists in the upstream repository?

Those tests all assert `md5: { $$exists: false }`, so I'm not sure what it would matter. Wouldn't PHPLIB 2.x just ignore the option (vs. a possible deprecation notice in 1.x)?

GromNaN: v2.x ignore the option.
- If not set: md5 field is not set
- If set to false: md5 field is not set ...

In v1.x, we talked about deprecating not setting `disableMD5: true`, but that was rejected because of the burden of having to set an option for something that is not used.

jmikola: I likely misunderstood the point of this change. Is it just being done to satisfy the [Util.php](https://github.com/mongodb/mongo-php-library/blob/v1.x/tests/UnifiedSpecTests/Util.php) check for unsupported options? If so, it'd be helpful to make a note of that in this comment (since I've just demonstrated how likely I am to misunderstand the purpose of this `unset()` ^_^).

No objections to merging after that.

GromNaN: After investigations, `disableMD5` option is [documented in the spec](https://github.com/mongodb/specifications/blob/9058491f2d5b5651913a56e373121a24d6498673/source/gridfs/gridfs-spec.md?plain=1#L281-L288). So I added the option to the list of allowed options, with a comment.

GromNaN: So I restored the option `disableMD5` that I removed from the allowed options in https://github.com/mongodb/mongo-php-library/pull/1398. Note that the spec document the option to create a bucket, not to call the `upload` method.

jmikola: If spec tests actually use a `disableMD5` option for upload operations, I think that's an oversight in the spec itself. The option should probably be added to [File Upload](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.md#file-upload) with the same note it has on the corresponding bucket option (that it's deprecated and transitional).

Defer to you if that's worth doing at this point.

tests/UnifiedSpecTests/Context.php
```diff
@@ -463,7 +463,7 @@ private function createSession(string $id, stdClass $o): void
 
     private function createBucket(string $id, stdClass $o): void
     {
-        Util::assertHasOnlyKeys($o, ['id', 'database', 'bucketOptions']);
+        Util::assertHasOnlyKeys($o, ['id', 'database', 'bucketOptions', 'disableMD5']);
```
jmikola:
```suggestion
        Util::assertHasOnlyKeys($o, ['id', 'database', 'bucketOptions']);
```

Was this change actually required for anything? This is the code path for `createEntities`. The UTF schema doesn't allow `disableMD5` here, so it would only appear within `bucketOptions` if at all.

That said, it doesn't even look like we have spec tests using `bucketOptions`.

tests/UnifiedSpecTests/Util.php
```diff
@@ -134,8 +134,8 @@ final class Util
             'delete' => ['id'],
             'downloadByName' => ['filename', 'revision'],
             'download' => ['id'],
-            'uploadWithId' => ['id', 'filename', 'source', 'chunkSizeBytes', 'metadata'],
-            'upload' => ['filename', 'source', 'chunkSizeBytes', 'metadata'],
+            'uploadWithId' => ['id', 'filename', 'source', 'chunkSizeBytes', 'disableMD5', 'metadata'],
```
jmikola, line 141: Consider a comment like "disableMD5 is unsupported by PHPLIB but still defined in the spec" above these two lines.

tests/UnifiedSpecTests/Context.php
```diff
@@ -499,7 +499,7 @@ private static function prepareCollectionOrDatabaseOptions(array $options): arra
 
     private static function prepareBucketOptions(array $options): array
     {
-        Util::assertHasOnlyKeys($options, ['bucketName', 'chunkSizeBytes', 'readConcern', 'readPreference', 'writeConcern']);
+        Util::assertHasOnlyKeys($options, ['bucketName', 'chunkSizeBytes', 'disableMD5', 'readConcern', 'readPreference', 'writeConcern']);
```
jmikola, line 502: Per my earlier comment about no tests using `bucketOptions`, I doubt this change is required. But if you'd like to keep it, I think it warrants the same comment I suggested below in `Util.php` so any future reader understands why this option is still here.

GromNaN: You are correct, this option is not necessary for the existing spec tests. But it should be allowed according to the spec.

```suggestion
        Util::assertHasOnlyKeys($options, ['bucketName', 'chunkSizeBytes', 'readConcern', 'readPreference', 'writeConcern']);
```

tests/UnifiedSpecTests/Util.php
GromNaN:
```suggestion
            // "disableMD5" is ignored but allowed for backward compatibility
            'uploadWithId' => ['id', 'filename', 'source', 'chunkSizeBytes', 'disableMD5', 'metadata'],
```

jmikola (approved):

PR #1504, opened 2024-10-29
PHPLIB-1568 Add `GridFS\Bucket::deleteByName(filename)` and `renameByName(filename, newFilename)`

GromNaN:
Fix PHPLIB-1568

## `GridFS\Bucket::deleteByName($filename)`
Use the `CollectionWrapper::deleteFileAndChunksByFilename()` introduced for the need of the `StreamWrapper::unlink()`. 
- https://github.com/mongodb/mongo-php-library/pull/1206

Implementation:

1. find file ids by file name
```js
db.files.find({filename: <filename>}, {projection: {_id: 1}})
```
3. delete files by ids
```js
db.files.deleteMany({_id: {$in: <ids>}})
```
5. delete chunks by file ids
```js
db.chunks.deleteMany({files_id: {$in: <ids>}})
```

If a DB error occurs during 1, nothing happened.
If a DB error occurs during 2 or 3, some chunks may become orphan and we cannot retry as the mapping chunks-id-to-file-name is missing.

There could be an issue if the list of ids is too big. In that case, we would have to work by batches.

## `GridFS\Bucket::renameByName($filename)`
Use the `CollectionWrapper::updateFilenameForFilename()` introduced for the need of `StreamWrapper::rename()`
- https://github.com/mongodb/mongo-php-library/pull/1207

Implementation:

1. update files with the file name, to set the new filename
```js
db.files.updateMany({filename: <filename>}, {$set: {filename: <newFilename>}})
```

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -284,9 +284,18 @@
     </MixedArgumentTypeCoercion>
   </file>
   <file src="src/GridFS/CollectionWrapper.php">
+    <InvalidNullableReturnType>
+      <code><![CDATA[int]]></code>
+    </InvalidNullableReturnType>
     <MixedAssignment>
       <code><![CDATA[$ids[]]]></code>
     </MixedAssignment>
+    <NullableReturnStatement>
+      <code><![CDATA[$this->filesCollection->updateMany(
+            ['filename' => $filename],
+            ['$set' => ['filename' => $newFilename]],
+        )->getMatchedCount()]]></code>
+    </NullableReturnStatement>
   </file>
   <file src="src/GridFS/ReadableStream.php">
     <MixedArgument>
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -286,11 +286,13 @@
   <file src="src/GridFS/CollectionWrapper.php">
     <InvalidNullableReturnType>
       <code><![CDATA[int]]></code>
+      <code><![CDATA[int]]></code>
     </InvalidNullableReturnType>
     <MixedAssignment>
       <code><![CDATA[$ids[]]]></code>
     </MixedAssignment>
     <NullableReturnStatement>
+      <code><![CDATA[$count]]></code>
       <code><![CDATA[$this->filesCollection->updateMany(
             ['filename' => $filename],
             ['$set' => ['filename' => $newFilename]],
--- a/src/GridFS/Bucket.php
+++ b/src/GridFS/Bucket.php
@@ -243,6 +243,19 @@ class Bucket
     }
 
     /**
+     * Delete all the revisions of a file name from the GridFS bucket.
+     *
+     * @param string $filename Filename
+     *
+     * @throws FileNotFoundException if no file could be selected
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function deleteByName(string $filename): void
+    {
+        $this->collectionWrapper->deleteFileAndChunksByFilename($filename);
+    }
+
+    /**
      * Writes the contents of a GridFS file to a writable stream.
      *
      * @param mixed    $id          File ID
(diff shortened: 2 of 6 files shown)


src/GridFS/CollectionWrapper.php
```diff
@@ -266,7 +263,7 @@ public function insertFile(array|object $file): void
     /**
      * Updates the filename field in the file document for all the files with a given filename.
      */
-    public function updateFilenameForFilename(string $filename, string $newFilename): ?int
+    public function updateFilenameForFilename(string $filename, string $newFilename): int
```
GromNaN, line 266: The class is internal, we can change the return type. This is always an `int`, an exception is thrown in case of error.

psalm-baseline.xml
```diff
@@ -284,9 +284,18 @@
...
+    </InvalidNullableReturnType>
     <MixedAssignment>
       <code><![CDATA[$ids[]]]></code>
     </MixedAssignment>
+    <NullableReturnStatement>
+      <code><![CDATA[$this->filesCollection->updateMany(
+            ['filename' => $filename],
+            ['$set' => ['filename' => $newFilename]],
+        )->getMatchedCount()]]></code>
+    </NullableReturnStatement>
```
GromNaN, line 300: Boths additions will be removed in v2.x, thanks to https://github.com/mongodb/mongo-php-library/pull/1454

src/GridFS/CollectionWrapper.php
```diff
@@ -150,9 +150,6 @@ public function findChunksByFileId(mixed $id, int $fromChunk = 0)
      */
     public function findFileByFilenameAndRevision(string $filename, int $revision): ?object
     {
-        $filename = $filename;
-        $revision = $revision;
```
GromNaN, line 154: I thought I had already removed that.

jmikola: Noted that this remained when https://github.com/mongodb/mongo-php-library/commit/1b29a6d71e5511ea46b596461442ca94069bdf42 introduced type hints on the method.

src/GridFS/Bucket.php
```diff
@@ -648,6 +661,24 @@ public function rename(mixed $id, string $newFilename)
...
+     * @param string $newFilename New filename
+     *
+     * @throws FileNotFoundException if no file could be selected
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function renameByName(string $filename, string $newFilename): void
+    {
+        $matchedCount = $this->collectionWrapper->updateFilenameForFilename($filename, $newFilename);
+
+        if (! $matchedCount) {
```
jmikola:
```suggestion
        if ($matchedCount === 0) {
```

If consistency with `StreamWrapper::rename()` matters. That method also has the following comment, which may be relevant here, too:

> If $count is null, the update is unacknowledged, the operation is considered successful.

tests/GridFS/BucketFunctionalTest.php
```diff
@@ -160,6 +160,41 @@ public function testDeleteStillRemovesChunksIfFileDoesNotExist($input, $expected
         $this->assertCollectionCount($this->chunksCollection, 0);
     }
 
+    public function testDeleteByName(): void
```
jmikola, line 163: Should this Pr block on spec changes for [DRIVERS-2807](https://jira.mongodb.org/browse/DRIVERS-2807) and [DRIVERS-2808](https://jira.mongodb.org/browse/DRIVERS-2808)? Also, will the PR(s) for those tickets include spec tests to be synced?

GromNaN: Spec tests added in https://github.com/mongodb/specifications/pull/1702

jmikola (approved):

PR #1508, opened 2024-10-30
PHPLIB-1569: Implement $$matchAsDocument and $$matchAsRoot

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1569
https://jira.mongodb.org/browse/PHPLIB-1577

POC for https://github.com/mongodb/specifications/pull/1706

Includes an extra test for numeric comparisons, which can be removed when PHPLIB-1577 is addressed.

### DIFF
--- a/tests/UnifiedSpecTests/Constraint/Matches.php
+++ b/tests/UnifiedSpecTests/Constraint/Matches.php
@@ -3,6 +3,7 @@
 namespace MongoDB\Tests\UnifiedSpecTests\Constraint;
 
 use LogicException;
+use MongoDB\BSON\Document;
 use MongoDB\BSON\Serializable;
 use MongoDB\BSON\Type;
 use MongoDB\Model\BSONArray;
@@ -25,10 +26,13 @@ use function is_float;
 use function is_int;
 use function is_object;
 use function ltrim;
+use function PHPUnit\Framework\assertInstanceOf;
 use function PHPUnit\Framework\assertIsBool;
 use function PHPUnit\Framework\assertIsString;
+use function PHPUnit\Framework\assertJson;
 use function PHPUnit\Framework\assertMatchesRegularExpression;
 use function PHPUnit\Framework\assertNotNull;
+use function PHPUnit\Framework\assertStringStartsWith;
 use function PHPUnit\Framework\assertThat;
 use function PHPUnit\Framework\containsOnly;
 use function PHPUnit\Framework\isInstanceOf;
@@ -39,6 +43,7 @@ use function range;
 use function sprintf;
 use function str_starts_with;
 use function strrchr;
+use function trim;
 
 /**
  * Constraint that checks if one value matches another.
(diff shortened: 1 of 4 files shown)


tests/UnifiedSpecTests/Constraint/Matches.php
```diff
@@ -263,6 +268,35 @@ private function assertMatchesOperator(BSONDocument $operator, $actual, string $
             return;
         }
 
+        if ($name === '$$matchAsDocument') {
```
GromNaN, line 271: This succession of `if` look like it should be a `switch($name)`.

tests/UnifiedSpecTests/Constraint/Matches.php
```diff
@@ -263,6 +268,35 @@ private function assertMatchesOperator(BSONDocument $operator, $actual, string $
...
+            assertInstanceOf(BSONDocument::class, $operator['$$matchAsDocument'], '$$matchAsDocument requires a BSON document');
```
GromNaN, line 272: A `MongoDB\Model\BSONDocument` instance is expected because of the default typeMap of PHPLIB. Isn't there a risk that a test will change the typeMap and obtain a `stdClass` or `array`?

tests/UnifiedSpecTests/Constraint/Matches.php
```diff
@@ -263,6 +268,35 @@ private function assertMatchesOperator(BSONDocument $operator, $actual, string $
...
+
+            if (! $constraint->evaluate($actualDocument, '', true)) {
+                self::failAt(sprintf('%s did not match: %s', (new Exporter())->shortenedExport($actual), $constraint->additionalFailureDescription(null)), $keyPath);
+            }
+
+            return;
+        }
+
+        if ($name === '$$matchAsRoot') {
+            $constraint = new Matches($operator['$$matchAsRoot'], $this->entityMap, allowExtraRootKeys: true);
```
GromNaN, line 291: Why `allowExtraRootKeys` is `true` for `$$matchAsRoot` and `false` for `$$matchAsDocument`?

jmikola: Per the UTF spec, [`$$matchAsDocument`](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.md#matchasdocument) does _not_ allow extra keys in the decoded JSON document.

[`$$matchAsRoot`](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.md#matchasroot) only exists to enable "allow extra keys" behavior when matching.

tests/UnifiedSpecTests/Constraint/Matches.php
jmikola: I originally thought this has to do with the PHPLIB typeMap but forgot that it is actually due to the `Matches::prepare()` method, which is used to normalize both the expected _and_ actual values.

tests/UnifiedSpecTests/Constraint/Matches.php
jmikola: I don't feel strongly about this since each if block returns. I'm not sure there's any technical difference, as [switch](https://www.php.net/manual/en/control-structures.switch.php) only discusses a benefit for complex conditions (not a variable reference) since it evaluates once before comparing with each `case`.

You're welcome to refactor this in a separate PR if you like, but modifying the entire method body in this PR would make the actual changes less visible.

GromNaN (approved):
LGTM.

jmikola:
Note: I'll rebase and merge this after https://github.com/mongodb/specifications/pull/1706 is closed.

PR #1515, opened 2024-11-04
Change dependabot to weekly runs for specifications submodule

alcaeus:
Reverts mongodb/mongo-php-library#1497

With spec fest over, we can go back to weekly updates. Note that manual updates can be triggered anytime we deem necessary.


GromNaN (approved):
Let's use manual update next time.

PR #1525, opened 2024-11-06
PHPLIB-1582: Support hint option for distinct command

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1582

### DIFF
--- a/src/Operation/Distinct.php
+++ b/src/Operation/Distinct.php
@@ -31,6 +31,7 @@ use function current;
 use function is_array;
 use function is_integer;
 use function is_object;
+use function is_string;
 use function MongoDB\create_field_path_type_map;
 use function MongoDB\is_document;
 
@@ -53,7 +54,13 @@ class Distinct implements Executable, Explainable
      *
      *  * comment (mixed): BSON value to attach as a comment to this command.
      *
-     *    This is not supported for servers versions < 4.4.
+     *    This is not supported for server versions < 4.4.
+     *
+     *  * hint (string|document): The index to use. Specify either the index
+     *    name as a string or the index key pattern as a document. If specified,
+     *    then the query system will only consider plans using the hinted index.
+     *
+     *    This is not supported for server versions < 7.1.
      *
      *  * maxTimeMS (integer): The maximum amount of time to allow the query to
      *    run.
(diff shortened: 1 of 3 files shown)


src/Operation/Distinct.php
```diff
@@ -175,6 +186,11 @@ private function createCommandDocument(): array
             $cmd['collation'] = (object) $this->options['collation'];
         }
 
+        if (isset($this->options['hint'])) {
+            /** @psalm-var string|array|object */
```
GromNaN: Do you really need this annotation? If it applies to `$cmd['hint']`, it should be `string|object`, without `array`

src/Operation/Distinct.php
```diff
@@ -83,6 +90,10 @@ public function __construct(private string $databaseName, private string $collec
             throw InvalidArgumentException::expectedDocumentType('"collation" option', $this->options['collation']);
         }
 
+        if (isset($this->options['hint']) && ! is_string($this->options['hint']) && ! is_array($this->options['hint']) && ! is_object($this->options['hint'])) {
```
GromNaN: The doc says `string|document`, the logical test should use `is_document`.

```suggestion
        if (isset($this->options['hint']) && ! is_string($this->options['hint']) && ! is_document($this->options['hint'])) {
```

src/Operation/Distinct.php
```diff
@@ -175,6 +186,11 @@ private function createCommandDocument(): array
             $cmd['collation'] = (object) $this->options['collation'];
         }
 
+        if (isset($this->options['hint'])) {
+            /** @psalm-var string|array|object */
+            $cmd['hint'] = is_array($this->options['hint']) ? (object) $this->options['hint'] : $this->options['hint'];
```
GromNaN, line 191: Isn't the `array` automatically converted to a BSON document? Or maybe that handles edge cases like `[]` or `[0 => 'foo']`.

tests/Operation/DistinctTest.php
```diff
@@ -42,6 +43,7 @@ public function testExplainableCommandDocument(): void
     {
         $options = [
             'collation' => ['locale' => 'fr'],
+            'hint' => '_id_',
```
GromNaN, line 46: The `hint` option is supported in `explain` command on server version < 7.1.0?

jmikola: I assume not. This is just a unit test, so we're only asserting that the option is included via the Explainable interface.

src/Operation/Distinct.php
jmikola: This is just for the edge case you highlighted.

src/Operation/Distinct.php
jmikola: Psalm did complain with out this. Good catch, though, as I can remove `array` here.

src/Operation/Distinct.php
jmikola: Good catch. I'll also make a separate ticket to migrate other `hint` option validation to use `is_document`, as it looks like we missed that when introducing PackedArray.

jmikola: Opened [PHPLIB-1587](https://jira.mongodb.org/browse/PHPLIB-1587) to track that.

GromNaN (approved):

PR #1526, opened 2024-11-07
Fix PHPUnit deprecation notices

jmikola:
This addresses several "Providing invalid named argument" notices. PHPUnit 11 will require keys in data providers to match method argument names.

### DIFF
--- a/tests/Operation/WatchFunctionalTest.php
+++ b/tests/Operation/WatchFunctionalTest.php
@@ -488,7 +488,7 @@ class WatchFunctionalTest extends FunctionalTestCase
     }
 
     #[DataProvider('provideCodecOptions')]
-    public function testNoChangeAfterResumeBeforeInsert(array $options): void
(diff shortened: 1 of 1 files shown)


tests/Operation/WatchFunctionalTest.php
```diff
@@ -488,7 +488,7 @@ public function testRewindMultipleTimesWithNoResults(): void
     }
 
     #[DataProvider('provideCodecOptions')]
-    public function testNoChangeAfterResumeBeforeInsert(array $options): void
+    public function testNoChangeAfterResumeBeforeInsert(array $options, Closure $getIdentifier): void
```
jmikola, line 491: Let me know if there's some annotation or other comment you'd like to see used to denote that this parameter is unused.

GromNaN: That could be `@phpstan-ignore constructor.unusedParameter`, but if PHPStan doesn't complain, that's fine not having it.

GromNaN (approved):

PR #1537, opened 2024-11-25
PHPLIB-1596 Fully describe `$search`, `$searchMeta` and `$vectorSearch` stages

GromNaN:
Fix PHPLIB-1596

The parameters of the `$search` and `$searchMeta` stages are currently not described. Changing the generic property object into a full list of named arguments would be a breaking change after the first release of this code, so it's now or never.

For the tests, you can see the deep structure of options is not described. That's something we an do later.

- [x] https://github.com/mongodb/mongo-php-library/pull/1540
- [x] https://github.com/mongodb/mongo-php-library/pull/1541

### DIFF
--- a/generator/composer.json
+++ b/generator/composer.json
@@ -4,7 +4,7 @@
     "repositories": [
         {
             "type": "path",
-            "url": "../",
+            "url": "..",
             "symlink": true
         }
     ],
--- a/generator/config/expressions.php
+++ b/generator/config/expressions.php
@@ -109,6 +109,10 @@ return $expressions + [
         'implements' => [ResolvesToAny::class],
         'acceptedTypes' => ['string'],
     ],
+    'searchOperator' => [
+        'returnType' => Type\SearchOperatorInterface::class,
+        'acceptedTypes' => [Type\SearchOperatorInterface::class, ...$bsonTypes['object']],
+    ],
     'geometry' => [
         'returnType' => Type\GeometryInterface::class,
         'acceptedTypes' => [Type\GeometryInterface::class, ...$bsonTypes['object']],
@@ -166,6 +170,14 @@ return $expressions + [
         'acceptedTypes' => [...$bsonTypes['object']],
     ],
     'GeoPoint' => [
+        'acceptedTypes' => [...$bsonTypes['object']],
+    ],
+
+    // Search
+    'searchPath' => [
+        'acceptedTypes' => ['string', 'array'],
+    ],
+    'searchScore' => [
         'acceptedTypes' => [...$bsonTypes['object']],
     ],
 ];
--- a/generator/config/search/autocomplete.yaml
+++ b/generator/config/search/autocomplete.yaml
@@ -1,0 +1,152 @@
+# $schema: ../schema.json
+name: autocomplete
+link: 'https://www.mongodb.com/docs/atlas/atlas-search/autocomplete/'
+type:
+    - searchOperator
+encode: object
+description: |
+    The autocomplete operator performs a search for a word or phrase that
+    contains a sequence of characters from an incomplete input string. The
+    fields that you intend to query with the autocomplete operator must be
+    indexed with the autocomplete data type in the collection's index definition.
+arguments:
+    -
+        name: path
+        type:
+            - searchPath
+    -
+        name: query
+        type:
+            - string
+    -
+        name: tokenOrder
+        optional: true
+        type:
+            - string # any|sequential
+    -
+        name: fuzzy
+        optional: true
+        type:
+            - object
+    -
+        name: score
+        optional: true
+        type:
+            - searchScore
+tests:
+    -
+        name: 'Basic'
+        link: 'https://www.mongodb.com/docs/atlas/atlas-search/autocomplete/#basic-example'
+        pipeline:
+            -
+                $search:
+                    autocomplete:
+                        query: 'off'
+                        path: 'title'
+            -
+                $limit: 10
+            -
+                $project:
+                    _id: 0
+                    title: 1
+
+    -
+        name: 'Fuzzy'
+        link: 'https://www.mongodb.com/docs/atlas/atlas-search/autocomplete/#fuzzy-example'
+        pipeline:
+            -
+                $search:
+                    autocomplete:
+                        query: 'pre'
+                        path: 'title'
+                        fuzzy:
+                            maxEdits: 1
+                            prefixLength: 1
+                            maxExpansions: 256
+            -
+                $limit: 10
+            -
+                $project:
+                    _id: 0
+                    title: 1
+
+    -
+        name: 'Token Order any'
+        link: 'https://www.mongodb.com/docs/atlas/atlas-search/autocomplete/#simple-any-example'
+        pipeline:
+            -
+                $search:
+                    autocomplete:
+                        query: 'men with'
+                        path: 'title'
+                        tokenOrder: 'any'
+            -
+                $limit: 4
+            -
+                $project:
+                    _id: 0
+                    title: 1
+
+    -
+        name: 'Token Order sequential'
+        link: 'https://www.mongodb.com/docs/atlas/atlas-search/autocomplete/#simple-sequential-example'
+        pipeline:
+            -
+                $search:
+                    autocomplete:
+                        query: 'men with'
+                        path: 'title'
+                        tokenOrder: 'sequential'
+            -
+                $limit: 4
+            -
+                $project:
+                    _id: 0
+                    title: 1
+
+    -
+        name: 'Highlighting'
+        link: 'https://www.mongodb.com/docs/atlas/atlas-search/autocomplete/#highlighting-example'
+        pipeline:
+            -
+                $search:
+                    autocomplete:
+                        query: 'ger'
+                        path: 'title'
+                    highlight:
+                        path: 'title'
+            -
+                $limit: 5
+            -
+                $project:
+                    score:
+                        $meta: 'searchScore'
+                    _id: 0
+                    title: 1
+                    highlights:
+                        $meta: 'searchHighlights'
+
+    -
+        name: 'Across Multiple Fields'
+        link: 'https://www.mongodb.com/docs/atlas/atlas-search/autocomplete/#search-across-multiple-fields'
+        pipeline:
+            -
+                $search:
+                    compound:
+                        should:
+                            -
+                                autocomplete:
+                                    query: 'inter'
+                                    path: 'title'
+                            -
+                                autocomplete:
+                                    query: 'inter'
+                                    path: 'plot'
+                        minimumShouldMatch: 1
+            -
+                $limit: 10
+            -
+                $project:
+                    _id: 0
+                    title: 1
+                    plot: 1
--- a/generator/config/search/compound.yaml
+++ b/generator/config/search/compound.yaml
@@ -1,0 +1,156 @@
+# $schema: ../schema.json
+name: compound
+link: 'https://www.mongodb.com/docs/atlas/atlas-search/compound/'
+type:
+    - searchOperator
+encode: object
+description: |
+    The compound operator combines two or more operators into a single query.
+    Each element of a compound query is called a clause, and each clause
+    consists of one or more sub-queries.
+arguments:
+    -
+        name: must
+        optional: true
+        type:
+            - searchOperator
+            - array # of searchOperator
+    -
+        name: mustNot
+        optional: true
+        type:
+            - searchOperator
+            - array # of searchOperator
+    -
+        name: should
+        optional: true
+        type:
+            - searchOperator
+            - array # of searchOperator
+    -
+        name: filter
+        optional: true
+        type:
+            - searchOperator
+            - array # of searchOperator
+    -
+        name: minimumShouldMatch
+        optional: true
+        type:
+            - int
+    -
+        name: score
+        optional: true
+        type:
+            - searchScore
+tests:
+    -
+        name: 'must and mustNot'
+        link: 'https://www.mongodb.com/docs/atlas/atlas-search/compound/#must-and-mustnot-example'
+        pipeline:
+            -
+                $search:
+                    compound:
+                        must:
+                            -
+                                text:
+                                    query: 'varieties'
+                                    path: 'description'
+                        mustNot:
+                            -
+                                text:
+                                    query: 'apples'
+                                    path: 'description'
+
+    -
+        name: 'must and should'
+        link: 'https://www.mongodb.com/docs/atlas/atlas-search/compound/#must-and-should-example'
+        pipeline:
+            -
+                $search:
+                    compound:
+                        must:
+                            -
+                                text:
+                                    query: 'varieties'
+                                    path: 'description'
+                        should:
+                            -
+                                text:
+                                    query: 'Fuji'
+                                    path: 'description'
+            -
+                $project:
+                    score:
+                        $meta: 'searchScore'
+
+    -
+        name: 'minimumShouldMatch'
+        link: 'https://www.mongodb.com/docs/atlas/atlas-search/compound/#minimumshouldmatch-example'
+        pipeline:
+            -
+                $search:
+                    compound:
+                        must:
+                            -
+                                text:
+                                    query: 'varieties'
+                                    path: 'description'
+                        should:
+                            -
+                                text:
+                                    query: 'Fuji'
+                                    path: 'description'
+                            -
+                                text:
+                                    query: 'Golden Delicious'
+                                    path: 'description'
+                        minimumShouldMatch: 1
+
+    -
+        name: 'Filter'
+        link: 'https://www.mongodb.com/docs/atlas/atlas-search/compound/#filter-examples'
+        pipeline:
+            -
+                $search:
+                    compound:
+                        must:
+                            -
+                                text:
+                                    query: 'varieties'
+                                    path: 'description'
+                        should:
+                            -
+                                text:
+                                    query: 'banana'
+                                    path: 'description'
+                        filter:
+                            -
+                                text:
+                                    query: 'granny'
+                                    path: 'description'
+
+    -
+        name: 'Nested'
+        link: 'https://www.mongodb.com/docs/atlas/atlas-search/compound/#nested-example'
+        pipeline:
+            -
+                $search:
+                    compound:
+                        should:
+                            -
+                                text:
+                                    query: 'apple'
+                                    path: 'type'
+                            -
+                                compound:
+                                    must:
+                                        -
+                                            text:
+                                                query: 'organic'
+                                                path: 'category'
+                                        -
+                                            equals:
+                                                value: true
+                                                path: 'in_stock'
+                        minimumShouldMatch: 1
--- a/generator/config/search/embeddedDocument.yaml
+++ b/generator/config/search/embeddedDocument.yaml
@@ -1,0 +1,155 @@
+# $schema: ../schema.json
+name: embeddedDocument
+link: 'https://www.mongodb.com/docs/atlas/atlas-search/embedded-document/'
+type:
+    - searchOperator
+encode: object
+description: |
+    The embeddedDocument operator is similar to $elemMatch operator.
+    It constrains multiple query predicates to be satisfied from a single
+    element of an array of embedded documents. embeddedDocument can be used only
+    for queries over fields of the embeddedDocuments
+arguments:
+    -
+        name: path
+        type:
+            - searchPath
+    -
+        name: operator
+        type:
+            - searchOperator
+    -
+        name: score
+        optional: true
+        type:
+            - searchScore
+tests:
+    -
+        name: 'Basic'
+        link: 'https://www.mongodb.com/docs/atlas/atlas-search/embedded-document/#index-definition'
+        pipeline:
+            -
+                $search:
+                    embeddedDocument:
+                        path: 'items'
+                        operator:
+                            compound:
+                                must:
+                                    -
+                                        text:
+                                            path: 'items.tags'
+                                            query: 'school'
+                                should:
+                                    -
+                                        text:
+                                            path: 'items.name'
+                                            query: 'backpack'
+                        score:
+                            embedded:
+                                aggregate: 'mean'
+            -
+                $limit: 5
+            -
+                $project:
+                    _id: 0
+                    items.name: 1
+                    items.tags: 1
+                    score:
+                        $meta: 'searchScore'
+
+    -
+        name: 'Facet'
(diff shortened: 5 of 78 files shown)


generator/composer.json
```diff
@@ -4,7 +4,7 @@
     "repositories": [
         {
             "type": "path",
-            "url": "../",
+            "url": "..",
```
jmikola, line 7: Was the trailing slash just redundant?

generator/src/OperatorTestGenerator.php
```diff
@@ -145,6 +147,8 @@ private function convertYamlTaggedValues(mixed $object): mixed
                 'bson_decimal128' => new Decimal128($value),
                 'bson_utcdatetime' => new UTCDateTime(is_numeric($value) ? $value : new DateTimeImmutable($value)),
                 'bson_binary' => new Binary(base64_decode($value)),
+                'bson_objectId' => new ObjectId($value),
+                'bson_uuid' => new Binary(pack('h*', str_replace('-', '', $value)), Binary::TYPE_UUID),
```
jmikola: Would [`hex2bin()`](https://www.php.net/manual/en/function.hex2bin.php) be more appropriate here? It's not immediately clear to me if the `h` format (low nibble first) for [`pack()`](https://www.php.net/pack) is correct.

I'll note that [`bin2hex()`](https://www.php.net/bin2hex) produces a string with the high nibble first.

generator/src/OperatorTestGenerator.php
```diff
@@ -158,6 +162,7 @@ private function convertYamlTaggedValues(mixed $object): mixed
         }
 
         if (is_object($object)) {
+            $object = clone $object;
```
jmikola, line 165: Was this a bug fix? It looks unrelated to implementing search operators.

tests/Builder/Stage/VectorSearchStageTest.php
```diff
@@ -0,0 +1,85 @@
...
+
+use MongoDB\Builder\Pipeline;
+use MongoDB\Builder\Query;
+use MongoDB\Builder\Stage;
+use MongoDB\Tests\Builder\PipelineTestCase;
+
+/**
+ * Test $vectorSearch stage
+ */
+class VectorSearchStageTest extends PipelineTestCase
```
jmikola, line 15: Are these files generated, or is that only `tests/Builder/Stage/Pipelines.php` (which has a header comment indicating as much). None of these appear to be automatically hidden in the code review, so I wonder if the repo config needs an update for changed paths.

tests/Builder/Search/Pipelines.php
```diff
@@ -0,0 +1,2814 @@
+<?php
+
+/**
+ * THIS FILE IS AUTO-GENERATED. ANY CHANGES WILL BE LOST!
```
jmikola, line 4: Noted this file is auto-generated but GitHub only suppressed it because of its size. The path may need to be added in the repo config.

tests/Builder/Search/InOperatorTest.php
```diff
@@ -0,0 +1,101 @@
...
+use DateTimeImmutable;
+use MongoDB\BSON\ObjectId;
+use MongoDB\BSON\UTCDateTime;
+use MongoDB\Builder\Pipeline;
+use MongoDB\Builder\Search;
+use MongoDB\Builder\Stage;
+use MongoDB\Tests\Builder\PipelineTestCase;
+
+/**
+ * Test in search
```
jmikola, line 16: If these descriptions are hand-written, it may be preferable to use the `$` prefix so they actually read as the operator names. This could be misread as a sentence, leaving the reader to ask, "test _what_ in the search?"

Consider:

> Test $in search operator

tests/Builder/Stage/VectorSearchStageTest.php
```diff
@@ -0,0 +1,85 @@
...
+
+namespace MongoDB\Tests\Builder\Stage;
+
+use MongoDB\Builder\Pipeline;
+use MongoDB\Builder\Query;
+use MongoDB\Builder\Stage;
+use MongoDB\Tests\Builder\PipelineTestCase;
+
+/**
+ * Test $vectorSearch stage
```
jmikola, line 13: I appreciate using `$` to prefix the stage names here.

generator/config/expressions.php
```diff
@@ -168,4 +172,12 @@
     'GeoPoint' => [
         'acceptedTypes' => [...$bsonTypes['object']],
     ],
+
+    // Search
+    'searchPath' => [
```
jmikola, line 177: Why do these keys start with lowercase letters? It looks inconsistent with the keys above.

Also, the comment on the line above may not be necessary since both keys start with "search". If the comment is intentional to distinguish these two array elements from those above, that's fine. I expect answering my first question above will clear up any misunderstanding.

generator/config/search/autocomplete.yaml
```diff
@@ -0,0 +1,152 @@
...
+            - searchPath
+    -
+        name: query
+        type:
+            - string
+    -
+        name: tokenOrder
+        optional: true
+        type:
+            - string #tokenOrder
```
jmikola: Does this comment add any value? I assume it may be referring to:

```
"tokenOrder": "any|sequential",
```

I assume you don't want to define an enum for that, nor enforce it client-side, but are the possible values documented anywhere in the builder?

generator/config/search/embeddedDocument.yaml
```diff
@@ -0,0 +1,155 @@
+# $schema: ../schema.json
+name: embeddedDocument
+link: 'https://www.mongodb.com/docs/atlas/atlas-search/embeddedDocument/'
```
jmikola: This looks like a 404. Also, I need to reiterate how annoying it is that our docs website redirects 404s to https://www.mongodb.com/docs/404/ instead of leaving the original URL as-is. I've reported this in the past and the feedback wasn't acted upon :disappointed: 

I think this should be https://www.mongodb.com/docs/atlas/atlas-search/embedded-document/

generator/config/search/compound.yaml
```diff
@@ -0,0 +1,156 @@
+# $schema: ../schema.json
+name: compound
+link: 'https://www.mongodb.com/docs/atlas/atlas-search/equals/'
```
jmikola:
```suggestion
link: 'https://www.mongodb.com/docs/atlas/atlas-search/compound/'
```

generator/config/search/equals.yaml
```diff
@@ -0,0 +1,104 @@
...
+        type:
+            - searchPath
+    -
+        name: value
+        type:
+            - binData
+            - bool
+            - date
+            - objectId
+            - 'null'
```
jmikola, line 21: Is this necessary to overcome some YAML behavior, where `null` would be interpreted as `~`?

generator/config/search/exists.yaml
```diff
@@ -0,0 +1,56 @@
+# $schema: ../schema.json
+name: exists
+link: 'https://www.mongodb.com/docs/atlas/atlas-search/equals/'
```
jmikola:
```suggestion
link: 'https://www.mongodb.com/docs/atlas/atlas-search/exists/'
```

generator/config/search/exists.yaml
```diff
@@ -0,0 +1,56 @@
...
+type:
+    - searchOperator
+encode: object
+description: |
+    The equals operator checks whether a field matches a value you specify.
```
jmikola: This looks like copypasta.

generator/config/search/facet.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - searchOperator # should be searchCollector
+encode: object
+description: |
+    The facet collector groups results by values or ranges in the specified
+    faceted fields and returns the count for each of those groups.
+arguments:
+    -
+        name: facets
+        type:
+            - object # of facetDefinition
```
jmikola: You previously used "of <type>" for an `array`. Was this intentional?

I assume it may be because it's a hash of facet names to definitions. Maybe change this to "# map of facetDefinition"?

generator/config/search/geoShape.yaml
```diff
@@ -0,0 +1,171 @@
...
+    geometry if indexShapes is set to true in the index definition.
+arguments:
+    -
+        name: path
+        type:
+            - searchPath
+    -
+        name: relation
+        type:
+            - string # contains | disjoint | intersects | within
```
jmikola, line 18: Curious why you enumerated the possible values here, but not for the tokenOrder string earlier.

generator/config/search/geoShape.yaml
```diff
@@ -0,0 +1,171 @@
...
+                $search:
+                    geoShape:
+                        relation: 'disjoint'
+                        geometry:
+                            type: 'Polygon'
+                            coordinates:
+                                -
+                                    -
+                                        - -161.323242
+                                        - 22.512557
```
jmikola: Would it be more readable to inline these, like `[ -161.323242, 22.512557 ]`? That might reduce a lot of lines in this file.

If there's a concern that it's generated and might conflict with future syncs, no worries leaving as-is. I'm not familiar with how these are imported.

generator/config/search/geoWithin.yaml
```diff
@@ -0,0 +1,117 @@
...
+        pipeline:
+            -
+                $search:
+                    geoWithin:
+                        geometry:
+                            type: 'Polygon'
+                            coordinates:
+                                -
+                                    -
+                                        - -161.323242
```
jmikola: Same inlining suggestion for this file.

jmikola (commented):
Taking a first pass and will leave more feedback on other YAML files later.

generator/config/search/autocomplete.yaml
GromNaN: Updating the comment, as I don't have implemented this enum for now.
```suggestion
            - string # any|sequential
```

generator/config/expressions.php
GromNaN: I think that when I did this initially, there was a distinction between “scalar” and “object” types in my mind. But that doesn't make sense. I need to harmonize this. It has no impact on the generated code.

generator/config/search/embeddedDocument.yaml
GromNaN:
```suggestion
link: 'https://www.mongodb.com/docs/atlas/atlas-search/embedded-document/'
```

generator/config/search/equals.yaml
GromNaN: Exactly, the unquoted `null` string is the NULL value in Yaml. But it seems that I can disable this behavior: https://github.com/symfony/yaml/blob/3ced3f29e4f0d6bce2170ff26719f1fe9aacc671/CHANGELOG.md#440

Let's stay safe and keep the quotes, if other scripts use this Yaml files.

generator/config/search/exists.yaml
GromNaN:
```suggestion
    The exists operator tests if a path to a specified indexed field name exists in a document.
```

generator/config/search/facet.yaml
GromNaN: Updated, thanks.
This comments are a kind of todo list to improve typing.
```suggestion
            - object # map of facetDefinition
```

generator/config/search/geoShape.yaml
GromNaN: Updating tokenOrder

generator/config/search/geoShape.yaml
GromNaN: It's generated by my `js2yaml.html` file. I've made a few editions in some places, so why not...

generator/src/OperatorTestGenerator.php
GromNaN: Thanks.
```suggestion
                'bson_uuid' => new Binary(hex2bin(str_replace('-', '', $value)), Binary::TYPE_UUID),
```

generator/src/OperatorTestGenerator.php
GromNaN: This is kind of but that doesn't affect the current code. When calling `$pipeline = $this->convertYamlTaggedValues($test->pipeline);`, the original `$test->pipeline` is modified, which is not expected.

I hit this issue because I hacked into this file to generate the pipeline of test methods with some dumb implementation. I did not commit that because that was not good.

tests/Builder/Search/Pipelines.php
GromNaN: I'll fix the .gitattribute config.

tests/Builder/Stage/VectorSearchStageTest.php
GromNaN: The test files are generated with almost empty method body, then I manually write every aggregation pipeline using the builder syntax. That why this files are not flagged as "generated".

generator/composer.json
GromNaN: That was not necessary.

GromNaN (commented):
Thanks for this first review. I'll commit the fixes.

generator/src/OperatorTestGenerator.php
jmikola: Oh, I originally missed that this was in the generator sources. Glad it's fixed and I see no need for a PHPLIB ticket to document this since the code is internal.

generator/src/OperatorTestGenerator.php
```diff
@@ -145,6 +147,8 @@ private function convertYamlTaggedValues(mixed $object): mixed
                 'bson_decimal128' => new Decimal128($value),
                 'bson_utcdatetime' => new UTCDateTime(is_numeric($value) ? $value : new DateTimeImmutable($value)),
                 'bson_binary' => new Binary(base64_decode($value)),
+                'bson_objectId' => new ObjectId($value),
+                'bson_uuid' => new Binary(hex2bin(str_replace('-', '', $value)), Binary::TYPE_UUID),
```
jmikola, line 151: Because I was curious, I went back and compared the original code you had with `pack('h*', ...)`. See below:

```php
<?php

// Initialized project with: composer require ramsey/uuid
require 'vendor/autoload.php';

use Ramsey\Uuid\Uuid;
use MongoDB\BSON\Binary;

$uuid = Uuid::uuid4();
var_dump($uuid);

$bytes = $uuid->getBytes();
$str = $uuid->toString();

$bin = new Binary($bytes, Binary::TYPE_UUID);
var_dump(bin2hex($bin));

$bin2 = new Binary(hex2bin(str_replace('-','',$str)), Binary::TYPE_UUID);
var_dump($bin == $bin2);

$bin3 = new Binary(pack('h*', str_replace('-','',$str)), Binary::TYPE_UUID);
var_dump($bin == $bin3);

$bin4 = new Binary(pack('H*', str_replace('-','',$str)), Binary::TYPE_UUID);
var_dump($bin == $bin4);
```

Output:

```
$ php foo.php 
class Ramsey\Uuid\Lazy\LazyUuidFromString#4 (2) {
  private ?Ramsey\Uuid\UuidInterface $unwrapped =>
  NULL
  private string $uuid =>
  string(36) "3a6224c2-f648-4978-b510-610c04fbbcd6"
}
string(32) "3a6224c2f6484978b510610c04fbbcd6"
bool(true)
bool(false)
bool(true)
```

So it looks like `h` (hex string, low nibble first) wouldn't have worked, but `H` (high nibble first) would have. `hex2bin()` is totally fine, though, and easier to read.

src/Builder/Encoder/OperatorEncoder.php
```diff
@@ -129,6 +132,34 @@ private function encodeAsObject(OperatorInterface $value): stdClass
             : $this->wrap($value, $result);
     }
 
+    private function encodeAsSearch(OperatorInterface $value): stdClass
+    {
+        assert($value instanceof SearchStage || $value instanceof SearchMetaStage);
```
jmikola: Is there a reason these methods use `assert()` instead of strict type hints?

src/Builder/Encoder/OperatorEncoder.php
```diff
@@ -129,6 +132,34 @@ private function encodeAsObject(OperatorInterface $value): stdClass
...
+            // Skip optional arguments. If they have a default value, it is resolved by the server.
+            if ($val === Optional::Undefined) {
+                continue;
+            }
+
+            // Merge the operator properties into the result object
+            if ($key === 'operator') {
+                $val = $this->recursiveEncode($val);
+                assert($val instanceof stdClass);
+                foreach (get_object_vars($val) as $subkey => $subval) {
```
jmikola: If `$val` is a stdClass, do you even need `get_object_vars()`? [Basic object iteration](https://www.php.net/manual/en/language.oop5.iterations.php) rules should apply. No objections if you feel this is more explicit/consistent.

Separate question: is the logic here also frequently applied in other operators? Or is it unique in its unwrapping and merging of the `operator` object into the parent context?

Is there any concern that operator keys (i.e. `$subkey` values) might conflict with a top-level `$key`?

generator/config/stage/search.yaml
```diff
@@ -3,16 +3,92 @@ name: $search
...
+        description: |
+            Operator to search with.  You can provide a specific operator or use
+            the compound operator to run a compound query with multiple operators.
+    -
+        name: index
+        optional: true
+        type:
+            - string
+        description: |
+            Name of the Atlas Search index to use. If omitted, defaults to default.
```
jmikola:
```suggestion
            Name of the Atlas Search index to use. If omitted, defaults to "default".
```

Would be helpful to clarify the string literal here.

generator/config/search/queryString.yaml
```diff
@@ -0,0 +1,35 @@
...
+        type:
+            - searchPath
+    -
+        name: query
+        type:
+            - string
+
+# The various example from the doc are variations of the "query" parameter
+# this is not pertinent for testing the aggregation builder, unless we create
+# a queryString builder.
```
jmikola, line 21: Let's not :wink:

generator/config/expressions.php
GromNaN: [PHPLIB-1598](https://jira.mongodb.org/browse/PHPLIB-1598)

tests/Builder/Search/InOperatorTest.php
GromNaN: In fact, the search operators are not prefixed by the dollar sign 🤦.
I'll do a following PR to add quotes as this will impact all test classes (or maybe remove the comment).

src/Builder/Encoder/OperatorEncoder.php
GromNaN: Indeed, fixed.

src/Builder/Encoder/OperatorEncoder.php
GromNaN: > If `$val` is a stdClass, do you even need `get_object_vars()`? [Basic object iteration](https://www.php.net/manual/en/language.oop5.iterations.php) rules should apply. No objections if you feel this is more explicit/consistent.

For the time being, in removed the `get_object_vars`.

> Separate question: is the logic here also frequently applied in other operators? Or is it unique in its unwrapping and merging of the `operator` object into the parent context?
> Is there any concern that operator keys (i.e. `$subkey` values) might conflict with a top-level `$key`?

I opened a PR to solve the need of merging the subproperties into the parent object: https://github.com/mongodb/mongo-php-library/pull/1540

GromNaN (commented):
Let's review https://github.com/mongodb/mongo-php-library/pull/1540 before going further.

src/Builder/Encoder/OperatorEncoder.php
GromNaN: Method removed.

tests/Builder/Search/Pipelines.php
GromNaN: There is this line in the `.gitattribute` file that should have detected this file.
```
/tests/Builder/*/Pipelines.php linguist-generated=true
```

generator/config/stage/search.yaml
```diff
@@ -3,16 +3,93 @@ name: $search
...
+            - object
+        description: |
+            Document that specifies the fields to sort the Atlas Search results by in ascending or descending order.
+    -
+        name: returnStoredSource
+        optional: true
+        type:
+            - bool
+        description: |
+            Flag that specifies whether to return the stored source document in the search results. If omitted, defaults to false.Flag that specifies whether to perform a full document lookup on the backend database or return only stored source fields directly from Atlas Search.
```
jmikola: Missing spacing between the sentences here. Perhaps a copy/paste error?

generator/config/stage/search.yaml
```diff
@@ -3,16 +3,93 @@ name: $search
...
+            Specifies the highlighting options for displaying search terms in their original context.
+    -
+        name: concurrent
+        optional: true
+        type:
+            - bool
+        description: |
+            Parallelize search across segments on dedicated search nodes.
+            If you don't have separate search nodes on your cluster, 
+            Atlas Search ignores this flag. If omitted, defaults to false.
```
jmikola, line 43: Elsewhere, I recall you specifying `default: <value>` in the YAML configs. Should that be done here? Likewise for the `index` option above and some other params below with default values.

generator/config/stage/searchMeta.yaml
```diff
@@ -3,16 +3,34 @@ name: $searchMeta
...
+        description: |
+            Operator to search with.  You can provide a specific operator or use
+            the compound operator to run a compound query with multiple operators.
+    -
+        name: index
+        optional: true
+        type:
+            - string
+        description: |
+            Name of the Atlas Search index to use. If omitted, defaults to default.
```
jmikola, line 25: Should `default:` be used here?

generator/config/stage/vectorSearch.yaml
```diff
@@ -0,0 +1,129 @@
...
+        description: |
+            This is required if numCandidates is omitted.
+            - false to run ANN search
+            - true to run ENN search
+    -
+        name: filter
+        optional: true
+        type:
+            - query
+        description: |
```
jmikola, line 46: Should these descriptions be empty? Likewise for `numCandidates` below.

generator/config/stage/search.yaml
GromNaN: The `default` option is only when we want a value to be set by default. This is only when a value is required, otherwise we rely on the default value server-side.

generator/config/stage/searchMeta.yaml
GromNaN: See https://github.com/mongodb/mongo-php-library/pull/1537#discussion_r1868405362

jmikola (approved):

PR #1540, opened 2024-11-27
Remove specific encoder for `Group` and `DollarGroup`, use property name map

GromNaN:
For every Operator class, we add a `PROPERTIES` constant that contains a map of PHP property name to the operator name.
This allows to remove 2 specific encoders:
- `$group` stage is now `[ '_id' => '_id', 'field' => null ]`. This means every property for the `field` object are encoded as properties of the operator object.
- `$near` operator is now `['geometry' => null, 'maxDistance' => '$maxDistance', 'minDistance' => '$minDistance']`. The property `$geometry` of the `geometry` object is set to the operator object properties. And the name of the other properties get the `$` prefix.

### DIFF
--- a/generator/README.md
+++ b/generator/README.md
@@ -19,17 +19,18 @@ The `generator/config/*.yaml` files contains the list of operators and stages th
 
 ### Arguments
 
-| Field | Type | Description |
-| `name` | `string` | The name of the argument. It can start with `$` when the aggregation operator needs it, but it will be trimmed from the class property name. |
-| `type` | list of `string` | The list of accepted types |
-| `description` | `string` | The description of the argument from MongoDB's documentation. |
-| `optional` | `boolean` | Whether the argument is optional or not. |
-| `valueMin` | `number` | The minimum value for a numeric argument. |
-| `valueMax` | `number` | The maximum value for a numeric argument. |
-| `variadic` | `string` | If sent, the argument is variadic. Defines the format `array` for a list or `object` for a map |
-| `variadicMin` | `integer` | The minimum number of arguments for a variadic parameter. |
-| `default` | `scalar` or `array` | The default value for the argument. |
-| `mergeObject` | `bool` | Default `false`. If `true`, the value must be an object and the properties of the value object are merged into the parent operator. `$group` stage uses it for the fields. |
+| Field         | Type                | Description                                                                                                                                                               |
+|---------------|---------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `name`        | `string`            | The name of the argument. If it starts with `$`, the dollar is trimmed from the class property name                                                                       |
+| `type`        | list of `string`    | The list of accepted types                                                                                                                                                |
+| `description` | `string`            | The description of the argument from MongoDB's documentation                                                                                                              |
+| `optional`    | `boolean`           | Whether the argument is optional or not                                                                                                                                   |
+| `valueMin`    | `number`            | The minimum value for a numeric argument                                                                                                                                  |
+| `valueMax`    | `number`            | The maximum value for a numeric argument                                                                                                                                  |
+| `variadic`    | `string`            | If sent, the argument is variadic. Defines the format `array` for a list or `object` for a map                                                                            |
+| `variadicMin` | `integer`           | The minimum number of arguments for a variadic parameter                                                                                                                  |
+| `default`     | `scalar` or `array` | The default value for the argument                                                                                                                                        |
+| `mergeObject` | `bool`              | Default `false`. If `true`, the value must be an object and the properties of the value object are merged into the parent operator. `$group` stage uses it for the fields |
 
 ### Test pipelines
 
--- a/generator/README.md
+++ b/generator/README.md
@@ -20,7 +20,7 @@ The `generator/config/*.yaml` files contains the list of operators and stages th
 ### Arguments
 
 | Field | Type | Description |
-| `name` | `string` | The name of the argument. It can start with `$`. |
+| `name` | `string` | The name of the argument. It can start with `$` when the aggregation operator needs it, but it will be trimmed from the class property name. |
 | `type` | list of `string` | The list of accepted types |
 | `description` | `string` | The description of the argument from MongoDB's documentation. |
 | `optional` | `boolean` | Whether the argument is optional or not. |
@@ -29,7 +29,7 @@ The `generator/config/*.yaml` files contains the list of operators and stages th
 | `variadic` | `string` | If sent, the argument is variadic. Defines the format `array` for a list or `object` for a map |
 | `variadicMin` | `integer` | The minimum number of arguments for a variadic parameter. |
 | `default` | `scalar` or `array` | The default value for the argument. |
-| `noName` | `bool` | Default `false`. If `true`, the value must be an object and the properties of the value object are merged into the parent operator. `$group` stage uses it for the fields. |
+| `mergeObject` | `bool` | Default `false`. If `true`, the value must be an object and the properties of the value object are merged into the parent operator. `$group` stage uses it for the fields. |
 
 ### Test pipelines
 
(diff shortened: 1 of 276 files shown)


generator/config/query/geoIntersects.yaml
```diff
@@ -3,12 +3,13 @@ name: $geoIntersects
...
 type:
     - fieldQuery
-encode: single
+encode: object
 description: |
     Selects geometries that intersect with a GeoJSON geometry. The 2dsphere index supports $geoIntersects.
 arguments:
     -
         name: geometry
+        noName: true
```
GromNaN: This change was not required. But this is for consistency with other geo operators.

src/Builder/Encoder/OperatorEncoder.php
```diff
@@ -55,7 +46,8 @@ private function encodeAsArray(OperatorInterface $value): stdClass
     {
         $result = [];
         /** @var mixed $val */
-        foreach (get_object_vars($value) as $val) {
+        foreach ($value::PROPERTIES as $prop => $name) {
```
GromNaN, line 48: Loops over existing array instead of converting object to array with `get_object_vars`. Must be more performant.

generator/README.md
```diff
@@ -10,17 +10,30 @@ Updating the generated code can be done only by modifying the code generator, or
...
+| `name` | `string` | The name of the argument. It can start with `$`. |
+| `type` | list of `string` | The list of accepted types |
+| `description` | `string` | The description of the argument from MongoDB's documentation. |
+| `optional` | `boolean` | Whether the argument is optional or not. |
+| `valueMin` | `number` | The minimum value for a numeric argument. |
+| `valueMax` | `number` | The maximum value for a numeric argument. |
+| `variadic` | `string` | If sent, the argument is variadic. Defines the format `array` for a list or `object` for a map |
+| `variadicMin` | `integer` | The minimum number of arguments for a variadic parameter. |
+| `default` | `scalar` or `array` | The default value for the argument. |
+| `noName` | `bool` | Default `false`. If `true`, the value must be an object and the properties of the value object are merged into the parent operator. `$group` stage uses it for the fields. |
```
alcaeus: I'll let @jmikola add his thoughts as well, but "pulling to the top" is usually called "hoisting". An example of this would be [JavaScript Hoisting](https://www.w3schools.com/js/js_hoisting.asp), where variable declarations are hoisted to the top of the block they appear in. Both `noName` and `hoist` would need explanations here, but I feel like `noName` is more ambiguous than `hoist`.

A different alternative is `mergeToTopLevel`, referring to the "merge" wording you used in the field description in the JSON schema.

generator/README.md
```diff
@@ -10,17 +10,30 @@ Updating the generated code can be done only by modifying the code generator, or
...
+3. Run the generator: `./generate`
 
 ## Configuration
 
 The `generator/config/*.yaml` files contains the list of operators and stages that are supported by the library.
 
+### Arguments
+
+| Field | Type | Description |
+| `name` | `string` | The name of the argument. It can start with `$`. |
```
alcaeus: I suppose a leading `$` is used when the parameter in the aggregation operator needs it, but it is not part of the name used in the generated factory? In that case, I would recommend specifying the last part, as it is important to know for people re-using these files.

generator/src/OperatorClassGenerator.php
```diff
@@ -61,83 +62,88 @@ public function createClass(GeneratorDefinition $definition, OperatorDefinition
         $namespace->addUse(Encode::class);
         $class->addConstant('ENCODE', new Literal('Encode::' . $operator->encode->name));
 
-        $constuctor = $class->addMethod('__construct');
+        $encodeNames = [];
+        $constructor = $class->addMethod('__construct');
         foreach ($operator->arguments as $argument) {
+            // Remove the leading $ from the argument name
+            $argName = ltrim($argument->name, '$');
+            $encodeNames[$argName] = $argument->noName ? null : $argument->name;
```
alcaeus: How about adding methods to the argument class for these:
* `name` would return the name as-is
* `getParameterName` would return the name as used in factories, properties, etc.
* `getEncodedName` would return the name, or `null` if `noName` has been set.

This would avoid duplicating the logic in various generator classes.

generator/src/OperatorClassGenerator.php
```diff
@@ -157,17 +163,21 @@ public function createClass(GeneratorDefinition $definition, OperatorDefinition
...
                 }
             }
 
             // Set property from constructor argument
-            $constuctor->addBody('$this->' . $argument->name . ' = $' . $argument->name . ';');
+            $constructor->addBody('$this->' . $argName . ' = $' . $argName . ';');
+        }
+
+        if ($encodeNames !== []) {
+            $class->addConstant('PROPERTIES', $encodeNames);
```
alcaeus, line 177: Adding a `@var array<string, string|null>` docblock here (in addition to the one added to `OperatorInterface` would resolve some psalm warnings we're getting later.

src/Builder/Type/OperatorInterface.php
```diff
@@ -9,8 +9,11 @@
  */
 interface OperatorInterface
 {
-    /** To be overridden by implementing classes */
+    /** @var Encode */
     public const ENCODE = Encode::Undefined;
```
alcaeus, line 13: I was wondering why we didn't use a typed class constant, until I saw that they were only [added in PHP 8.3](https://wiki.php.net/rfc/typed_class_constants).

alcaeus (commented):
Nice way of cleaning up some more confusing encoding types, I like it. I have a few smaller suggestions around wording and logic encapsulation.

generator/src/OperatorClassGenerator.php
GromNaN: It's added to `OperatorInterface::PROPERTIES`, that solved the issues.

generator/README.md
GromNaN: `hoist` is too unfamiliar to me. I'll change to `mergeObject`.

generator/README.md
GromNaN: Added

generator/src/OperatorClassGenerator.php
GromNaN: Added `$propName` property.

generator/README.md
jmikola: I'm fine with `mergeObject` or `mergeUp`. I read the "JavaScript Hoisting" article, putting aside my historic disdain for [W3Fools](https://www.w3fools.com/), but it didn't seem entirely relevant as it was talking about variable declarations and scopes.

generator/README.md
```diff
@@ -10,17 +10,30 @@ Updating the generated code can be done only by modifying the code generator, or
...
 ## Configuration
 
 The `generator/config/*.yaml` files contains the list of operators and stages that are supported by the library.
 
+### Arguments
+
+| Field | Type | Description |
+| `name` | `string` | The name of the argument. It can start with `$` when the aggregation operator needs it, but it will be trimmed from the class property name. |
+| `type` | list of `string` | The list of accepted types |
+| `description` | `string` | The description of the argument from MongoDB's documentation. |
```
jmikola: For single-phrase descriptions, I think you can omit the trailing period. Periods make sense when there are multiple sentences in a description.

I'll refrain from commenting on the other lines here, but there are some inconsistencies.

Also, this is a very minor nit.

src/Builder/Encoder/OperatorEncoder.php
```diff
@@ -68,60 +59,29 @@ private function encodeAsArray(OperatorInterface $value): stdClass
...
+            $val = $value->$prop;
+
             // Skip optional arguments. If they have a default value, it is resolved by the server.
             if ($val === Optional::Undefined) {
                 continue;
             }
 
-            $result->{$key} = $this->recursiveEncode($val);
+            if ($name === null) {
```
jmikola, line 79: Assuming this is the place where `mergeObject: true` is handled, a comment here would be helpful.

jmikola (approved):
Two small comments, but defer to you to address.

src/Builder/Encoder/OperatorEncoder.php
GromNaN: 👍🏻 Comment added.

generator/README.md
GromNaN: Updated.

PR #1541, opened 2024-11-28
Replace `OperatorInterface::getOperator()` with `OperatorInterface::NAME` constant

GromNaN:
- Remove the `Encode::FlatObject` specific encoding by setting `null` for the operator name.
- Replace `OperatorInterface::getOperator()` with `OperatorInterface::NAME` constant. This doesn't need to be dynamic.

### DIFF
--- a/generator/config/expression/case.yaml
+++ b/generator/config/expression/case.yaml
@@ -3,7 +3,8 @@ name: $case
 link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/switch/'
 type:
     - switchBranch
-encode: flat_object
+encode: object
+wrapObject: false
 description: |
     Represents a single case in a $switch expression
 arguments:
--- a/src/Builder/Encoder/OperatorEncoder.php
+++ b/src/Builder/Encoder/OperatorEncoder.php
@@ -37,7 +37,7 @@ final class OperatorEncoder extends AbstractExpressionEncoder
         return match ($value::ENCODE) {
             Encode::Single => $this->encodeAsSingle($value),
             Encode::Array => $this->encodeAsArray($value),
-            Encode::Object, Encode::FlatObject => $this->encodeAsObject($value),
+            Encode::Object => $this->encodeAsObject($value),
             default => throw new LogicException(sprintf('Class "%s" does not have a valid ENCODE constant.', $value::class)),
         };
     }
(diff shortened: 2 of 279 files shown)


generator/config/expression/case.yaml
```diff
@@ -3,7 +3,8 @@ name: $case
 link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/switch/'
 type:
     - switchBranch
-encode: flat_object
+encode: object
+noName: true
```
jmikola: Per #1540, should this be `mergeObject`?

GromNaN: This is not the same property. Here, we don't wrap the properties into an object with the operator name as single property. `{ $operator: { ... properties }}` vs `{ ... properties }`. Maybe `wrapObject: false`

src/Builder/Encoder/OperatorEncoder.php
```diff
@@ -108,10 +106,19 @@ private function encodeAsSingle(OperatorInterface $value): stdClass
         throw new LogicException(sprintf('Class "%s" does not have a single property.', $value::class));
     }
 
+    /**
+     * Wrap the result in an object if the operator has a name.
+     * The operator name is NULL, and the result is returned as is
+     * when wrapObject is false in the YAML configuration of the operator.
+     */
     private function wrap(OperatorInterface $value, mixed $result): stdClass
```
jmikola, line 115: The current `mixed $result` type hint conflicts with the `stdClass` return type. I assume this method is already used elsewhere, so it's not feasible to simply change the parameter type.

Would it be easier to just inline the necessary logic in `encodeAsObject()` instead of modifying it use this existing method?

GromNaN: Done, I also added a validation in `OperatorDefinition` to validate the `wrapObject` is set to false only on `object` encoding.

src/Builder/Encoder/OperatorEncoder.php
```diff
@@ -108,10 +112,17 @@ private function encodeAsSingle(OperatorInterface $value): stdClass
         throw new LogicException(sprintf('Class "%s" does not have a single property.', $value::class));
     }
 
+    /**
+     * Wrap the result in an object if the operator has a name.
+     * The operator name is NULL, and the result is returned as is
+     * when wrapObject is false in the YAML configuration of the operator.
```
jmikola: I think this comment no longer applies, as you now assert that `$value::NAME` is a string.

jmikola (approved):
Code LGTM, but please correct the comment on `private function wrap()` before merging.

PR #1542, opened 2024-11-28
Mark builder classes as final and internal

GromNaN:
Most of the classes don't need to be used outside of the factory methods. Mark them as internal and final. This can be changed later if necessary.
This allows us to make changes as needed to this classes.


jmikola (commented):
Is this technically a BC break since we've already released a 1.x version with these classes? I wouldn't object to moving forward with this anyway, but just wanted to confirm.

GromNaN:
This code have not been released yet in the library. Only as pre-version in `mongodb/builder`.

jmikola (approved):
In the future, it may be best to isolate any changes to generated files in a separate commit. GitHub collapses the contents of those files by default, but they still pollute the PR view.

PR #1543, opened 2024-11-28
Remove unused Projection operator

GromNaN:
The class `MongoDB\Builder\Projection\ElemMatchOperator` was generated by an early version of the lib. But this category of operators was removed, replaced by `MongoDB\Builder\Query\ElemMatchOperator`.


jmikola (approved):
Was this never documented?

Also, since some of the removed files were auto-generated, is there anything in place to detect that generated files are obsolete? In this case, I noted a configuration file was never removed, so presumably there's no longer anything in place that would produce these files.

GromNaN:
There is nothing to detect generated files that should be removed. I can hack something that deletes all files that contains "THIS FILE IS AUTO-GENERATED. ANY CHANGES WILL BE LOST!" before running the `generate` command. 
The projection operators were removed by https://github.com/mongodb/mongo-php-builder/pull/54.

PR #1551, opened 2024-12-04
PHPLIB-1600 Accept `stdClass` objects in `BSONDocument` constructor

### LINKED ISSUE #1549: BSONDocument doesn't accept an object in its constructor
It should accept it, to be consistent with its parent

GromNaN:
Fixes PHPLIB-1600
Fixes https://github.com/mongodb/mongo-php-library/issues/1549

Only `stdClass` objects are allowed, because other types of objects can have private/protected properties and we don't what to define the rule for this properties (keep or skip).

### DIFF
--- a/src/Model/BSONDocument.php
+++ b/src/Model/BSONDocument.php
@@ -52,10 +52,10 @@ class BSONDocument extends ArrayObject implements JsonSerializable, Serializable
      * by default.
      *
      * @see https://php.net/arrayobject.construct
-     * @param object|array<string, mixed> $input
+     * @param array<string, mixed>|stdClass $input
      * @psalm-param class-string<ArrayIterator<string,mixed>>|class-string<ArrayObject<string,mixed>> $iteratorClass
      */
-    public function __construct(array|object $input = [], int $flags = ArrayObject::ARRAY_AS_PROPS, string $iteratorClass = ArrayIterator::class)
+    public function __construct(array|stdClass $input = [], int $flags = ArrayObject::ARRAY_AS_PROPS, string $iteratorClass = ArrayIterator::class)
     {
         parent::__construct($input, $flags, $iteratorClass);
     }
--- a/src/Model/BSONDocument.php
+++ b/src/Model/BSONDocument.php
@@ -52,10 +52,10 @@ class BSONDocument extends ArrayObject implements JsonSerializable, Serializable
      * by default.
      *
      * @see https://php.net/arrayobject.construct
-     * @param array<string, mixed> $input
+     * @param object|array<string, mixed> $input
      * @psalm-param class-string<ArrayIterator<string,mixed>>|class-string<ArrayObject<string,mixed>> $iteratorClass
      */
-    public function __construct(array $input = [], int $flags = ArrayObject::ARRAY_AS_PROPS, string $iteratorClass = ArrayIterator::class)
+    public function __construct(array|object $input = [], int $flags = ArrayObject::ARRAY_AS_PROPS, string $iteratorClass = ArrayIterator::class)
     {
         parent::__construct($input, $flags, $iteratorClass);
     }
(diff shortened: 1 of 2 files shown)


src/Model/BSONDocument.php
```diff
@@ -52,10 +52,10 @@ public function __clone()
      * by default.
      *
      * @see https://php.net/arrayobject.construct
-     * @param array<string, mixed> $input
+     * @param object|array<string, mixed> $input
      * @psalm-param class-string<ArrayIterator<string,mixed>>|class-string<ArrayObject<string,mixed>> $iteratorClass
      */
-    public function __construct(array $input = [], int $flags = ArrayObject::ARRAY_AS_PROPS, string $iteratorClass = ArrayIterator::class)
+    public function __construct(array|object $input = [], int $flags = ArrayObject::ARRAY_AS_PROPS, string $iteratorClass = ArrayIterator::class)
```
GromNaN: I'll have to dig this error.
```
ERROR: PossiblyInvalidArgument
at /Users/jerome/Develop/mongo-php-library/src/Model/BSONDocument.php:60:29
Argument 1 of ArrayObject::__construct expects (array<TKey:ArrayObject as array-key, TValue:ArrayObject as mixed>)|null|object, but possibly different type array<string, mixed>|object provided (see https://psalm.dev/092)
        parent::__construct($input, $flags, $iteratorClass);

```

tests/Model/BSONDocumentTest.php
```diff
@@ -23,6 +23,22 @@ public function testConstructorDefaultsToPropertyAccess(): void
...
+    }
+
+    public function testConstructorWithClassObject(): void
+    {
+        $document = new BSONDocument(new class () {
+            public string $foo = 'bar';
+            protected string $baz = 'qux';
+        });
+        $this->assertSame('bar', $document->foo);
+        $this->assertObjectNotHasProperty('baz', $document);
```
jmikola: This is essentially a test of ArrayObject itself. I think it'd be preferable to test `bsonSerialize()` (note: `jsonSerialize()` has the same implementation), so we can see what would actually get serialized to BSON.

My concern is that we should preserve the same encoding behavior as the extension, which will only encodes public properties.

See: https://3v4l.org/47VCv#v8.4.1

I think it's bit strange that the encoded names for protected and private properties still end up in output from `ArrayObject::getArrayCopy()`, and are preserved when casting back to an object. It's certainly odd that the stdClass instance resulting from that cast ends up with non-public properties, but that's likely an oddity of PHP itself.

If we can get away with only accepting `stdClass|array`, that might circumvent this issue.

src/Model/BSONDocument.php
jmikola: This may be related: https://github.com/vimeo/psalm/issues/9082

GromNaN: Thanks for checking. So I think we can ignore this error.

GromNaN: This psalm issue was solved by moving to `array|stdClass`.

tests/Model/BSONDocumentTest.php
GromNaN: Thanks, I followed your suggestion on allowing only `array|stdClass`. That covers the [request](https://github.com/mongodb/mongo-php-library/issues/1549#issuecomment-2517509705).

tests/Model/BSONDocumentTest.php
```diff
@@ -23,6 +23,15 @@ public function testConstructorDefaultsToPropertyAccess(): void
         $this->assertSame('bar', $document->foo);
     }
 
+    public function testConstructorWithStandardObject(): void
+    {
+        $object = (object) ['foo' => 'bar'];
+        $document = new BSONDocument($object);
+        $this->assertEquals(ArrayObject::ARRAY_AS_PROPS, $document->getFlags());
+        $this->assertSame('bar', $document->foo);
```
jmikola: I think these two lines can be omitted, as they seem redundant alongside `testConstructorDefaultsToPropertyAccess()` above. All you really need to test is that a stdClass class can yield the expected `bsonSerialize()` output.

jmikola (approved):
LGTM although I think you can remove two lines from the test case.

PR #1558, opened 2025-01-03
Fix Aggregation builder for `$search` stage, option `count` is a document

GromNaN:
https://www.mongodb.com/docs/atlas/atlas-search/counting/#std-label-count-ref

Also adding tests for the properties of `$search` stage.

### DIFF
--- a/generator/config/stage/search.yaml
+++ b/generator/config/stage/search.yaml
@@ -45,7 +45,7 @@ arguments:
         name: count
         optional: true
         type:
-            - string
+            - object
         description: |
             Document that specifies the count options for retrieving a count of the results.
     -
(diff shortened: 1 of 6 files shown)


generator/config/stage/search.yaml
```diff
@@ -45,7 +45,7 @@ arguments:
         name: count
         optional: true
         type:
-            - string
+            - object
```
GromNaN, line 48: Fixed bug.

jmikola: Should there be a JIRA ticket for this fix?

generator/config/stage/search.yaml
```diff
@@ -116,3 +116,150 @@ tests:
...
+                    text:
+                        path: 'title'
+                        query: 'war'
+                    sort:
+                        score:
+                            $meta: 'searchScore'
+                        released: 1
+                    searchBefore: 'CJ6kARoGELqvp4MIGgkpACDA3U8BAAA='
+    -
+        name: 'Count results'
```
jmikola, line 206: I assume this "Count results" example is what revealed the bug, but why are so many other examples added by this PR? Was `search.yaml` just very outdated?

jmikola (approved):
Fix looks fine but two questions.

generator/config/stage/search.yaml
GromNaN: This is not released yet.

generator/config/stage/search.yaml
GromNaN: When I implemented `$search` I thought that the tests on each search operator would be enough to cover the stage. But the test cases on all the parameters accepted by the stage were missing, so I included them.

generator/config/stage/search.yaml
jmikola: Thanks for the reminder. :smile:

In the future, including the original JIRA issue for the feature would be preferable to establish context from both the git history and JIRA issue (assuming the git bot is working again). Typically, the commits without JIRA links are usually CI fixes or Markdown edits.

PR #1567, opened 2025-01-15
PHPLIB-1587 Validate hint option using `is_document`

GromNaN:
Fix PHPLIB-1587

Reject `PackedArray` for `hint` option.

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -250,7 +250,7 @@ OUTPUT;
      */
     protected static function getInvalidHintValues(): array
     {
-        return [123, 3.14, true, PackedArray::fromPHP(['hint'])];
+        return [123, 3.14, true, PackedArray::fromPHP([])];
     }
 
     /**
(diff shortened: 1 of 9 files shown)


tests/TestCase.php
```diff
@@ -250,7 +250,7 @@ protected static function getInvalidDocumentCodecValues(): array
      */
     protected static function getInvalidHintValues(): array
     {
-        return [123, 3.14, true];
+        return [123, 3.14, true, PackedArray::fromPHP(['hint'])];
```
jmikola:
```suggestion
        return [123, 3.14, true, PackedArray::fromPHP([])];
```

Assuming this passes all tests just the same, I don't think a value matters.

jmikola (approved):
Defer to you about using a simpler value in `getInvalidHintValues()`.

tests/TestCase.php
GromNaN: Indeed, it can be confusing to put a value.

PR #1569, opened 2025-01-23
PHPLIB-1609 Don't mandate the use of "distanceField" in `$geoNear`

GromNaN:
Fix PHPLIB-1609 see DRIVERS-3071

The `$geoNear` distanceField argument is now optional for queries on non-timeseries collections.


src/Builder/Stage/FactoryTrait.php
```diff
@@ -265,8 +265,8 @@ public static function fill(
      * Default: false.
      */
     public static function geoNear(
-        string $distanceField,
```
GromNaN, line 268: The parameter was put first because required parameters must be before the optional ones. Changing the order of the parameters would be a breaking change one the library is released.

jmikola: I worry that the server being able to make such changes at will (where we might require a BC break to maintain consistency) is going to inevitably become a problem down the line.

Had you not caught this before a release, how could [PHPLIB-1609](https://jira.mongodb.org/browse/PHPLIB-1609) have been implemented?

jmikola (approved):
Change looks fine but I'm concerned about how we might get screwed over by a similar change down the line.

PR #1570, opened 2025-01-23
PHPLIB-1598 Use named types for enums

GromNaN:
Fix PHPLIB-1598

When a field accepts a specific list of values, I use a named type. It will be possible to improve the parameter types later by specifying the list of accepted values.

### DIFF
--- a/generator/config/expressions.php
+++ b/generator/config/expressions.php
@@ -131,45 +131,36 @@ return $expressions + [
     ],
 
     // @todo add enum values
-    'Granularity' => [
+    'granularity' => [
         'acceptedTypes' => [...$bsonTypes['string']],
     ],
-    'FullDocument' => [
+    'fullDocument' => [
         'acceptedTypes' => [...$bsonTypes['string']],
     ],
-    'FullDocumentBeforeChange' => [
+    'fullDocumentBeforeChange' => [
         'acceptedTypes' => [...$bsonTypes['string']],
     ],
-    'AccumulatorPercentile' => [
+    'accumulatorPercentile' => [
         'acceptedTypes' => [...$bsonTypes['string']],
     ],
-    'WhenMatched' => [
+    'whenMatched' => [
         'acceptedTypes' => [...$bsonTypes['string']],
     ],
-    'WhenNotMatched' => [
+    'whenNotMatched' => [
         'acceptedTypes' => [...$bsonTypes['string']],
     ],
 
     // @todo create specific model classes factories
-    'OutCollection' => [
+    'outCollection' => [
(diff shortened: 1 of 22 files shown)


src/Builder/Stage/BucketAutoStage.php
```diff
@@ -49,24 +49,24 @@ final class BucketAutoStage implements StageInterface, OperatorInterface
...
-     * @param Optional|Document|Serializable|array|stdClass $granularity A string that specifies the preferred number series to use to ensure that the calculated boundary edges end on preferred round numbers or their powers of 10.
+     * @param Optional|string $granularity A string that specifies the preferred number series to use to ensure that the calculated boundary edges end on preferred round numbers or their powers of 10.
      * Available only if the all groupBy values are numeric and none of them are NaN.
      */
     public function __construct(
         Type|ExpressionInterface|stdClass|array|bool|float|int|null|string $groupBy,
         int $buckets,
         Optional|Document|Serializable|stdClass|array $output = Optional::Undefined,
-        Optional|Document|Serializable|stdClass|array $granularity = Optional::Undefined,
+        Optional|string $granularity = Optional::Undefined,
```
GromNaN, line 69: The type was incorrect. `granularity` must be a string: https://www.mongodb.com/docs/manual/reference/operator/aggregation/bucketAuto/

generator/config/expressions.php
```diff
@@ -131,45 +131,36 @@
...
-    'FillOut' => [
-        'acceptedTypes' => [...$bsonTypes['object']],
-    ],
-    'SortSpec' => [
-        'acceptedTypes' => [...$bsonTypes['object']],
-    ],
-    'Window' => [
-        'acceptedTypes' => [...$bsonTypes['object']],
-    ],
-    'GeoPoint' => [
```
jmikola, line 172: Just want to confirm that "FillOut", "Window", and "GeoPoint" were never used.

jmikola (approved):
IIRC, in a previous Atlas Search PR you opted _not_ to use enums since that might become a hindrance if something changed upstream before this library was able to apply the change.

In the event the server _removed_ a supported string value, might that lead of a situation in the library where we'd be stuck with it out of BC obligations?

Certainly not for this implementation, since everything is just a string; however, I'm concerned with what happens if you do start modeling the server API more closely. I'm very sensitive to how that might backfire on us (similar to https://github.com/mongodb/mongo-php-library/pull/1569).

generator/config/expressions.php
GromNaN: If they were used, the generator would fail.

PR #1571, opened 2025-01-23
PHPLIB-1597 Accept a $-prefixed string anywhere an expression is accepted

GromNaN:
Fix PHPLIB-1597

The string representation of a field path is simpler than using the `FieldPath` factory.

Before:
```php
            Stage::group(
                _id: Expression::fieldPath('sizes'),
                averagePrice: Accumulator::avg(
                    Expression::numberFieldPath('price'),
                ),
            ),
```

After:
```php
            Stage::group(
                _id: '$sizes'
                averagePrice: Accumulator::avg('$price'),
            ),
```

### DIFF
--- a/generator/config/expressions.php
+++ b/generator/config/expressions.php
@@ -13,6 +13,7 @@ use MongoDB\CodeGenerator\Definition\PhpObject;
 use MongoDB\Model\BSONArray;
 use stdClass;
 
+use function in_array;
 use function ucfirst;
 
 $bsonTypes = [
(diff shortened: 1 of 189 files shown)


generator/config/expressions.php
```diff
@@ -45,6 +46,11 @@
 $expressions = [];
 $resolvesToInterfaces = [];
 foreach ($bsonTypes as $name => $acceptedTypes) {
+    // an expression can be a string with a $-prefixed field name
+    if (! in_array('string', $acceptedTypes)) {
+        $acceptedTypes[] = 'string';
```
jmikola: Assuming this is the only code change in this PR, what about this change allows $-prefixed strings specifically? It looks like you're just accepting _any_ kind of string whenever an expression is expected.

Side note: I couldn't confirm if this was really the only code change in the PR because GitHub's PR view is not correctly labeling all generated files. In the future it'd be preferable to segregate changes to generated files into their own commit.

GromNaN: I added validation of `$string`, but that's something the server would do.

> Failed to optimize pipeline :: caused by :: $abs only supports numeric types, not string

jmikola (approved):

GromNaN:
I had to change the order of conditions to please psalm. https://psalm.dev/r/4e4cd7b846

PR #1578, opened 2025-01-28
Lock drivers-evergreen-tools version

GromNaN:
Lock evergreen tools version, as the commit https://github.com/mongodb-labs/drivers-evergreen-tools/commit/50d13abef3759ad76bd594cb67fc8cc01e6648bb broke the build.

This can be reverted once fixed by https://github.com/mongodb/mongo-php-library/pull/1577

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -67,7 +67,7 @@ jobs:
           python-version: '3.13'
 
       - id: setup-mongodb
-        uses: mongodb-labs/drivers-evergreen-tools@master
+        uses: mongodb-labs/drivers-evergreen-tools@6a0686e0194d30b0a259404839c3e8bf9be12a5b
         with:
           version: ${{ matrix.mongodb-version }}
           topology: ${{ matrix.topology }}
--- a/.evergreen/config/functions.yml
+++ b/.evergreen/config/functions.yml
@@ -119,6 +119,10 @@ functions:
             cp -R ${PROJECT_DIRECTORY}/ $DRIVERS_TOOLS
           else
             git clone https://github.com/mongodb-labs/drivers-evergreen-tools.git --depth 1 $DRIVERS_TOOLS
+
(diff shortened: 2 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -67,7 +67,7 @@ jobs:
           python-version: '3.13'
 
       - id: setup-mongodb
-        uses: mongodb-labs/drivers-evergreen-tools@master
+        uses: mongodb-labs/drivers-evergreen-tools@6a0686e0194d30b0a259404839c3e8bf9be12a5b
```
jmikola, line 70: Does this actually work? My understanding from @alcaeus' comments in the meeting this morning was that we couldn't use a commit hash here and would need to resort to a submodule.

GromNaN: The only issue it that we cannot use dependabot to track new versions.
> For each action in the file, Dependabot checks the action’s reference (typically a version number or commit identifier associated with the action)
https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot#about-dependabot-version-updates-for-actions

I tried [previously](https://github.com/mongodb/mongo-php-library/network/updates/944321909). 
image

alcaeus (approved):
LGTM until we have a better fix.

alcaeus:
@GromNaN can you make the same change in PHPC please?

GromNaN:
The fix is ready: https://github.com/mongodb-labs/drivers-evergreen-tools/pull/584

jmikola:
> The fix is ready: https://github.com/mongodb-labs/drivers-evergreen-tools/pull/584

@GromNaN: Does this mean that this PR is no longer necessary? I noticed that my PHPC PRs started passing after the above fix was merged to drivers-evergreen-tools.

GromNaN:
Yes, this PR is no longer necessary.

PR #1580, opened 2025-02-07
PHPLIB-1617 Accept a Pipeline instance in aggregate and watch methods

GromNaN:
Fix PHPLIB-1617

In v1.x, we could not change the signature of some methods to accept a `Pipeline` instance, so in https://github.com/mongodb/mongo-php-library/pull/1383 I added a workaround.

But for v2.0.0, we allow ourselves this breaking change to simplify the API.

### DIFF
--- a/tests/Collection/BuilderCollectionFunctionalTest.php
+++ b/tests/Collection/BuilderCollectionFunctionalTest.php
@@ -6,6 +6,7 @@ use MongoDB\Builder\Expression;
 use MongoDB\Builder\Pipeline;
 use MongoDB\Builder\Query;
 use MongoDB\Builder\Stage;
+use PHPUnit\Framework\Attributes\TestWith;
 
 use function iterator_to_array;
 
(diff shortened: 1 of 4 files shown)


tests/Collection/BuilderCollectionFunctionalTest.php
```diff
@@ -27,8 +30,10 @@ public function testAggregate(): void
                 buckets: 2,
             ),
         );
-        // Extract the list of stages for arg type restriction
-        $pipeline = iterator_to_array($pipeline);
+
+        if ($pipelineAsArray) {
+            $pipeline = iterator_to_array($pipeline);
+        }
```
jmikola, line 36: I don't disagree that this tests the underlying behavior, but it seems a bit backwards that this operates by converting the Pipeline _back_ into an array, instead of creating a list of stages and then conditionally converting _that_ to a pipeline.

The latter approach seems more straightforward.

jmikola (approved):
I'll defer to you about using a list of stages as the default code path for tests and then only conditionally converting that into a Pipeline object (presumably with the `...` operator, just as is done in the helper methods).

tests/Collection/BuilderCollectionFunctionalTest.php
GromNaN: You're right, I hadn't thought of it like that.  https://github.com/mongodb/mongo-php-library/pull/1596

PR #1584, opened 2025-02-07
PHPLIB-1613 Load drivers-evergreen-tools using submodule

GromNaN:
Fix PHPLIB-1613

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
+[submodule "drivers-evergreen-tools"]
+	path = tests/drivers-evergreen-tools
+	url = https://github.com/mongodb-labs/drivers-evergreen-tools
 [submodule "specifications"]
 	path = tests/specifications
 	url = https://github.com/mongodb/specifications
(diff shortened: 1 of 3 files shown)


.gitmodules
```diff
@@ -1,3 +1,6 @@
+[submodule "drivers-tools"]
+	path = tests/drivers-tools
```
jmikola: I realize we use this for both Evergreen _and_ GitHub CI, but would it be preferable to keep the path consistent with the repository name?

jmikola (approved):

GromNaN:
I think there is a configuration of dependabot to do to ensure we get distinct PR for each submodule. The solution might be to create groups: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/optimizing-pr-creation-version-updates

GromNaN:
I'm merging the PR as the last dependabot PR #1587 is failing and I suspect the commit https://github.com/mongodb-labs/drivers-evergreen-tools/commit/31d5280ac32110207d4586e0007f21686bca3cd0

We will figure the best dependabot configuration in an other PR.

PR #1585, opened 2025-02-07
PHPLIB-1619: Pass ReadPreference via options array for executeCommand()

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1619

Passing a ReadPreference directly was deprecated in PHPC 1.21 (PHPC-2489) and removed in 2.0 (PHPC-2485).


GromNaN (approved):
Thank you for fixing it so quickly.

PR #1605, opened 2025-02-24
PHPLIB-1492: Support sort option for updateOne and replaceOne

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1492

Addresses both single-statement updates and bulk writes.

This will depend on https://github.com/mongodb/mongo-php-driver/pull/1794 being merged.


src/Operation/BulkWrite.php
```diff
@@ -91,6 +91,14 @@ class BulkWrite implements Executable
      *  * upsert (boolean): When true, a new document is created if no document
      *    matches the query. The default is false.
      *
+     * Supported options for replaceOne and updateOne operations:
+     *
+     *  * sort (document): Determines which document the operation modifies if
+     *    the query selects multiple documents.
+     *
+     *    This is not supported for server versions < 8.0 and will result in an
+     *    exception at execution time if used.
```
GromNaN, line 100: The meaning of "exception at execution" wasn't clear to me on first reading. I would have understood better with "runtime exception" or "server error". But we can keep that as it's consistent with the rest of the project.

src/Operation/BulkWrite.php
```diff
@@ -413,6 +425,14 @@ private function validateOperations(array $operations, ?DocumentCodec $codec, En
                         throw InvalidArgumentException::expectedDocumentType(sprintf('$operations[%d]["%s"][2]["collation"]', $i, $type), $args[2]['collation']);
                     }
 
+                    if (isset($args[2]['sort']) && ! is_document($args[2]['sort'])) {
+                        throw InvalidArgumentException::expectedDocumentType(sprintf('$operations[%d]["%s"][2]["sort"]', $i, $type), $args[2]['sort']);
+                    }
+
+                    if (isset($args[2]['sort']) && $args[2]['multi']) {
+                        throw new InvalidArgumentException(sprintf('"sort" option cannot be used with $operations[%d]["%s"]', $i, $type));
+                    }
```
GromNaN, line 434: Noted that the `multi` option cannot `true` for `replaceOne`, that why the validation is only for `updateMany` and `updateOne`.

GromNaN (approved):
LGTM

alcaeus (approved):

src/Operation/BulkWrite.php
jmikola: Aye, this language is used throughout the API documentation (both actual docs and these comments) to distinguish between exceptions raised during construction and during a call to `execute()`. I wouldn't object to changing it, but it should be done across the board if so.

GromNaN (approved):

PR #1608, opened 2025-02-25
PHPLIB-1635 BuilderEncoder accepts an instance of encoder instead of a class string

GromNaN:
Fix PHPLIB-1635

Provides better dependency injection capability.


src/Builder/Encoder/ExpressionEncoder.php
GromNaN, line 1: @alcaeus I also removed the interface and replaced the abstract class with a trait. That simplifies inheritance chain. This was useful for the constructor, but since only internal classes are instantiated by class name, I don't think we need them.

src/Builder/BuilderEncoder.php
```diff
@@ -67,7 +67,7 @@ public function canEncode(mixed $value): bool
         return (bool) $this->getEncoderFor($value)?->canEncode($value);
     }
 
-    public function encode(mixed $value): stdClass|array|string|int
+    public function encode(mixed $value): mixed
```
jmikola: Is `mixed` necessary here? I noticed it's absent on Encoder interface, which only specifies:

```
* @return mixed
* @psalm-return BSONType
```

I also noted that the original return type prior to this PR corresponds to the psalm-template implementation.

src/Builder/Encoder/PipelineEncoder.php
```diff
@@ -32,7 +34,7 @@
 
         $encoded = [];
         foreach ($value->getIterator() as $stage) {
-            $encoded[] = $this->encoder->encodeIfSupported($stage);
+            $encoded[] = $this->recursiveEncode($stage);
```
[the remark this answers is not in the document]
jmikola: Shouldn't the return type be derived from `@psalm-return` on `RecursiveEncode::recursiveEncode()`? If not, does this need to be ignored in the baseline?

jmikola (approved):
A couple of questions but I think the only change necessary may be a baseline addition for the psalm warning. No need to hold that up with a second review unless you want @alcaeus to take a look.

src/Builder/BuilderEncoder.php
GromNaN: Good point, that's what bothers me. There's no constraint on the return type of the custom encoders. So technically the `BuilderEncoder` could accept anything an `Encoder` would accept.

src/Builder/Encoder/PipelineEncoder.php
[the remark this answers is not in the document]
GromNaN: That's already ignored in the baseline, but the first time, GitHub stills shows it.

src/Builder/BuilderEncoder.php
GromNaN: I reverted to `Type|stdClass|array|string|int`, as this is expected.

PR #1610, opened 2025-02-26
PHPLIB-1634: Support encoding DateTimeInterface instances in aggregation builder

alcaeus:
PHPLIB-1634

With a new encoder, we only need to add DateTimeInterface to be accepted for the `date` type, as well as `any`. In addition, the stage factory needs manual updating, as well as some return types in the expression encoders.

### DIFF
--- a/src/Builder/Stage.php
+++ b/src/Builder/Stage.php
@@ -22,9 +22,9 @@ final class Stage
      *
      * @see https://www.mongodb.com/docs/manual/reference/operator/aggregation/match/
      *
-     * @param QueryInterface|FieldQueryInterface|Type|stdClass|array<array-key, mixed>|bool|float|int|string|DateTimeInterface|null ...$queries The query predicates to match
+     * @param DateTimeInterface|QueryInterface|FieldQueryInterface|Type|stdClass|array<array-key, mixed>|bool|float|int|string|null ...$queries The query predicates to match
      */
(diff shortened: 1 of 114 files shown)


src/Builder/Stage.php
```diff
@@ -21,9 +22,9 @@ final class Stage
      *
      * @see https://www.mongodb.com/docs/manual/reference/operator/aggregation/match/
      *
-     * @param QueryInterface|FieldQueryInterface|Type|stdClass|array<array-key,mixed>|bool|float|int|string|null ...$queries The query predicates to match
+     * @param QueryInterface|FieldQueryInterface|Type|stdClass|array<array-key, mixed>|bool|float|int|string|DateTimeInterface|null ...$queries The query predicates to match
      */
-    public static function match(QueryInterface|FieldQueryInterface|Type|stdClass|array|bool|float|int|string|null ...$queries): MatchStage
+    public static function match(QueryInterface|FieldQueryInterface|Type|stdClass|array|bool|float|int|string|DateTimeInterface|null ...$queries): MatchStage
```
alcaeus: This sent me on a short chase as I couldn't figure out why the match stage itself wasn't updated automatically. I forgot why we had to manually write this factory, but just leaving this here as a reminder for myself for the next time I have to make changes to the types accepted for `$match`.

GromNaN: nit: the types are sorted. I would keep the same order as in generated files.

```suggestion
    public static function match(DateTimeInterface|QueryInterface|FieldQueryInterface|Type|stdClass|array|bool|float|int|string|null ...$queries): MatchStage
```

GromNaN (approved):

PR #1613, opened 2025-02-26
Pre-init all sub-encoders of the builder encoder

GromNaN:
- Use `WeakReference` to prevent circular reference
- Avoid having the same encoder initialized multiple times for all the stage and operator classes

The overhead of instantiating all these objects in the constructor is offset by the fact that we don't have to check each call.

image

### DIFF
--- a/src/Builder/Encoder/RecursiveEncode.php
+++ b/src/Builder/Encoder/RecursiveEncode.php
@@ -11,6 +11,7 @@ use WeakReference;
 use function get_object_vars;
 use function is_array;
 
+/** @internal */
 trait RecursiveEncode
 {
     /** @param WeakReference<Encoder> $encoder */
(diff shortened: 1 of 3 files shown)


src/Builder/Encoder/RecursiveEncode.php
```diff
@@ -4,15 +4,18 @@
...
 
 use function get_object_vars;
 use function is_array;
 
+/** @internal */
 trait RecursiveEncode
 {
-    final public function __construct(protected readonly BuilderEncoder $encoder)
+    /** @param WeakReference<Encoder> $encoder */
+    final public function __construct(private readonly WeakReference $encoder)
```
alcaeus, line 18: Is there a way to enforce the type of the class the `WeakReference` is pointing to? Would that entail checking the type of `$encoder->get()` and manually throwing a `TypeError` if it's wrong?

I don't think we need to do this here since this is an internal trait, mainly just asking out of curiosity.

alcaeus (approved):

src/Builder/Encoder/RecursiveEncode.php
GromNaN: You would like native support for generics, that doesn't exist.
Also, the `WeakReference` class is final, we cannot extend it to restrict the type.

PR #1619, opened 2025-02-28
Remove obsolete `symfony/polyfill-php80` and `symfony/polyfill-php81`

OskarStark:
PHP 8.1 is already required


GromNaN (approved):
Thank you @OskarStark

PR #1621, opened 2025-02-28
Remove unneeded dependencies

remicollet:
- hash is always present (cannot be disabled)
- json is always present since 8.0
- polyfills no more needed as 8.1 is required

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,13 +11,9 @@
     ],
     "require": {
         "php": "^8.1",
-        "ext-hash": "*",
-        "ext-json": "*",
         "ext-mongodb": "^1.21.0",
         "composer-runtime-api": "^2.0",
-        "psr/log": "^1.1.4|^2|^3",
-        "symfony/polyfill-php80": "^1.27",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -11,13 +11,9 @@
     ],
     "require": {
         "php": "^8.1",
-        "ext-hash": "*",
-        "ext-json": "*",
         "ext-mongodb": "^1.21.0",
         "composer-runtime-api": "^2.0",
-        "psr/log": "^1.1.4|^2|^3",
-        "symfony/polyfill-php80": "^1.27",
-        "symfony/polyfill-php81": "^1.27"
```
jmikola, line 20: These were also removed in https://github.com/mongodb/mongo-php-library/pull/1619, but there doesn't appear to be any conflict here.

GromNaN: Maybe because Git is smart enough to detect that's the exact same changes.

GromNaN (approved):

GromNaN:
Thank you @remicollet

PR #1627, opened 2025-03-03
Support creating release branches when releasing new versions

alcaeus:
This PR adds branching logic to the release workflow to support creating new non-patch releases from development branches without having to manually push the release branch.

In summary:
- Non-patch releases (i.e. `A.B.0`), including pre-releases can be created from either the `vA.B` release branch or the `vA.x` development branch.
- Patch releases (i.e. `A.B.C`), including pre-releases have to be created from the `vA.B` release branch
- When creating a release from the development branch, the release branch will be created and pushed. Note that this may lead to a failure if the release branch already exists.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -57,7 +57,9 @@ jobs:
       - name: "Store version numbers in env variables"
         run: |
           echo RELEASE_VERSION=${{ inputs.version }} >> $GITHUB_ENV
+          echo RELEASE_VERSION_WITHOUT_STABILITY=$(echo ${{ inputs.version }} | awk -F- '{print $1}') >> $GITHUB_ENV
           echo RELEASE_BRANCH=v$(echo ${{ inputs.version }} | cut -d '.' -f-2) >> $GITHUB_ENV
+          echo DEVELOPMENT_BRANCH=v$(echo ${{ inputs.version }} | cut -d '.' -f-1).x >> $GITHUB_ENV
 
       - name: "Ensure release tag does not already exist"
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -57,7 +57,9 @@ jobs:
       - name: "Store version numbers in env variables"
         run: |
           echo RELEASE_VERSION=${{ inputs.version }} >> $GITHUB_ENV
+          echo RELEASE_VERSION_WITHOUT_STABILITY=$(echo ${{ inputs.version }} | awk -F- '{print $1}') >> $GITHUB_ENV
```
alcaeus, line 60: This is necessary as we need to check for a patch release. The best way is to cut the suffix (e.g. `-RC0`) from the version and only keep the major, minor, and patch numbers.

.github/workflows/release.yml
```diff
@@ -90,7 +111,7 @@ jobs:
           EOL
 
       - name: "Create draft release"
-        run: echo "RELEASE_URL=$(gh release create ${{ inputs.version }} --target ${{ github.ref_name }} --title "${{ inputs.version }}" --notes-file release-message --draft)" >> "$GITHUB_ENV"
+        run: echo "RELEASE_URL=$(gh release create ${{ inputs.version }} --target ${{ env.RELEASE_BRANCH }} --title "${{ inputs.version }}" --notes-file release-message --draft)" >> "$GITHUB_ENV"
```
alcaeus, line 114: Using the (potentially newly created) release branch ensures that GitHub shows the correct "x commits to <branch>" when looking at a release on GitHub.

.github/workflows/release.yml
```diff
@@ -66,16 +68,35 @@ jobs:
...
       # Preliminary checks done - commence the release process
       #
 
+      # Create the new release branch if we're releasing from the vA.x
+      # development branch
+      - name: "Create new release branch"
+        if: ${{ github.ref_name != env.RELEASE_BRANCH }}
+        run: |
+          git checkout -b ${{ env.RELEASE_BRANCH }}
+          git push origin ${{ env.RELEASE_BRANCH }}
```
GromNaN: For reproducibility, if the branch already exists (because the job has already been launched but failed when calling the GitHub API, for example), then the `git push` will fail. 
We can accept this and consider deleting the branch manually in this case. Or you can check whether the branch exists and is up-to-date.

GromNaN (commented):
Could you please add some tests? Joking aside, it's not easy to review without tests to make sure there aren't any special cases left out. The amount of logic embedded in Yaml is borderline, so if it grows, we'll have to think of a separate script.

.github/workflows/release.yml
alcaeus: Good point.

What we can do is that we check out all branches (similar to the merge-up workflow) and switch to that branch in certain instances. Those can be releasing a non-patch release from the development branch when the release branch already exists instead of creating the new branch. The only issue with that is that instead of using the release workflow from that branch, it runs the release workflow from whatever branch was used. This shouldn't be a big issue in most cases, but it's important to keep in mind.

alcaeus:
@GromNaN agree with the amount of logic in the workflows. If it makes you feel any better, the Java driver has been using [very similar logic](https://github.com/mongodb/mongo-java-driver/blob/d29c080f7b77b2b46e208cae4b4d7204b93623d2/.github/workflows/release.yml#L71-L87) for a while now, apparently I just forgot to also add this to the PHP driver when I created it for Java.

.github/workflows/release.yml
alcaeus: Turns out, we're already using this kind of logic in the Laravel integration: https://github.com/mongodb/laravel-mongodb/blob/e373350f436596402a2b51cd7a8fe68fdd2df848/.github/workflows/release.yml#L44-L67

With that in mind, I'd update the logic here to be the same as in the Laravel integration, and improve on that in a separate step. We can also consider extracting the logic to drivers-github-tools, which already holds some driver-specific actions.

.github/workflows/release.yml
jmikola: Noted that this is using the `-F sepstring` syntax from [awk(1p)](https://man7.org/linux/man-pages/man1/awk.1p.html). It wasn't immediately clear to me due to the omitted space, but https://askubuntu.com/a/342850 helped clarify.

.github/workflows/release.yml
```diff
@@ -66,12 +68,31 @@ jobs:
...
+      # For patch releases (A.B.C where C != 0), we expect the release to be
+      # triggered from the A.B maintenance branch
+      - name: "Fail if patch release is created from wrong release branch"
+        if: ${{ !endsWith(env.RELEASE_VERSION_WITHOUT_STABILITY, '.0') && env.RELEASE_BRANCH != github.ref_name }}
         run: |
           echo '❌ Release failed due to branch mismatch: expected ${{ inputs.version }} to be released from ${{ env.RELEASE_BRANCH }}, got ${{ github.ref_name }}' >> $GITHUB_STEP_SUMMARY
           exit 1
 
+      # For non-patch releases (A.B.C where C == 0), we expect the release to
+      # be triggered from the A.x maintenance branch or A.x development branch
```
jmikola:
```suggestion
      # be triggered from the A.B maintenance branch or A.x development branch
```

GromNaN (approved):

PR #1630, opened 2025-03-04
PHPLIB-1627: BulkWriteCommand API

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1627

Requires https://github.com/mongodb/mongo-php-driver/pull/1818

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,11 @@ on:
       - "feature/*"
 
 env:
-  DRIVER_VERSION: "stable"
+  # TODO: change to "stable" once 2.0.0 is released
+  # DRIVER_VERSION: "stable"
+  # TODO: change to "mongodb/mongo-php-driver@v2.x" once mongodb/mongo-php-driver#1790 is merged
+  # DRIVER_VERSION: "mongodb/mongo-php-driver@v2.x"
+  DRIVER_VERSION: "jmikola/mongo-php-driver@2.x-bulkwrite"
 
 jobs:
   phpunit:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,9 +14,7 @@ on:
 env:
   # TODO: change to "stable" once 2.0.0 is released
   # DRIVER_VERSION: "stable"
-  # TODO: change to "mongodb/mongo-php-driver@v2.x" once mongodb/mongo-php-driver#1790 is merged
-  # DRIVER_VERSION: "mongodb/mongo-php-driver@v2.x"
-  DRIVER_VERSION: "jmikola/mongo-php-driver@2.x-bulkwrite"
+  DRIVER_VERSION: "mongodb/mongo-php-driver@v2.x"
 
 jobs:
   phpunit:
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,8 @@
         "php": "^8.1",
         "ext-mongodb": "^2.0",
         "composer-runtime-api": "^2.0",
-        "psr/log": "^1.1.4|^2|^3"
+        "psr/log": "^1.1.4|^2|^3",
+        "symfony/polyfill-php85": "^1.32"
     },
     "require-dev": {
         "doctrine/coding-standard": "^12.0",
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
     ],
     "require": {
         "php": "^8.1",
-        "ext-mongodb": "^2.0",
+        "ext-mongodb": "^2.1",
         "composer-runtime-api": "^2.0",
         "psr/log": "^1.1.4|^2|^3",
         "symfony/polyfill-php85": "^1.32"
--- a/src/BulkWriteCommandBuilder.php
+++ b/src/BulkWriteCommandBuilder.php
@@ -67,24 +67,21 @@ final readonly class BulkWriteCommandBuilder
         );
     }
 
-    public function withCollection(Collection $collection): self
+    public function deleteMany(array|object $filter, ?array $options = null): self
     {
-        /* Prohibit mixing Collections associated with different Manager
-         * objects. This is not technically necessary, since the Collection is
-         * only used to derive a namespace and encoding options; however, it
-         * may prevent a user from inadvertently mixing writes destined for
-         * different deployments. */
-        if ($this->manager !== $collection->getManager()) {
-            throw new InvalidArgumentException('$collection is associated with a different MongoDB\Driver\Manager');
+        $filter = $this->builderEncoder->encodeIfSupported($filter);
+
+        if (isset($options['collation']) && ! is_document($options['collation'])) {
+            throw InvalidArgumentException::expectedDocumentType('"collation" option', $options['collation']);
         }
 
-        return new self(
-            $this->bulkWriteCommand,
-            $this->manager,
-            $collection->getNamespace(),
-            $collection->getBuilderEncoder(),
-            $collection->getCodec(),
-        );
+        if (isset($options['hint']) && ! is_string($options['hint']) && ! is_document($options['hint'])) {
+            throw InvalidArgumentException::expectedDocumentOrStringType('"hint" option', $options['hint']);
+        }
+
+        $this->bulkWriteCommand->deleteMany($this->namespace, $filter, $options);
+
+        return $this;
     }
 
     public function deleteOne(array|object $filter, ?array $options = null): self
@@ -100,23 +97,6 @@ final readonly class BulkWriteCommandBuilder
         }
 
         $this->bulkWriteCommand->deleteOne($this->namespace, $filter, $options);
-
-        return $this;
-    }
-
-    public function deleteMany(array|object $filter, ?array $options = null): self
-    {
-        $filter = $this->builderEncoder->encodeIfSupported($filter);
-
-        if (isset($options['collation']) && ! is_document($options['collation'])) {
-            throw InvalidArgumentException::expectedDocumentType('"collation" option', $options['collation']);
-        }
-
-        if (isset($options['hint']) && ! is_string($options['hint']) && ! is_document($options['hint'])) {
-            throw InvalidArgumentException::expectedDocumentOrStringType('"hint" option', $options['hint']);
-        }
-
-        $this->bulkWriteCommand->deleteMany($this->namespace, $filter, $options);
 
         return $this;
     }
@@ -175,6 +155,36 @@ final readonly class BulkWriteCommandBuilder
         return $this;
     }
 
+    public function updateMany(array|object $filter, array|object $update, ?array $options = null): self
+    {
+        $filter = $this->builderEncoder->encodeIfSupported($filter);
+        $update = $this->builderEncoder->encodeIfSupported($update);
+
+        if (! is_first_key_operator($update) && ! is_pipeline($update)) {
+            throw new InvalidArgumentException('Expected update operator(s) or non-empty pipeline for $update');
+        }
+
+        if (isset($options['arrayFilters']) && ! is_array($options['arrayFilters'])) {
+            throw InvalidArgumentException::invalidType('"arrayFilters" option', $options['arrayFilters'], 'array');
+        }
+
+        if (isset($options['collation']) && ! is_document($options['collation'])) {
+            throw InvalidArgumentException::expectedDocumentType('"collation" option', $options['collation']);
+        }
+
+        if (isset($options['hint']) && ! is_string($options['hint']) && ! is_document($options['hint'])) {
+            throw InvalidArgumentException::expectedDocumentOrStringType('"hint" option', $options['hint']);
+        }
+
+        if (isset($options['upsert']) && ! is_bool($options['upsert'])) {
+            throw InvalidArgumentException::invalidType('"upsert" option', $options['upsert'], 'boolean');
+        }
+
+        $this->bulkWriteCommand->updateMany($this->namespace, $filter, $update, $options);
+
+        return $this;
+    }
+
     public function updateOne(array|object $filter, array|object $update, ?array $options = null): self
     {
         $filter = $this->builderEncoder->encodeIfSupported($filter);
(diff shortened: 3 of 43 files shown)


src/BulkWriteCommandBuilder.php
```diff
@@ -0,0 +1,215 @@
...
+        );
+    }
+
+    public function withCollection(Collection $collection): self
+    {
+        $this->namespace = $collection->getNamespace();
+        $this->builderEncoder = $collection->getBuilderEncoder();
+        $this->codec = $collection->getCodec();
+
+        return $this;
```
GromNaN: Usually [wither functions](https://dev.to/fredbouchery/dx-setter-wither-and-mutants-29e6) create a new instance of the object. I would apply this pattern and return a new instance of the object.
The state is stored in the `BulkWriteCommand` that is not cloned.

src/Collection.php
```diff
@@ -753,6 +753,16 @@ public function findOneAndUpdate(array|object $filter, array|object $update, arr
         return $operation->execute(select_server_for_write($this->manager, $options));
     }
 
+    public function getBuilderEncoder(): BuilderEncoder
```
GromNaN: We can avoid this new getters by creating a `Collection::createBulkWrite` method that gives the `BulkWriteCommandBuilder` class a closure to access its options.

See https://3v4l.org/L70mG

```php
class Collection {
    public function __construct(private array $options) {}
    
    public function createBulkWrite()
    {
        $bulk = new BulkWrite(
            static function (Collection $collection) {
                return $collection->options;
            },
        );
        
        return $bulk->withCollection($this);
    }
}

class BulkWrite {
    private string $collectionName;
    
    /** @internal The closure cannot be created outside of the collection class */
    public function __construct(private Closure $getCollectionOptions) {}
    
    public function withCollection(Collection $collection)
    {
        $collectionOptions = ($this->getCollectionOptions)($collection);
        $this->collectionName = $collectionOptions['name'];
        
        return $this;
    }
}
```

src/BulkWriteCommandBuilder.php
jmikola: My initial reservation with creating multiple builder instances is that users could overlook that the underlying BulkWriteCommand is shared among them, and end up appending operations in an order they don't expect or trying to execute the objects multiple times (prohibited by libmongoc).

But I think the more important point raised in the blog post is that by _not_ doing so, and mixing mutability with a fluent API, we risk users staging operations their on an unexpected collection (arguably worse, since it'd execute successfully).

src/Collection.php
alcaeus: I don't necessarily mind the new accessors - being able to fetch the codec or builder encoder from a collection isn't negative, and I prefer this over passing a closure to the builder to break encapsulation.

src/BulkWriteCommandBuilder.php
```diff
@@ -0,0 +1,226 @@
...
+        /* Prohibit mixing Collections associated with different Manager
+         * objects. This is not technically necessary, since the Collection is
+         * only used to derive a namespace and encoding options; however, it
+         * may prevent a user from inadvertently mixing writes destined for
+         * different deployments. */
+        if ($this->manager !== $collection->getManager()) {
+            throw new InvalidArgumentException('$collection is associated with a different MongoDB\Driver\Manager');
+        }
+
+        return new self(
```
GromNaN: Do you allow extending this class? In that case, you should use `static` here and in all return types. Otherwise the class should be final.

src/Client.php
```diff
@@ -189,6 +192,32 @@ final public function addSubscriber(Subscriber $subscriber): void
...
+     * Executes multiple write operations.
+     *
+     * @see ClientBulkWrite::__construct() for supported options
+     * @param string $databaseName Database name
+     * @param array  $options      Additional options
+     * @throws UnsupportedException if options are unsupported on the selected server
+     * @throws InvalidArgumentException for parameter/option parsing errors
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function bulkWrite(BulkWriteCommand|BulkWriteCommandBuilder $bulk, array $options = []): ?BulkWriteCommandResult
```
GromNaN: When is `null` returned? I think this should be mentioned in the comment.

src/Collection.php
GromNaN: Ok for these accessors, it's true that they're useful. In an other context I need to get the codec to make an update command.

src/BulkWriteCommandBuilder.php
jmikola: Introducing it as `final` seems preferable.

src/Client.php
jmikola: Currently, `executeBulkWriteCommand()` returns `null` when an unacknowledged write concern is used. 

The alternative would be to return BulkWriteCommandResult object where `isAcknowledged()` returns `false` and no other methods can be called. That's what we do in the legacy `executeBulkWrite()` API.

I'm leaning towards a `null` object since it's consistent with the spec's instruction on _not_ populating `partialResult` on BulkWriteCommandException when zero successful writes have occurred. In the legacy API, the result object could just be empty.

jmikola: I'm now wondering if the current design of having `executeBulkWriteCommand()` return a nullable BulkWriteCommandResult is optimal. Interestingly, the [spec](https://github.com/mongodb/specifications/blob/master/source/crud/bulk-write.md) doesn't explicitly suggest that when it presents a signature for `Client::bulkWrite()`, but the section on [modeling unacknowledged results](https://github.com/mongodb/specifications/blob/master/source/crud/bulk-write.md#unacknowledged-results) ends up suggesting this when it links to the [CRUD spec's guidance](https://github.com/mongodb/specifications/blob/master/source/crud/crud.md#write-results) on modeling unacknowledged results. I was also influenced by libmongoc, which leaves the result null in these cases.

The spec is clear about BulkWriteCommandException having an optional value for the partial write result, so I was planning to use nullable BulkWriteCommandResult for its `$partialResult` property (and getter); however, I'm curious if a BulkWriteCommandResult object in an unacknowledged state would also be preferable there.

In [PHPC-2144](https://jira.mongodb.org/browse/PHPC-2144), we changed PHPC 2.0 to remove the nullable return types for fields on WriteResult when the write was unacknowledged. Instead, we just throw if getters are accessed and the result object is empty (i.e. isAcknowledged() returns false). That made PHPC consistent with the PHPLIB API. My memory is a bit fuzzy, but I remember the nullable return values being more of an annoyance for static analysis. And in the general case where users don't use `w:0`, they can probably ignore `isAcknowledged()` and just assume that exceptions won't be thrown when they access the result.

alcaeus: In general, I think it would be preferable to return something other than `null` in all cases, as long as users can deduce what happen (e.g. unacknowledged writes). Returning `null` in some cases creates one "special" condition that users need to check for instead of handling a result in all cases that didn't error.

GromNaN: The method `Collection::bulkWrite()` always returns an instance of `BulkWriteResult`, I would do the same for `Client::bulkWrite()`.

jmikola: > I think it would be preferable to return something other than null in all cases

How do you feel about the verbose result fields (e.g. `BulkWriteCommandResult::getInsertResults()`) returning a nullable Document? If we have the regular getters (e.g. `getInsertedCount()`) throw when `isAcknowledged()` is false, we can do the same with the others in relation to a `hasVerboseResults()` field. That is one approach suggested in the [bulkWrite spec](https://github.com/mongodb/specifications/blob/master/source/crud/bulk-write.md#summary-vs-verbose-results), although it's this is a level removed from the `Client::bulkWrite()` API so it's less important.

I also didn't have qualms about using a nullable BulkWriteCommandResult for `BulkWriteCommandException::getPartialResult()`, which is directly suggested in the spec.

src/Collection.php
GromNaN: To be exhaustive, we'll need unit tests for these new getters.

src/BulkWriteCommandBuilder.php
```diff
@@ -0,0 +1,226 @@
...
+use MongoDB\Codec\Encoder;
+use MongoDB\Driver\BulkWriteCommand;
+use MongoDB\Driver\Manager;
+use MongoDB\Exception\InvalidArgumentException;
+
+use function is_array;
+use function is_bool;
+use function is_string;
+
+final readonly class BulkWriteCommandBuilder
```
GromNaN: To debate the name of this class. I find it long and a bit technical. How about `BulkWriteBuilder` or even shorter `MongoDB\BulkWrite`?

"command" and "builder" are concepts that don't need to be expressed.

This is important as it's the entrypoint of this feature.

src/Client.php
jmikola: `Client::bulkWrite()` always returns a BulkWriteCommandResult now.

I've left `BulkWriteCommandException::getPartialResult()` as `?BulkWriteCommandResult`, since the result is intended to be unset. While `isAcknowledged()` could also be used in that capacity to distinguish an "empty" result, it wouldn't be semantically correct.

I've also left `?Document` for the optional verbose fields.

src/BulkWriteCommandBuilder.php
jmikola: Happy to revisit this before we settle on an API.

I agree that a Builder suffix is superfluous. I'd be open to moving this within the `MongoDB\Builder` namespace if you don't think that'd be problematic with all of the other classes there.

I think having "ClientBulkWrite" or a "Command" suffix may be relevant, though, as I'm worried about the need to distinguish this from the `Collection::bulkWrite()` API. I suppose we have more flexibility here, but I felt compelled to do so in PHPC (it doesn't sit well with me, but I didn't see an alternative).

GromNaN: `ClientBulkWrite` is a good name. It gives the scope.

src/Client.php
GromNaN: So you can update the return type of this method.
```suggestion
    public function bulkWrite(BulkWriteCommand|BulkWriteCommandBuilder $bulk, array $options = []): BulkWriteCommandResult
```

src/ClientBulkWrite.php
```diff
@@ -0,0 +1,241 @@
...
+        if (isset($options['upsert']) && ! is_bool($options['upsert'])) {
+            throw InvalidArgumentException::invalidType('"upsert" option', $options['upsert'], 'boolean');
+        }
+
+        $this->bulkWriteCommand->updateOne($this->namespace, $filter, $update, $options);
+
+        return $this;
+    }
+
+    public function withCollection(Collection $collection): self
```
GromNaN, line 230: You can add the upcoming `#[NoDiscard]` attribute on this method, as it returns a new instance. @alcaeus approved the [RFC](https://wiki.php.net/rfc/marking_return_value_as_important), so he would agree.

This would prevent people from thinking this is mutating the object.

jmikola: Should that also be done for `createWithCollection()`, since it also returns a new instance? `#[NoDiscard]` seems like it'd be appropriate for _any_ factory method.

jmikola: I assume this will also require baselining the UndefinedAttributeClass error in Psalm.

src/ClientBulkWrite.php
```diff
@@ -0,0 +1,246 @@
...
+
+        $this->bulkWriteCommand->deleteOne($this->namespace, $filter, $options);
+
+        return $this;
+    }
+
+    public function insertOne(array|object $document, mixed &$id = null): self
+    {
+        if ($this->codec) {
+            $document = $this->codec->encode($document);
```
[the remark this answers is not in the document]
GromNaN: The type in `DocumentCodec` should be relaxed to `@psalm-template ObjectType of object|array`

https://github.com/mongodb/mongo-php-library/blob/ae6a3a53d507fe38e447dfde4e77c8efe5724925/src/Codec/DocumentCodec.php#L27

src/ClientBulkWrite.php
```diff
@@ -0,0 +1,246 @@
...
+        }
+
+        // Capture the document's _id, which may have been generated, in an optional output variable
+        /** @var mixed */
```
GromNaN: This `@var` annotation should not be necessary as you added the stub for `BulkWriteCommand`

src/ClientBulkWrite.php
GromNaN: We'll add it later: this new attribute will be in the next version of PHP, in November. We can expect it in the polyfill https://github.com/symfony/polyfill/pull/522

alcaeus: Since adding `#[NoDiscard]` later could be seen as a BC break, I think it's better to add the attribute now and silence the Psalm errors. Unless people are actively trying to read this particular attribute, it will have no effect until PHP 8.5 is released. Even then, the polyfill will allow libraries (such as Psalm) to look for the attribute on earlier PHP versions and provide the same functionality.

src/ClientBulkWrite.php
GromNaN: if fact it's necessary for psalm.

src/ClientBulkWrite.php
```diff
@@ -0,0 +1,246 @@
...
+use MongoDB\Driver\BulkWriteCommand;
+use MongoDB\Driver\Manager;
+use MongoDB\Exception\InvalidArgumentException;
+use stdClass;
+
+use function is_array;
+use function is_bool;
+use function is_string;
+
+final readonly class ClientBulkWrite
```
GromNaN: `readonly` class is not supported on PHP 8.1

src/ClientBulkWrite.php
GromNaN: Attribute added.

src/ClientBulkWrite.php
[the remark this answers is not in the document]
GromNaN: Apply the same change as https://github.com/mongodb/mongo-php-library/pull/1687

GromNaN:
The spec tests need to be updated with https://github.com/mongodb/specifications/commit/63e382ee72e63ea65dfb3dd8eff7af7bb7fbb568

alcaeus:
> The spec tests need to be updated with [mongodb/specifications@63e382e](https://github.com/mongodb/specifications/commit/63e382ee72e63ea65dfb3dd8eff7af7bb7fbb568)

The previous failure was unrelated, but I'll update the spec tests once https://github.com/mongodb/mongo-php-driver/pull/1822 is merged up to v2.x.

alcaeus (approved):

src/Operation/ClientBulkWriteCommand.php
```diff
@@ -0,0 +1,95 @@
...
+     * @throws DriverRuntimeException for other driver errors (e.g. connection errors)
+     */
+    public function execute(Server $server): BulkWriteCommandResult
+    {
+        $inTransaction = isset($this->options['session']) && $this->options['session']->isInTransaction();
+        if ($inTransaction && isset($this->options['writeConcern'])) {
+            throw UnsupportedException::writeConcernNotSupportedInTransaction();
+        }
+
+        $options = array_filter($this->options, fn ($value) => isset($value));
```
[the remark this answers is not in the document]
GromNaN: I agree with copilot on this one.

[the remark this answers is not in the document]
alcaeus: Applied.

composer.json
```diff
@@ -13,7 +13,8 @@
         "php": "^8.1",
         "ext-mongodb": "^2.0",
```
GromNaN:
```suggestion
        "ext-mongodb": "^2.1",
```

.github/workflows/tests.yml
```diff
@@ -12,7 +12,9 @@ on:
       - "feature/*"
 
 env:
-  DRIVER_VERSION: "stable"
+  # TODO: change to "stable" once 2.0.0 is released
```
GromNaN, line 15:
```suggestion
  # TODO: change to "stable" once 2.1.0 is released
```

GromNaN (approved):
LGTM. Great work @jmikola

PR #1653, opened 2025-04-02
PHPLIB-1659: Fix dropping collections after successful tests

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1659

cb33c1c92b5d5cead5f18996db5825cc6064e81f introduced onNotSuccessfulTest(), but that is the inverse logic. The internal TestCase::status() method seems to be the only replacement for TestCase::hasFailed(), which was removed in PHPUnit 10.


GromNaN (approved):
Good point. 
The Throwable import must be removed to fix the CS.

PR #1659, opened 2025-04-07
draft: augment yaml definitions with test schema

nirinchev:
This is very much a POC, but I figured I'd open up a PR to get early feedback on the direction we're heading in.

For context, the devtools team wants to use the yaml definitions from the PHP driver to build an autocomplete experience for the aggregation pipelines in mongosh/compass. We've built a tool to convert the yaml to typescript definitions and it seems to be working well for the base case, but to polish it, we'd need to enhance the yaml definitions to allow us to generate more precise type definitions.

Here's a list of the high-level changes we're proposing (will be updated over time):
- Update schema.json to remove duplicate `$comment` fields (since most json parsers are not happy about it).
- Update all yaml definitions to add schema information for the test cases.
	- The schema shape matches [`SimplifiedSchema`](https://github.com/mongodb-js/mongodb-schema/blob/6f22cb72aaaf97a0bbbb6ae92a4fdadc4290db67/src/schema-analyzer.ts#L307) from the `mongodb-schema` package. This is a little bit verbose since it supports multiple types per field, even though the docs rarely have heterogeneous schemas.
	- Schema is generated automatically by crawling the docs website at the link pointed to by the test and then using the `mongodb-schema` package to infer a probabilistic schema from the example collection.
- Removed a few tests that no longer seem to point to anything in the docs - e.g. `firstN:Null and Missing Values`.


generator/config/accumulator/accumulator.yaml
```diff
@@ -1,6 +1,6 @@
 # $schema: ../schema.json
 name: $accumulator
-link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/accumulator/'
```
GromNaN, line 3: Is it possible to keep quotes around strings? This style changes make the diff hard to read. Also, I use quotes by default around every string to prevent unexpected parsing rule from Yaml.

generator/config/accumulator/percentile.yaml
```diff
@@ -25,7 +25,7 @@ arguments:
     -
         name: p
         type:
-            - resolvesToArray # of resolvesToNumber
```
GromNaN, line 28: I was keeping this comments as a note for later, as we would like to extend type definition for generic lists and hashes.

generator/config/accumulator/stdDevSamp.yaml
```diff
@@ -16,28 +16,70 @@ arguments:
...
+        name: Use in $group Stage
+        link: https://www.mongodb.com/docs/manual/reference/operator/aggregation/stdDevSamp/#use-in--group-stage
         pipeline:
             -
                 $sample:
                     size: 100
             -
                 $group:
-                    _id: ~
+                    _id: null
```
GromNaN: Noted that both `~` and `null` can be used according to the [spec](https://yaml.org/spec/1.2-old/spec.html#id2805071).

generator/config/expression/convert.yaml
```diff
@@ -35,26 +35,27 @@ arguments:
...
                         $convert:
-                            input: '$price'
-                            to: 'decimal'
-                            onError: 'Error'
-                            onNull: !bson_decimal128 '0'
+                            input: $price
+                            to: decimal
+                            onError: Error
+                            onNull:
+                                bytes: !!binary AAAAAAAAAAAAAAAAAABAMA==
```
GromNaN: This new notation seems less legible to me.

generator/config/accumulator/firstN.yaml
```diff
@@ -24,99 +24,122 @@ arguments:
             A positive integral expression that is either a constant or depends on the _id value for $group.
 tests:
     -
-        name: 'Null and Missing Values'
-        link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/firstN/#null-and-missing-values'
```
GromNaN, line 28: > Removed a few tests that no longer seem to point to anything in the docs - e.g. "firstN:Null and Missing Values".

This example is in the "Behavior" part of the docs.

https://www.mongodb.com/docs/manual/reference/operator/aggregation/firstN/#null-and-missing-values

generator/config/schema.json
```diff
@@ -52,17 +52,17 @@
...
+                        "array: parameters are encoded as an array of values in the order they are defined by the spec",
+                        "object: parameters are encoded as an object with keys matching the parameter names",
+                        "single: get the single parameter value"
+                    ],
                     "type": "string",
                     "enum": [
                         "array",
                         "object",
-                        "single",
-                        "search"
```
GromNaN, line 65: Good catch, this value is not used.

generator/config/schema.json
```diff
@@ -214,8 +253,123 @@
...
+                        "Null",
+                        "ObjectId",
+                        "BSONRegExp",
+                        "String",
+                        "BSONSymbol",
+                        "Timestamp",
+                        "Undefined",
+                        "Document",
+                        "Number"
+                    ],
```
GromNaN: We'll have to converge on the same type names for operator return types and argument types. I'm fine with changing the names in [`expressions.php`](https://github.com/mongodb/mongo-php-library/blob/ae6a3a53d507fe38e447dfde4e77c8efe5724925/generator/config/expressions.php#L19-L38) to be consistent with other tools.

generator/config/schema.json
```diff
@@ -214,8 +253,123 @@
...
+        "SimplifiedSchemaBaseType": {
+            "additionalProperties": false,
+            "properties": {
+                "bsonType": {
+                    "enum": [
+                        "Array",
+                        "Binary",
+                        "Boolean",
+                        "Code",
+                        "CodeWScope",
```
GromNaN, line 355: Is the `CodeWScope` type specific to mongosh?

GromNaN (commented):
Very good idea to have the schema of the source collection.

generator/config/accumulator/accumulator.yaml
nirinchev: Hey, looks like the quotes usage is inconsistent in these files - I'd be inclined to define a prettier rule and reformat them in a separate PR to make this one less noisy if that'd make sense to you. The reason for these getting removed is that the way we're adding the schema information is parsing the yaml, fetching the schema from the docs, then writing it, which results in the file getting reformatted based on the tool conventions (we use https://www.npmjs.com/package/js-yaml here).

generator/config/accumulator/percentile.yaml
nirinchev: Yeah, that makes sense - same as above, those got removed due to the yaml parser not preserving comments, but will try and find a way to avoid that.

generator/config/stage/project.yaml
```diff
@@ -11,7 +11,7 @@ arguments:
         name: specification
         type:
             - expression
-        variadic: object
+        variadic: object # XXX: This should somehow allow explicit typings for { fieldName: 1|0 }
```
GromNaN: The type is "expression", because any expression is accepted to compute a value.

generator/config/schema.json
nirinchev: This is the deprecated code_w_s bson type. Realistically, I don't expect any of the more advanced types to be found in the docs, but I generated the schema from the type definitions and opted not to manually edit it to make it easier to maintain if it ever changes. If we wanted to clean it up, we can definitely remove the deprecated types.

generator/config/expression/convert.yaml
nirinchev: Agreed, will look at tweaking the generator to ensure it preserves the old notation.

generator/config/schema.json
jmikola: @GromNaN: For additional context, PHPC [conditionally encodes Code or CodeWScope](https://github.com/mongodb/mongo-php-driver/blob/1.21.0/src/phongo_bson_encode.c#L248-L252) based on whether the MongoDB\BSON\Javascript object was constructed with a non-null `$scope` parameter (see: [`php_phongo_javascript_init`](https://github.com/mongodb/mongo-php-driver/blob/1.21.0/src/BSON/Javascript.c#L48-L52)).

generator/config/stage/project.yaml
addaleax: True. For IDE integrations/user support it's still valuable to be able to say "while this can be an arbitrary document, frequently it will have keys that match document keys and values which are either 0 or 1"

generator/config/accumulator/accumulator.yaml
addaleax: Yeah, at least `prettier` supports YAML out of the box, so that feels like a great way to remove any formatting preference ambiguity

generator/config/stage/project.yaml
GromNaN: So we need an additional type for the projection. Maybe an enum:
```
0: Exclude
1: Include
```

And this parameter accepts an union of `ProjectEnum|expression`

generator/config/accumulator/lastN.yaml
```diff
@@ -12,7 +12,7 @@ description: |
 arguments:
     - name: input
       type:
-          - resolvesToArray
+          - expression
```
nirinchev: This was conflicting with the docs, which say that it can be any expression. It was particularly problematic in the `$group` example since the actual input field doesn't need to be an array, it only becomes one in the context of a `$group`, which is hard to statically enforce. I've relaxed the requirement in the interest of reducing the false negatives, at the cost of potentially accepting more false positives.

generator/config/expression/avg.yaml
```diff
@@ -18,11 +18,9 @@ tests:
       pipeline:
           - $project:
                 quizAvg:
-                    $avg:
-                        - $quizzes
+                    $avg: $quizzes
                 labAvg:
-                    $avg:
-                        - $labs
+                    $avg: $labs
```
nirinchev: These seem to have been errors, right? At least in the docs, those are field expressions rather than arrays.

generator/config/query/eq.yaml
```diff
@@ -9,7 +9,7 @@ description: |
 arguments:
     - name: value
       type:
-          - any
+          - expression
```
nirinchev, line 12: Not sure why this was any previously, but happy to revert if expression is incorrect here.

generator/config/accumulator/count.yaml
```diff
@@ -10,28 +10,64 @@ description: |
...
+      pipeline:
+          - $group:
+                _id: $state
+                countNumberOfDocumentsForState:
+                    $count: {}
+      schema:
+          cakeSales:
+              _id:
+                  types:
+                      - bsonType: Number
```
GromNaN, line 24: Should we use a less verbose syntax?
```suggestion
                      - Number
```

Is this Camelcase name a cross-driver convention? I used lowercase: https://github.com/mongodb/mongo-php-library/blob/75fa63ea5c72e735efbc005ea5b6de8df7c45ec0/generator/config/expressions.php#L19-L38

nirinchev: I believe the format here was chosen to match the format in the [mongodb-schema package](https://github.com/mongodb-js/mongodb-schema), while the casing there is likely intended to match the documentation: https://www.mongodb.com/docs/manual/reference/bson-types/.

GromNaN:
Yaml files moved to https://github.com/mongodb/mql-specifications

PR #1664, opened 2025-04-10
Fix `readPreference` option in benchmark

GromNaN:
Passing a `ReadPreference` to `executeCommand` is deprecated. The driver version 2 requires an array of options.


alcaeus:
evergreen keep-definitions

alcaeus (approved):
LGTM

PR #1669, opened 2025-04-10
PHPLIB-1662 Update GHA test jobs to newer ubuntu version

GromNaN:
Fix PHPLIB-1662

- MongoDB version < 6.0 are not available on ubuntu 22.04.
- Add build on MongoDB 8.0 using ubuntu 24.04

We could run legacy versions of the server in docker, as done in the [laravel package](https://github.com/mongodb/laravel-mongodb/blob/5.3/.github/workflows/build-ci.yml).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -42,7 +42,7 @@ jobs:
             php-version: "8.1"
             mongodb-version: "6.0"
             topology: "sharded_cluster"
-          - os: "ubuntu-22.04"
+          - os: "ubuntu-24.04"
             php-version: "8.1"
             mongodb-version: "8.0"
             topology: "server"
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -23,7 +23,7 @@ jobs:
       fail-fast: true
       matrix:
         os:
-          - "ubuntu-20.04"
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yml
```diff
@@ -23,36 +23,36 @@ jobs:
       fail-fast: true
       matrix:
         os:
-          - "ubuntu-20.04"
+          - "ubuntu-22.04"
```
alcaeus, line 26: WDYT of running MongoDB 8.0 tests on Ubuntu 24.04? That would give us an extra two years assuming GitHub removes support at roughly the same time frame as they did in this instance.

alcaeus (commented):
> We could run legacy versions of the server in docker

I don't think that's necessary, as we test with all versions in Evergreen. The GHA pipeline is mainly here to give us and external contributors quick feedback, and the latest stable release is probably a better reference point anyways.

.github/workflows/tests.yml
GromNaN: I think that's what I did, but did not write explicitly.

GromNaN: Fixed for one missing.

alcaeus (approved):

PR #1681, opened 2025-05-06
DEVPROD-17509 - Replace perf.send

MAhmadShah:
The perf.send functionality in evergreen is no longer maintained and is not the preferred method of sending performance data to the signal processing service. This PR updates the evergreen yaml file to instead send the data down stream using the preferred performance-monitoring-api.corp.mongodb.com/raw_perf_results/cedar_report end point.

Changes were tested in [this patch](https://spruce.mongodb.com/version/681a45ffb6a3ca000747c3be/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### DIFF
--- a/.evergreen/config/test-tasks.yml
+++ b/.evergreen/config/test-tasks.yml
@@ -36,6 +36,35 @@ tasks:
           TOPOLOGY: "server"
           MONGODB_VERSION: "v6.0-perf"
       - func: "run benchmark"
-      - command: perf.send
+      - command: shell.exec
         params:
-          file: src/benchmark/.phpbench/results.json
+          script: |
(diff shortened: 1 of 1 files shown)


GromNaN:
@alcaeus we don't update the drivers-evergreen-tools gitmodule in the branch, should we merge this PR in 1.21 or 2.0?

alcaeus (approved):

alcaeus:
@GromNaN since Devprod wants to eventually remove the perf.send command from evergreen, we have to make this change in all branches that still get built on the regular. I'm not too worried about anything older than 1.21, but there may be the occasional build in 1.21.

GromNaN (approved):

PR #1687, opened 2025-05-09
PHPLIB-1679 Call `DocumentCodec::encode()` only on objects

GromNaN:
Fix PHPLIB-1679

- Check if the input document is an object before calling `DocumentCodec::encode()`
- Never validate the value returned by `DocumentCodec::encode()` as the type `MongoDB\BSON\Document` is enforced.

### DIFF
--- a/src/Operation/BulkWrite.php
+++ b/src/Operation/BulkWrite.php
@@ -35,6 +35,7 @@ use function count;
 use function current;
 use function is_array;
 use function is_bool;
+use function is_object;
 use function key;
 use function MongoDB\is_document;
 use function MongoDB\is_first_key_operator;
@@ -305,7 +306,7 @@ final class BulkWrite
                 case self::INSERT_ONE:
                     // $args[0] was already validated above. Since DocumentCodec::encode will always return a Document
                     // instance, there is no need to re-validate the returned value here.
-                    if ($codec) {
+                    if ($codec && is_object($args[0])) {
                         $operations[$i][$type][0] = $codec->encode($args[0]);
                     }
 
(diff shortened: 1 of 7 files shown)


GromNaN:
I should add tests, probably in `CodecCollectionFunctionalTest`.

src/Operation/BulkWrite.php
```diff
@@ -305,7 +306,7 @@ private function validateOperations(array $operations, ?DocumentCodec $codec, En
                 case self::INSERT_ONE:
                     // $args[0] was already validated above. Since DocumentCodec::encode will always return a Document
                     // instance, there is no need to re-validate the returned value here.
-                    if ($codec) {
+                    if ($codec && is_object($args[0])) {
```
GromNaN: Alternatively, we can throw a `MongoDB\Exception\UnsupportedValueException` when the value is not an object, instead of skipping the codec.

alcaeus:
I don't think we should add the error checks here. The template annotation on the `DocumentCodec` requires an object type as input for `encode`, but the idea is that the codecs accept any value and throw the `invalidEncodableValue` exception themselves if they cannot encode the value - that is unless one uses `encodeIfSupported`.

I wonder if we should instead document the parameter as `mixed` in the `encode` and `decode` methods in `DocumentCodec` (and potentially in `Encoder` and `Decoder` as well) as the methods can be called with the wrong input type but will simply throw an exception if it's not of the expected type. Not sure if we can represent that with static analysis annotations.

GromNaN:
What I understand is that calling `DocumentCodec::encode()` with something other than an object of the class it can encode is an error: the `DocumentCodec` implementation must throw an exception at runtime in this case. But by no means a normal use case. `UnsupportedValueException` is a child of `LogicException`, meaning that's an error in the code.

> `LogicException` that represents error in the program logic. This kind of exception should directly lead to a fix in your code.

The [`@psalm-param ObjectType $value`](https://github.com/mongodb/mongo-php-library/blob/cca72334df754c85ce2f61e080f593d10b9d30c3/src/Codec/DocumentCodec.php#L40) annotation clearly shows the intention. I don't see why we should make static analysis less capable of detecting errors by not specifying the expected value type.

Throwing the error before or inside `DocumentCodec::encode()` will result in the same exception to be thrown.

PR #1693, opened 2025-05-20
PHPLIB-1656: Stop testing with MongoDB 4.0

alcaeus:
PHPLIB-1656

Support for MongoDB 4.0 will be dropped in a future extension release, but we can stop testing with MongoDB 4.0 already.


GromNaN (approved):
I renamed the PR as you dropped a specific error handling for MongoDB 4.0 in the **aggregate** operation.

> Cleanup/refactor any code that isn't needed anymore (code needed to support server version 4.0)

alcaeus:
Good point - with that in mind we may want to wait until connecting to a 4.0 server is rejected by libmongoc. I'll move this into draft until such time.

GromNaN (approved):

alcaeus:
@GromNaN dropped the last commit, so this PR now only drops testing without making any changes to the operations themselves. These changes will be made after we upgrade to libmongoc 2.1 in the extension.

PR #1706, opened 2025-05-26
PHPLIB-1674: Fix timeouts in CSE custom endpoint test

alcaeus:
PHPLIB-1674

This should speed up our builds as each build is currently spending ~4 minutes waiting for timeouts.


GromNaN (approved):
Good use of [.invalid](https://en.wikipedia.org/wiki/.invalid).

PR #1723, opened 2025-06-04
PHPLIB-1378 Replace example printfs with a callable

paulinevos:
PHPLIB-1378

To make it easier for users to work off the examples without having to replace all the print statements


alcaeus (approved):
LGTM. Failing tests were fixed in #1712, so rebasing on top of v2.x should fix the broken build.

PR #1730, opened 2025-07-07
Report coverage and test results for GitHub Actions runs

alcaeus:
In this first PR, we start reporting coverage and test results to codecov.io to get baseline coverage information. I'll add coverage reporting to Evergreen in a separate pull request as this requires changes to the toolchain first.

Note that running tests in coverage mode with xdebug is significantly slower, so we should keep an eye on whether the runtime of tests is still acceptable.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -83,7 +83,7 @@ jobs:
           php-ini-values: "zend.assertions=1"
 
       - name: "Run PHPUnit"
-        run: "vendor/bin/phpunit --configuration phpunit.evergreen.xml"
+        run: "vendor/bin/phpunit --configuration phpunit.evergreen.xml --coverage-clover coverage.xml"
         env:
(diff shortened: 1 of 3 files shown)


.github/workflows/tests.yml
```diff
@@ -81,6 +83,23 @@ jobs:
           php-ini-values: "zend.assertions=1"
 
       - name: "Run PHPUnit"
-        run: "vendor/bin/phpunit"
+        run: "vendor/bin/phpunit --configuration phpunit.evergreen.xml --coverage --coverage-clover coverage.xml"
```
GromNaN: I don't think the `--coverage` option exists.

alcaeus: That's what you get for trusting AI. Removed that, only `--coverage-clover` is required.

paulinevos (approved):

GromNaN (approved):

PR #1732, opened 2025-07-11
PHPLIB-1654: Support new `$concatArrays` and `$setUnion` aggregation accumulators

paulinevos:
Closes PHPLIB-1654

Does what it says on the tin.

### DIFF
--- a/tests/Builder/Accumulator/ConcatArraysAccumulatorTest.php
+++ b/tests/Builder/Accumulator/ConcatArraysAccumulatorTest.php
@@ -1,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MongoDB\Tests\Builder\Accumulator;
+
+use MongoDB\Builder\Accumulator;
+use MongoDB\Builder\Expression;
+use MongoDB\Builder\Pipeline;
+use MongoDB\Builder\Stage;
+use MongoDB\Tests\Builder\PipelineTestCase;
+
+/**
+ * Test $concatArrays accumulator
+ */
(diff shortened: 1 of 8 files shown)


tests/Builder/Accumulator/ConcatArraysAccumulatorTest.php
```diff
@@ -0,0 +1,33 @@
...
+use MongoDB\Tests\Builder\PipelineTestCase;
+
+/**
+ * Test $concatArrays accumulator
+ */
+class ConcatArraysAccumulatorTest extends PipelineTestCase
+{
+    public function testWarehouseCollection(): void
+    {
+        $concatArrays = Accumulator::concatArrays(...);
```
alcaeus, line 20: Ha, our code from discussing different ways to write agg pipelines found its way in here. I'll let you decide whether you want to keep it this way or inline it.

alcaeus (approved):
LGTM with a minor comment.

Looking at the coverage, it's missing tests for exceptions when providing invalid arguments. We can leave those out for the time being and I'd rather add some kind of semi-automated testing like we have for operation classes (see [this code](https://github.com/mongodb/mongo-php-library/blob/ae31f00855de01af72e174d949b34c1c82ca893d/tests/Operation/CountDocumentsTest.php#L27-L39) for an example).

tests/Builder/Accumulator/ConcatArraysAccumulatorTest.php
paulinevos: I liked this :)

PR #1736, opened 2025-07-24
Upload evergreen test results to codecov.io

alcaeus:
In order to better use the flaky test feature of codecov.io, this PR adds a step to upload the test result file to codecov.io, as is already done for GitHub Actions.

### DIFF
--- a/.evergreen/config/functions.yml
+++ b/.evergreen/config/functions.yml
@@ -139,6 +139,7 @@ functions:
               --disable-search \
               --fail-on-error \
               --token ${CODECOV_TOKEN} \
+              --flag "${MONGODB_VERSION}-${TOPOLOGY}" \
               --file test-results.xml || \
               echo "Skipping codecov test result upload due to error"
 
--- a/.evergreen/config/functions.yml
+++ b/.evergreen/config/functions.yml
@@ -126,6 +126,21 @@ functions:
     - command: attach.results
       params:
         file_location: "${DRIVERS_TOOLS}/results.json"
+    - command: shell.exec
+      params:
+        working_dir: src
+        script: |
+          ${PREPARE_SHELL}
+          curl -Os https://cli.codecov.io/latest/linux/codecov
+          sudo chmod +x codecov
+          [[ -f test-results.xml ]] && \
+            ./codecov upload-process \
+              --report-type test_results \
+              --disable-search \
+              --fail-on-error \
+              --token ${CODECOV_TOKEN} \
+              --file test-results.xml || \
+              echo "Skipping codecov test result upload due to error"
 
   "bootstrap mongo-orchestration":
     - command: shell.exec
--- a/.evergreen/config/functions.yml
+++ b/.evergreen/config/functions.yml
@@ -131,19 +131,21 @@ functions:
         working_dir: src
         script: |
           ${PREPARE_SHELL}
-          curl -Os https://cli.codecov.io/latest/linux/codecov
-          curl -Os https://cli.codecov.io/latest/linux/codecov.SHA256SUM
-          shasum -a 256 -c codecov.SHA256SUM
-          sudo chmod +x codecov
-          [[ -f test-results.xml ]] && \
+          if [ -f test-results.xml ]; then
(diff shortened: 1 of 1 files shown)


.evergreen/config/functions.yml
```diff
@@ -126,6 +126,22 @@ functions:
     - command: attach.results
       params:
         file_location: "${DRIVERS_TOOLS}/results.json"
+    - command: shell.exec
+      params:
+        working_dir: src
+        script: |
+          ${PREPARE_SHELL}
+          curl -Os https://cli.codecov.io/latest/linux/codecov
```
jmikola: Since this isn't be pulled from a repository, it'd be prudent to [integrity check the binary](https://docs.codecov.com/docs/the-codecov-cli#integrity-checking-the-binary). I noted this didn't apply to the earlier GitHub Action integration (https://github.com/mongodb/mongo-php-library/commit/ae31f00855de01af72e174d949b34c1c82ca893d) since you just reference their action (`codecov/test-results-action`).

.evergreen/config/functions.yml
```diff
@@ -143,6 +159,13 @@ functions:
           LOAD_BALANCER=${LOAD_BALANCER} \
           REQUIRE_API_VERSION=${REQUIRE_API_VERSION} \
           bash ${DRIVERS_TOOLS}/.evergreen/run-orchestration.sh
+    - command: shell.exec
+      params:
+        script: |
+          printf "\n" >> mo-expansion.yml
+          printf "MONGODB_VERSION: '%s'\n" "${MONGODB_VERSION}" >> mo-expansion.yml
+          printf "TOPOLOGY: '%s'\n" "${TOPOLOGY}" >> mo-expansion.yml
+          cat mo-expansion.yml
```
jmikola: Why was this necessary? `$MONGODB_VERSION` and `$TOPOLOGY` should already be available in the environment, since they're referenced in the previous `shell.exec` command. The inclusion of `cat` here also looks like it may be leftover debugging.

Based on the comment below, `expansions.update` was only necessary to add 
`MONGODB_URI` and `CRYPT_SHARED_LIB_PATH`, which are written in [`drivers_orchestration.py`](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/orchestration/drivers_orchestration.py).

.evergreen/config/functions.yml
```diff
@@ -126,6 +126,22 @@ functions:
...
+        script: |
+          ${PREPARE_SHELL}
+          curl -Os https://cli.codecov.io/latest/linux/codecov
+          sudo chmod +x codecov
+          [[ -f test-results.xml ]] && \
+            ./codecov upload-process \
+              --report-type test_results \
+              --disable-search \
+              --fail-on-error \
+              --token ${CODECOV_TOKEN} \
```
jmikola, line 143: Does this reference a variable from the Evergreen project config? I didn't see this when taking a look at the PHPLIB 2.x project settings.

.evergreen/config/functions.yml
alcaeus: Added the `shasum` check for the time being. I can add the gpg signature check for the checksum file as well, but with evergreen sharing hosts between runs I'm not sure if we should be messing with the keychain.

.evergreen/config/functions.yml
alcaeus: `MONGODB_VERSION` and `TOPOLOGY` are passed as variables to the `bootstrap mongo-orchestration` function only, so they are no longer available later in the run. Adding them to the expansion file created by the script allows us to reuse the values later.

.evergreen/config/functions.yml
alcaeus: For some reason uploading the test result file worked even without a token. I have added it to the repo variables nonetheless.

jmikola: OK. I think it's worth defining the variable, especially since the argument isn't quoted. If it resolves to nothing that could lead to some unexpected behavior for the `--token` option.

jmikola (approved):

PR #1738, opened 2025-07-29
PHPLIB-1695 Add `$createObjectId` operator to the agg builder

GromNaN:
Fix PHPLIB-1695

The documentation is not available yet, so I invented an example. To be updated once the documentation is published.

### DIFF
--- a/generator/config/expression/createObjectId.yaml
+++ b/generator/config/expression/createObjectId.yaml
@@ -1,0 +1,17 @@
+# $schema: ../schema.json
+name: $createObjectId
+link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/createObjectId/'
+type:
+  - resolvesToObjectId
+encode: object
+description: |
+  Returns a random object ID
+tests:
+  -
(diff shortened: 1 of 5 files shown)


generator/config/expression/createObjectId.yaml
```diff
@@ -0,0 +1,17 @@
...
+  Returns a random object ID
+tests:
+  -
+    name: 'Example'
+    link: 'https://www.mongodb.com/docs/manual/reference/operator/aggregation/createObjectId/#example'
+    pipeline:
+      -
+        $project:
+          objectId:
+            $createObjectId: {}
```
GromNaN, line 17: From the [server source](https://github.com/10gen/mongo/pull/38667/files#diff-c5d4272bc19eeeccbbf1f5f93ab06568424f2bf37d8f335fc809ed501a7a1512R4876):
> `$createObjectId` only accepts the empty object as argument

alcaeus (approved):
I love how little work it is to add a new operator 🎉

PR #1742, opened 2025-08-08
PHPLIB-1702: Always consult server encryptedFieldsMap when dropping collections

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1702

PR #1745, opened 2025-08-12
PHPLIB-1702: Always consult server encryptedFieldsMap when dropping collections with `autoEncryption` enabled

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1702

This is a rebase of https://github.com/mongodb/mongo-php-library/pull/1742 on v1.21 with additional modifications from @GromNaN's code review.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -83,6 +83,8 @@ class Client
 
     private WriteConcern $writeConcern;
 
+    private bool $autoEncryptionEnabled;
+
     /**
      * Constructs a new Client instance.
      *
@@ -134,6 +136,7 @@ class Client
         $this->uri = $uri ?? self::DEFAULT_URI;
         $this->builderEncoder = $driverOptions['builderEncoder'] ?? new BuilderEncoder();
         $this->typeMap = $driverOptions['typeMap'];
+        $this->autoEncryptionEnabled = isset($driverOptions['autoEncryption']['keyVaultNamespace']);
 
         $driverOptions = array_diff_key($driverOptions, ['builderEncoder' => 1, 'typeMap' => 1]);
 
@@ -258,7 +261,7 @@ class Client
      */
     public function getCollection(string $databaseName, string $collectionName, array $options = []): Collection
     {
-        $options += ['typeMap' => $this->typeMap, 'builderEncoder' => $this->builderEncoder];
+        $options += ['typeMap' => $this->typeMap, 'builderEncoder' => $this->builderEncoder, 'autoEncryptionEnabled' => $this->autoEncryptionEnabled];
 
         return new Collection($this->manager, $databaseName, $collectionName, $options);
     }
@@ -273,7 +276,7 @@ class Client
      */
     public function getDatabase(string $databaseName, array $options = []): Database
     {
-        $options += ['typeMap' => $this->typeMap, 'builderEncoder' => $this->builderEncoder];
+        $options += ['typeMap' => $this->typeMap, 'builderEncoder' => $this->builderEncoder, 'autoEncryptionEnabled' => $this->autoEncryptionEnabled];
 
         return new Database($this->manager, $databaseName, $options);
     }
--- a/src/functions.php
+++ b/src/functions.php
@@ -201,13 +201,8 @@ function get_encrypted_fields_from_driver(string $databaseName, string $collecti
  * @see Database::dropCollection()
  * @return array|object|null
  */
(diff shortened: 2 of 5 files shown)


tests/Functions/GetEncryptedFieldsFromServerFunctionalTest.php
```diff
@@ -0,0 +1,94 @@
...
+use MongoDB\Collection;
+use MongoDB\Database;
+use MongoDB\Driver\ClientEncryption;
+use MongoDB\Driver\WriteConcern;
+use MongoDB\Tests\FunctionalTestCase;
+
+use function iterator_count;
+use function str_repeat;
+
+class GetEncryptedFieldsFromServerFunctionalTest extends FunctionalTestCase
```
jmikola, line 16: @GromNaN: Per your suggestion in https://github.com/mongodb/mongo-php-library/pull/1742#discussion_r2263727275 I consolidated everything into a single test file named after the function. Even though the tests are utilizing the Collection and Database methods, I think we have a case for this given that it's addressing a bug fix in the underlying function.

GromNaN (approved):

alcaeus (approved):

tests/Functions/GetEncryptedFieldsFromServerFunctionalTest.php
```diff
@@ -0,0 +1,93 @@
...
+        $this->database->createEncryptedCollection(
+            $this->getCollectionName(),
+            $this->clientEncryption,
+            'local',
+            null,
+            ['encryptedFields' => ['fields' => []]],
+        );
+
+        // createEncryptedCollection should create three collections
+        $this->assertCount($originalNumCollections + 3, $this->database->listCollectionNames());
```
GromNaN: If the collection already exists before running the test, this assertion fails.

src/Client.php
```diff
@@ -83,6 +83,8 @@ class Client
 
     private WriteConcern $writeConcern;
 
+    private bool $autoEncryptionEnabled;
```
GromNaN, line 86: @alcaeus @jmikola This new property will not be documented in Database and Collection options.

tests/Functions/GetEncryptedFieldsFromServerFunctionalTest.php
jmikola: Is this only a concern when aborting the test locally (or retrying after a failure)?

No objections to dropping it before this first assertion.

src/Client.php
```diff
@@ -134,6 +136,7 @@ public function __construct(?string $uri = null, array $uriOptions = [], array $
         $this->uri = $uri ?? self::DEFAULT_URI;
         $this->builderEncoder = $driverOptions['builderEncoder'] ?? new BuilderEncoder();
         $this->typeMap = $driverOptions['typeMap'];
+        $this->autoEncryptionEnabled = isset($driverOptions['autoEncryption']['keyVaultNamespace']);
```
jmikola, line 139: Noted that `keyVaultNamespace` is a required `autoEncryption` option.

[`phongo_manager_set_auto_encryption_opts`](https://github.com/mongodb/mongo-php-driver/blob/2.1.1/src/phongo_client.c#L1137) doesn't seem to check about `keyVaultNamespace` being unset, so I assume we'd rely on libmongoc(rypt) raising an error since it's documented as required. In any event, it's sensible to check for it directly instead of just seeing if `autoEncryption` is set.

src/functions.php
```diff
@@ -201,13 +201,8 @@ function get_encrypted_fields_from_driver(string $databaseName, string $collecti
  * @see Database::dropCollection()
  * @return array|object|null
  */
-function get_encrypted_fields_from_server(string $databaseName, string $collectionName, Manager $manager, Server $server)
+function get_encrypted_fields_from_server(string $databaseName, string $collectionName, Server $server)
 {
-    // No-op if the encryptedFieldsMap autoEncryption driver option was omitted
-    if ($manager->getEncryptedFieldsMap() === null) {
```
jmikola, line 207: Ideally, it seems like we'd want the Manager to be able to report whether autoEncryption is enabled.

GromNaN (approved):

src/Client.php
jmikola: Created [PHPC-2615](https://jira.mongodb.org/browse/PHPC-2615) to allow us to fetch this from the Manager directly.

PR #1757, opened 2025-08-19
PHPLIB-1708 Cast empty KMS provider into an object

GromNaN:
Fix PHPLIB-1708

For some KMS providers, all options can be omitted. The authentication is done using the env var or the system.
https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.md#credentialproviders

The approach in https://github.com/doctrine/mongodb-odm/pull/2801 is not sufficient because Symfony is [not able to dump the empty object instance into the container](https://github.com/symfony/symfony/blob/7ed3a4b57cdf66932417c02cc634eadcc65e4772/src/Symfony/Component/DependencyInjection/Dumper/PhpDumper.php#L1986).
 

    Unable to dump a service container if a parameter is an object or a resource, got "stdClass".

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -119,12 +119,8 @@ class Client
             throw InvalidArgumentException::invalidType('"typeMap" driver option', $driverOptions['typeMap'], 'array');
         }
 
-        if (isset($driverOptions['autoEncryption']['keyVaultClient'])) {
-            if ($driverOptions['autoEncryption']['keyVaultClient'] instanceof self) {
-                $driverOptions['autoEncryption']['keyVaultClient'] = $driverOptions['autoEncryption']['keyVaultClient']->manager;
-            } elseif (! $driverOptions['autoEncryption']['keyVaultClient'] instanceof Manager) {
-                throw InvalidArgumentException::invalidType('"keyVaultClient" autoEncryption option', $driverOptions['autoEncryption']['keyVaultClient'], [self::class, Manager::class]);
-            }
+        if (isset($driverOptions['autoEncryption']) && is_array($driverOptions['autoEncryption'])) {
+            $driverOptions['autoEncryption'] = $this->prepareEncryptionOptions($driverOptions['autoEncryption']);
         }
 
         if (isset($driverOptions['builderEncoder']) && ! $driverOptions['builderEncoder'] instanceof Encoder) {
@@ -213,13 +209,7 @@ class Client
      */
     public function createClientEncryption(array $options)
     {
-        if (isset($options['keyVaultClient'])) {
-            if ($options['keyVaultClient'] instanceof self) {
-                $options['keyVaultClient'] = $options['keyVaultClient']->manager;
-            } elseif (! $options['keyVaultClient'] instanceof Manager) {
-                throw InvalidArgumentException::invalidType('"keyVaultClient" option', $options['keyVaultClient'], [self::class, Manager::class]);
-            }
-        }
+        $options = $this->prepareEncryptionOptions($options);
 
         return $this->manager->createClientEncryption($options);
     }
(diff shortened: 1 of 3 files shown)


GromNaN:
Fixed in Doctrine MongoDB ODM Bundle by setting a DI `Definition` of `stdClass` in the parameters instead of an empty `stdClass` object. https://github.com/doctrine/DoctrineMongoDBBundle/pull/897/commits/953613d40fbcccc43871aa419c8243615f54aba8

src/Client.php
```diff
@@ -457,6 +447,28 @@ public function watch(array $pipeline = [], array $options = [])
         return $operation->execute($server);
     }
 
+    private function formatEncryptionOptions(array $options): array
```
jmikola:
```suggestion
    private function prepareAutoEncryptionOptions(array $options): array
```

I'm not sure if there's prior art in PHPLIB, but I use "prepare" extensively in PHPC for this sort of thing. Up to you, though.

I do like including "AutoEncryption" in the name here since it refers to that driver option.

src/Client.php
```diff
@@ -457,6 +447,28 @@ public function watch(array $pipeline = [], array $options = [])
...
+            } elseif (! $options['keyVaultClient'] instanceof Manager) {
+                throw InvalidArgumentException::invalidType('"keyVaultClient" option', $options['keyVaultClient'], [self::class, Manager::class]);
+            }
+        }
+
+        // The server requires an empty document for automatic credentials.
+        if (isset($options['kmsProviders']) && is_array($options['kmsProviders'])) {
+            foreach ($options['kmsProviders'] as $name => $provider) {
+                if ($provider === []) {
+                    $options['kmsProviders'][$name] = Document::fromPHP([]);
```
GromNaN: Is it better to cast the empty array value like this?
```suggestion
                    $options['kmsProviders'][$name] = (object) $provider;
```

jmikola: Any reason to use `Document::fromPHP([])` instead of `(object) []` or `new stdClass`? PHPC is going to end up [encoding this as BSON anyway](https://github.com/mongodb/mongo-php-driver/blob/9eff31bd994a988743044a6cad727e3ea3a9f619/src/phongo_client.c#L1249).

tests/ClientTest.php
```diff
@@ -37,6 +37,18 @@ public function testConstructorAutoEncryptionOpts(): void
         new Client(static::getUri(), [], ['autoEncryption' => $autoEncryptionOpts]);
     }
 
+    #[DoesNotPerformAssertions]
+    public function testConstructorEmptyKmsProvider(): void
```
jmikola, line 41: Would this test have failed without the above patch? If not, is anything being tested at all?

src/Client.php
```diff
@@ -119,12 +119,8 @@ public function __construct(?string $uri = null, array $uriOptions = [], array $
...
         }
 
-        if (isset($driverOptions['autoEncryption']['keyVaultClient'])) {
-            if ($driverOptions['autoEncryption']['keyVaultClient'] instanceof self) {
-                $driverOptions['autoEncryption']['keyVaultClient'] = $driverOptions['autoEncryption']['keyVaultClient']->manager;
-            } elseif (! $driverOptions['autoEncryption']['keyVaultClient'] instanceof Manager) {
-                throw InvalidArgumentException::invalidType('"keyVaultClient" autoEncryption option', $driverOptions['autoEncryption']['keyVaultClient'], [self::class, Manager::class]);
-            }
+        if (isset($driverOptions['autoEncryption']) && is_array($driverOptions['autoEncryption'])) {
+            $driverOptions['autoEncryption'] = $this->formatEncryptionOptions($driverOptions['autoEncryption']);
```
jmikola:
```suggestion
            $driverOptions['autoEncryption'] = $this->prepareAutoEncryptionOptions($driverOptions['autoEncryption']);
```

Leaving this suggestion in case you accept the other as-is.

src/Client.php
GromNaN: Ok for "prepare", but no for "AutoEncryption": the method `createClientEncryption` isn't specific to auto encryption.

src/Client.php
GromNaN: The method must be moved to please the pedantic order.

tests/ClientTest.php
GromNaN: Yes:
> MongoDB\Driver\Exception\EncryptionException: expected BSON document for field: gcp

src/Client.php
GromNaN: I have no idea what's the more efficient. Changed to `new stdClass`

jmikola (approved):

PR #1761, opened 2025-08-20
Remove calls to `ReflectionProperty::setAccessible(true)`

GromNaN:
> As of PHP 8.1.0, calling this method has no effect; all properties are accessible by default. ([doc](https://www.php.net/manual/en/reflectionproperty.setaccessible.php))

A deprecation message is triggered as of PHP 8.5: https://github.com/php/php-src/pull/19273


jmikola (approved):
Noted that v1.21 requires 8.1+, so there's no concern with supporting older PHP versions. Glad to see this method removed, as it never quite made sense to me.

PR #1762, opened 2025-08-20
Use `Stringable` interface explicitly

GromNaN:
The [Stringable](https://php.watch/versions/8.0/stringable) interface was added in PHP 8.0. It makes it explicit that the class implements the magic method `__toString`.

### DIFF
--- a/tests/Model/BSONIteratorTest.php
+++ b/tests/Model/BSONIteratorTest.php
@@ -19,7 +19,7 @@ class BSONIteratorTest extends TestCase
     #[DataProvider('provideTypeMapOptionsAndExpectedDocuments')]
     public function testValidValues(?array $typeMap, array $expectedDocuments): void
     {
-        $binaryString = implode(array_map(
+        $binaryString = implode('', array_map(
             fn ($input) => (string) Document::fromPHP($input),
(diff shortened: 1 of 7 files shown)


tests/Model/BSONIteratorTest.php
```diff
@@ -19,7 +19,7 @@ class BSONIteratorTest extends TestCase
     #[DataProvider('provideTypeMapOptionsAndExpectedDocuments')]
     public function testValidValues(?array $typeMap, array $expectedDocuments): void
     {
-        $binaryString = implode(array_map(
+        $binaryString = implode('', array_map(
```
jmikola, line 22: [`implode()`](https://www.php.net/implode) suggests this alternative syntax is still supported, provided the separator string is empty and named arguments aren't utilized. No objection here, but I'm curious what prompted this change. Was it tooling?

jmikola (approved):

tests/Model/BSONIteratorTest.php
GromNaN: That's prompted by rector. I expect the alternative syntax could be deprecated one day.

PR #1766, opened 2025-08-26
PHPLIB-1710: Remove dead local `.evergreen` config

paulinevos:
As it's covered by the `drivers-evergreen-tools` submodule now.


jmikola:
I think there are some additional changes that can be applied and tested:

 * `.evergreen/config/functions.yml` refers to the local `x509gen` directory several times. Those references can be changed to `tests/drivers-evergreen-tools/.evergreen/x509gen`.
 * `.evergreen/orchestration` only contains a single `basic.json` file for a sharded cluster, and that doesn't appear to be referenced anywhere.
 * `.evergreen/auth_aws` doesn't appear to be referenced.

paulinevos:
@jmikola What I don't understand re: the `x509gen` references is that they're from the same exact commit as references to `x509gen` in the submodule a few lines later. That makes me feel like maybe there's a reason that points to the local ones

paulinevos:
@jmikola Oh actually, they aren't references to the local one at all:
```
cd ${DRIVERS_TOOLS}/.evergreen/csfle
```

I'll see if they're referenced anywhere else and otherwise remove

.github/workflows/tests.yml
```diff
@@ -70,7 +70,7 @@ jobs:
         with:
           python-version: '3.13'
 
-      - name: "Setup MongoDB"
+      - name: "Set up MongoDB"
```
jmikola, line 73: Was this change intentional? The old name was consistent with [PHPC](https://github.com/mongodb/mongo-php-driver/blob/v2.x/.github/workflows/tests.yml#L60). I don't feel strongly either way, just wanted to point that out.

jmikola (approved):
Thanks, I missed the `cd` command in `functions.yml`.

Changes LGTM. I'll defer to you on the `tests.yml` change.

And please revise the PR title before squash-merging, since the PR addresses more than just OCSP.

.github/workflows/tests.yml
paulinevos: It was, I just changed it because of grammatical correctness, as "setup" is a noun and "to set up" is the verb. It's just a peeve I have 🫣

PR #1794, opened 2025-11-05
PHPLIB-1689 Make the Atlas Search exception more specific

GromNaN:
Fix PHPLIB-1689

Conditions imported from https://github.com/mongodb/laravel-mongodb/blob/a2b4ab86dfc9248050b2592d9830773ac335774e/src/Schema/Builder.php#L386-L395

For the test, the strategy is to check that the exception throw is always a `SearchNotSupportedException` for operations related to search. But since the server are many inconsistent behaviors depending on the versions, we don't assume if the operation will work or not.


src/Exception/SearchNotSupportedException.php
```diff
@@ -0,0 +1,31 @@
...
+    public static function isSearchNotSupportedError(Throwable $e): bool
+    {
+        if (! $e instanceof ServerException) {
+            return false;
+        }
+
+        return in_array($e->getCode(), [
+            59,      // MongoDB 4 to 6, 7-community: no such command: 'createSearchIndexes'
+            40324,   // MongoDB 4 to 6: Unrecognized pipeline stage name: '$listSearchIndexes'
+            115,     // MongoDB 7-ent: Search index commands are only supported with Atlas.
```
GromNaN: Unfortunately, this is too broad. These error codes can be used for errors unrelated to Atlas Search.

jmikola: IIUC, your comment isn't just about [error code 115](https://www.mongodb.com/docs/manual/reference/error-codes/#mongodb-error-115), but more generally applies to all of the codes for unrecognized commands/operators.

Perhaps this would require inspecting the server-side error message for a relevant string? Although that might take us back to an anti-pattern in driver development.

GromNaN:
Since mongo 6.0.11-ent, the `$listSearchIndexes` stage returns nothing, while the previous versions threw a server error. This makes it more complicated to test and detect.

jmikola (commented):
@GromNaN: Was there anything in this PR you wanted to keep, or should it all be abandoned (given the difficulty in error detection)?

GromNaN:
I found a more reliable way to detect when Search indexes are supported.
And I made the test pass weither search indexes are supported or not: if there is an exception it must be a `SearchIndexNotSupportedException`.

Also fixed the exception for update and drop search index.

src/Exception/SearchNotSupportedException.php
```diff
@@ -0,0 +1,55 @@
...
+
+        return match ($exception->getCode()) {
+            // MongoDB 8: Using Atlas Search Database Commands and the $listSearchIndexes aggregation stage requires additional configuration.
+            31082 => true,
+            // MongoDB 7: $listSearchIndexes stage is only allowed on MongoDB Atlas
+            6047401 => true,
+            // MongoDB 7-ent: Search index commands are only supported with Atlas.
+            115 => true,
+            // MongoDB 4 to 6, 7-community
+            59 => match ($exception->getMessage()) {
```
paulinevos, line 37: match inception 🤯

src/Exception/SearchNotSupportedException.php
```diff
@@ -0,0 +1,55 @@
...
+                'no such command: \'createSearchIndexes\'' => true,
+                'no such command: \'updateSearchIndex\'' => true,
+                'no such command: \'dropSearchIndex\'' => true,
+                default => false,
+            },
+            // MongoDB 4 to 6
+            40324 => match ($exception->getMessage()) {
```
paulinevos, line 44: Did you trawl through all the old versions' source code to find all these error codes? 😭

src/Operation/Aggregate.php
```diff
@@ -233,7 +235,15 @@ public function execute(Server $server): CursorInterface
             $this->createCommandOptions(),
         );
 
-        $cursor = $this->executeCommand($server, $command);
+        try {
+            $cursor = $this->executeCommand($server, $command);
+        } catch (ServerException $exception) {
+            if (SearchNotSupportedException::isSearchNotSupportedError($exception)) {
```
paulinevos, line 241: Nice solution

tests/Exception/SearchNotSupportedExceptionTest.php
```diff
@@ -0,0 +1,102 @@
...
+class SearchNotSupportedExceptionTest extends FunctionalTestCase
+{
+    #[DoesNotPerformAssertions]
+    public function testListSearchIndexesNotSupportedException(): void
+    {
+        $collection = $this->createCollection($this->getDatabaseName(), $this->getCollectionName());
+
+        try {
+            $collection->listSearchIndexes();
+        } catch (SearchNotSupportedException) {
```
paulinevos, line 20: Basically are you doing it like these because it's too complex to skip if it _is_ supported? Can't we just do the inverse of the check you do in `skipIfSearchIndexIsNotSupported` and expect an exception?

src/Exception/SearchNotSupportedException.php
GromNaN: I tried many server versions using mongo-orchestration, [m](https://www.npmjs.com/package/m) and mongo-launch; Atlas M0 and Atlas Local Docker.
Then our test matrix helped validate all the versions.

tests/Exception/SearchNotSupportedExceptionTest.php
GromNaN: This is a way to test even if `skipIfSearchIndexIsNotSupported` is incomplete. Here, I test each operation one by one to ensure that if there is an exception, it is properly wrapped.

I noticed that not all server versions are consistent. As an example in mongo `6.0.11-ent`, the `$listSearchIndexes` stage returns an empty list but `createSearchIndexes` returns an error.

src/Exception/SearchNotSupportedException.php
GromNaN: A decision tree 🎄

src/Operation/Aggregate.php
jmikola: @GromNaN: if the snippet is always a conditional check and throw, do you want to consider a single method that conditionally NOPs? Then you can replace these lines with `throw SearchNotSupportedException::proxy($exception)`.

I don't mind keeping this explicit, though.

src/Operation/DropSearchIndex.php
```diff
@@ -72,6 +73,10 @@ public function execute(Server $server): void
         } catch (CommandException $e) {
             // Drop operations are idempotent. The server may return an error if the collection does not exist.
             if ($e->getCode() !== self::ERROR_CODE_NAMESPACE_NOT_FOUND) {
+                if (SearchNotSupportedException::isSearchNotSupportedError($e)) {
```
jmikola, line 76: Noted that [CommandException](https://www.php.net/manual/en/class.mongodb-driver-exception-commandexception.php) extends ServerException.

tests/FunctionalTestCase.php
```diff
@@ -434,13 +433,22 @@ protected function skipIfServerVersion(string $operator, string $version, ?strin
         }
     }
 
-    protected function skipIfAtlasSearchIndexIsNotSupported(): void
+    protected function skipIfSearchIndexIsNotSupported(): void
```
jmikola, line 436: I assume this change was done to allow search index tests to run against local Atlas instances, correct? Therefore, the URI check was insufficient.

tests/Exception/SearchNotSupportedExceptionTest.php
```diff
@@ -0,0 +1,102 @@
...
+            // If an exception is thrown because Atlas Search is not supported,
+            // then the test is successful because it has the correct exception class.
+        }
+    }
+
+    #[DoesNotPerformAssertions]
+    public function testSearchStageNotSupportedException(): void
+    {
+        // The server returns an empty result if the search index does not exist.
```
jmikola, line 29: IIRC, didn't you ask the server to reconsider returning empty results? Just curious if there's an upstream ticket where this behavior could potentially change in the future.

jmikola (approved):
Some feedback/questions but not pushing for any changes.

src/Operation/Aggregate.php
GromNaN: The first line of the exception backtrace is where the exception is thrown, and in many environments it's the only line that is displayed. Thats why I think it's better to throw the exception in the catch, to provide more information of the context.

tests/FunctionalTestCase.php
GromNaN: Exactly. When this method was created, atlas local did not exist. And search is not available outside of atlas.

tests/Exception/SearchNotSupportedExceptionTest.php
GromNaN: I remember something about that. We'll see about ignoring the new error in the test in this case, because you can't create a search index in an environment that doesn't support it.

PR #1695, opened 2025-05-20
PHPLIB-1680: Remove obsolete wire version checks for MongoDB 4.2

alcaeus:
PHPLIB-1680

PR #1801, opened 2025-11-26
PHPLIB-1680: Remove pre-4.2 wire version checks and test code

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1680

Supersedes https://github.com/mongodb/mongo-php-library/pull/1695

### DIFF
--- a/src/Operation/FindAndModify.php
+++ b/src/Operation/FindAndModify.php
@@ -55,8 +55,6 @@ final class FindAndModify implements Explainable
 {
     private const WIRE_VERSION_FOR_HINT = 9;
 
-    private const WIRE_VERSION_FOR_UNSUPPORTED_OPTION_SERVER_SIDE_ERROR = 8;
-
     private array $options;
 
     /**
@@ -227,12 +225,6 @@ final class FindAndModify implements Explainable
      */
     public function execute(Server $server): array|object|null
     {
-        /* Server versions >= 4.2.0 raise errors for unsupported update options.
-         * For previous versions, the CRUD spec requires a client-side error. */
-        if (isset($this->options['hint']) && ! server_supports_feature($server, self::WIRE_VERSION_FOR_UNSUPPORTED_OPTION_SERVER_SIDE_ERROR)) {
-            throw UnsupportedException::hintNotSupported();
-        }
-
         /* CRUD spec requires a client-side error when using "hint" with an
          * unacknowledged write concern on an unsupported server. */
         if (
--- a/tests/FunctionalTestCase.php
+++ b/tests/FunctionalTestCase.php
@@ -469,10 +469,6 @@ abstract class FunctionalTestCase extends TestCase
 
     protected function skipIfClientSideEncryptionIsNotSupported(): void
     {
-        if (version_compare($this->getFeatureCompatibilityVersion(), '4.2', '<')) {
-            $this->markTestSkipped('Client Side Encryption only supported on FCV 4.2 or higher');
-        }
-
         if (static::getModuleInfo('libmongocrypt') === 'disabled') {
             $this->markTestSkipped('Client Side Encryption is not enabled in the MongoDB extension');
         }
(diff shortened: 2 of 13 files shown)


tests/Operation/UpdateFunctionalTest.php
GromNaN: Please run phpcbf.

GromNaN (changes requested):
There is an error with [PHPUnit tests (ubuntu-22.04, 8.1, 6.0, sharded_cluster)](https://github.com/mongodb/mongo-php-library/actions/runs/19707362304/job/56458331008?pr=1801#logs)

```
UnexpectedValueException: Could not determine server storage engine
```

src/Operation/FindAndModify.php
```diff
@@ -227,12 +225,6 @@ public function __construct(private string $databaseName, private string $collec
      */
     public function execute(Server $server): array|object|null
     {
-        /* Server versions >= 4.2.0 raise errors for unsupported update options.
-         * For previous versions, the CRUD spec requires a client-side error. */
-        if (isset($this->options['hint']) && ! server_supports_feature($server, self::WIRE_VERSION_FOR_UNSUPPORTED_OPTION_SERVER_SIDE_ERROR)) {
-            throw UnsupportedException::hintNotSupported();
```
alcaeus, line 233: Nit: Is this method still used after we remove everything? Not sure if we consider this public API, but we can remove the method if we consider it private and it's no longer used.

alcaeus (changes requested):
One nit, and a request to run phpcbf. Changes themselves LGTM.

src/Operation/FindAndModify.php
jmikola: It's still used by the Delete operation: https://github.com/mongodb/mongo-php-library/blob/2.1.2/src/Operation/Delete.php#L126-L133

I don't feel strongly about removing it after we raise the minimum server version to 4.4+, but I'm amenable to leaving it around and deprecated since it's technically public.

tests/FunctionalTestCase.php
```diff
@@ -491,16 +487,18 @@ protected function skipIfGeoHaystackIndexIsNotSupported(): void
...
-        }
+        switch ($this->getPrimaryServer()->getType()) {
+            case Server::TYPE_STANDALONE:
+                $this->markTestSkipped('Transactions are not supported on standalone servers');
+                break;
 
-        if ($this->isShardedCluster()) {
-            $this->markTestSkipped('Transactions are only supported on FCV 4.2 or higher');
-        }
+            case Server::TYPE_RS_PRIMARY:
```
GromNaN, line 495: The condition on `$this->getPrimaryServer()->getType() === Server::TYPE_RS_PRIMARY`. I read the following comment, but I don't get the reason of this update.

alcaeus: Previous behaviour was to skip on a standalone server, on sharded clusters (not sure why this was unconditional, it should've only skipped on sharded clusters < 4.2), and on replica sets with a storage engine other than WiredTiger. Since we're now always on 4.2 and higher, we can skip on standalones, check for WiredTiger on a primary, and always support sharded clusters, hence the change.

alcaeus (approved):
LGTM.

tests/FunctionalTestCase.php
jmikola: > not sure why this was unconditional, it should've only skipped on sharded clusters < 4.2

That stood out as an error to me (i.e. never running these tests on sharded clusters), but it went away with this change so I didn't call it out.

The new logic assumes WiredTiger on sharded clusters since we can't ask `mongos` directly (similar to what was already done for casual consistency checks). We have logic elsewhere (or did at one point) to punch through to individual servers in the sharded cluster configuration but that seems like overkill given we know exactly what environments we test on (and the only options going forward are in-memory and WT).

GromNaN (approved):

PR #1808, opened 2025-12-15
Draft PR for SBOM generation in release

ekovalets:
This PR is aimed to present a conceptual version of sbom generation during release


.github/actions/setup-sbom/action.yml
```diff
@@ -0,0 +1,22 @@
...
+    description: "The directory where composer.json is located"
+    required: false
+    default: "."
+
+runs:
+  using: composite
+  steps:
+    - name: Install latest Composer
+      shell: bash
+      run: curl -sS https://getcomposer.org/installer | php -- --install-dir=. --filename=composer
```
GromNaN: Composer is already installed with `shivammathur/setup-php`: https://github.com/mongodb/mongo-php-library/blob/v2.x/.github/actions/setup/action.yml

.github/workflows/release-with-sbom.yml
```diff
@@ -0,0 +1,240 @@
...
+          driver-version: ${{ env.DRIVER_VERSION }}
+          working-directory: '.'
+
+      - name: "Generate/Update composer.lock"
+        id: composer-lock
+        run: |
+          echo "Resolving dependencies and generating composer.lock..."
+          composer update --lock --no-install --ignore-platform-reqs
+          echo "composer.lock generated with resolved versions"
+        continue-on-error: true
```
GromNaN: Our setup can install dependencies and use cache. https://github.com/mongodb/mongo-php-library/blob/004a3c022ca20d922125cd800768fbff543eb63e/.github/actions/setup/action.yml#L53-L57

.github/workflows/release-with-sbom.yml
```diff
@@ -0,0 +1,240 @@
...
+          aws_secret_id: ${{ secrets.AWS_SECRET_ID }}
+
+      - name: "Prepare release message"
+        run: |
+          cat > release-message <<'EOL'
+          ${{ format(env.default-release-message, inputs.version, inputs.jira-version-number) }}
+          EOL
+
+      - name: "Create draft release"
+        run: echo "RELEASE_URL=$(gh release create ${{ inputs.version }} --target ${{ github.ref_name }} --title "${{ inputs.version }}" --notes-file release-message --draft)" >> "$GITHUB_ENV"
```
GromNaN: We already have a release workflow: https://github.com/mongodb/mongo-php-library/blob/v2.x/.github/workflows/release.yml

.github/workflows/sbom.yml
```diff
@@ -0,0 +1,114 @@
...
+             <(cat $SBOM_FILE | jq -r "$JQ_NORMALIZER"); then
+             
+             echo "HAS_CHANGES=false" >> $GITHUB_OUTPUT
+             echo "No changes detected in sbom.json"
+          else
+             echo "HAS_CHANGES=true" >> $GITHUB_OUTPUT
+             echo "Changes detected in sbom.json"
+          fi
+
+      - name: Create Pull Request
```
jmikola: IIRC, we last talked about _not_ creating a PR and only attaching the generated SBOM to the release assets. 

Quoting the [#php-driver thread](https://mongodb.slack.com/archives/C0V31PR2T/p1765313970318239?thread_ts=1764788758.215839&cid=C0V31PR2T):

>> (from @addaleax) For our tools, we don't commit the SBOM to the repository. We generate them as part of regular CI on mainline commits, as part of PRs, and as part of the release flow; they are being passed to Kondukto for scanning, and for release builds, the resulting augmented SBOM files are made part of the release artifacts.
>
> To that end, I expect we might be better off not having SBOMs committed to PHP repos at all and merely generated and attached as release assets. IIRC, the PHPLIB SBOM only exists because it's an empty template that never needed to be updated. And the PHPC SBOM is generated manually whenever we bump our libmongocrypt and libmongoc dependencies.

And from an earlier comment in that thread:

> I suggested incorporating SBOM regeneration into the release workflow (e.g. above the "[Create release tag](https://github.com/mongodb/mongo-php-library/blob/98c020d2b7f4a9d325031d9ecf6912c295fda5e4/.github/workflows/release.yml#L116C7-L120C53)" step). Alternatively, if the SBOM doesn't really need to exist in the repo, we can just generate it at any point prior to "[Upload SBOM as release artifact](https://github.com/mongodb/mongo-php-library/blob/98c020d2b7f4a9d325031d9ecf6912c295fda5e4/.github/workflows/release.yml#L177C16-L177C47)").

ekovalets: @jmikola during the call we were talking about adding the generated sbom directly to repo during generation - recall we were discussion if github actions has correct access and if project setup change is required.

for now the plan is to get sbom into the repo. (as we do for other driver and products and where our customer take if from now)

if you don't want it to be auto added to the repo, we can return to PR creation approach or I'm open for suggestion provided that we need to store sbom in the repo.

.github/workflows/release-with-sbom.yml
ekovalets: @GromNaN yeah, I've created a separate one for debug - I'll move the code into the release one once we agree on top level approach.

.github/actions/setup-sbom/action.yml
ekovalets: There were issues with old composer version - the cyclonedx didn't work with it.

can you update the composer in the environment?

PR #1809, opened 2025-12-16
PHPLIB-1728: Encryption flag in driver metadata

paulinevos:
Prepends `iue`  for `In-Use Encryption` to platform metadata string in MongoDB handshake. This will allow for easier detection of encryption usage in higher level libraries.

This PR also extracts the options array into `DriverOptions` in order to simplify testing the I/O of options as we do quite a bit of transformation on them.

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -141,6 +141,7 @@ class Client implements Stringable
         $driverOptions = array_diff_key($driverOptions, ['builderEncoder' => 1, 'typeMap' => 1]);
 
         $this->manager = new Manager($uri, $uriOptions, $driverOptions);
+
         $this->readConcern = $this->manager->getReadConcern();
         $this->readPreference = $this->manager->getReadPreference();
         $this->writeConcern = $this->manager->getWriteConcern();
--- a/src/Client.php
+++ b/src/Client.php
@@ -17,13 +17,8 @@
 
 namespace MongoDB;
 
-use Composer\InstalledVersions;
 use Iterator;
-use MongoDB\BSON\Document;
-use MongoDB\BSON\PackedArray;
-use MongoDB\Builder\BuilderEncoder;
 use MongoDB\Builder\Pipeline;
-use MongoDB\Codec\Encoder;
 use MongoDB\Driver\BulkWriteCommand;
 use MongoDB\Driver\BulkWriteCommandResult;
 use MongoDB\Driver\ClientEncryption;
@@ -38,9 +33,9 @@ use MongoDB\Driver\WriteConcern;
 use MongoDB\Exception\InvalidArgumentException;
 use MongoDB\Exception\UnexpectedValueException;
 use MongoDB\Exception\UnsupportedException;
-use MongoDB\Model\BSONArray;
-use MongoDB\Model\BSONDocument;
+use MongoDB\Model\AutoEncryptionOptions;
 use MongoDB\Model\DatabaseInfo;
+use MongoDB\Model\DriverOptions;
 use MongoDB\Operation\ClientBulkWriteCommand;
 use MongoDB\Operation\DropDatabase;
 use MongoDB\Operation\ListDatabaseNames;
@@ -48,25 +43,13 @@ use MongoDB\Operation\ListDatabases;
 use MongoDB\Operation\Watch;
 use stdClass;
 use Stringable;
-use Throwable;
 
 use function array_diff_key;
-use function is_array;
-use function is_string;
 
+/** @psalm-import-type stage from Builder\Pipeline */
 class Client implements Stringable
 {
     public const DEFAULT_URI = 'mongodb://127.0.0.1/';
-
-    private const DEFAULT_TYPE_MAP = [
-        'array' => BSONArray::class,
-        'document' => BSONDocument::class,
-        'root' => BSONDocument::class,
-    ];
-
-    private const HANDSHAKE_SEPARATOR = '/';
-
-    private static ?string $version = null;
 
     private Manager $manager;
 
@@ -76,14 +59,9 @@ class Client implements Stringable
 
     private string $uri;
 
-    private array $typeMap;
-
-    /** @psalm-var Encoder<array|stdClass|Document|PackedArray, mixed> */
-    private readonly Encoder $builderEncoder;
-
     private WriteConcern $writeConcern;
 
-    private bool $autoEncryptionEnabled;
+    private DriverOptions $driverOptions;
 
     /**
      * Constructs a new Client instance.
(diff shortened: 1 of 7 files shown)


src/ValueObject/AutoEncryptionOptions.php
```diff
@@ -0,0 +1,71 @@
+<?php
+
+namespace MongoDB\ValueObject;
```
GromNaN: This class should move to the `MongoDB\Model` namespace, which contains similar classes.

src/ValueObject/DriverOptions.php
```diff
@@ -0,0 +1,147 @@
...
+        public readonly Encoder $builderEncoder,
+        public readonly array $autoEncryption,
+        public readonly array $driver,
+        private readonly array $miscOptions,
+    ) {
+    }
+
+    public function toArray(): array
+    {
+        return array_filter([
```
GromNaN: Some [options](https://www.php.net/manual/en/mongodb-driver-manager.construct.php#mongodb-driver-manager.construct-driveroptions) are boolean: `disableClientPersistence`.

You must filter only the `null` values.

src/Model/DriverOptions.php
```diff
@@ -0,0 +1,147 @@
...
+    ];
+
+    private const HANDSHAKE_SEPARATOR = '/';
+
+    private function __construct(
+        public readonly array $typeMap,
+        public readonly Encoder $builderEncoder,
+        public readonly array $autoEncryption,
+        public readonly array $driver,
+        private readonly array $miscOptions,
```
GromNaN, line 48: Should it be more specific.
Not `$managerOptions` because `$autoEncryption` is a manager option.

src/ValueObject/DriverOptions.php
```diff
@@ -0,0 +1,147 @@
...
+        'document' => BSONDocument::class,
+        'root' => BSONDocument::class,
+    ];
+
+    private const HANDSHAKE_SEPARATOR = '/';
+
+    private function __construct(
+        public readonly array $typeMap,
+        public readonly Encoder $builderEncoder,
+        public readonly array $autoEncryption,
```
GromNaN: This should be nullable. We don't send the `autoEncryption` to the manager if it's not enabled.

src/ValueObject/DriverOptions.php
```diff
@@ -0,0 +1,147 @@
...
+    public static function fromArray(array $options): self
+    {
+        $options += ['typeMap' => self::DEFAULT_TYPE_MAP];
+
+        self::ensureValidArrayOptions($options);
+
+        $autoEncryption = match (isset($options[self::KEY_AUTO_ENCRYPTION])) {
+            true => AutoEncryptionOptions::fromArray($options[self::KEY_AUTO_ENCRYPTION])->toArray(),
+            false => []
+        };
```
GromNaN: I find the usage of `match` abusive here.
```suggestion
        $autoEncryption = null;
        if (isset($options[self::KEY_AUTO_ENCRYPTION])) {
            $autoEncryption = AutoEncryptionOptions::fromArray($options[self::KEY_AUTO_ENCRYPTION])->toArray();
        }
```

src/Model/DriverOptions.php
```diff
@@ -0,0 +1,147 @@
...
+        self::ensureValidArrayOptions($options);
+
+        $autoEncryption = match (isset($options[self::KEY_AUTO_ENCRYPTION])) {
+            true => AutoEncryptionOptions::fromArray($options[self::KEY_AUTO_ENCRYPTION])->toArray(),
+            false => []
+        };
+
+        return (new self(
+            typeMap: $options[self::KEY_TYPE_MAP],
+            builderEncoder: $options[self::KEY_BUILDER_ENCODER] ?? new BuilderEncoder(),
```
GromNaN, line 85: You don't validate that `builderEncoder` is an instance of `BuilderEncoder` before using it for a typed parameter. This will result in an unspecific `TypeError`.

src/Client.php
```diff
@@ -441,15 +409,15 @@ public function watch(array $pipeline = [], array $options = []): ChangeStream
...
+            $options['typeMap'] = $this->driverOptions->typeMap;
         }
 
         $operation = new Watch($this->manager, null, null, $pipeline, $options);
 
         return $operation->execute($server);
     }
 
-    private static function getVersion(): string
+    public static function getVersion(): string
```
GromNaN: We don't add methods to the public API unless it's necessary. You can move it to the new DriverOptions class and keep it private.

GromNaN (changes requested):
> This PR also extracts the options array into DriverOptions in order to simplify testing the I/O of options as we do quite a bit of transformation on them.

This sounds like a good idea, and this has no impact on users, since these classes are internal and cannot be accessed from public APIs.

But I question whether an undefined and a null option have the same meaning in this context.

src/ValueObject/DriverOptions.php
paulinevos: Good catch

src/Model/DriverOptions.php
paulinevos: I figured no, as if we get more specific we are tying it more to server implementation. I figured that as much as possible we want to pass options as is. In my mind it's just any options that we aren't transforming in some way.

src/Model/DriverOptions.php
paulinevos: I validate that it's an instance of `Encoder` in `ensureValidArrayOptions`. There's a test case for it too

paulinevos:
> But I question whether an undefined and a null option have the same meaning in this context.

There were some test cases re: this in `ClientTest` so I ensured that those passed and that that behavior didn't change. As far as other potential nullables, I'm not too sure...

src/Model/DriverOptions.php
```diff
@@ -0,0 +1,163 @@
...
+    public function toArray(): array
+    {
+        return array_filter(
+            [
+                'typeMap' => $this->typeMap,
+                'builderEncoder' => $this->builderEncoder,
+                'autoEncryption' => $this->autoEncryption,
+                'driver' => $this->driver,
+            ] + $this->miscOptions,
+            fn ($option) => $option !== null,
```
GromNaN:
```suggestion
            static fn ($option) => $option !== null,
```

src/Model/DriverOptions.php
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        return new self(
+            typeMap: $this->typeMap,
+            builderEncoder: $this->builderEncoder,
+            autoEncryption: $this->autoEncryption,
+            driver: $mergedDriver,
+            miscOptions: $this->miscOptions,
+        );
+    }
```
GromNaN: Creating a second instance of `DriverOptions` does not seem efficient to me, especially since this method is always called in the factory. A simple assignment  is enough.
```suggestion
        $this->driver = $mergedDriver;
    }
```

GromNaN (commented):
Tests are failing. Do you need help for the spec tests?

src/Model/DriverOptions.php
paulinevos: The thing is that then I'd have to change the `readonly` and visibility (which is fine, I can make it private I guess. Nothing is directly accessing it as of now)

GromNaN: If you don't use CPP for `$driver`, you can alter its value before setting the property.

src/Model/DriverOptions.php
```diff
@@ -0,0 +1,163 @@
...
+        public readonly Encoder $builderEncoder,
+        public readonly ?array $autoEncryption,
+        public readonly array $driver,
+        private readonly array $miscOptions,
+    ) {
+    }
+
+    public static function fromArray(array $options): self
+    {
+        $options += ['typeMap' => self::DEFAULT_TYPE_MAP];
```
alcaeus: Should this be `self::KEY_TYPE_MAP`?
```suggestion
        $options += ['self::KEY_TYPE_MAP' => self::DEFAULT_TYPE_MAP];
```

paulinevos:
> Tests are failing. Do you need help for the spec tests?

@GromNaN I might do actually. I can't run tests locally for 8.1 cause pie is throwing an error when I install the extension so I gotta dig into that I think.

src/Client.php
```diff
@@ -17,13 +17,10 @@
...
 namespace MongoDB;
 
-use Composer\InstalledVersions;
 use Iterator;
-use MongoDB\BSON\Document;
-use MongoDB\BSON\PackedArray;
-use MongoDB\Builder\BuilderEncoder;
 use MongoDB\Builder\Pipeline;
-use MongoDB\Codec\Encoder;
+use MongoDB\Builder\Stage;
```
GromNaN: I don't think this class is used in the `Client` class. PHPCS may be confused because you declare the `stage` type.

src/Client.php
```diff
@@ -38,36 +35,24 @@
...
 use MongoDB\Operation\Watch;
 use stdClass;
 use Stringable;
-use Throwable;
 
 use function array_diff_key;
-use function is_array;
-use function is_string;
 
+/** @psalm-type stage = StageInterface|array<string,mixed>|stdClass */
```
GromNaN:
```suggestion
/** @@psalm-import-type stage from Builder\Pipeline */
```

src/Client.php
paulinevos: I think this was actually my fault cause I typed it like that first before I found the psalm-type

GromNaN (approved):

PR #1810, opened 2025-12-17
PHPLIB-1688: Updating Release pipeline with SBOM generation steps

ekovalets:
This PR updated the mongo-php-library release pipeline with automatic sbom.json file generation on release, which then being added to the released branch.

### DIFF
--- a/.github/actions/update-sbom/action.yml
+++ b/.github/actions/update-sbom/action.yml
@@ -1,0 +1,31 @@
+name: Generate SBOM
+description: Generates CycloneDX SBOM using CycloneDX PHP Composer plugin
+inputs:
+  output-file:
+    description: "Output filename for the SBOM"
+    required: false
+    default: "sbom.json"
+runs:
+  using: composite
+  steps:
+    - name: Allow CycloneDX plugin
+      shell: bash
+      run: composer config allow-plugins.cyclonedx/cyclonedx-php-composer true
+    - name: Install CycloneDX plugin
+      shell: bash
+      run: composer require --dev cyclonedx/cyclonedx-php-composer --ignore-platform-reqs
+    - name: Generate SBOM
+      shell: bash
+      working-directory: ${{ inputs.working-directory }}
+      run: |
+        echo "Generating SBOM for 'php' project..."
+        composer CycloneDX:make-sbom --output-file=${{ inputs.output-file }} --output-format=json --spec-version=1.5
+    - name: Validate SBOM presence
+      shell: bash
+      run: |
+        if [ ! -f "${{ inputs.output-file }}" ]; then
+          echo "Error: SBOM file not found"
+          exit 1
+        fi
+        
+        echo "SBOM file validated: ${{ inputs.output-file }}"
--- a/.github/actions/update-sbom/action.yml
+++ b/.github/actions/update-sbom/action.yml
@@ -5,27 +5,95 @@ inputs:
     description: "Output filename for the SBOM"
     required: false
     default: "sbom.json"
+outputs:
+  HAS_CHANGES:
+    description: "Whether the SBOM has meaningful changes compared to the existing version"
+    value: ${{ steps.check_changes.outputs.HAS_CHANGES }}
 runs:
   using: composite
   steps:
     - name: Allow CycloneDX plugin
       shell: bash
       run: composer config allow-plugins.cyclonedx/cyclonedx-php-composer true
+
     - name: Install CycloneDX plugin
       shell: bash
       run: composer require --dev cyclonedx/cyclonedx-php-composer
+
     - name: Generate SBOM
+      id: generate-sbom
       shell: bash
-      working-directory: ${{ inputs.working-directory }}
       run: |
         echo "Generating SBOM for 'php' project..."
-        composer CycloneDX:make-sbom --output-file=${{ inputs.output-file }} --output-format=json --spec-version=1.5
+        composer CycloneDX:make-sbom --output-file=sbom-new.json --output-format=json --spec-version=1.5
+
     - name: Validate SBOM presence
       shell: bash
       run: |
-        if [ ! -f "${{ inputs.output-file }}" ]; then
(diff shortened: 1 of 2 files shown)


.github/actions/update-sbom/action.yml
```diff
@@ -0,0 +1,31 @@
...
+    default: "sbom.json"
+runs:
+  using: composite
+  steps:
+    - name: Allow CycloneDX plugin
+      shell: bash
+      run: composer config allow-plugins.cyclonedx/cyclonedx-php-composer true
+    - name: Install CycloneDX plugin
+      shell: bash
+      run: composer require --dev cyclonedx/cyclonedx-php-composer --ignore-platform-reqs
```
alcaeus: Is there a reason you're using `ignore-platform-reqs`? If we want to ignore a specific requirement, we should use `ignore-platform-req`, e.g. `--ignore-platform-req=php+` if we're using a newer version of PHP than is supported by the plugin.

.github/workflows/release.yml
```diff
@@ -48,12 +51,6 @@ jobs:
       - name: "Create release output"
         run: echo '🎬 Release process for version ${{ inputs.version }} started by @${{ github.triggering_actor }}' >> $GITHUB_STEP_SUMMARY
 
-      - name: "Generate token and checkout repository"
-        uses: mongodb-labs/drivers-github-tools/secure-checkout@v3
-        with:
-          app_id: ${{ vars.APP_ID }}
-          private_key: ${{ secrets.APP_PRIVATE_KEY }}
-
```
alcaeus, line 56: Why was this removed? The "Create and push new release branch" step pushes a branch in line 91, which requires an appropriate token if I'm not mistaken.

.github/workflows/release.yml
```diff
@@ -93,6 +90,84 @@ jobs:
...
+          php-version: ${{ env.PHP_VERSION }}
+          driver-version: ${{ env.DRIVER_VERSION }}
+          working-directory: '.'
+        continue-on-error: true
+
+      - name: "Generate/Update composer.lock"
+        id: composer-lock
+        run: |
+          echo "Resolving dependencies and generating composer.lock..."
+          composer update --no-install --ignore-platform-reqs
```
alcaeus: This step should never use `ignore-platform-reqs`. This could install dependencies that we don't actually want to install or lead to an invalid dependency chain.

.github/workflows/release.yml
```diff
@@ -93,6 +90,84 @@ jobs:
...
+      - name: "Commit SBOM changes"
+        if: steps.sbom_status.outputs.HAS_CHANGES == 'true'
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git add ${{ env.SBOM_FILE }}
+          git commit -m "chore: Update SBOM for release ${{ inputs.version }}"
+          git push
+          echo "📦 SBOM updated and committed" >> $GITHUB_STEP_SUMMARY
+        continue-on-error: true
```
alcaeus: Please use the `drivers-github-tools/setup` action to set git config: (https://github.com/mongodb/mongo-php-driver/blob/a2109ca2730584022ecba44dc7eab65e9cfe397a/.github/workflows/release.yml#L70-L75). The push should also happen with the correct credentials set through the secure-checkout action that you removed previously. Last but not least, a push to a stable branch will lead to a merge-up pull request; we should discuss how to handle this. For example, in PHPC we ensure that newer branches are up-to-date with older branches, so in this case we'll want to mark the branch as merged up using `strategy=ours` to avoid conflicts or overwriting a newer SBOM. Please see the corresponding step in PHPC: https://github.com/mongodb/mongo-php-driver/blob/a2109ca2730584022ecba44dc7eab65e9cfe397a/.github/workflows/release.yml#L118-L127

.github/workflows/release.yml
```diff
@@ -93,6 +90,84 @@ jobs:
...
+
+      - name: "Generate SBOM"
+        id: generate-sbom
+        if: steps.composer-lock.outcome == 'success'
+        uses: ./.github/actions/update-sbom
+        with:
+          php-version: ${{ env.PHP_VERSION }}
+          working-directory: '.'
+          output-file: ${{ env.SBOM_FILE }}
+          output-format: 'json'
```
alcaeus: The new action only defines an `output-file` input, so the others should not be needed.

.github/workflows/release.yml
```diff
@@ -93,6 +90,84 @@ jobs:
           git checkout -b ${RELEASE_BRANCH}
           git push origin ${RELEASE_BRANCH}
 
+      #
+      # Preliminary checks done - generate SBOM before tagging
+      #
+      - name: Checkout repository (Base Branch)
+        uses: actions/checkout@v5
+        with:
+          ref: ${{ github.event.pull_request.base.ref || github.ref }}
```
alcaeus: Since this workflow is only called manually on a branch, `github.event.pull_request.base.ref` will never be defined, so this should be changed. Ideally, we should rely on the secure-checkout action that we used in this workflow.

alcaeus (changes requested):

.github/actions/update-sbom/action.yml
ekovalets: removed

.github/workflows/release.yml
ekovalets: this was moved to later stages, but now I moved it up - before sbom gen.

.github/workflows/release.yml
ekovalets: replaced whole with secure checkout

.github/workflows/release.yml
ekovalets: removed

.github/workflows/release.yml
ekovalets: remove unnecessary args

GromNaN:
The project is on hold. Create a new pull request if the project restarts.

PR #1829, opened 2026-01-26
PHPLIB-1770 The unknown command name is not quoted in 5.0.32-enterprise

GromNaN:
Fix PHPLIB-1770

The error message does not contain the quotes when running in evergreen.

### DIFF
--- a/src/Exception/SearchNotSupportedException.php
+++ b/src/Exception/SearchNotSupportedException.php
@@ -36,8 +36,11 @@ final class SearchNotSupportedException extends ServerException
             // MongoDB 4 to 6, 7-community
             59 => match ($exception->getMessage()) {
                 'no such command: \'createSearchIndexes\'' => true,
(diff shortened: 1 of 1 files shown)


src/Exception/SearchNotSupportedException.php
```diff
@@ -36,8 +36,11 @@ public static function isSearchNotSupportedError(Throwable $exception): bool
             // MongoDB 4 to 6, 7-community
             59 => match ($exception->getMessage()) {
                 'no such command: \'createSearchIndexes\'' => true,
+                'no such command: createSearchIndexes' => true,
```
jmikola, line 39: LGTM, and given this is 5.0.x I doubt the inconsistency is worth raising to the server team. I'd suggest adding a comment somewhere here to explain why these `match` cases are being duplicated, though.

jmikola (approved):

PR #1831, opened 2026-02-02
Skip large encryption tests on mongocryptd

GromNaN:
Encryption tests sending large payloads fail on mongocryptd 8.2.4, 8.0.18, and 7.0.29 due to SERVER-118428

To be reverted by DRIVERS-3382

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -20,6 +20,9 @@ return RectorConfig::configure()
     ])
     ->withPhpSets(php74: true)
     ->withComposerBased(phpunit: true)
+    // Error with StaticCallOnNonStaticToInstanceCallRector
+    // https://github.com/rectorphp/rector/issues/9608
+    ->withSkipPath(__DIR__ . '/tests/Builder/BuilderEncoderTest.php')
     ->withRules([
         ChangeSwitchToMatchRector::class,
     ])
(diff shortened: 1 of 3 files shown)


jmikola (approved):

tests/SpecTests/ClientSideEncryptionSpecTest.php
```diff
@@ -566,6 +566,8 @@ static function (self $test, Collection $collection, array $document): void {
     #[DataProvider('provideBSONSizeLimitsAndBatchSplittingTests')]
     public function testBSONSizeLimitsAndBatchSplitting(Closure $test): void
     {
+        $this->markTestSkipped('Encryption tests sending large payloads fail on some mongocryptd versions (See DRIVERS-3382)');
+
```
jmikola, line 570: [DRIVERS-3382](https://jira.mongodb.org/browse/DRIVERS-3382) noted that the issue only affects server versions since 8.2.4, 8.0.18, and 7.0.29.

If it wasn't intended to skip earlier server versions (e.g. 6.0), you could gate this behind a server version check. But no objection to keeping it as-is.

rector.php
```diff
@@ -20,6 +20,9 @@
     ])
     ->withPhpSets(php74: true)
     ->withComposerBased(phpunit: true)
+    // Error with StaticCallOnNonStaticToInstanceCallRector
+    // https://github.com/rectorphp/rector/issues/9608
+    ->withSkipPath(__DIR__ . '/tests/Builder/BuilderEncoderTest.php')
     ->withRules([
```
GromNaN, line 26: Applied the same patch as https://github.com/mongodb/mongo-php-library/commit/3b4d0289ef6c85531862ba4f888601bbee637b4b#diff-d933a2ad57daa43419b483d3f1ddedaec7aa44933007fead9f9437390f596f4a since the issue is not resolved.

PR #1834, opened 2026-02-03
PHPLIB-1698: Remove reference to test-atlas-data-lake task

jmikola:
https://jira.mongodb.org/browse/PHPLIB-1698

This was addressed in the original PR (#1778) but missed when merging to v2.1 and up.

I noted the reference is still in the [v2.0](https://github.com/mongodb/mongo-php-library/blob/v2.x/.evergreen/config/templates/test-variant/phpc.yml) branch, but I assume that's intentional as it was already unmaintained at the time this change was implemented (and should no longer be tested in Evergreen).


GromNaN (approved):
Good catch

PR #1835, opened 2026-02-04
PHPLIB-1647: Add `Database::getGridFSBucket()`

alcaeus:
PHPLIB-1647

This applies the same renaming to `selectGridFSBucket` as we've already done with `selectCollection` and `Client::selectDatabase` in PHPLIB-1185. Note that we're not deprecating the old method yet -- this will be done in a future release as part of PHPLIB-1553.

### DIFF
--- a/src/Database.php
+++ b/src/Database.php
@@ -423,6 +423,25 @@ class Database implements Stringable
     }
 
     /**
+     * Returns a GridFS bucket instance.
+     *
+     * @see Bucket::__construct() for supported options
+     * @param array $options Bucket constructor options
+     * @throws InvalidArgumentException for parameter/option parsing errors
+     */
+    public function getGridFSBucket(array $options = []): Bucket
+    {
+        $options += [
+            'readConcern' => $this->readConcern,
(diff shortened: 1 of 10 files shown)


src/Database.php
```diff
@@ -576,14 +595,7 @@ public function selectCollection(string $collectionName, array $options = []): C
      */
```
GromNaN, line 595: A depreciation annotation should be added. There is no urgent need to remove the old method, but having two identical methods creates confusion about which one to use.

alcaeus: I applied the same principle we applied with `selectDatabase` and `selectCollection`, for which there's PHPLIB-1553 to deprecate them.
We can schedule that ticket as well -- only reason I picked this one up is because it was already in the 2.2.0 fixVersion and I figured it would be a quick change to make before the release. We can deprecate all the methods in 2.3.0, as that will involve also replacing usages of `selectDatabase` and `selectCollection` in our tests.

GromNaN: Ok then...

GromNaN (approved):

PR #1846, opened 2026-02-18
PHPLIB-1714: implement backoff for withTransaction

alcaeus:
PHPLIB-1714

This implements the backoff algorithm as described in the convenient transaction API. The PR is best reviewed commit-by-commit:
- The first commit refactors the `withTransaction` class to extract the error checking and commit logic into smaller methods that will throw if the transaction loop is to be aborted or return if it is to be retried
- The second commit implements the backoff logic
- The third commit implements the prose test for the backoff logic, including a change in `withTransaction` itself to allow injecting a fixed jitter for testing.

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -756,10 +756,12 @@
   </file>
   <file src="src/Operation/WithTransaction.php">
     <InvalidOperand>
+      <code><![CDATA[$backoffMs * 1e6]]></code>
       <code><![CDATA[$this->getJitter() * min(self::BACKOFF_INITIAL * (1.5 ** ($transactionAttempt - 1)), self::BACKOFF_MAX)]]></code>
       <code><![CDATA[1.5 ** ($transactionAttempt - 1)]]></code>
+      <code><![CDATA[hrtime(true) + ($backoffMs * 1e6)]]></code>
       <code><![CDATA[self::BACKOFF_INITIAL * (1.5 ** ($transactionAttempt - 1))]]></code>
-      <code><![CDATA[time() + ($backoffMs / 1000)]]></code>
+      <code><![CDATA[self::MAX_TIME * 1e9]]></code>
     </InvalidOperand>
   </file>
   <file src="src/UpdateResult.php">
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="6.14.2@bbd217fc98c0daa0a13aea2a7f119d03ba3fc9a0">
+<files psalm-version="6.14.3@d0b040a91f280f071c1abcb1b77ce3822058725a">
   <file src="examples/atlas_search.php">
     <MixedArgument>
       <code><![CDATA[$document['name']]]></code>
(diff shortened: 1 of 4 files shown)


src/Operation/WithTransaction.php
```diff
@@ -52,73 +74,158 @@ public function __construct(callable $callback, private array $transactionOption
...
      *
-     * @param int $startTime The time the transaction was started
+     * @param int $startTime The time the transaction loop was started
+     * @param int $backoffMs The amount of time that will be spent backing off before the next retry attempt, in milliseconds
      */
-    private function isTransactionTimeLimitExceeded(int $startTime): bool
+    private function isTransactionTimeLimitExceeded(int $startTime, int $backoffMs = 0): bool
     {
-        return time() - $startTime >= 120;
+        return time() + ($backoffMs / 1000) - $startTime >= self::MAX_TIME;
```
GromNaN: The monotonic `hrtime` should be used, in case the computer time is updated during the transaction.

psalm-baseline.xml
```diff
@@ -760,6 +754,14 @@
       <code><![CDATA[$this->changeStreamOptions['startAfter']]]></code>
     </MixedReturnStatement>
   </file>
+  <file src="src/Operation/WithTransaction.php">
+    <InvalidOperand>
+      <code><![CDATA[$this->getJitter() * min(self::BACKOFF_INITIAL * (1.5 ** ($transactionAttempt - 1)), self::BACKOFF_MAX)]]></code>
+      <code><![CDATA[1.5 ** ($transactionAttempt - 1)]]></code>
+      <code><![CDATA[self::BACKOFF_INITIAL * (1.5 ** ($transactionAttempt - 1))]]></code>
+      <code><![CDATA[time() + ($backoffMs / 1000)]]></code>
```
GromNaN: Haven't figured out how to resolve these issues?

> Cannot process ints and floats in strict binary operands mode, please cast explicitly

src/Operation/WithTransaction.php
```diff
@@ -52,73 +74,158 @@ public function __construct(callable $callback, private array $transactionOption
...
-                }
+            // Commit was successful, indicate to break out of the transaction loop
+            return true;
+        }
+    }
 
-                // Commit was successful
-                break;
-            }
+    private function computeBackoffMs(int $transactionAttempt): int
```
GromNaN, line 204: It would make sense to unit test this method.

psalm-baseline.xml
alcaeus: That's not the issue. The issue is that psalm's strict operand mode is perfectly find and useful, except when it comes to mixing ints and floats.  Especially in this case, I'd rather just not deal with adding various type casts that aren't necessary in the first place, as operations like `<float> ** <int>` is perfectly safe.

src/Operation/WithTransaction.php
alcaeus: I can, especially since we already allow injecting a closure that overrides the random jitter behaviour. I'll access the method through reflection for testing.

src/Operation/WithTransaction.php
alcaeus: Updated

GromNaN (approved):

PR #1851, opened 2026-02-26
PHPLIB-1798 Extract MQL definition into a submodule for builder generator

GromNaN:
> [!NOTE]  
> This PR stay in draft while we work on the instable [mongodb/mql-specifications](https://github.com/mongodb/mql-specifications) project.


tests/Collection/BuilderCollectionFunctionalTest.php
```diff
@@ -83,8 +85,7 @@ public function testBulkWriteUpdateMany(): void
             [
                 'updateMany' => [
                     Query::query(x: Query::gt(1)),
-                    // @todo Use Builder when update operators are supported by PHPLIB-1507
-                    ['$set' => ['x' => 3]],
+                    new Update(Update::set(x: 3)),
```
GromNaN: A single update operator should be accepted for brevity.
```suggestion
                    Update::set(x: 3),
```

src/Builder/Encoder/UpdateEncoder.php
```diff
@@ -0,0 +1,68 @@
...
+    /** @return array<string, mixed> */
+    public function encode(mixed $value): array
+    {
+        if (! $this->canEncode($value)) {
+            throw UnsupportedValueException::invalidEncodableValue($value);
+        }
+
+        $encoded = [];
+        foreach ($value->update as $operator) {
+            assert($operator instanceof UpdateInterface);
```
GromNaN: This `assert` should not be necessary. `Update::$update` type is `UpdateInterface[]`

tests/Builder/Update/AddToSetOperatorTest.php
```diff
@@ -0,0 +1,35 @@
...
+    public function testAddToArray(): void
+    {
+        $update = Update::addToSet(tags: 'accessories');
+
+        $this->assertSameUpdate(Pipelines::AddToSetAddToArray, $update);
+    }
+
+    public function testUseEachModifier(): void
+    {
+        $update = Update::addToSet(tags: ['$each' => ['camera', 'electronics', 'accessories']]);
```
GromNaN, line 25: This test should use a builder for `$each`

tests/Builder/Update/CurrentDateOperatorTest.php
```diff
@@ -0,0 +1,31 @@
...
+            Update::set(...['cancellation.reason' => 'user request'], status: 'D'),
+        );
+        /*
+        $update = new Update(
+            Update::currentDate('lastModified'),
+            Update::currentDate('cancellation.date', type: 'timestamp'),
+            Update::set('cancellation.reason', 'user request'),
+            Update::set('status', 'D'),
+        );
+        */
```
GromNaN: This is a different API design that would make sense for update operators.
It moves the field path as an argument instead of an array key. That enables using a FieldPath classes.

tests/Builder/Update/Pipelines.php
GromNaN, line 1: We should rename this enum to `Expected` for all types, as it's not pipeline.

src/Builder/Encoder/UpdateEncoder.php
alcaeus: I can also add the error to the baseline, but for some reason psalm was complaining.

tests/Builder/Update/AddToSetOperatorTest.php
alcaeus: There is no builder for `$each` at this time. We can add it, but we'll have to introduce a new expression type to allow using it in the update operators that accept it.

tests/Builder/Update/Pipelines.php
alcaeus: Would it make sense to move the expected documents into the tests they're used in? That would make it a little easier to find them.

GromNaN: As constant on the test class, yes.
Initially I used a distinct enum to simplify code generation. But I'm now confident enough to update the generated test case.

GromNaN: To be done in a distinct PR of course.

src/Builder/Encoder/UpdateEncoder.php
alcaeus: Removed, turns out if wasn't necessary after all 🤷

tests/Builder/Update/CurrentDateOperatorTest.php
alcaeus: I removed this example, we can always return to this later.

tests/Builder/Update/AddToSetOperatorTest.php
alcaeus: Created PHPLIB-1807 to track adding `Update::each()`

tests/Builder/Update/Pipelines.php
alcaeus: See PHPLIB-1808.

tests/Builder/Update/CurrentDateOperatorTest.php
GromNaN: This is a design decision that we can't change later.
I feel the current API not usable as you have to use `...['field.path' => ]` everywhere.

alcaeus: I pushed a new commit that allows for this kind of API for `$currentDate`, `$set`, and `$setOnInsert`. I'll double-check to see where else we may want to leverage this kind of API.

tests/Builder/Update/IncOperatorTest.php
```diff
@@ -0,0 +1,21 @@
...
+use MongoDB\Tests\Builder\UpdateTestCase;
+
+/**
+ * Test $inc update
+ */
+class IncOperatorTest extends UpdateTestCase
+{
+    public function testIncrementFields(): void
+    {
+        $update = Update::inc(...['quantity' => -2, 'metrics.orders' => 1]);
```
GromNaN, line 17: Using variadics does not make sense.
```suggestion
        $update = Update::inc(['quantity' => -2, 'metrics.orders' => 1]);
```

src/Builder/Update.php
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    /**
+     * Sets the value of a field.
+     *
+     * @see https://www.mongodb.com/docs/manual/reference/operator/update/set/
+     */
+    public static function set(string|bool|int|stdClass|array|DateTimeInterface|Type|float|null ...$field): SetOperator
+    {
+        if (count($field) === 2 && array_is_list($field) && is_string($field[0])) {
```
GromNaN: `$field[0]` should also accept a `FieldPathInterface`.

tests/Builder/Update/CurrentDateOperatorTest.php
GromNaN: Thanks, I'm still concerned by keeping the variadic API.

alcaeus: Created PHPLIB-1809 to track the separate API, so we can discuss potential issues with offering two APIs, as well as potentially applying the same change to pipeline builders.

tests/Builder/Update/IncOperatorTest.php
alcaeus: I'd leave this as is for consistency with other builders, and handle this as part of PHPLIB-1809.

src/Builder/Update.php
alcaeus: This API was removed, but we discussed that any field names should always be treated as unprefixed field paths. Not sure how we would handle this, as we're technically accepting `array<UnprefixedFieldPath|string, mixed>` and would ideally want to represent this in the MQL definitions.

alcaeus (approved):

PR #1885, opened 2026-04-29
PHPLIB-1624 Add `$rankFusion` stage and `$sigmoid` expression

GromNaN:
Fix PHPLIB-1624

- [x] Requires https://github.com/mongodb/mql-specifications/pull/32

### DIFF
--- a/tests/Builder/Stage/RankFusionStageTest.php
+++ b/tests/Builder/Stage/RankFusionStageTest.php
@@ -1,0 +1,54 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MongoDB\Tests\Builder\Stage;
+
+use MongoDB\Builder\Pipeline;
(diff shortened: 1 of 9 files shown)


tests/Builder/Stage/RankFusionStageTest.php
```diff
@@ -0,0 +1,56 @@
...
+                            )
+                        )
+                    )
+                ),
+                combination: object(
+                    weights: object(
+                        searchPlot: 0.6,
+                        searchGenre: 0.4,
+                    )
+                ),
```
GromNaN, line 44: Usage of `object` is breaking the type system.

alcaeus: I presume this would be fixed by #1886?

paulinevos (approved):

PR #1893, opened 2026-05-06
Add submodule validation workflow

GromNaN:
Validates submodule references on every PR targeting `v*.*` or `feature/*` branches.

## Checks

1. **Commit is on the upstream branch** — the referenced commit must be an ancestor of the configured branch (or the remote's default branch if not set in `.gitmodules`)
2. **No regression** — the commit must not be older than what the target branch already points to

- Requires https://github.com/mongodb-labs/drivers-github-tools/pull/103


GromNaN:
Interestingly, when the submodule is updated on the target branch, the new action fails because it's behind. Not a big deal because this action is only run when a submodule is updated, and the solution is just to rebase the PR.

.github/workflows/submodules.yml
```diff
@@ -0,0 +1,33 @@
...
+      - ".gitmodules"
+      - ".github/workflows/submodules.yml"
+      - "tests/drivers-evergreen-tools"
+      - "tests/specifications"
+      - "generator/mql-specifications"
+  pull_request:
+    branches:
+      - "v*.*"
+      - "feature/*"
+    paths:
```
alcaeus: Limiting this to paths would prevent us from making this workflow as required.

alcaeus (commented):
Workflow LGTM. We should consider removing the paths restriction to be able to block pull requests on inconsistent submodules.

alcaeus (approved):

PR #1894, opened 2026-05-06
DRIVERS-3088: Support new search operators and argument

paulinevos:
Part of PHPLIB-1738 [DRIVERS-3088](https://jira.mongodb.org/browse/DRIVERS-3088)

Depends on: https://github.com/mongodb/mql-specifications/pull/37

### DIFF
--- a/tests/Builder/Search/HasRootOperatorTest.php
+++ b/tests/Builder/Search/HasRootOperatorTest.php
@@ -1,0 +1,87 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MongoDB\Tests\Builder\Search;
+
+use MongoDB\Builder\Pipeline;
+use MongoDB\Builder\Search;
+use MongoDB\Builder\Stage;
+use MongoDB\Tests\Builder\PipelineTestCase;
+
+use function MongoDB\object;
+
+/**
+ * Test hasRoot search
+ */
+class HasRootOperatorTest extends PipelineTestCase
+{
+    public function testCompoundQuery(): void
+    {
+        $pipeline = new Pipeline(
+            Stage::search(
+                Search::compound(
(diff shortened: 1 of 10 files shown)


paulinevos:
Diff check will be resolved after merging the other thingamabob

tests/Builder/Search/HasRootOperatorTest.php
```diff
@@ -0,0 +1,85 @@
...
+                        Search::hasRoot(
+                            Search::wildcard(
+                                path: 'description',
+                                query: '*network*',
+                                allowAnalyzedField: true,
+                            ),
+                        ),
+                    ],
+                ),
+                returnScope: ['path' => 'funding_rounds'],
```
alcaeus: Nit: one of these tests could use the `object` function to ensure we're also accepting object instances for the parameter:
```suggestion
                returnScope: object(path: 'funding_rounds'),
```

alcaeus (commented):
Code and tests LGTM with a small nit. Waiting for mql-spec submodule update before approving.

tests/specifications
GromNaN, line 1: This submodule update is unrelated.

tests/Builder/Search/HasRootOperatorTest.php
GromNaN: @paulinevos you resolved this thread but did not apply the requested change.

GromNaN (changes requested):

tests/Builder/Search/HasRootOperatorTest.php
paulinevos: Something went wrong there with pushing I think

GromNaN (approved):

PR #1895, opened 2026-05-08
Fix flaky session test

paulinevos:
Cluster time advancement is needed for snapshot sessions, since initialData uses an internal client that does not gossip its cluster time to test entities. Without advancement, a snapshot session may establish its atClusterTime before the initial data is visible, causing snapshot reads to return no results.

### DIFF
--- a/tests/UnifiedSpecTests/UnifiedTestRunner.php
+++ b/tests/UnifiedSpecTests/UnifiedTestRunner.php
@@ -457,7 +457,7 @@ final class UnifiedTestRunner
         foreach ($createEntities ?? [] as $entity) {
             $session = $entity->session ?? null;
             if ($session === null) {
-                return false;
+                continue;
             }
 
             if (($session->sessionOptions?->snapshot ?? false) === true) {
--- a/tests/UnifiedSpecTests/UnifiedTestRunner.php
+++ b/tests/UnifiedSpecTests/UnifiedTestRunner.php
@@ -189,7 +189,7 @@ final class UnifiedTestRunner
         $context = $this->createContext();
 
         if (isset($initialData)) {
-            $this->prepareInitialData($initialData, $context, $this->isAdvanceClusterTimeNeeded($test->operations));
+            $this->prepareInitialData($initialData, $context, $this->isAdvanceClusterTimeNeeded($test->operations, $createEntities));
         }
 
         /* If an EntityMap observer has been configured, assign the Context's
@@ -435,8 +435,12 @@ final class UnifiedTestRunner
 
     /**
      * Work around potential MigrationConflict errors on sharded clusters.
+     *
+     * Cluster time advancement is also needed for snapshot sessions, since initialData uses an internal client that
+     * does not gossip its cluster time to test entities. Without advancement, a snapshot session may establish its
+     * atClusterTime before the initial data is visible, causing snapshot reads to return no results.
      */
-    private function isAdvanceClusterTimeNeeded(array $operations): bool
+    private function isAdvanceClusterTimeNeeded(array $operations, ?array $createEntities = null): bool
     {
         if (! in_array($this->getPrimaryServer()->getType(), [Server::TYPE_MONGOS, Server::TYPE_LOAD_BALANCER], true)) {
             return false;
@@ -447,6 +451,17 @@ final class UnifiedTestRunner
                 case 'startTransaction':
                 case 'withTransaction':
                     return true;
+            }
+        }
+
+        foreach ($createEntities ?? [] as $entity) {
+            $session = $entity->session ?? null;
+            if ($session === null) {
+                return false;
+            }
+
+            if (($session->sessionOptions?->snapshot ?? false) === true) {
+                return true;
             }
         }
 


tests/UnifiedSpecTests/UnifiedTestRunner.php
```diff
@@ -435,8 +435,12 @@ private function prepareInitialData(array $initialData, Context $context, bool $
...
+     *
+     * Cluster time advancement is also needed for snapshot sessions, since initialData uses an internal client that
+     * does not gossip its cluster time to test entities. Without advancement, a snapshot session may establish its
+     * atClusterTime before the initial data is visible, causing snapshot reads to return no results.
      */
-    private function isAdvanceClusterTimeNeeded(array $operations): bool
+    private function isAdvanceClusterTimeNeeded(array $operations, ?array $createEntities = null): bool
     {
         if (! in_array($this->getPrimaryServer()->getType(), [Server::TYPE_MONGOS, Server::TYPE_LOAD_BALANCER], true)) {
             return false;
```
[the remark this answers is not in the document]
paulinevos: OOS.

tests/UnifiedSpecTests/UnifiedTestRunner.php
```diff
@@ -450,6 +454,12 @@ private function isAdvanceClusterTimeNeeded(array $operations): bool
             }
         }
 
+        foreach ($createEntities ?? [] as $entity) {
+            if (($entity->session?->sessionOptions?->snapshot ?? false) === true) {
```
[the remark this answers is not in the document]
paulinevos: Please check this again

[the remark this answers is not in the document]
paulinevos: Should be fine. Warnings don't fail the tests

GromNaN:
The test is still [failing](https://github.com/mongodb/mongo-php-library/actions/runs/25561330404/job/75033412615?pr=1895) 😢 
```
1) MongoDB\Tests\UnifiedSpecTests\UnifiedSpecTest::testSessions with data set "sessions/snapshot-sessions: Find operation with snapshot" (MongoDB\Tests\UnifiedSpecTests\UnifiedTestCase Object (...))
Failed asserting that expected value matches actual value.
$actual count is 0, expected 1
--- Expected
+++ Actual
@@ @@
-MongoDB\Model\BSONArray Object #150220 (
-    0 => MongoDB\Model\BSONDocument Object #150203 (
-        '_id' => 1,
-        'x' => 11,
-    ),
-)
+MongoDB\Model\BSONArray Object #150202 ()
```

tests/UnifiedSpecTests/UnifiedTestRunner.php
```diff
@@ -450,6 +454,17 @@ private function isAdvanceClusterTimeNeeded(array $operations): bool
             }
         }
 
+        foreach ($createEntities ?? [] as $entity) {
+            $session = $entity->session ?? null;
+            if ($session === null) {
+                return false;
```
alcaeus: This should be a `continue` -- there could still be an entity with a snapshot session later in the list that might necessitate a cluster time advancement.

tests/UnifiedSpecTests/UnifiedTestRunner.php
```diff
@@ -435,8 +435,12 @@ private function prepareInitialData(array $initialData, Context $context, bool $
 
     /**
      * Work around potential MigrationConflict errors on sharded clusters.
+     *
+     * Cluster time advancement is also needed for snapshot sessions, since initialData uses an internal client that
+     * does not gossip its cluster time to test entities. Without advancement, a snapshot session may establish its
+     * atClusterTime before the initial data is visible, causing snapshot reads to return no results.
      */
-    private function isAdvanceClusterTimeNeeded(array $operations): bool
+    private function isAdvanceClusterTimeNeeded(array $operations, ?array $createEntities = null): bool
```
alcaeus, line 443: I suspect the nullable array is because we have tests without entities, in which case we get `null`? Would it make sense to normalise that to an empty array before we pass it around?

tests/UnifiedSpecTests/UnifiedTestRunner.php
GromNaN: This is the reason the test is still failing after this PR.

The `createEntities` array in `snapshot-sessions.json` starts with `client`, `database`, and `collection` entries before the `session` entries:
```
[{client: {id: "client0", ...}}, {database: {...}}, {collection: {...}}, {session: {snapshot: true}}, ...]
```

When the loop reaches the first entity (`{client: {...}}`), `$entity->session` is `null` (the entity has no `session` key), so it hits `return false` and exits immediately — never reaching the actual snapshot sessions.

This should be `continue` (as @alcaeus noted) so the loop keeps checking remaining entities.

GromNaN: Fixed, thanks!

GromNaN:
I've also changed the target branch to `v2.3`.

GromNaN (approved):

PR #1916, opened 2026-06-05
PHPLIB-1844: Support range in PHP doc types

paulinevos:
Fix PHPLIB-1844

### DIFF
--- a/generator/src/Definition/ArgumentDefinition.php
+++ b/generator/src/Definition/ArgumentDefinition.php
@@ -35,6 +35,8 @@ final class ArgumentDefinition
         public string|null $minVersion = null,
         public int|null $minItems = null,
         public int|null $maxItems = null,
+        public int|null $valueMin = null,
+        public int|null $valueMax = null,
         mixed ...$ignoredOtherArgs,
     ) {
         assert($this->optional === false || $this->default === null, 'Optional arguments cannot have a default value');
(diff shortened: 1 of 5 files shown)


GromNaN (approved):
Perfect!

generator/src/OperatorGenerator.php
```diff
@@ -135,6 +136,15 @@ final protected function getAcceptedTypes(ArgumentDefinition $arg): stdClass
             $nativeTypes = ['mixed'];
         }
 
+        if ($arg->valueMin !== null || $arg->valueMax !== null) {
+            $min = $arg->valueMin !== null ? (string) $arg->valueMin : 'min';
+            $max = $arg->valueMax !== null ? (string) $arg->valueMax : 'max';
```
GromNaN: Nitpick: I find this easier to read:
```suggestion
            $min = (string) ($arg->valueMin ?? 'min');
            $max = (string) ($arg->valueMax ?? 'max');
```

generator/src/Definition/ArgumentDefinition.php
```diff
@@ -35,6 +35,8 @@ public function __construct(
         public string|null $minVersion = null,
         public int|null $minItems = null,
         public int|null $maxItems = null,
+        public int|null $valueMin = null,
+        public int|null $valueMax = null,
```
GromNaN, line 39: We can validate `$valueMin <= $valueMax` when loading the config.

generator/src/OperatorGenerator.php
paulinevos: agree

generator/src/Definition/ArgumentDefinition.php
paulinevos: You just mean here in the constructor, right? Seems to be the most logical place to me seeing as we don't have validators and such before this stage

GromNaN: Yes, perfect.

GromNaN (approved):

PR #1926, opened 2026-06-15
PHPLIB-1789: Add autoEmbed type in VectorSearchIndexShape

GromNaN:
Adds `type: 'autoEmbed'` to the `VectorSearchIndexShape` Psalm type in `Collection.php`, with the following fields:

- Required: `type`, `modality` (`'text'`), `path`, `model` (string)
- Optional: `numDimensions`, `quantization` (`'float'|'scalar'|'binary'|'binaryNoRescore'`), `similarity`, `indexingMethod`, `hnswOptions`

The `model` field is typed as `string` rather than a string enum to avoid having to update the type every time a new model is released.

Two Psalm type tests added in `SearchIndexShapes.php`.

Jira: https://jira.mongodb.org/browse/PHPLIB-1789

### DIFF
--- a/src/Collection.php
+++ b/src/Collection.php
@@ -109,6 +109,7 @@ use function strlen;
  * @psalm-type VectorSearchIndexShape = array{
  *     fields: list<
  *         array{type: 'vector', path: string, numDimensions: int, similarity: 'euclidean'|'cosine'|'dotProduct', quantization?: 'none'|'scalar'|'binary', indexingMethod?: 'flat'|'hnsw', hnswOptions?: array{maxEdges?: int, numEdgeCandidates?: int}} |
+ *         array{type: 'autoEmbed', modality: 'text', path: string, model: string, numDimensions?: int, quantization?: 'float'|'scalar'|'binary'|'binaryNoRescore', similarity?: 'euclidean'|'cosine'|'dotProduct', indexingMethod?: 'flat'|'hnsw', hnswOptions?: array{maxEdges?: int, numEdgeCandidates?: int}} |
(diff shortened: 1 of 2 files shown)


src/Collection.php
```diff
@@ -109,6 +109,7 @@
  * @psalm-type VectorSearchIndexShape = array{
  *     fields: list<
  *         array{type: 'vector', path: string, numDimensions: int, similarity: 'euclidean'|'cosine'|'dotProduct', quantization?: 'none'|'scalar'|'binary', indexingMethod?: 'flat'|'hnsw', hnswOptions?: array{maxEdges?: int, numEdgeCandidates?: int}} |
+ *         array{type: 'autoEmbed', modality: 'text', path: string, model: string, numDimensions?: int, quantization?: 'float'|'scalar'|'binary'|'binaryNoRescore', similarity?: 'euclidean'|'cosine'|'dotProduct', indexingMethod?: 'flat'|'hnsw', hnswOptions?: array{maxEdges?: int, numEdgeCandidates?: int}} |
```
paulinevos, line 112: Can we multiline these types of shapes with a lot of different fields? Would make it easier to read the diff

paulinevos (approved):

src/Collection.php
GromNaN: I have mixed feelings about it, because it also makes the docblock very, very long.

paulinevos: True, but now it's also long, but horizontally, and hard to read 😅 But I see what you mean. It's not aesthetically pleasing but will make it easier to see what's changed.

PR #1959, opened 2026-08-18
Reduce round trips in GridFS delete and chunk lookups

GromNaN:
`Bucket::delete()` issued a find command on the files collection only to check whether the file existed, before the delete commands on the `files` and `chunks` collections. This uses the deleted count reported by the files collection delete instead, removing that extra round trip.

`deleteFileAndChunksById()` now returns null when the write is unacknowledged, since a deleted count is not available in that case, and `delete()` only raises `FileNotFoundException` when the count is known to be `0`. This preserves the existing behavior for unacknowledged write concerns.

This also skips the `n >= 0` filter when reading chunks from the start of a file, since that clause never excludes any document and was sent on every read.

### DIFF
--- a/src/GridFS/Bucket.php
+++ b/src/GridFS/Bucket.php
@@ -210,10 +210,9 @@ class Bucket
      */
     public function delete(mixed $id): void
     {
-        $file = $this->collectionWrapper->findFileById($id);
-        $this->collectionWrapper->deleteFileAndChunksById($id);
+        $deletedCount = $this->collectionWrapper->deleteFileAndChunksById($id);
 
-        if ($file === null) {
+        if ($deletedCount === 0) {
             throw FileNotFoundException::byId($id, $this->getFilesNamespace());
         }
     }
--- a/src/GridFS/CollectionWrapper.php
+++ b/src/GridFS/CollectionWrapper.php
@@ -95,11 +95,16 @@ final class CollectionWrapper
 
     /**
      * Deletes a GridFS file and related chunks by ID.
+     *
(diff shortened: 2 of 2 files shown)


paulinevos (approved):

GromNaN:
Compared this implementation against the Node.js, Python, and Java drivers' GridFS `delete` by id.

| Driver | Not found detection | Order | Orphaned chunks cleanup |
|---|---|---|---|
| PHP (this PR) | `deletedCount` from files delete | delete files, then chunks | Always deleted |
| Node.js | `deletedCount` from `deleteOne` | delete files, then chunks | Always deleted (explicit comment: "Delete orphaned chunks before returning FileNotFound") |
| Python (PyMongo) | `deleted_count` from `delete_one` | delete files, then chunks | Always deleted |
| Java (sync) | `DeleteResult.getDeletedCount()` | delete files, then chunks | Always deleted |
| Java (reactive) | same | delete files, check count, then chunks | Not deleted if count is 0 (stops early) |

This change aligns the PHP driver with the majority pattern used by Node.js, Python, and the Java sync driver: delete the files document first, then the chunks, and detect a missing file from the delete result's count instead of a separate find. The Java reactive streams driver is the outlier, it skips the orphaned chunks cleanup when the file document is not found, which looks like an inconsistency within the Java driver itself rather than a pattern worth following.

GromNaN:
This change is compliant with [DRIVERS-3618](https://jira.mongodb.org/browse/DRIVERS-3618)

PR #1960, opened 2026-08-18
Require ext-mongodb 2.4.0 for PHPLIB 2.4.0

paulinevos:
Bump the extension constraint and update the Evergreen build tasks: the "lowest" task now compiles the just-released 2.4.0, and "next-stable" tracks the new v2.4 branch.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
     ],
     "require": {
         "php": "^8.1",
-        "ext-mongodb": "^2.3",
+        "ext-mongodb": "^2.4",
         "composer-runtime-api": "^2.0",
         "psr/log": "^1.1.4|^2|^3",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -11,7 +11,7 @@
     ],
     "require": {
         "php": "^8.1",
-        "ext-mongodb": "^2.3",
+        "ext-mongodb": "^2.4",
```
GromNaN, line 14: Noted this update is triggered by the [releasing instructions](https://github.com/mongodb/mongo-php-library/blob/v2.x/RELEASING.md#update-extension-requirement-new-minor-versions-only).

Even if there is nothing in this release that require an upgrade of the extension. I'm fine keeping the driver and library versions in sync.

GromNaN (approved):

PR #1966, opened 2026-08-27
PHPLIB-1879: Support Client Backpressure with baseBackoffMS

paulinevos:
Part of DRIVERS-3535

By implementing Prose5.

### DIFF
--- a/src/Operation/WithTransaction.php
+++ b/src/Operation/WithTransaction.php
@@ -19,10 +19,10 @@ use function usleep;
 final class WithTransaction
 {
     /** Initial backoff time in ms */
-    private const BACKOFF_INITIAL = 5;
+    private const BACKOFF_INITIAL = 5.0;
 
     /** Maximum backoff time in ms */
-    private const BACKOFF_MAX = 500;
+    private const BACKOFF_MAX = 500.0;
 
     /** Default transaction timeout in seconds */
     private const MAX_TIME = 120;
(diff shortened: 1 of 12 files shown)


src/Operation/WithTransaction.php
```diff
@@ -202,7 +202,9 @@ private function commitTransaction(Session $session, int $startTime, int $transa
 
     private function computeBackoffMs(int $transactionAttempt): int
     {
-        return (int) floor($this->getJitter() * min(self::BACKOFF_INITIAL * (1.5 ** ($transactionAttempt - 1)), self::BACKOFF_MAX));
+        $exponentialBackoffMs = (float) self::BACKOFF_INITIAL * 1.5 ** (float) $transactionAttempt;
+
+        return (int) floor($this->getJitter() * min($exponentialBackoffMs, (float) self::BACKOFF_MAX));
```
GromNaN: I don't think any of the `(float)` cast are necessary. PHP convers to float automatically.
```suggestion
        return (int) floor($this->getJitter() * min([$exponentialBackoffMs](self::BACKOFF_INITIAL * (1.5 ** $transactionAttempt)), self::BACKOFF_MAX));
```

You can append `.0` to the constants to store them as float; but I'm not sure that gains any performance.

GromNaN:
We should pin the rector version. It finds upgrades (relevant in this case) at random time.

src/Operation/WithTransaction.php
paulinevos: No, Github security bot and psalm don't like that. But still applied your suggestion to spare us the cast

GromNaN:
Rebased on top of `v2.x`. Dropped the "Pin rector version" commit, as it is now redundant with the rector version already pinned on `v2.x`.

GromNaN (approved):

PR #1986, opened 2026-09-16
Fix dangling GridFS stream

alcaeus:
This PR fixes a rare issue with GridFS uploads. When uploading from a stream, `Bucket::uploadFromStream` relies on the `$destination` file pointer to go out of scope and closed, which then triggers the write to the databases. However, Xdebug can end up holding an additional reference to this file pointer, which then causes the write to be skipped.

To fix this, we wrap the entire write/return logic in a try-finally block to ensure the file pointer gets closed explicitly instead of relying on garbage collection to handle it for us.

Note: I haven't added a test for this as it is hard to reproduce. This appeared locally in my Doctrine ODM test suite as I run with Xdebug in `develop` mode by default. I confirmed locally that this patch fixes the failures I encountered.


alcaeus:
Test failures are unrelated.

